PR #3, opened 2022-06-09
Polyfill.io + Browser Compat Data + Browserslist data crunch

voxpelli:
To get an insight into whether a feature can work across browsers thanks to a polyfill or not I have started to run Browser Compat Data against Polyfill.io data.

It consists of three data sources:

1. The browser target is defined as a browserslist definition which is then resolved using the [`browserslist`](https://www.npmjs.com/package/browserslist) module
2. [polyfill.io](https://polyfill.io/v3/) data from their [`polyfill-library`](https://www.npmjs.com/package/polyfill-library) module
3. Browser Compat Data from [`@mdn/browser-compat-data`](https://github.com/mdn/browser-compat-data)

It also contains mappings of identifiers and data between these three sources and matching/filtering the other sources with the browser target from the first one.

### What primarily remains to be done:

- [ ] Do a useful correlation of the data and output
- [ ] Decide on a set of browsers to run through it

### What may not be covered?

Only polyfills available through polyfill.io are covered. There are other approaches that may make a feature usable / polyfilled. Such as:

* Through the use of a [ponyfill](https://ponyfill.com/)
* Through transpilation like Babel, Autoprefixer etc.

### DIFF
--- a/polyfill-data-crunch/.editorconfig
+++ b/polyfill-data-crunch/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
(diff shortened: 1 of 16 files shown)


li09882815965-gif (approved):
888h.co

PR #8, opened 2022-12-09
Add starter file for feature grouping effort

ddbeck:
This seeds a directory for bringing the work of work of [ddbeck/common-web-platform-feature](https://github.com/ddbeck/common-web-feature-mockup/) over here.

This week, Chris (I don't know his GitHub username, unfortunately) posed an interesting question to the WebDX CG call, which was, badly paraphrased: **what is a web feature?** Coincidentally, I was also wondering about this question myself a few weeks previous and had written myself an answer. I didn't think anyone would be interested in it, but since Chris asked, I've tidied it up and used it to start the README for the feature grouping work.

cc: @atopal

### DIFF
--- a/feature-group-definitions/README.md
+++ b/feature-group-definitions/README.md
@@ -27,6 +27,8 @@ It also means that a feature does not represent a random or arbitrary selection 
 For example, a group of JavaScript’s `Symbol` built-in object, the CSS `border-color` property, and the HTTP `POST` method is not a web feature, but all of the CSS border properties (individually or as a group) might be.
 
 The definition leaves the size the subset undefined on purpose. Logical subsets may be composed of a single specific feature or hundreds of them. For example, the [range syntax](https://drafts.csswg.org/mediaqueries-4/#mq-ranges) in Media Queries may be a web feature, but so would Media Queries as a whole.
+
+
 ### Addressable
 
 _Addressable_ means that a feature must be identifiable and distinguishable from other features.
--- a/feature-group-definitions/README.md
+++ b/feature-group-definitions/README.md
@@ -1,0 +1,63 @@
+# Feature groups
+
+This directory houses the experiment to provide structured definitions groups of web platform features.
+At inception, it consists of this README file only, but it should soon contain data and tools.
+See [ddbeck/common-web-feature-mockup](https://github.com/ddbeck/common-web-feature-mockup/) for more information.
+
+
+## What is a web platform feature?
+
+For the purposes of this project, consider this defintion of a web platform feature:
+
+**A _feature_ (or _feature group_) is a logical subset of all the addressable behaviors and interfaces that the web platform, as mediated by browsers and related tools, exposes to web developers.**
+
+You could come up with many reasonable definitions for "web platform feature",
+so this definition is surely not authoritative.
+But it might be useful for understanding the scope of this project.
+
+However, it's a mouthful. Take a look at each piece:
+
+
+### A logical subset
+
+_A logical subset_ means that all web technologies together are not a feature;
+all of the web platform features taken together **are** the web.
+It also means that a feature does not represent a random or arbitrary selection of web technologies.
+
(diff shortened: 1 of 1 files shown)


feature-group-definitions/README.md
```diff
@@ -0,0 +1,63 @@
+# Feature groups
+
+This directory houses the experiment to provide structured definitions groups of web platform features.
+At inception, it consists of this README file only, but it should soon contain data and tools.
+See [ddbeck/common-web-feature-mockup](https://github.com/ddbeck/common-web-feature-mockup/) for more information.
+
+
+## What is a web platform feature?
+
+For the purposes of this project, consider this defintion of a web platform feature:
```
ddbeck, line 10: On discussion with @atopal, I ought to come back to this to emphasize:

- The descriptive nature of the definition (i.e., that we're trying to capture things which are commonly true of features of the web, not provide a checklist or set of requirements)
- The narrow application of the definition (i.e., we're using this definition as part of an effort to understand web developers, their needs, and the ways they understand the web, which, by necessity, is an approximation of a great many individual perspectives)
- The incompleteness of this definition (i.e., that we're most concerned with the web developer audience and that other definitions might be more useful to specification authors or implementers or technical writers)

zcorpan:
IMO, we could think of features as trees where the root is a big chunk of things (e.g. SVG) and the leafs are the most specific subfeatures that we don't need to further divide into subfeatures (e.g. SVG vector-effect: non-scaling-stroke).

This means we can organize features and if necessary split or add child features in the taxonomy.

Given that feature lists such as caniuse or https://chromestatus.com/features range from broad to specific, it seems reasonable that we should be able to address "features" at different levels of granularity.

feature-group-definitions/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### A logical subset
+
+_A logical subset_ means that all web technologies together are not a feature;
+all of the web platform features taken together **are** the web.
+It also means that a feature does not represent a random or arbitrary selection of web technologies.
+
+For example, a group of JavaScript’s `Symbol` built-in object, the CSS `border-color` property, and the HTTP `POST` method is not a web feature, but all of the CSS border properties (individually or as a group) might be.
+
+
```
tidoust, line 29: Since this keeps coming back in discussions, it could be useful to be explicit that features may be at different granularity levels. I suggest the following text but feel free to update or to use some other example.

```suggestion
The definition leaves the size the subset undefined on purpose. Logical subsets may be composed of a single specific feature or hundreds of them. For example, the [range syntax](https://drafts.csswg.org/mediaqueries-4/#mq-ranges) in Media Queries may be a web feature, but so would Media Queries as a whole.
```

feature-group-definitions/README.md
```diff
@@ -0,0 +1,82 @@
...
+### Addressable
+
+_Addressable_ means that a feature must be identifiable and distinguishable from other features.
+
+For example, “a form” is not a web feature, but the HTML `<form>` element, the `HTMLFormElement` API, or a group of both (“forms”) might be.
+
+
+### Behaviors and interfaces
+
+_Behaviors and interfaces_ means that a feature must represent the things that the web platform can do, not specific instances of those behaviors and interface in use.
```
tidoust, line 39:
```suggestion
_Behaviors and interfaces_ means that a feature must represent the things that the web platform can do, not specific instances of those behaviors and interfaces in use.
```

tidoust (approved):

feature-group-definitions/README.md
ddbeck: I like this addition a lot. I think granularity is something implied in some of the other points, but never mentioned explicitly—it's good to highlight it here. 👍

ddbeck:
> IMO, we could think of features as trees where the root is a big chunk of things (e.g. SVG) and the leafs are the most specific subfeatures that we don't need to further divide into subfeatures (e.g. SVG vector-effect: non-scaling-stroke).

@zcorpan This is a good remark. I agree that we'll be treating these things as a tree, though I didn't want commit to a specific structure just yet (e.g., maybe there's multiple trees that aren't necessarily connected—right now the mockup doesn't put everything under a single tree, at least). I think François's suggested text covers the granularity question nicely, until we've got a stricter sense of things.

PR #16, opened 2022-12-20
Check feature spec URL against web-specs

foolip:
Closes https://github.com/web-platform-dx/feature-set/issues/12.


foolip:
Turning this into draft, after https://github.com/web-platform-dx/feature-set/pull/17 lands it will need to be a bit different, probably a lint and not part of the build step.

ddbeck (approved):

PR #21, opened 2022-12-23
Define built web-features package in its own directory

foolip:
It needs its own package.json and README. Even if they can be derived by
scripts, it's still convenient to use this structure. It's similar to
webref, and also the structure that Lerna expects. But it's also
possible to just run `npm publish` from the subdirectory.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,8 +17,8 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run build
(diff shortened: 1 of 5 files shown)


foolip:
I tried running `npm publish` from this branch, and it published without any prompt, so I guess the package exists now: https://www.npmjs.com/package/web-features

ddbeck (approved):
I like modeling after webref. That makes sense. 👍

PR #22, opened 2022-12-23
Add links to caniuse and a script to show mapping completion

foolip:


foolip:
This depends on https://github.com/web-platform-dx/feature-set/pull/21.

ddbeck (approved):

PR #23, opened 2022-12-23
Add 20 more features

foolip:
This is a mix of things that are in caniuse and things that aren't.


foolip:
This depends on https://github.com/web-platform-dx/feature-set/pull/22.

feature-group-definitions/css-has.yml
```diff
@@ -0,0 +1,2 @@
+spec: https://drafts.csswg.org/selectors-4/#relational
```
ddbeck, line 1: I'd prefer `has.yml` (or perhaps `has-pseudo-class.yml`) for this file (and similar for the other `css-*.yml` files below). More context here in https://github.com/web-platform-dx/feature-set/issues/24#issuecomment-1363913244.

feature-group-definitions/css-nesting.yml
```diff
@@ -0,0 +1,2 @@
+spec: https://drafts.csswg.org/css-nesting-1/
```
ddbeck, line 1: `nested-rules.yml` or `nested-styles.yml`. More fodder for the the naming convention discussion, but I think this PR is having me think of features as nouns and gerunds as cheating. 😄

feature-group-definitions/css-scoping.yml
```diff
@@ -0,0 +1 @@
+spec: https://drafts.csswg.org/css-cascade-6/#scope-atrule
```
ddbeck, line 1: Hmm, `scoped-rules.yml`? `scoped-styles.yml`? Even `scopes.yml`, alone?

feature-group-definitions/navigation-api.yml
```diff
@@ -0,0 +1 @@
+spec: https://wicg.github.io/navigation-api/
```
ddbeck, line 1: This and Sanitizer have boxed us in with one-word-and-API titles. This is probably fine.

feature-group-definitions/offscreencanvas.yml
```diff
@@ -0,0 +1,2 @@
+spec: https://html.spec.whatwg.org/multipage/#the-offscreencanvas-interface
```
ddbeck, line 1: `offscreen-canvas` as two words maybe? I suspect that "offscreen canvas" (the concept) and its interface names aren't quite the same thing (supported by the quick and dirty test showing that there are more Stack Overflow questions with "offscreen canvas" in the title than "OffscreenCanvas").

ddbeck (approved):
A few naming ideas for you, though not blockers.

feature-group-definitions/css-has.yml
foolip: I quite like the approach of using the shortest reasonable name and making them longer to disambiguate. So just "has" seems good to me here.

feature-group-definitions/css-nesting.yml
foolip: One point in favor of "nesting" is that the feature is known as such in [Sass](https://sass-lang.com/guide), and is also called this in [State of CSS](https://2022.stateofcss.com/en-US/usage/).

feature-group-definitions/css-scoping.yml
foolip: The syntax is `@scope` so something with the word "scope" is also an option. Although that doesn't seem like the best choice in the case of `@layer` so I think consistency won't be possible everywhere.

feature-group-definitions/css-nesting.yml
ddbeck: Oh cool—let's go with `nesting` then. I think names in common use ought to outweigh formalisms (even if I really, really like formalisms).

feature-group-definitions/css-scoping.yml
ddbeck: `scope` is good too! The main thing I want to steer away from is the `css` prefix. I'd rather we choose names that are familiar and match, as best we can, the names people actually use for those things.

(I also expect to add `shortName`, `longName`, `description`, and such so we have access to other words for presentation and search.)

feature-group-definitions/offscreencanvas.yml
foolip: Looks like it is sometimes referred to as "offscreen canvas", so sure! Example: https://webrtc.ventures/2022/02/offscreen-canvas-api/

feature-group-definitions/navigation-api.yml
foolip: I think it is fine if we have to use -api to disambiguate, but just `navigation` here actually seems reasonable. There's a `window.navigation` property which can be used like `navigation.navigate(url)`. So I think I'll drop the suffix actually.

foolip:
I've done some renaming, and the 29 features in total are now:

- anchor-positioning
- border-image
- broadcast-channel
- cascade-layers
- container-queries
- content-visibility
- custom-properties
- dialog
- flexbox
- fullscreen
- has
- idle-detection
- image-set
- import-maps
- leading-trim
- line-clamp
- navigation
- nesting
- offscreen-canvas
- sanitizer
- scope
- sticky-positioning
- subgrid
- temporal
- webcodecs
- webdriver-bidi
- webtransport
- webusb
- webvtt

An "as short as possible without being ambiguous" guideline would probably push us to make these even shorter:
  - anchor-positioning &rarr; anchor
  - cascade-layers &rarr; layer
  - container-queries &rarr; container
  - flexbox &rarr; flex
  - sticky-positioning &rarr; sticky

When to use the "web" prefix is also going to get complicated, but what we have so far seems OK.

foolip:
@ddbeck do you want to review my recent renaming and idle musings above?

ddbeck:
@foolip This all looks really nice.


> An "as short as possible without being ambiguous" guideline would probably push us to make these even shorter:
> - anchor-positioning &rarr; anchor
> - cascade-layers &rarr; layer
> - container-queries &rarr; container
> - flexbox &rarr; flex
> - sticky-positioning &rarr; sticky

I think these are all good (at least for now). I think some of them have issues shortened (e.g., "flexbox" seems like the conventional name and "anchor" might confuse with `<a>` tag)—I can live with what's here now. 😄

ddbeck:
(Oh, do you have a sense of how you want to handle merges here? Do you want to merge your own PRs?)

foolip:
@ddbeck thanks for the review! I'm happy for you to merge when you feel like it, if I really don't want something merge in the current state I'd make it draft. In this case I had additional thoughts after you reviewed so I wanted to see what you had to say about that first. I agree that neither flexbox or anchor-positioning are improved by shortening further.

I'll go ahead and merge this now. Hopefully I've made some mistake so we can learn how to handle renaming :D

PR #25, opened 2022-12-23
Add a schema for the data so far and validate against it

foolip:


foolip:
This depends on https://github.com/web-platform-dx/feature-set/pull/22.

ddbeck (approved):
Thank you!

PR #29, opened 2023-01-04
Make the caniuse script quieter for testing

ddbeck:
While playing with bringing things over from the mockup, I found that it was nice to have less output from successful `npm test` runs. This is, admittedly, a minor annoyance—consider it optional.


foolip:
I'd like to see this fixed too. The way that I've usually fixed this in other projects is to use some kind of logger that can be configured to be silent when testing. That way the log sites don't all have to check the "log level", which is a boolean in this case.

In BCD we have a custom logger, but a simple one using winston can be found here:
https://github.com/foolip/mdn-bcd-collector/blob/main/lib/logger.ts

WDYT about doing something similar here?

ddbeck:
That's a good idea. I figured I ought to do something like that (and use an arg parsing library but that's a different can of worms) but I didn't have a library in mind. Glad to take this suggestion. 👍

foolip (approved):

PR #30, opened 2023-01-04
Deduplicate node version in GitHub Actions

ddbeck:
This has the added benefit of working with nodenv.

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file


foolip (approved):
Neat!

foolip:
But it seems like I'm not allowed to merge this... can you try?

foolip:
Hmm, now I could. Might have been a network issue.

PR #37, opened 2023-01-12
Add popover feature

foolip:
Added to HTML here:
https://github.com/whatwg/html/pull/8221
https://github.com/whatwg/html/pull/8717


foolip:
It's landed in the spec again.

ddbeck (approved):
Thanks!

PR #38, opened 2023-01-12
Add initial docs with naming conventions

ddbeck:
This is to give form to some of the things we've discussed so far in https://github.com/web-platform-dx/feature-set/issues/24 and https://github.com/web-platform-dx/feature-set/issues/23.

I don't expect this to be the last word on the subject, as we create more feature descriptions and understand how they all fit together.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -1,0 +1,40 @@
+# Project documentation
+
+## Feature guidelines
+
+### Identifiers
+
+Feature identifiers must contain only lowercase alphanumeric characters (`a`-`z` and `0-9`) plus the `-` character (hyphen or minus sign) as a word separator.
+
+Feature authors should (in descending order of priority):
+
+- Prefer identifiers to known to be in widespread use by web developers.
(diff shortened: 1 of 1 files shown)


docs/readme.md
```diff
@@ -0,0 +1,40 @@
...
+  - 👎 Not recommended: `offscreencanvas` (as in `OffscreenCanvas`)
+  - 👍 Recommended: `grid`
+  - 👎 Not recommended: `display-grid` (as in `display: grid`)
+
+- Prefer shorter identifiers to longer identifiers, as long as they're unique and unamibguous.
+
+  - 👍 Recommended: `has`
+  - 👎 Not recommended: `has-pseudo-class`
+
+Feature identifiers may use common suffixes (such as `-api`) to resolve naming conflicts.
```
foolip, line 40: We may need to revisit this when we have a real conflict. I've tried but can form no opinion about whether this is good guidance without some real examples.

foolip (approved):

PR #44, opened 2023-01-19
Generate JSON Schema definitions from types

ddbeck:
This fell out of playing around with expanding the YAML with constituent features data. I don't know if this is a good idea, but it is _an_ idea. This will be modestly more interesting after #40.


foolip:
@ddbeck I'm not sure I understand what this does. Is it purely refactoring, or will it catch errors that weren't caught before?

ddbeck:
@foolip Sorry, I did a lousy job of contextualizing this.

Right now, we've got the schema (in a broad sense) defined in two places: a JSON Schema file and a TypeScript interface in `index.ts`. On another PR, I changed one and not the other and thought to myself: surely these ought to be the same? And if they ought to be the same, must we remember to make sure they're the same?

So this PR does a few things:

- It adds an `npm run` command to generate JSON Schema from the TypeScript interface definition in `index.ts`. The generated bits go into a fully-generated `defs.schema.json` file (unfortunately, the top-most object can't be generated because TypeScript lacks a way to restrict keys to the corresponding `patternProperties` in JSON Schema).
- It adds annotations to `index.ts` to supply extra details to the JSON Schema generator that TypeScript doesn't have a direct analogue for (i.e., descriptions and string formats). 
- It changes the way the schema and schema test script work, to use the `defs.schema.json` file.

Your comment does suggest one additional step, which I have not done (but could do): test that a schema derived from `index.ts` does not diverge with the `defs.schema.json` file. This would actually catch the mistake of not updating the schema when `index.ts` changes (or vice versa).

foolip:
Thanks @ddbeck! If I understand the setup in BCD correctly that generates TypeScript definitions from the schema using json-schema-to-typescript, and this PR is generating the schema from TypeScript using ts-json-schema-generator.

I don't have any opinions about which direction is better, so this seems fine :)

foolip (approved):

PR #46, opened 2023-01-23
Add usage metrics to schema

foolip:


foolip:
@ddbeck this is the very simplest way I could come up with to link a feature to a usage metric. Firefox also has [use counters](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/use-counters.html), although I don't know where the results are published. My idea was that if we add those, we'd just allow `usage` to be a URL or an array of URLs. But maybe it would be better to have `chrome_usage_url`, and later a `firefox_usage_url`?

ddbeck:
A couple thoughts:

- Perhaps this could be named `usage-stats` or something else more descriptive? I could imagine `usage` alone being confused for something like a documentation link.
- Perhaps the `usage-stats` (or whatever it comes to be called) could be an array of URLs? I'd expect the URLs themselves to be somewhat self-explanatory. It seems pretty unlikely that a chromestatus URL is going to report out Firefox usage.

foolip:
How about `usage_stats` with an underscore? Matching the style of BCD seems better in case we move things between the projects, or mix them into a third data set.

ddbeck:
@foolip Underscores works for me (and I have a vague recollection that we came to that conclusion previously). That works for me. Do you want to change it in this PR? And did you have any thoughts on the array idea?

foolip:
@ddbeck I've made the change. On arrays, I've followed BCD's model of allowing a string or an array of strings. I've also made it easy to see how we'd make the same change for specs, but am not making the change in this PR.

ddbeck (approved):

PR #47, opened 2023-01-23
Alternate initial feature support definition

ddbeck:
Noodling inspired by @foolip's https://github.com/web-platform-dx/feature-set/pull/40#discussion_r1070823059 (and a discussion with Kadir).

## What is this?

This is an alternative approach to defining features that distinguishes between what constitutes that feature (some syntax, an interface, methods, elements, etc.) and the sources we use to determine whether those things, taken together, are supported.

## Why would we do this?

The main benefit of this approach is authorability and reviewability: a spec author or browser engineer doesn't need to know anything about BCD's idiosyncrasies to write or update a feature definition. This makes it easier to collaborate on high-level feature groups.

A secondary benefit is that we might support more than one mapping without paying a direct penalty in terms of feature definition size and complexity (that is, adding a second source doesn't imply a doubly-long feature YAML file). I've illustrated this with a _strictly hypothetical and extremely naïve_ WPT mapping (though I know about 👉👈 this much about WPT test cases and results, so I couldn't tell you if it's a _good_ mpaping).

## What other work does this imply?

In a better world, the `.map.yml` file would be unnecessary. For example, BCD might have an API (or, more likely, a companion library) that can take a "natural" name (e.g., `CSSImportRule.prototype.layerName`) and return the compat data out of BCD's internal structure (e.g., `api.CSSImportRule.layerName`). We could write that library, but in the near term, we don't have to commit to the increased scope. Instead, we write the mappings ourselves (which is what my other proposal does anyway, more or less).


foolip:
I've given some thought to grouping WPT by feature as well, to allow showing the tests for a specific feature on wpt.fyi. https://github.com/foolip/wpt/commit/ebbbffcfac3e770697a4c21cc6d35acd9850a499 is the approach I tinkered with then. (A lot is missing to actually have the wpt.fyi feature.)

I believe that grouping by high-level feature is tractable and that the most practical way to do it is with metadata in the WPT repo itself. However, one could get started with out-of-band metadata as well, it would just be more work to maintain over time.

I think that mapping at a more granular level is going to be far more work, as identifying which tests are for `CSSImportRule.prototype.layerName` requires first searching for the string "layerName" and then checking if it's really accessing it on a `CSSImportRule` instance.

This is all to say that I think the level of indirection being suggested here, with more granular "constituent features" that are in turn mapped to other data sources, will be more work and my inclination is to start simpler, without an indirection.

dontcallmedom:
re figuring out if an actual API end point is used in a test, I had done some toying in that space with https://github.com/dontcallmedom/wpt-coverage-tracker (see also https://lists.w3.org/Archives/Public/public-test-infra/2020OctDec/0002.html )

ddbeck:
Thanks for the feedback, y'all. I've cherry picked the formatting work as https://github.com/web-platform-dx/feature-set/pull/61. But otherwise it sounds like this is not the way forward, so I'm going to close this PR.

PR #78, opened 2023-03-01
Add five new web platform feature groups

ddbeck:
Hi @captainbrosset and @AnthumChris! Thanks for taking the time to review some of this. Hopefully this won't take a ton of time, but feel free to limit the time you spend on this to a level that's comfortable for your schedule and interest.

This PR adds five new features to the feature set. I'd like your first impressions, as someone who knows about web technologies in general (though perhaps not specifically _these_ features), but does not know anything about the specifics of feature-set.

**Please note that this is an experiment.** So far, Philip and I are the only ones to have reviewed new features. We don't yet know what other reviewers might need to make good use of reviewing time. Finding the stumbling blocks is as important as coming up with good features and definitions for them.

Third, I've written some optional prompts in the next section to help you understand what I'm looking for, but feel free to go off script. 🙂 

## Feedback hints

I'm looking for feedback both substantive and procedural. Here are some prompts which may guide you:

- While you review, make a note of any questions you have. What's obvious and familiar? What's unclear, unknown, or confusing to you?
- Look at the file names. Do you recognize a feature? Does the file name give you an idea about what feature I'm talking about?
- Compare each file name to the file's contents. Does content seem closely related to the name? If you recognized a feature by name, are the contents consistent with your expectations about what you'd find inside?
- Look at the file's `compat_features`. I don't expect you're a [browser-compat-data](https://github.com/mdn/browser-compat-data) expert, but does the list of related features look plausible?
- If you have independent knowledge of the feature, are you surprised by anything you see? Are you surprised by anything you did _not_ see?
- What would you expect to see next? Any features you're curious about?


feature-group-definitions/grid.yml
```diff
@@ -0,0 +1,30 @@
+spec: https://drafts.csswg.org/css-grid-3/
+caniuse: css-grid
+compat_features:
```
captainbrosset, line 3: I think we're missing the `place-content`, `place-items`, `place-self`, `align-content`, and `justify-content` properties.
You've got `self` and `items` for both `justify` and `align` directions, but you also need `content` (which applies to a grid container element and applies to the alignment "subjects" in the grid, aka the row and column tracks). And you need `place` too, which is the short-hand for `align` and `justify`.

feature-group-definitions/grid.yml
```diff
@@ -0,0 +1,30 @@
...
+  - css.properties.align-items
```
captainbrosset, line 4: Can we add the `minmax` function too? I don't know where it is in BCD right now though. Maybe it's already included in the `repeat` function which you've already added.
Also, the `auto-fit` and `auto-fill` values should be in there somewhere (they probably already are in the `repeat` function in fact).

feature-group-definitions/grid.yml
```diff
@@ -0,0 +1,30 @@
...
+  - css.properties.align-self
```
captainbrosset, line 5: (sorry if I keep adding general comments on unrelated lines of code, I just didn't want to add everything in one big comment, to allow for discussion threads).

How do we think about subgrid? Technically, it is *just* a property value you can use with `grid-template-rows/columns`, but it's considered by most developers who know about grid as an entirely separate feature.

In this group, would it just naturally be included by `css.properties.grid-template-*`? Or should be it appear as its own subgroup, or something else?

captainbrosset: Same comment for masonry, which is also a value that can be used.

feature-group-definitions/font-face.yml
```diff
@@ -0,0 +1,8 @@
+spec: https://drafts.csswg.org/css-fonts-5/
```
captainbrosset, line 1: To me, font-face isn't really a feature. At least I've never referred to it as such. As a web dev, I want to load a custom font, use one that I've purchased, or use a freely available web font.

So, to me at least, the feature here is "Custom font" (there are probably better names for it).

Using the at-font-face rule is the way to make it work from CSS, but you can also use the (upcoming) Local Font Access API (https://chromestatus.com/feature/6234451761692672).

Also, the compat_features here should probably include the FontFace interface: https://developer.mozilla.org/en-US/docs/Web/API/FontFace.

Unless there's a parent group that's called "Custom fonts" that contains this group, and another one called local-fonts, and another called font-face-js-access, or something.

I guess I'm trying to understand what the right granularity here is.

feature-group-definitions/custom-elements.yml
```diff
@@ -0,0 +1,14 @@
+# Tests don't allow the more specific https://html.spec.whatwg.org/multipage/custom-elements.html (yet)
+spec: https://html.spec.whatwg.org/multipage/
+caniuse: custom-elementsv1
+compat_features:
+  - css.at-rules.containe
+  - css.properties.container-nam
+  - api.CustomElementRegistr
+  - api.CustomElementRegistry.builtin_element_suppor
+  - api.CustomElementRegistry.defin
```
captainbrosset, line 9:
```suggestion
  - api.CustomElementRegistry.define
```

feature-group-definitions/custom-elements.yml
```diff
@@ -0,0 +1,14 @@
...
+  - api.Window.customElement
+  - css.selectors.define
```
captainbrosset, line 11:
```suggestion
  - css.selectors.defined
```

feature-group-definitions/custom-elements.yml
```diff
@@ -0,0 +1,14 @@
...
+  - css.selectors.hos
+  - css.selectors.host-contex
+  - css.selectors.par
```
captainbrosset, line 14: These also seem to be cut off.

captainbrosset: Haha, now that I look at the file content in more details, every line seems to have a missing character.

feature-group-definitions/custom-elements.yml
```diff
@@ -0,0 +1,14 @@
+# Tests don't allow the more specific https://html.spec.whatwg.org/multipage/custom-elements.html (yet)
+spec: https://html.spec.whatwg.org/multipage/
```
captainbrosset, line 2: Should this link to the #custom-elements anchor on that page?

captainbrosset (commented):
Love where this is going. I left a few comments which, I hope, will be useful.
In terms of next steps, I'd be curious to see what a "web components" group would look like (since you've already done custom elements and shadow dom, it seems like a good next step).

feature-group-definitions/custom-elements.yml
ddbeck: Oh wow, I have no idea how that happened. I also managed to copy and paste some container queries features from spreadsheet too. 🤦  I'll fix this shortly (then come back to your other insightful comments, after giving Chris a little more time to chime in).

feature-group-definitions/custom-elements.yml
foolip: That would pass the lint at least. But I think we should change the lint to allow for links like https://html.spec.whatwg.org/multipage/custom-elements.html. The extra complexity is checking that those links actually work.

feature-group-definitions/font-face.yml
foolip: I think a common term for this feature is just "web fonts", which is also what https://caniuse.com/fontface has in the description.

feature-group-definitions/task-scheduling.yml
```diff
@@ -0,0 +1,13 @@
+spec: https://wicg.github.io/scheduling-apis/
```
foolip, line 1: How about calling this feature just "scheduler"? That's the simplest name for it I can think or. Or "scheduling", if we think that "idle detection" is a precedent.

foolip (approved):

feature-group-definitions/grid.yml
captainbrosset: Sounds great to me.

feature-group-definitions/grid.yml
foolip: If the data is structured correctly the question becomes: should we treat aliases or prefixes as full support, and making the feature eligible for baseline, or should we treat it like an important note that should be called out when support is summarized? I'd lean towards the latter.

ddbeck: That's a good question! I've opened https://github.com/web-platform-dx/feature-set/issues/86 to preserve it (and to raise some closely related questions).

ddbeck:
Thanks for the reviews, y'all. I'm going to merge this and dig into some of the consequences. 👍

PR #88, opened 2023-03-09
Add initial script to generate a release PR

ddbeck:
OK, this got to be rather more complex than I expected. In the interests of keeping review sizes manageable, this is part one of the effort to automate packages releases. Some highlights:

- This creates a new script for managing releases. You invoke it with `npx ts-node ./scripts/release.ts init`. Right now you could do that manually, but later we'll set up a GitHub Action to do this on some schedule or logic. If you uncomment the line containing `fork` you can test this against my fork without creating noise on this repo, if you care to do so.

- The only implemented part here is the versioning and PR generation command called `init`. It does the following:

  1. Runs various preflight checks to make sure we're doing some sensible (tools available, auth, base branch, etc.).
  2. Gets the JSON from currently released package, builds a fresh new one (via `npm run build`) and diffs them (pretty like, with help from `jq`).
  3. Bumps the version (patch by default), makes a commit, and creates a PR with a suitably scary (though currently incorrect) description and a diff.

- You can see an example result here: https://github.com/ddbeck/feature-set/pull/6.


scripts/release.ts
```diff
@@ -0,0 +1,265 @@
...
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+import winston from "winston";
+import yargs from "yargs";
+
+const loggerTransport = new winston.transports.Console({ level: "silly" });
```
foolip, line 17: Silly? What does this even mean? 😆

scripts/release.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  // Push release branch
+  logger.info("Pushing release branch");
+  const pushCmd = `git push --set-upstream origin ${releaseBranch}`;
+  run(pushCmd);
+
+  // Create PR
+  logger.info(`Creating PR for ${version}`);
+  const title = `📦 Release web-features@${version}`;
+  const reviewer = "ddbeck";
```
foolip, line 120: 👍

scripts/release.ts
```diff
@@ -0,0 +1,265 @@
...
+function run(cmd: string) {
+  execSync(cmd, { stdio: "inherit" });
+}
+
+function diffJson(): string {
+  const temporaryDir = mkdtempSync(join(tmpdir(), "web-features-"));
+
+  execSync("npm install web-features", {
+    cwd: temporaryDir,
+    encoding: "utf-8",
```
foolip, line 166: Is `encoding` needed in other calls to `execSync` as well?

scripts/release.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  try {
+    const result = execSync(
+      `diff --unified "${prettyReleasedJsonFp}" "${prettyPreparedJsonFp}"`,
+      { encoding: "utf-8" }
+    );
+    rmSync(temporaryDir, { recursive: true });
+    return result;
+  } catch (err) {
+    if (err.status === 1) {
```
foolip, line 197: Can you comment what this all is about?

foolip (approved):

scripts/release.ts
ddbeck: Really, it's superfluous here. Without a `stdio` option (or capturing the result of `execSync` and doing something with it), the output of `npm install web-features` disappears. I ought to either drop the encoding or let it rip with `stdio: "inherit"`. I've decided to go with the latter, at least for now.

As it stands, this whole thing is mostly a fancy shell script rather than a quiet, well-behaved standalone utility. If we find this thing performs reliably, we can suppress the output from the commands it runs, or make it toggle-able with a flag.

scripts/release.ts
ddbeck: Good idea. Added a comment with 36c9de6.

PR #93, opened 2023-03-10
Add more compat features (background Fetch + off-screen canvas)

ddbeck:
Wrapping up my work week with a few more features out of my spreadsheet: Background Features and Offscreen Canvas. These are starting to get long (I've got longer ones still to add) and is getting me curious about how these will become reviewable.

One maybe obvious idea: confirm that the features listed exist in BCD?


foolip (approved):
I think we'll need wildcards or some way to list an interface with all of its members, but still notice when we update BCD if the list of matched features changes. One way to think about that is like a lockfile, where we put both the simple-ish input and the expanded output under source control, and ensure they stay in sync through CI.

ddbeck:
> One way to think about that is like a lockfile, where we put both the simple-ish input and the expanded output under source control, and ensure they stay in sync through CI.

That's an excellent idea. I'll create an issue out of it. 👍

PR #107, opened 2023-03-23
More release automation

ddbeck:
This PR adds to the release script:

- An `update` command to rebase (and, optionally, bump the version number)
- A `publish` command to build and tag a release, then publish to npm

I sort of let this branch get away from me a little with various refactorings along the way—I'll do a self-review to call out some important bits.

The follow up to this is to set up GitHub Actions workflows to actually run `init | update | publish` in response to various events.

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -20,8 +20,6 @@ const logger = winston.createLogger({
 });
 
 const pullTitleBase = `📦 Release web-features@`;
-let targetRepo = "web-platform-dx/feature-set";
-// targetRepo = "ddbeck/feature-set"; // Uncomment this line to point this at a fork for testing
 
 const packages = {
   "web-features": fileURLToPath(
@@ -32,6 +30,11 @@ const packages = {
 yargs(process.argv.slice(2))
   .scriptName("release")
   .usage("$0 <cmd> [args]")
+  .option("target-repo", {
+    describe: "Select upstream GitHub repository",
+    nargs: 1,
+    default: "web-platform-dx/feature-set",
+  })
   .command({
     command: "init <semverlevel>",
     describe: "Start a new release pull request",
@@ -122,7 +125,7 @@ function init(args) {
     `--body-file=-`,
     `--base="main"`,
     `--head="${releaseBranch}"`,
-    `--repo="${targetRepo}"`,
+    `--repo="${args.targetRepo}"`,
   ].join(" ");
   execSync(pullRequestCmd, {
     input: body,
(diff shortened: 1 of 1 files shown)


scripts/release.ts
```diff
@@ -33,53 +30,68 @@ const packages = {
...
     builder: (yargs) => {
-      return yargs
-        .positional("pr", {
-          describe: "the PR to rebase and update",
-        })
-        .demandOption("pr");
+      return yargs.option("dry-run", {
+        type: "boolean",
+        describe: "Do everything short of publishing",
+        default: true,
```
ddbeck, line 81: Safety first!

scripts/release.ts
```diff
@@ -33,53 +30,68 @@ const packages = {
...
 function init(args) {
-  preflight();
+  preflight({ expectedBranch: "main" });
 
   const diff = diffJson();
 
   // Start a release branch
   // Convention borrowed from https://github.com/w3c/webref/blob/60ebf71b9d555c523975cfefb08f5420d12b7293/tools/prepare-release.js#L164-L165
-  const releaseBranch = `release-${new Date()
+  const releaseBranch = `release/web-features/${new Date()
```
ddbeck, line 94: This groups the branches generated by this process. I thought this might help avoid name collisions (which I kept tab completing my way into 🙄).

scripts/release.ts
```diff
@@ -93,71 +105,156 @@ function init(args) {
...
+  try {
+    run(`git rebase ${args.base}`);
+  } catch (err) {
+    logger.error("Rebasing failed. Abandoning PR.");
+    run(`git rebase --abort`);
+    run(
+      `gh pr comment --repo="${args.targetRepo}" "${args.pr}" --body="😱 Rebasing failed. Closing this PR. 😱"`
+    );
+    run(`gh pr close --repo="${args.targetRepo}" "${args.pr}"`);
+    process.exit(1);
```
ddbeck, line 195: If the rebase fails, it closes the PR. My thinking here is that instead of being clever, we can wait for (or force) a new release PR instead.

scripts/release.ts
```diff
@@ -93,71 +105,156 @@ function init(args) {
...
+  let publishCmd = `npm publish`;
+  if (args.dryRun) {
+    publishCmd = `${publishCmd} --dry-run`;
+  }
+  execSync(publishCmd, { cwd: packages["web-features"], stdio: "inherit" });
 }
 
 function run(cmd: string) {
   execSync(cmd, { stdio: "inherit" });
 }
```
ddbeck, line 243: This file isn't very well organized—the functions are in no particular order. But it made the diff even harder to follow. Perhaps something to tidy up either just before merging, or in a separate PR.

scripts/release.ts
```diff
@@ -93,71 +105,156 @@ function init(args) {
...
+  const body = [
+    readFileSync(bodyFile, { encoding: "utf-8" }),
+    "```diff",
+    diff,
+    "```",
+  ].join("\n");
+  return body;
 }
 
 function update(args) {
```
ddbeck, line 165: Here's where the rebase-and-bump stuff happens.

PR #110, opened 2023-03-24
Experiment with lockfiles for compat feature lists

ddbeck:
@foolip Today, I took some time to experiment with the idea logged in https://github.com/web-platform-dx/feature-set/issues/96: allowing a shorthand notation for compat features (e.g., a wildcard for descendant features) and recording the actual features via a lockfile.

This isn't really ready to be used yet (I've skipped over the tedious bit of crawling BCD), since I wanted to check that this overall approach made sense before putting more effort into it.

Some initial notes:

- This generates locks when it constructs the full data structure in `index.ts`. Truthfully, I'm not sure if this is the right place for it. *Should* it happen at this point? Or should creating and validating locks be a separate thing, while `index.ts` merely takes advantage of the presence of lock data, if it's available?
- This generates a lockfile for each feature YAML file. That will double the number of files, but it may be easier for people to reason about when reviewing changes. (It was also easier for me to implement as a quick Friday afternoon experiment, though that's not a real concern.)
- The hash for each lockfile is based on the totality of the feature YAML file. That is, _any_ change in the values of the YAML file should invalidate its corresponding lockfile. This is a little forward looking, but maybe a wrong idea. Should we pick-and-choose what data counts?


foolip:
I've pondered this a bit and I think we might benefit from expanding this slightly. The value of expanding the lists of features is to make it clear to reviewers what BCD features are actually included, but that's not the only relevant information. If there are notes, `partial_implementation`, or similar that factor into the derived status, that's also important to surface.

So I wonder if we should write files that capture something like this:

```yaml
status: baseline
notes: "doThing() doesn't work in Chrome"
support:
  chrome: 84
  firefox: 50
  safari: 13.1
lockstuff:
  - api.SomeThing.doThing
     notes:
       chrome: "Doesn't work"
  - api.SomeThing.happen_event
```

Whether that's all in a single file, or if we only put the `lockstuff` stuff in a file, I'm not sure.

But in any case, this would be like a build output, and we want to make it very easy to review this output and ensure it doesn't accidentally change. The most straightforward way of doing this is by putting the files in the repo, but it could also be an elaborate CI setup.

foolip:
@ddbeck is this ready for review again? Any tradeoffs to highlight?

ddbeck:
@foolip This is ready for another review. The idea now is to split authored data (`*.yml`, a feature definition) from computed data (`*.lockfile.yml`, a lockfile—though we still need a better name). The computed files contain all of the data that gets put into the published package.

The intended procedure is this:

1. Read the feature definition.
2. For each key in the feature definition, either:

   a. Pass it through as-is to the lockfile, or
   b. Compute some result (e.g., expand BCD keys or determine a simplified support status) and then put it in the lockfile.

3. Use the lockfile data (only) to construct and export the complete features object.

3. If the feature definition and the lockfile are materially the same (that is, making a JSON-ish comparison without comments or excluded fields as equal), delete the lockfile as needless duplication.

Some hopes I have for this approach:

- Editorial overrides will have a clear home: in feature definition files. For instance, we could make it such that simplified support statuses in feature definitions overrule computed results (e.g., when you have a showstopper bug in an implementation).
- Material changes to the data we publish become directly reviewable. For example, if changes in BCD cause a feature's baseline status to change, we'll see that in changes to the lockfiles. Likewise, if we change the definition of a feature and it changes the baseline status, that will also show in changes to the lockfiles.
- Unwanted changes make themselves more easily known. For example, if a refactor produces lockfile changes, we don't need a separate test to know that. Similarly, immaterial changes become obvious and safer to make. For example, changing a comment, source style, or some private editorial field won't affect the lockfile, so we can make such changes more freely.
- Taken together, reviewing release PRs will be safer: the in-description JSON diff will be more informative than substantive—the data changes will have been visible to the people most likely to have knowledge and a stake in the results specific to a particular feature.

Some drawbacks:

- There will be more files to review. Eventually, every feature will have two files. I've tried to reduce that be eliminating files when there'd be no difference between the two, but that slows rather than eliminates the proliferation of files. An alternative might be a monolithic lockfile.
- Right now, the files live alongside their definitions. I imagine there might be some frustration with attempts to manually edit generated files. I'm not sure there's a good fix for that.

foolip:
> Unwanted changes make themselves more easily known. For example, if a refactor produces lockfile changes, we don't need a separate test to know that. Similarly, immaterial changes become obvious and safer to make. For example, changing a comment, source style, or some private editorial field won't affect the lockfile, so we can make such changes more freely.

I love this part.

foolip (commented):
This approach makes sense to build. Before we can land it `expandCompat` needs to be implemented, and we need a more accurate name for what these files are. In a way they're just build output that we're tracking because it's convenient, so one option is to just put it in a `dist/` or `built/` directory. If we put files alongside the originals, maybe `feature.dist.yml`? If we make the output JSON instead of YAML it could just be `feature.json`, but I don't think that helps make it clear (at all) what the source of truth is.

ddbeck:
Thank you, Philip! I'll proceed to develop this more. Good idea about a `built` directory—I'd like to see how that "feels" as well. 👍

PR #111, opened 2023-03-24
Fix formatting test

ddbeck:
`npm test` was _applying_ the formatter, not checking whether prettier was satisfied. This PR fixes the mistake and applies YAML formatting to the one file currently out of compliance.


foolip (approved):

foolip:
Looks like prettier isn't a dependency.

PR #120, opened 2023-03-30
Add two JavaScript features: `array-group` and `array-at`

ddbeck:
Trying to get into the habit of adding _something_ on a regular basis. These are the last of the (non-huge) APIs in my spreadsheet from the mockup.

This adds:

- Array grouping (this one is boring and straightforward)
- Array relative indexing

The latter entry has two interesting quirks:

- It consists of one compat feature. A group of one is perhaps a degenerate case. But I suspect its relationship to other groups (including, eventually, an `Array` group) might merit its inclusion. Any thoughts on this?
- The spec URL limitations continue to be a problem. Needs loosening sooner rather than later, I think.


feature-group-definitions/array-relative-indexing.yml
```diff
@@ -0,0 +1,4 @@
+spec: https://tc39.es/ecma262/multipage/
```
foolip, line 1: How about we call this features just `array-at`?

foolip (approved):

feature-group-definitions/array-grouping.yml
```diff
@@ -0,0 +1,4 @@
+spec: https://tc39.es/proposal-array-grouping/
```
foolip, line 1: Naming consistency with https://github.com/web-platform-dx/feature-set/pull/128 would be great here.

If that stays `array-flat-and-flatmap` then this could be `array-group-and-grouptomap`, and if you rename it `array-flat` then this could be `array-group`.

Have short names that seem like they include only part of the feature isn't great, but it's OK I think.

ddbeck: `array-group` seems like a good choice. I had a look over at Stack Overflow and I think this might also be reasonably called "group by" but "by" doesn't appear in the method names and there's [some interesting background](https://github.com/tc39/proposal-array-grouping#why-group-and-not-groupby) as to why it wasn't.

ddbeck:
> It consists of one compat feature. A group of one is perhaps a degenerate case. But I suspect its relationship to other groups (including, eventually, an `Array` group) might merit its inclusion. Any thoughts on this?

For posterity, Philip and I discussed this on a call earlier today. We decided that single-feature groups are OK: they're significant in their relationship to other groups and may be necessary to tell the story of a group evolution (e.g., Array acquiring new methods which eventually just become routine parts of the Array API).

PR #122, opened 2023-03-31
Allow more permissive spec URLs

ddbeck:
This changes the spec test to allow spec URL tests to pass when:

- A spec URL ~~starts with a URL~~ is known to `web-specs` (i.e., is one of `web-specs` `url` or `pages` strings, excluding hash fragments)
- A spec URL is specifically allowlisted

Merging this will unblock https://github.com/web-platform-dx/feature-set/pull/116.

### DIFF
--- a/scripts/specs.ts
+++ b/scripts/specs.ts
@@ -16,7 +16,7 @@ const defaultAllowlist: allowlistItem[] = [
 
 function isOK(url: URL, allowlist: allowlistItem[] = defaultAllowlist) {
     for (const specUrl of specUrls) {
-        if (specUrl.protocol !== url.protocol || specUrl.origin !== url.origin) {
+        if (specUrl.origin !== url.origin) {
             continue;
         }
         if (url.pathname.startsWith(specUrl.pathname)) {
--- a/scripts/specs.ts
+++ b/scripts/specs.ts
@@ -1,18 +1,37 @@
+import assert from "node:assert/strict";
+
 import webSpecs from 'web-specs' assert { type: 'json' };
 
 import features from '../index.js';
 
-const specUrls = new Set(webSpecs.map(spec => spec.nightly.url));
+const specUrls = webSpecs.map(spec => new URL(spec.nightly.url));
+
+function isOK(url: URL) {
+    for (const specUrl of specUrls) {
+        if (specUrl.protocol !== url.protocol || specUrl.origin !== url.origin) {
+            continue;
+        }
+        if (url.pathname.startsWith(specUrl.pathname)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+function testIsOK() {
+    assert.ok(isOK(new URL("https://tc39.es/ecma262/multipage/")));
+    assert.ok(isOK(new URL("https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.at")));
+    assert.ok(!isOK(new URL("https://typo.csswg.org/css-anchor-position-1/#anchoring")));
+};
+testIsOK();
 
 let checked = 0;
 let errors = 0;
 
 for (const [id, data] of Object.entries(features)) {
     const url = new URL(data.spec);
-    // Remove any #fragment from end of URL.
-    url.hash = '';
-    if (!specUrls.has(url.toString())) {
-        console.error(`URL for ${id} not in web-specs: ${url}`);
+    if (!isOK(url)) {
+        console.error(`URL for ${id} not in web-specs: ${url.toString()}`);
         errors++;
     }
     checked++;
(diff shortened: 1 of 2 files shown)


scripts/specs.ts
```diff
@@ -1,18 +1,54 @@
...
+const exceptions: exception[] = [
+    [
+        "https://wicg.github.io/navigation-api/",
+        "This spec is moving to WHATWG HTML. Remove this exception when https://github.com/whatwg/html/pull/8502 merges."
+    ],
+];
+
+function isOK(url: URL, allowlist: exception[] = exceptions) {
+    for (const specUrl of specUrls) {
+        if (specUrl.protocol !== url.protocol || specUrl.origin !== url.origin) {
```
foolip, line 19: The origin includes the scheme, so this can be simplified:

```suggestion
        if (specUrl.origin !== url.origin) {
```

scripts/specs.ts
```diff
@@ -1,18 +1,54 @@
+import assert from "node:assert/strict";
+
 import webSpecs from 'web-specs' assert { type: 'json' };
 
 import features from '../index.js';
 
-const specUrls = new Set(webSpecs.map(spec => spec.nightly.url));
+const specUrls = webSpecs.map(spec => new URL(spec.nightly.url));
+
+type exception = [url: string, message: string];
```
foolip, line 9: I suggest avoiding the word exception as it seems like something that would be thrown.

```suggestion
type allowlistItem = [url: string, message: string];
```

scripts/specs.ts
```diff
@@ -1,18 +1,54 @@
...
+const exceptions: exception[] = [
```
foolip, line 10:
```suggestion
const defaultAllowlist: allowlistItem[] = [
```

scripts/specs.ts
```diff
@@ -1,18 +1,54 @@
...
+    [
+        "https://wicg.github.io/navigation-api/",
+        "This spec is moving to WHATWG HTML. Remove this exception when https://github.com/whatwg/html/pull/8502 merges."
+    ],
+];
+
+function isOK(url: URL, allowlist: exception[] = exceptions) {
```
foolip, line 17:
```suggestion
function isOK(url: URL, allowlist: allowlistItem[] = defaultAllowlist) {
```

scripts/specs.ts
```diff
@@ -1,18 +1,54 @@
...
+    for (const specUrl of specUrls) {
+        if (specUrl.protocol !== url.protocol || specUrl.origin !== url.origin) {
+            continue;
+        }
+        if (url.pathname.startsWith(specUrl.pathname)) {
```
foolip, line 22: This will also allow some plausible incorrect spec URLs to persist, like https://w3c.github.io/gamepad/extensions.html if it is folded into https://w3c.github.io/gamepad/. And some less plausible broken URLs like https://w3c.github.io/gamepad/extensions.htmltypo.

The problem is HTML multipage of course. I think we should have the rule that if `specUrl.pathname` ends with "/multipage/" then anything goes, and unfortunately we won't catch broken URLs there. For other URLs, the pathname must match exactly, but `url.hash` can be whatever.

foolip (changes requested):

scripts/specs.ts
ddbeck: I think eb63176 does what you're describing. Want to take another look?

> if `specUrl.pathname` ends with "/multipage/" then anything goes, and unfortunately we won't catch broken URLs there

Correct. Previously, I considered confirming that linked URLs (that don't exactly match `specUrl`) and hashes actually exist, but I know from experience that sort of thing is too slow to run all the time. If we're worried about this, we could put in a more thorough check and have it run only sometimes (e.g., have it run on some schedule and open an issue when it fails, or run only on dependabot upgrades to `web-specs`).

foolip: I've filed https://github.com/w3c/browser-specs/issues/931 about making it easier for us to check URLs.

scripts/specs.ts
```diff
@@ -1,18 +1,59 @@
...
+function isOK(url: URL, allowlist: allowlistItem[] = defaultAllowlist) {
+    for (const specUrl of specUrls) {
+        if (specUrl.origin !== url.origin) {
+            continue;
+        }
+        if (specUrl.pathname === url.pathname) {
+            // that is, specUrl and url are the same, with the exception of the hash or query (`search`) string
+            return true;
+        }
+        if (specUrl.pathname.includes("/multipage") && url.pathname.startsWith(specUrl.pathname)) {
```
foolip, line 26:
```suggestion
        if (specUrl.pathname.includes("/multipage/") && url.pathname.startsWith(specUrl.pathname)) {
```

foolip (approved):

PR #128, opened 2023-04-11
Add `array-flat` and `class-syntax` groups

ddbeck:
This PR adds two more groups to cover JavaScript-related groups already in caniuse. Some highlights here:

- Specific spec URLs continue to be a problem; see also https://github.com/web-platform-dx/feature-set/pull/122.
- Like what's already come up with web components, the `class-syntax` group is another vanguard for a big group-of-groups: es6.
- One _could_ have more specific spec URLs, though it would start to duplicate what BCD already provides. Possibly an opportunity to hoist data from BCD in conjunction WIP on https://github.com/web-platform-dx/feature-set/pull/110.


feature-group-definitions/array-flat-and-flatmap.yml
```diff
@@ -0,0 +1,6 @@
+caniuse: array-flat
```
foolip, line 1: Any reason to not match caniuse here and call this feature `array-flat`? This also has implications for https://github.com/web-platform-dx/feature-set/pull/120 if the answer is yes.

feature-group-definitions/class-syntax.yml
```diff
@@ -0,0 +1,9 @@
+caniuse: es6-class
```
foolip, line 1: There are many ways to name this feature, but yeah I think `class-syntax` works well.

foolip (approved):

feature-group-definitions/array-flat-and-flatmap.yml
ddbeck: `array-flat` is now consistent with `array-group`. See https://github.com/web-platform-dx/feature-set/pull/120#discussion_r1168897684.

PR #134, opened 2023-04-17
Add baseline status to `offscreen-canvas`

ddbeck:
This PR adds handcrafted baseline status information to the `offscreen-canvas` feature. Some notes:

- This feature is not yet baseline, but will be in the future. The one point holding this back is the lack of a second Safari release. We don't really have data that _says_ this explicitly, however.
- I've added Edge as an allowed browser for representing the status. This lines up with the visual work on this and the user research about what browsers web developers care most about.
- I split the context loss API to its own feature group. Because I predicted otherwise in https://github.com/ddbeck/common-web-feature-mockup/issues/7, Safari did _not_ ship the `contextlost` and `contextrestored` events in 16.4.
- I removed the `filter` entry entirely. Like context loss, this is a new addition to 2D rendering context objects (not just offscreen canvas). It's not widely supported and may end up being grouped with other 2D rendering API changes.

cc: @atopal


foolip (changes requested):

feature-group-definitions/canvas-context-lost.yml
```diff
@@ -0,0 +1,16 @@
+spec: https://html.spec.whatwg.org/multipage/
+# spec: https://html.spec.whatwg.org/multipage/webappapis.html#context-lost-steps
```
foolip, line 2:
```suggestion
spec: https://html.spec.whatwg.org/multipage/webappapis.html#context-lost-steps
```

foolip (approved):

PR #135, opened 2023-04-17
Add baseline status to `subgrid`

ddbeck:
This one is pretty straightforward: not baseline yet.


feature-group-definitions/subgrid.yml
```diff
@@ -1,5 +1,9 @@
 spec: https://drafts.csswg.org/css-grid-2/#subgrids
 caniuse: css-subgrid
+status:
+  is_baseline: false
+  support:
+    firefox: "71"
```
foolip, line 6: It's also in Safari.

ddbeck: I have no idea how I missed that. Fixing.

PR #142, opened 2023-04-21
Add Fetch Priority feature

foolip:
The feature is both the support in Fetch and the HTML attributes.

A BCD entry for the Request parameter is missing.

### DIFF
--- a/feature-group-definitions/fetch-priority.yml
+++ b/feature-group-definitions/fetch-priority.yml
@@ -1,0 +1,12 @@
+spec:
+  - https://fetch.spec.whatwg.org/#ref-for-dom-requestinit-priority
+  - https://html.spec.whatwg.org/multipage/semantics.html#attr-link-fetchpriority
+  - https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-fetchpriority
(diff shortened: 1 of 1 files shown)


feature-group-definitions/fetch-priority.yml
```diff
@@ -0,0 +1,12 @@
+spec:
```
foolip, line 1: OK, looks like we don't allow features with multiple spec links yet. @ddbeck is it time to make that change then? Do you think we need to designate one the URLs the main one?

foolip: I checked the specs again and think we should actually use these two URLs:
https://fetch.spec.whatwg.org/#request-priority
https://html.spec.whatwg.org/multipage/urls-and-fetching.html#fetch-priority-attributes

The first is the concept of fetch priority, and the second is about the HTML attributes that set priority. The BCD entries then point at the API surface directly.

foolip: I'll add the entries in https://github.com/mdn/browser-compat-data/pull/19634 now.

ddbeck (approved):

PR #147, opened 2023-04-24
Allow multiple specification URLs

foolip:


ddbeck (changes requested):
Oh good my change in https://github.com/web-platform-dx/feature-set/pull/44 has really confused matters.

You'll need something along the lines of this in `index.ts`

```typescript
/** Web platform feature */
export interface FeatureData {
    /** Specification URLs */
    spec: spec_url | [spec_url, spec_url, ...spec_url[]];
```

and a corresponding `spec_url` type to make it such that `npm run generate-schema-defs` creates the desired output.

This suggests a follow-up task for me: write some test that fails if regenerating the schema creates a diff.

ddbeck (approved):
@foolip I pushed some changes. If you're happy with this, then I am.

foolip (commented):
Thanks @ddbeck, looks good!

PR #155, opened 2023-04-30
Add Baseline status to `grid` (and add `grid-animation`)

ddbeck:
This adds Baseline status for `grid`. I'll self-review with some implementation notes.

### DIFF
--- a/feature-group-definitions/grid-animation.yml
+++ b/feature-group-definitions/grid-animation.yml
@@ -1,0 +1,11 @@
+spec: https://drafts.csswg.org/css-grid-2/
+# status:
+#   is_baseline: false # waiting on next Safari major version
+#   support:
+#     chrome: "107"
+#     edge: "107"
+#     firefox: "76"
+#     safari: "16.1"
+compat_features:
+  - css.properties.grid-template-columns.animation
+  - css.properties.grid-template-rows.animation
--- a/feature-group-definitions/grid-animation.yml
+++ b/feature-group-definitions/grid-animation.yml
@@ -1,4 +1,4 @@
-spec: https://drafts.csswg.org/css-grid-2/
+spec: https://drafts.csswg.org/css-grid-2/#track-sizing
 # status:
 #   is_baseline: false # waiting on next Safari major version
 #   support:
--- a/feature-group-definitions/grid.yml
+++ b/feature-group-definitions/grid.yml
@@ -1,5 +1,13 @@
 spec: https://drafts.csswg.org/css-grid-3/
 caniuse: css-grid
+status:
+  is_baseline: true
+  since: "2020-06-02" # since Firefox 77
+  support:
+    chrome: "66"
+    edge: "79"
+    firefox: "76"
+    safari: "12.1"
 compat_features:
   - css.properties.align-content.grid_context
   - css.properties.align-items.grid_context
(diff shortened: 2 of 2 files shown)


feature-group-definitions/grid-animation.yml
```diff
@@ -0,0 +1,11 @@
+spec: https://drafts.csswg.org/css-grid-2/
```
ddbeck, line 1: Chrome (107) and Safari (16) only recently implemented animation for grid. If they were included in `grid`, it would  prevent baseline status for grid, which seemed wrong. I separated the animation features into their own group, shown in this file.

feature-group-definitions/grid-animation.yml
```diff
@@ -0,0 +1,11 @@
...
+# status:
+#   is_baseline: false # waiting on next Safari major version
+#   support:
+#     chrome: "107"
+#     edge: "107"
+#     firefox: "76"
+#     safari: "16.1"
```
ddbeck, line 8: I commented out the baseline status, since I was skeptical we'd want to allowlist this just yet—it's one of those "will be baseline soon, but not yet" features (pending Safari 17, I suppose). I did the work though, so I thought I'd show what _would_ be.

feature-group-definitions/grid.yml
```diff
@@ -1,13 +1,24 @@
 spec: https://drafts.csswg.org/css-grid-3/
 caniuse: css-grid
+status:
+  is_baseline: true
+  since: "2020-06-02" # since Firefox 77
```
ddbeck, line 5: This might look a little funny.

Firefox mostly implements `grid` from version 52. Alternative names and a few exceptional features delays full, unambiguous support to 76. The main problem is `repeat` values for `grid-template-{rows,columns}`: there are multiple notes and partial implementations.

Were it not for `repeat` then Edge 80 would mark the baseline cutoff, four months earlier.

feature-group-definitions/grid.yml
```diff
@@ -22,15 +33,19 @@ compat_features:
...
-  - css.properties.place-items
-  - css.properties.place-self
-  - css.properties.row-gap
+  - css.properties.justify-content.grid_context
+  - css.properties.justify-items.grid_context
+  - css.properties.justify-self.grid_context
+  - css.properties.place-content.grid_context
+  - css.properties.place-items.grid_context
+  - css.properties.place-self.grid_context
+  - css.properties.row-gap.grid_context
```
ddbeck, line 49: The diff here fills in descendant features (e.g., adding `fit-content`) or targets the `grid_context` entries.

The features omitted from this list are:

- `subgrid`
- `masonry`
- grid animation features (see above)

feature-group-definitions/grid-animation.yml
foolip: Is https://drafts.csswg.org/css-grid-2/ here instead of https://drafts.csswg.org/css-grid-3/ intentional? Since there's not specific section for it, using the same URL as the `grid` feature makes the most sense to me.

foolip (approved):

feature-group-definitions/grid-animation.yml
ddbeck: It really needs to be _this_:

```suggestion
spec: https://drafts.csswg.org/css-grid-2/#track-sizing
```

The level 3 spec doesn't address the animation of the properties (it doesn't cover the properties at all).

PR #157, opened 2023-05-01
Update baseline statuses

ddbeck:
I've gotten a bit better at (manually) exercising the baseline status, so I went through everything we already had to make sure it was correct. Some changes fell out of that. I'll self-review with detailed commentary.

### DIFF
--- a/feature-group-definitions/canvas-context-lost.yml
+++ b/feature-group-definitions/canvas-context-lost.yml
@@ -5,7 +5,6 @@ status:
   support:
     chrome: "99"
     edge: "99"
-    firefox: "105"
 compat_features:
   - api.CanvasRenderingContext2D.isContextLost
   - api.HTMLCanvasElement.contextlost_event
--- a/feature-group-definitions/cascade-layers.yml
+++ b/feature-group-definitions/cascade-layers.yml
@@ -3,9 +3,10 @@ caniuse: css-cascade-layers
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4007
 status:
   is_baseline: true
-  since: 2022-03-15
+  since: "2022-09-19"
   support:
     chrome: "99"
+    edge: "99"
     firefox: "97"
     safari: "15.4"
 compat_features:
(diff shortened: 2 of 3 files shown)


feature-group-definitions/canvas-context-lost.yml
```diff
@@ -5,7 +5,6 @@ status:
   support:
     chrome: "99"
     edge: "99"
-    firefox: "105"
```
ddbeck, line 8: Firefox supports only the events on `OffscreenCanvas` and none of the rest of the API.

feature-group-definitions/container-queries.yml
```diff
@@ -4,5 +4,17 @@ usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4165
...
+  - api.CSSContainerRule
+  - api.CSSContainerRule.containerName
+  - api.CSSContainerRule.containerQuery
+  - css.at-rules.container
+  - css.properties.contain
+  - css.properties.contain.inline-size
+  - css.properties.contain.style
+  - css.properties.container
+  - css.properties.container-name
+  - css.properties.container-type
```
ddbeck, line 20: I don't know if this feature list is good, to be honest. Safari is missing the CSSOM bit. If we relaxed this requirement, then we'd be looking at Safari 17 as the baseline-effective version.

feature-group-definitions/cascade-layers.yml
```diff
@@ -3,9 +3,10 @@ caniuse: css-cascade-layers
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4007
 status:
   is_baseline: true
-  since: 2022-03-15
+  since: "2022-09-19"
```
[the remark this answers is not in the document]
foolip: Ah, that's a curious property of baseline dates I hadn't thought about. It's right, but I think takes a bit of explanation if these dates ever are shown to developers.

feature-group-definitions/container-queries.yml
foolip: I think it's fine to not consider CSSOM part of CQ, it matters very little in practice.

ddbeck: OK, in that case this is a coming-soon feature, which we don't want to show just yet. I'll drop the CSSOM bits from the feature list but update-and-comment-out the baseline data.

PR #159, opened 2023-05-02
Document baseline indication with data from `web-features`

ddbeck:
@foolip You suggested this last week, to make explicit what we intend with the current schema.

(It's likely that we'll want to switch browsers to booleans or something, but it seems ill-advised to change the schema _now_, before we've received feedback from our first consumers.)

### DIFF
--- a/packages/web-features/README.md
+++ b/packages/web-features/README.md
@@ -24,7 +24,7 @@ For baseline iconography, follow this procedure for each feature:
 
 For browser support iconography (that is, browser logos and checkmarks and Xs), follow this procedure, for each feature and browser (Chrome, Edge, Firefox, and Safari):
 
-0. **Do not** show a version number, whether one is provided.
+1. **Do not** show a version number, whether one is provided.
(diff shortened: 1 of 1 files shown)


packages/web-features/README.md
```diff
@@ -11,3 +11,20 @@ npm install web-features
...
+
+For baseline iconography, follow this procedure for each feature:
+
+1. If `status.is_baseline` is `true`, then show the baseline (🟢 affirmative, green check) badge.
+1. If `status.is_baseline` is `false`, then show the non-baseline (🟡 caution, yellow X) badge.
+1. If `status.is_baseline` is `undefined`, then **do not** show any baseline or non-baseline badge.
+
+For browser support iconography (that is, browser logos and checkmarks and Xs), follow this procedure, for each feature and browser (Chrome, Edge, Firefox, and Safari):
+
+0. **Do not** show a version number, whether one is provided.
```
foolip, line 27: Why start at zero? Just 1-1-1-1 like above?

foolip (approved):

PR #160, opened 2023-05-03
`border-image`: Add compat_features and baseline status

ddbeck:
This adds compat features and a baseline status to `border-image`.

This one is interesting because the data says this enjoys longstanding support, but its inclusion into Interop 2023 makes me doubt it. The only thing in BCD itself which hints at this is lengthy notes on Firefox's support for `border-image` and `border-image-slice`. But [the Interop dashboard](https://wpt.fyi/interop-2023?feature=interop-2023-cssborderimage) graph makes it look like Chrome and Safari lagged significantly, but no notes indicate why that might be. 🤔


ddbeck:
@foolip Thank you! I opened a PR against BCD for the bug fix. Merging now.

PR #166, opened 2023-05-09
`offscreen-canvas`: suppress baseline status (for now)

ddbeck:
We don't have baseline display assets or guidelines for "coming soon" features, so this hides the status until we've sorted that out. This is similar to the way we've handled `container-queries` and `grid-animation`.


atopal (approved):
LGTM!

PR #183, opened 2023-05-15
Add WebHID feature

foolip:


feature-group-definitions/webhid.yml
```diff
@@ -0,0 +1,36 @@
+spec: https://wicg.github.io/webhid/
```
ddbeck, line 1: It might be nice to have a TODO comment or an issue noting that this is a… conflicted specification. That is, Apple and Mozilla are explicitly _not_ prototyping/implementing the spec. That's an interesting bucket that I imagine we'll get to eventually and it'd be nice to have a way to come back to it when we're ready to add it to the schema.

feature-group-definitions/webhid.yml
```diff
@@ -0,0 +1,36 @@
...
+  - api.HID.disconnect_event
+  - api.HID.getDevices
+  - api.HID.requestDevice
+  - api.HIDConnectionEvent
+  - api.HIDConnectionEvent.HIDConnectionEvent
+  - api.HIDConnectionEvent.device
+  - api.HIDDevice
+  - api.HIDDevice.close
+  - api.HIDDevice.collections
+  # - api.HIDDevice.forget
```
ddbeck, line 21: I'm guessing you omitted this one because it landed several versions later. What do you think of including it anyway, and advancing the Chrome and Edge version numbers? This would make the group more complete to the contemporary API without strongly altering the compatibility story.

ddbeck (approved):
No blockers here, but I do have a couple little suggestions.

feature-group-definitions/webhid.yml
foolip: I think the best way to capture that would be a `standards_positions` field that can link to issues in https://github.com/mozilla/standards-positions or https://github.com/WebKit/standards-positions.

feature-group-definitions/webhid.yml
foolip: It's a good question. Should we be more liberal about updating groups for features that are "new"? I'm inclined to say no actually, and think we should handle this situation the same as we would given a new feature that ships in all browsers around the same time, and then gets some small later. I think the feature as a whole should still use the older launch date.

What do about api.HIDDevice.forget then? I don't know. We have this question sprinkled all over our YAML files now.

ddbeck: > we should handle this situation the same as we would given a new feature that ships in all browsers around the same time

OK, that's a fair point. I think the overriding thing ought to be developer expectations: if we get information that `forget` is part of the popular understanding of WebHID—even if it arrived later—then it should be part of the main group. But absent that information, it does seem wise to treat it like anything else.

And I'd be probably be less reluctant to split things, if we had already worked out group composition (e.g., "WebHID (kitchen sink)" = "WebHID" + "forget"). So let's keep this as is, then sort out composition so we don't have to have these conversations as often. 😄

feature-group-definitions/webhid.yml
ddbeck: Tracked in https://github.com/web-platform-dx/feature-set/issues/186

feature-group-definitions/webhid.yml
foolip: I agree that developer expectations trumps all.

PR #200, opened 2023-05-23
Add WebVTT BCD features

foolip:


foolip:
@dontcallmedom I think that in practice, support for `<track>` is the same thing as basic support for WebVTT, as it's the format that all browsers support, and always have.

dontcallmedom:
should `html.elements.track` be listed in the compat features then?

foolip:
I actually don't think so, no, if we need to create a feature I'd put `<track>` and `HTMLTrackElement` as its own "text tracks" feature.

ddbeck:
I have approximately zero domain knowledge here, so forgive me if either of you have thought of this already: do web developers think of WebVTT as distinct from `<track>` and `TextTrack`? Or is it possible we've got some more complex relationship like…

```mermaid
flowchart LR
    A[Tracks] --> B[Text tracks]
    A --> C[WebVTT]
```

Before lumping or splitting, it'd be kinda nice to confirm that somebody's thought about this from a developer perspective.

foolip:
I think the features are recognizable as separate, but `TextTrackCue` and related generic text track stuff are abstract and never used on their own.

It's a bit like `Element` and `HTMLElement`, where the only kind of element most web developers deal with is an HTML element, so the line between them won't be clear.

So I think we should have a "text tracks" feature, but probably hidden somehow. We'll probably want to do the same for the ancient parts of ``, it always needs an image format to do anything, but it's still not identical to GIF/JPEG/whatever support.

ddbeck (approved):

PR #202, opened 2023-05-23
Rename leading-trim to text-box-trim

foolip:
The spec link hasn't changed, but the property was renamed: https://github.com/w3c/csswg-drafts/commit/67822bd3c1e02cd187abbc3856a5c28fa14c8431


foolip:
@ddbeck this removes a feature, do you think we should create the mechanism for feature aliases as part of this? Or just bump the major version?

ddbeck (approved):

ddbeck:
@foolip Since we're pre-1.0, we can make changes and increment the minor version freely (i.e., the next release would be 0.5.0). I think we ought to figure out aliases (and offer some API that follows them automatically) at some point, but it doesn't _need_ to be for this PR.

That said, if we wanted to stake out a way of annotating feature aliases now, it might be nice. I think you've got three options for this PR:

1. Stake out a place for aliases/formerly known as-es/nées or however you want to call it, in this PR or another PR that blocks this one.
2. Add a comment to https://github.com/web-platform-dx/feature-set/issues/91, describing this rename and perhaps what you might've noted this alias (e.g., in this file, in another file, etc.)
3. Do nothing.

I'm fine with any one of these and if there's no action (or you go on leave before you've had a chance to take action), then I'll merge on the presumption of option 3.

foolip:
@ddbeck I went ahead with an `alias` property. I'm on my phone (not working) and it's not convenient to update the schema. It should be a very straightforward change however, if you want to make it and bump the version I can approve it, before landing this.

ddbeck:
Works for me. I just pushed a commit for the schema change (like you said, straightforward) and I'll deal with the version bump on the next release. Thank you, @foolip!

foolip (commented):
Thanks, looks good!

PR #215, opened 2023-05-31
Mark masonry a child feature of grid

foolip:


foolip:
Closing this since Masonry may ultimately end up not a subfeature of Grid at all, see https://github.com/w3c/csswg-drafts/issues/9041.

PR #274, opened 2023-08-01
Add feature: `media-query-range-syntax`

ddbeck:

- Are any pieces missing?


ddbeck:
@MrHBS I need to do some work to improve the documentation around this, but this repo is bigger than support or Baseline alone. We're also cataloguing web platform features in general, so we can follow their progress (e.g., before they're Baseline) and link them together (e.g., to show where there's a new addition to an existing feature, or what specs or web platfomr tests they relate to).

foolip (approved):
I've filled in more details now, merging.

PR #275, opened 2023-08-01
Link `grid-animation` to caniuse

ddbeck:


ddbeck:
Interesting! It's in the `caniuse-lite` package, but the package brings everything over from https://github.com/Fyrd/caniuse including features marked as `"shown": false`. I'll open an issue on `caniuse-lite` to see if they'll omit those entries or expose the `shown` property.

Alternatively, though less conveniently, we could pull in caniuse's data directly.

foolip:
@ddbeck with https://github.com/browserslist/caniuse-lite/issues/119 resolved, can this be revisited?

PR #282, opened 2023-08-07
`import-maps`: add compat feature

ddbeck:


ddbeck:
Thanks, François! That makes sense—I expect any such relaxations will result in subfeatures in BCD, so this seems OK for now.

PR #285, opened 2023-08-08
`line-clamp`: add compat features

ddbeck:


foolip (approved):
I think we should do exactly what you've done here. A single feature for prefixed and unprefixed properties. Even if the unprefixed form ships with significant differences to prefixed, developers will think about it as one feature.

Right now, we'd call this supported with an important note about requiring a prefix.

PR #286, opened 2023-08-08
`custom-properties`: add compat features

ddbeck:

- Are any pieces missing?


foolip (approved):
I think this makes sense as a group. We'll also need a group for registered custom properties, which is the `@property` rule and `CSS.registerProperty()` static method.

PR #291, opened 2023-08-11
Add `modulepreload` feature for `<link rel="modulepreload">`

ddbeck:

- Are any pieces missing?


foolip (approved):

foolip:
@ddbeck can you review the `name` I added before merging this? It's definitely the most code-y name so far.

ddbeck:
Yeah, I think this is fine, for now. We'll probably want a guideline for attributes, once we hear from consumers about this kind of thing.

PR #292, opened 2023-08-11
Add View Transitions feature

ddbeck:

- Are any pieces missing?


atopal (approved):
LGTM!

PR #293, opened 2023-08-11
Add asynchronous clipboard APIs feature

ddbeck:

- Are any pieces missing?


foolip:
> Is this a reasonable identifier for the feature?

There's a creative typo in it, so I'll take the liberty of fixing that :)

foolip (approved):
I think this grouping make sense. Note that `navigator.clipboard` with members other than `read()` shipped earlier than `read()`, but I think it's still a reasonable grouping, and far enough in the past that it doesn't make sense to split apart now.

Issue #91, opened 2023-03-10
Consider handling feature identifier changes or aliases


ddbeck:
Inspired by discussion on https://github.com/web-platform-dx/feature-set/pull/89.

If the feature set is based on real-world usage by web developers, then some features may need to change identifiers from time to time or accommodate multiple known names. Some possible scenarios:

- A feature is folded into a closely related feature. For instance, `subgrid` is subsumed by `grid`. At least one feature is removed and the definition of another feature changes, triggering some sort of versioning event (either at the data or package level). 
- A feature has some name but official or popular usage goes in another direction or multiple directions (e.g., `es6` and `es2015`).
- A feature is named erroneously (e.g. `felxbox`) and we ship with the erroneous name.

Some questions emerge:

- How do we communicate such changes to consumers? (e.g., as release notes, as structured data, or something else)
- How much should we assist consumers in following changes? (e.g., do we maintain a formerly-known-as list for each feature? do we duplicate data access under multiple names?)
- How long should we assist consumers in following changes? (e.g., do we follow a deprecation schedule for outgoing features? do we maintain historic names in perpetuity?)

The answers to some of these questions might influence how we handle versioning (if we understand versions to be a kind of identifier for a feature).

We ought to resolve this issue before a 1.0 release.

foolip:
My concrete proposal here is a new `alias` property which is a string. In case we make multiple mistakes for the same feature we'll need an array of strings, but it's not certain that we ever will.

The API can wait until later, and might also including looking features up by BCD and caniuse identifiers.

ddbeck:
#202 stakes out the optional `alias` property for at least the simple case of renaming something.

foolip:
There's some additional discussion in https://github.com/web-platform-dx/web-features/issues/729. @romainmenke points out that `is` might refer to multiple things, and I sketched out how I think we could handle alias reuse (removal followed by later addition) if we really have to in the future.

romainmenke:
The `alias` approach does offer a way for a person to intervene and resolve issues by hand.
But it is a cumbersome process. You would have to scan the entire feature set and try to find the entry that has the old identifier in an `alias` field.

Maybe all features could have a stable machine readable id?
Something like a uuid. This would never change, even if a feature is renamed.

Having such an id would make it easier for 3rd parties (like myself) to reference web-features without having to worry too much about churn.

foolip:
[...]

PR #295, opened 2023-08-14
Add `overflow` shorthand feature (with `clip` values)

ddbeck:
This corresponds to https://caniuse.com/css-overflow

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/overflow-shorthand.yml
+++ b/feature-group-definitions/overflow-shorthand.yml
@@ -1,0 +1,15 @@
+spec: https://drafts.csswg.org/css-overflow-3/#propdef-overflow
+caniuse: css-overflow
+compat_features:
+  - css.properties.overflow-x
+  - css.properties.overflow-x.clip
+  - css.properties.overflow-y
+  - css.properties.overflow-y.clip
+  - css.properties.overflow
+  - css.properties.overflow.clip
+  - css.properties.overflow.multiple_keywords
+  # BCD appears to represent `<overflow>` values as both
+  # 1) keyword subfeatures on properties (above) and
+  # 2) a type (below)
+  - css.types.overflow
+  - css.types.overflow.clip


tidoust:
[CSS Overflow Module Level 4](https://drafts.csswg.org/css-overflow-4/#overflow-control) extends `overflow` to also apply to "replaced elements". That's still being worked upon, I think, and seems at risk.

Assuming this goes through, the definitions of the CSS properties would remain unchanged, the scope of their application would change. I'm wondering how we could represent that evolution once css-overflow-4 becomes the base level. Would we just update the URL to `https://drafts.csswg.org/css-overflow-4/#propdef-overflow`?

ddbeck:
@tidoust I think the evolution would look something like this:

1. When the first implementer ships `overflow` for replaced elements, then I'd expect a new feature to appear in BCD for the relevant properties (e.g., `css.properties.overflow.applies_to_replaced_elements`).
2. At that point, we'd create a feature here (something like `overflow-for-replaced-elements`). We'd indicate it has some relationship to the `overflow-shorthand` (see https://github.com/web-platform-dx/web-features/issues/217 for more on that question).
3. The march of time happens and browsers eventually implement the feature.
4. A day comes when the `overflow-for-replaced-elements` feature is mature enough (and developers have ceased to distinguish it) that we merge the feature into the `overflow-shorthand` feature. (I imagine we'll have a guideline to cover this, something like a year after a feature has achieved some very high level of support.) See also: https://github.com/web-platform-dx/web-features/issues/91.

Does that feel like a practical approach? Or do you think we need to do some future-proofing today?

tidoust:
> Does that feel like a practical approach? Or do you think we need to do some future-proofing today?

That works! I wasn't clear that there could be a more "abstract" (in the sense of not being a 1-to-1 mapping to a property value) entry in BCD such as `css.properties.overflow.applies_to_replaced_elements`. That seems a good approach to me.

PR #299, opened 2023-08-14
Add HTML media capture feature

ddbeck:
This corresponds to https://caniuse.com/html-media-capture.

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/media-capture.yml
+++ b/feature-group-definitions/media-capture.yml
@@ -1,0 +1,5 @@
+spec: https://w3c.github.io/html-media-capture/
+caniuse: html-media-capture
+compat_features:
+  - api.HTMLInputElement.capture
+  - html.elements.input.capture


tidoust (commented):
Content looks good to me.

There are different media capture mechanism on the web: HTML Media Capture, [Media Capture and Streams](https://www.w3.org/TR/mediacapture-streams/), [Media Capture from DOM Elements](https://www.w3.org/TR/mediacapture-fromelement/), and possibly [MediaStream Recording](https://w3c.github.io/mediacapture-record/).

The `html-` prefix may be warranted in this case to clarify that the feature here is HTML Media Capture?

ddbeck:
> The `html-` prefix may be warranted in this case to clarify that the feature here is HTML Media Capture?

Yes. This seems somewhat distinct from the capture APIs. It's a attribute that tells (AFAICT only mobile) browsers to treat the camera as if it were a file to upload. It's rather narrowly designed, as compared to the other APIs.

PR #301, opened 2023-08-15
Add `:default` pseudo class feature

ddbeck:
Corresponding to https://caniuse.com/css-default-pseudo

- Are any pieces missing?


foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/CSS/:default#browser_compatibility doesn't match https://caniuse.com/css-default-pseudo especially well, but I'm confident they represent the same API surface and the differences are far back enough that it doesn't matter.

We may have to sort out these differences at some point for caniuse to fully migrate to web-features, but definitely not today.

PR #302, opened 2023-08-15
Add `:indeterminate` pseudo class feature

ddbeck:
Corresponds to https://caniuse.com/css-indeterminate-pseudo

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/indeterminate.yml
+++ b/feature-group-definitions/indeterminate.yml
@@ -3,7 +3,6 @@ spec:
   - https://html.spec.whatwg.org/multipage/semantics-other.html#selector-indeterminate
 caniuse: css-indeterminate-pseudo
 compat_features:
-  - api.HTMLInputElement.indeterminate
   - css.selectors.indeterminate
   - css.selectors.indeterminate.checkbox
   - css.selectors.indeterminate.progress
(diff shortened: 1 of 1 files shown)


feature-group-definitions/indeterminate.yml
```diff
@@ -0,0 +1,10 @@
+spec:
+  - https://drafts.csswg.org/selectors-4/#indeterminate
+  - https://html.spec.whatwg.org/multipage/semantics-other.html#selector-indeterminate
+caniuse: css-indeterminate-pseudo
+compat_features:
+  - api.HTMLInputElement.indeterminate
```
foolip, line 6: It's a hard call whether the `indeterminate` IDL property should be listed here. It doesn't *only* exist to provide state for the pseudo-class but has some other effects as well. If we ignore the name being the same, we could argue it has the same relationship to `:indeterminate` as the `progress` element's `value` attribute, which is also part of the definition in https://html.spec.whatwg.org/multipage/semantics-other.html#selector-indeterminate.

I fall on the side of not listing it here, since I don't know how to draw the line following that kind of reasoning. A reason to list it would be to assign it to _some_ feature, but we're not yet working to categorize all of BCD, and I think "form elements" may well be a better category for it.

ddbeck: > we could argue it has the same relationship to `:indeterminate` as the `progress` element's `value` attribute

This is exactly right. I think I got tripped up because they share a name. I'll drop this, as suggested.

```suggestion
```

PR #303, opened 2023-08-15
Add `scrollbar-{color,width}` properties feature

ddbeck:
Adds `scrollbar-color` and `scrollbar-width`, corresponding to https://caniuse.com/css-scrollbar

### DIFF
--- a/feature-group-definitions/scrollbar-styles.yml
+++ b/feature-group-definitions/scrollbar-styles.yml
@@ -1,0 +1,5 @@
+spec: https://drafts.csswg.org/css-scrollbars-1/
+caniuse: css-scrollbar
+compat_features:
+  - css.properties.scrollbar-color
+  - css.properties.scrollbar-width


foolip (commented):
This is a bit like `let` and `const` as a single feature to match caniuse. Maybe it makes sense, but I'm more inclined to treat them as separate features.

ddbeck:
I'm not sure this is just a case of following caniuse. I think there are other examples of treating these as closely linked. Some examples:

- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling
- https://css-tricks.com/the-current-state-of-styling-scrollbars-in-css/

And they're the only things of consequence in the spec.

That said, I'm OK with considering them separately, but then I'm not sure what to do about caniuse. Just not acknowledge it? Or open an issue for somehow communicating to caniuse that a _caniuse_ feature can span multiple web-feature features?

foolip:
OK, let's keep this as a single feature then. I don't think the name `scrollbar-styles` works though. At least to browser folks, I think "styles" means the stuff inside stylesheets, so these are "styles":
```css
.scrollable {
  -ms-overflow-style: auto;
  scrollbar-color: rebeccapurple green;
  scrollbar-width: thin;
}
```

Support for `scrollbar-color` and `scrollbar-width` isn't "styles" in this sense. I think`scrollbar-styling` is a better name, but https://github.com/web-platform-dx/web-features/blob/main/docs/readme.md talk about noun phrases. Is it fine?

foolip:
https://wpt.fyi/interop-2024 has a Scrollbar Styling focus area that includes `scrollbar-width` and `scrollbar-gutter`, but not `scrollbar-color`. This makes it fairly likely that some browser will ship support for `scrollbar-width` but not `scrollbar-color`.

PR #308, opened 2023-08-18
Add `:is()` pseudo-class feature

ddbeck:
Corresponds to https://caniuse.com/css-matches-pseudo

- Are any pieces missing?


foolip (approved):
My only doubt is whether to group this with `:where()` since it was enabled at the same time, but I think we should generally not do that. Eventually the feature introduction is so far back that it just seems random to group them.

foolip:
@ddbeck can you merge this together with https://github.com/web-platform-dx/web-features/pull/386?

PR #310, opened 2023-08-18
Add `unicode-bidi` feature

ddbeck:
Corresponds to https://github.com/Fyrd/caniuse/blob/main/features-json/css-unicode-bidi.json

This is a slightly odd feature. On caniuse, this is redirected to a search, though the data still exists in Fryd/caniuse. Moreover, the spec and MDN docs both advise against web developers actually using this feature. This is likely to be our first case of a discouraged feature.

See also: https://github.com/web-platform-dx/feature-set/issues/311

---

- Are any pieces missing?


ddbeck:
Thanks to https://github.com/web-platform-dx/web-features/pull/411 this feature (rightly) fails the caniuse test. Closing as unneeded.

PR #313, opened 2023-08-18
Add `tabindex` global attribute feature

ddbeck:
Corresponds to https://caniuse.com/tabindex-attr

- Are any pieces missing?


foolip (approved):
This is pretty straightfoward. Old but yes it's a feature.

PR #314, opened 2023-08-18
Add `Element.scrollIntoView()` feature

ddbeck:
Corresponds to https://caniuse.com/scrollintoview

- Are any pieces missing?


foolip (approved):
I double checked the web platform only has one `scrollIntoView()` method, so no need to disambiguate.

PR #320, opened 2023-08-24
Add `Object.entries()` feature

ddbeck:
Corresponds to https://caniuse.com/object-entries

Now that I've done something like this a few times: I wonder if it would be a good idea to a have shorthand for "there is a one-to-one correspondence between this a BCD key and a caniuse key."

---

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/object-entries.yml
+++ b/feature-group-definitions/object-entries.yml
@@ -1,0 +1,4 @@
+spec: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.entries
+caniuse: object-entries
+compat_features:
+  - javascript.builtins.Object.entries


foolip:
From BCD it looks like `Object.entries()` and `Object.values()` shipped together, which makes sense. But if we make this `object-entries-values`, it no longer matches the caniuse features.

Whether we group them or not, I think this entry should be marked as "no longer interesting" in some way, given how long ago it became interoperable.

foolip:
Also, does this have a meaningful parent feature? `Object` isn't a feature worth distinguishing, but it *does* group other features that are worthwhile, including recent ones like `Object.groupBy()`.

ddbeck:
> Whether we group them or not, I think this entry should be marked as "no longer interesting" in some way, given how long ago it became interoperable.

@foolip I've been thinking about this a bit and I'm guessing that we'll have a bunch of boring/historic/single-entry features, if we're really aiming for wide coverage with respect to caniuse. At the risk of saying something that goes against things I've suggested previously: maybe we should allow a `caniuse-` prefix, where we're making a group that exists just to map to caniuse?

ddbeck:
> Does caniuse accept PRs to remove features or classify them as "no longer interesting"? Seems like something to maybe try to fix in caniuse itself?

I talked to Alexis about this earlier in the week. He said that except in rare cases (e.g., a feature is never implemented and abandoned and its continued presence on caniuse confuses people), he doesn't remove features. He doesn't want to break existing links that depend on those features.

He did say he might be open to redirecting to new feature groups in the future, but I think in the near-term we'll have to just live with these.

> Also, what's caniuse's benefit when it provides two entries? Seems like the manually maintained entry could go?
> 
> * https://caniuse.com/object-entries
> * https://caniuse.com/mdn-javascript_builtins_object_entries

I forgot to ask about this specifically. I'm not sure what his opinion is on these. I suppose once have a high-level of caniuse _and_ BCD coverage in web-features, we'd be able to produce a comprehensive map of duplicate entries like this (e.g., we'd have a generalized answer to the question of "when does caniuse and BCD refer to the same thing?").

> And if in the future caniuse also imports webfeature groups, would that mean there would be a third one?
> 
>  * https://caniuse.com/webfeatures-object-entries

No, the idea behind the `caniuse` key in web-features is to say that web-features's `object-entries` feature is the _same thing as_ caniuse's `object-entries`. If caniuse imported the feature, it'd know that it's just adding new information to a feature that already exists on caniuse.

foolip:
I think the conclusion here is that we shouldn't add this to web-features as a standalone feature. @ddbeck can you note this in the caniuse burndown list and close this?

PR #322, opened 2023-08-24
Add `input` event feature

ddbeck:
Corresponds to https://caniuse.com/input-event

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/input-event.yml
+++ b/feature-group-definitions/input-event.yml
@@ -2,3 +2,10 @@ spec: https://w3c.github.io/uievents/#event-type-input
 caniuse: input-event
 compat_features:
   - api.HTMLElement.input_event
+  - api.InputEvent
+  - api.InputEvent.InputEvent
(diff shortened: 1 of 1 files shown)


feature-group-definitions/input-event.yml
```diff
@@ -0,0 +1,4 @@
+spec: https://w3c.github.io/uievents/#event-type-input
+caniuse: input-event
+compat_features:
+  - api.HTMLElement.input_event
```
foolip, line 4: I think we should also include these features from https://developer.mozilla.org/en-US/docs/Web/API/InputEvent#browser_compatibility:

api.InputEvent
api.InputEvent.InputEvent
api.InputEvent.data
api.InputEvent.dataTransfer
api.InputEvent.getTargetRanges
api.InputEvent.inputType
api.InputEvent.isComposing

(But not api.InputEvent.inputType.insertFromPasteAsQuotation.)

foolip (approved):

PR #323, opened 2023-08-24
Add `<input type="file" accept="…">` feature

ddbeck:
Corresponds to https://caniuse.com/input-file-accept

- Are any pieces missing?

### DIFF
--- a/feature-group-definitions/input-file-accept-attribute.yml
+++ b/feature-group-definitions/input-file-accept-attribute.yml
@@ -1,0 +1,5 @@
+spec: https://html.spec.whatwg.org/multipage/input.html#attr-input-accept
+caniuse: input-file-accept
+compat_features:
+  - api.HTMLInputElement.accept
+  - html.elements.input.accept


foolip:
@ddbeck what would the Baseline low date for this be? I suspect it's too old for us to bother distinguishing from general forms support.

ddbeck:
I need to run the script on this one, but I _think_ it might actually be a threshold feature for Chromium Edge, which is in-range for us. Writing down various options if it's not:

1. Use an editorial override to set a reasonable sounding historical date (e.g., we go look at the feature's history and pick a reasonable looking date in our expert judgment).
2. Indicate an out-of-bounds low condition with a string instead of a date (`"out-of-bounds"`).
3. Do the above, but with a really fake date (`"1900-02-29"`, Excel style, perhaps).
4. Don't provide a date (i.e., `baseline: "high"` without a date means "since approximately forever").
5. Have a special value for `baseline`: `"hierloom"`, `"high"`, `"low"` and `false`.

ddbeck:
Oh no, I was wrong. It's Edge 12 after all. So we do need to do something about this, though I don't know if needs to be in this PR necessarily.

| Key                                                                                                                 | Baseline | Low since date | Versions                                                                                                           |
| :------------------------------------------------------------------------------------------------------------------ | :------: | :------------- | ------------------------------------------------------------------------------------------------------------------ |
| **input-file-accept-attribute**                                                                                     |  ✅ High  | 2015-07-28     | Chrome 1<br>Chrome Android 18<br>Edge 12 🔑💎<br>Firefox 1<br>Firefox for Android 4<br>Safari 1<br>Safari on iOS 1 |
| `api.HTMLInputElement.accept`                                                                                       |  ✅ High  | 2015-07-28     | Chrome 1<br>Chrome Android 18<br>Edge 12 🔑💎<br>Firefox 1<br>Firefox for Android 4<br>Safari 1<br>Safari on iOS 1 |
| [`html.elements.input.accept`](https://developer.mozilla.org/docs/Web/HTML/Attributes/accept#browser_compatibility) |  ✅ High  | 2015-07-28     | Chrome 1<br>Chrome Android 18<br>Edge 12 🔑💎<br>Firefox 1<br>Firefox for Android 4<br>Safari 1<br>Safari on iOS 1 |

🔑💎 marks a release that contributes to the Baseline low date.<br>

foolip:
@ddbeck for cases like this where we're on the edge, how about we allow for "draft" or "private" features that we don't publish in the package? We can then revisit when we're looking at forms support. I think either that or just closing this as a not-worthwhile caniuse entry are the best options.

ddbeck:
OK, let's defer this one for now. Thanks!

PR #324, opened 2023-08-24
Add `maxlength` attributes feature

ddbeck:
🖐 On hold. This ought to be merged with a forms feature. 🖐

---

Corresponds to https://caniuse.com/maxlength

- Are any pieces missing?


captainbrosset:
My 2 cents: I don't really consider `maxlength` a feature of the web platform myself. I consider `input` or `textarea` as features, and I see `maxlength` as a way to customize those features. So it's a property of those features.

ddbeck:
> I think we should just treat this as part of a "forms" feature, yet to be defined.

OK. I've put this into draft for now, until I have a place to put these things.

ddbeck:
This is now part of a draft feature, so it'll probably get swept up eventually. Close this one.

PR #331, opened 2023-08-28
Add CSS page breaking feature

ddbeck:
Corresponds to https://caniuse.com/css-page-break

This one is very messy. A quick tour of both the caniuse and MDN pages for these features suggests that the data here is out-of-date and/or disorganized. I'm fine with letting this feature through, but when we generate a status for it, I think we're in for some surprises and updating BCD.

---

- Are any pieces missing?


feature-group-definitions/page-break.yml
```diff
@@ -0,0 +1,31 @@
+spec:
+  - https://drafts.csswg.org/css-multicol-1/#break-before-break-after-break-inside
```
tidoust, line 2: This is really just a redirect to the CSS Fragmentation Module, which defines the actual properties. I suggest to replace with (or at least to add):

```suggestion
  - https://drafts.csswg.org/css-break/#breaking-controls
```

tidoust (commented):
It feels weird to have a feature about *page* break and to include values that are for multi-column breaks.

Also, no [`orphans` and `widows`](https://drafts.csswg.org/css-break/#widows-orphans)?

foolip:
@mstensho do you have thoughts on how to organize fragmentation feature's in a way that aligns with web developers think about them and that roughly matches how browsers have evolved?

mstensho:
The data there is certainly outdated, indeed. Chrome has supported the ```avoid``` value on break-before/break-after for some time already, for instance.

Treating the three CSS properties (```break-after```, ```break-before```, ```break-inside```) together makes sense. Caniuse refers to the CSS 2.1 page-break-* properties, though. Would be better to refer to the break-* properties.

The entry should be fragmentation type agnostic (printing / multicol), and should prefer referring to css-break (rather than css-multicol, css-page or CSS 2.1).

What about orphans and widows? There is an entry for the orphans property: https://caniuse.com/mdn-css_properties_orphans , but no corresponding one for widows. There is a common "CSS widows & orphans" entry, though: https://caniuse.com/mdn-css_properties_orphans . Should all of this be merged into one "breaking feature", or would that be considered "too big"?

ddbeck:
OK, I've made a few changes to this feature's ID, name, and description. That said, I have low confidence about what to name this. I couldn't find a lot of evidence that developers think about `break-` as a general thing, so I've tried my best here. It might be that we choose a "bad" name and wait for developer feedback to revise it.

feature-group-definitions/column-breaks.yml
```diff
@@ -0,0 +1,15 @@
+name: Column breaks
+description: The `break-after`, `break-before`, `break-inside` CSS properties control where columns start or end.
```
foolip, line 2: Since `break-after` appears in the other description too, can you say something about the context this feature refers to? It's structurally similar to `gap` in grid and flex contexts, but I don't think we have good wording to copy from.

I think the context can be described as the use of `columns` or `column-count`.

feature-group-definitions/page-breaks.yml
```diff
@@ -0,0 +1,18 @@
+name: Page breaks
+description: The `break-after`, `break-before`, `break-inside` CSS properties (along with `page-break-` aliases) control where pages start and end. Also known as pagination or page breaking.
```
foolip, line 2: Here, I suspect there's no syntax to point to for context. But if this is only for printing, perhaps mention that?

feature-group-definitions/column-breaks.yml
```diff
@@ -0,0 +1,15 @@
+name: Column breaks
+description: In columnar layouts (created by the `columns` or `column-count` CSS properties), the `break-after`, `break-before`, `break-inside` properties control where columns start or end.
```
ddbeck, line 2: This one does something a little funny, setting the context before describing the thing. It does make its description distinct from the other breaking feature though, which I think makes this in the spirit of the rule if not the letter of it.

feature-group-definitions/page-breaks.yml
```diff
@@ -0,0 +1,18 @@
+name: Page breaks
+description: The `break-after`, `break-before`, `break-inside` CSS properties (along with `page-break-` aliases) control where printed pages start and end. Also known as pagination or page breaking.
```
ddbeck, line 2: Now specifies "printed" in the description.

feature-group-definitions/column-breaks.yml
foolip: Makes sense to me!

feature-group-definitions/page-breaks.dist.yml
```diff
@@ -0,0 +1,33 @@
+# Generated from: page-breaks.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Page breaks
+description: The `break-after`, `break-before`, `break-inside` CSS properties (along with `page-break-` aliases) control where printed pages start and end. Also known as pagination or page breaking.
+spec: https://drafts.csswg.org/css-break-3/#breaking-controls
+caniuse: css-page-break
```
foolip, line 7: https://caniuse.com/css-page-break says partial support across the board, except good old Presto of course. Can you add a comment about the difference and if there's anything we need TODO?

feature-group-definitions/column-breaks.yml
```diff
@@ -0,0 +1,15 @@
...
+spec: https://drafts.csswg.org/css-break-3/#breaking-controls
+caniuse: css-page-break
```
foolip, line 4: This shouldn't be here, right?

feature-group-definitions/page-breaks.dist.yml
ddbeck: Ah, this. Caniuse says everything is partially implemented. Some of this is just out-of-date (e.g., lack of support for `break-` properties)—I can open a PR against caniuse for that.

But because nothing is shown as being fully supported on caniuse except Presto, we can't be sure what Alexis meant by fully-supported. I've filed https://github.com/Fyrd/caniuse/issues/7045.

The good news here is that I believe we still ought to show a positive status. If caniuse says partially supported, it's still _equally_ partially implemented across all browsers. They seem to have the same set of limitations.

foolip:
@mstensho do you think the state of this PR now makes sense, treating page breaks and column breaks as distinct features?

feature-group-definitions/page-breaks.dist.yml
ddbeck: > The good news here is that I believe we still ought to show a positive status. If caniuse says partially supported, it's still _equally_ partially implemented across all browsers. They seem to have the same set of limitations.

I spoke too soon. 😞 

Alexis tested this again and showed that the avoid value is not consistently supported by browsers. I submitted https://github.com/mdn/browser-compat-data/pull/22987 to update the BCD and changed the list of compat keys here to match caniuse.

This makes me think that the multicolumn breaks might also be broken, but I couldn't actually find bugs to that effect, so I'm letting that stuff stand.

foolip: Oh my, things are rarely simple. I've reviewed and merged that PR.

mstensho:
> @mstensho do you think the state of this PR now makes sense, treating page breaks and column breaks as distinct features?

Works for me.

foolip:
Thanks @mstensho! @ddbeck can you apply the directory rename and merge this?

PR #332, opened 2023-08-28
Add `font-variant-alternates` feature

ddbeck:
Corresponds to https://caniuse.com/font-variant-alternates

I think caniuse might be slightly wrong about the support story here. It doesn't call out some [differences in Safari noted by BCD](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-alternates#browser_compatibility). Another one that may require work when we generate statuses.

<details>
<summary>This is a new feature. Expand for ideas for reviewing it.</summary>

- Are any pieces missing?

</details>


foolip:
I updated the spec link to to use the section heading https://drafts.csswg.org/css-fonts-4/#font-variant-alternates-prop. When the heading is for a single property this is a nicer landing location IMHO.

PR #333, opened 2023-08-28
Add motion path feature

ddbeck:
Corresponds to https://caniuse.com/css-motion-paths

I think we might need an "also known as" field for this kind of thing. I wonder if this name will stick, long-term, without it being the name of a property directly.

<details>
<summary>This is a new feature. Expand for ideas for reviewing it.</summary>

- Are any pieces missing?

</details>

### DIFF
--- a/feature-group-definitions/motion-path.yml
+++ b/feature-group-definitions/motion-path.yml
@@ -1,0 +1,14 @@
+spec: https://drafts.fxtf.org/motion-1/
+caniuse: css-motion-paths
+compat_features:
+  - css.properties.offset-anchor
+  - css.properties.offset-distance
+  - css.properties.offset-path
+  - css.properties.offset-path.basic-shape
+  - css.properties.offset-path.coord-box
+  - css.properties.offset-path.path
+  - css.properties.offset-path.ray
+  - css.properties.offset-position
+  - css.properties.offset-position.normal
+  - css.properties.offset-rotate
+  - css.properties.offset


Elchi3 (commented):
Thanks Daniel! This PR was useful to me when I wanted to find relevant BCD data entries for "CSS Motion Path" in order to update them all for Chrome 116. I think the property list is complete. However, because MDN also documents CSS types separately sometimes, there are also detailed docs on the [ray()](https://developer.mozilla.org/en-US/docs/Web/CSS/ray) function. The BCD data points you could include for that are:
- css.types.ray
- css.types.ray.position
- css.types.ray.size

Also, there are some new [`<basic-shape>`](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape) functions that are relevant here, but I'm not sure if it would be worth including them. The chromestatus entry mentions them with Motion Path, however: https://chromestatus.com/feature/5124394449371136.

dipikabh:
Hello Daniel. For the sake of completeness, perhaps the feature list can/should also include `css.properties.offset-path.url` – based on the syntax in the [motion path specification](https://drafts.fxtf.org/motion/#offset-path-property) and corresponding [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/offset-path#values).

foolip:
For this feature I think we'll ultimately need notes and a concept of partial support before the recent improvements in implementations. Maybe we'll need to create a subgroup for "basic support", but for now I suggest just listing all of the compat entries, and dealing with this problem as a manual override, since we'd compute the wrong versions based on the data directly.

@ddbeck can you include the css.types.ray entries and css.properties.offset-path.url?

ddbeck:
The reason I did not include `css.types.ray` or `css.properties.offset-path.url` is that neither is well-supported, unlike the rest of the features. This would cause us to break with the linked caniuse feature pretty significantly. That is, compare these two views of support:

- https://developer.mozilla.org/en-US/docs/Web/CSS/ray#browser_compatibility
- https://caniuse.com/css-motion-paths

I'm OK with making that judgment—I don't take it as a given that caniuse is always right—but I think we ought to go into it intentionally.

foolip (approved):

feature-group-definitions/motion-path.yml
```diff
@@ -0,0 +1,14 @@
+spec: https://drafts.fxtf.org/motion-1/
+caniuse: css-motion-paths
+compat_features:
+  - css.properties.offset-anchor
+  - css.properties.offset-distance
+  - css.properties.offset-path
+  - css.properties.offset-path.basic-shape
+  - css.properties.offset-path.coord-box
+  - css.properties.offset-path.path
+  - css.properties.offset-path.ray
```
foolip, line 10: Do you want to exclude this given https://developer.mozilla.org/en-US/docs/Web/CSS/offset-path#browser_compatibility? In fact it seems like css.properties.offset-path.path is the only subentry that we should include if we want this to be "old supported stuff"...

ddbeck:
> How about filing an issue (or draft PR) for the newer bits of motion path so that we don't lose sight of this? With that, I think it's fine to merge this with the current scope.

Yes, OK. I am going to remove the features that prevent this feature from showing as fully-supported as it does on caniuse and park them in https://github.com/web-platform-dx/web-features/issues/445.

foolip (approved):

PR #340, opened 2023-09-04
Rename `feature-set` to `web-features` in one more place

foolip:
These are the last instances of `feature-set` in the repo.


ddbeck (approved):
Nice catch. Thank you!

PR #386, opened 2023-10-12
Add `:where()` pseudo-class feature

foolip:
This goes well with https://github.com/web-platform-dx/web-features/pull/308.


ddbeck (approved):
Yes, I very much like handling these separately. They're not obviously tightly related and a quick search of "is and where selectors" mostly yields blog posts talking about them because they're both new.

PR #389, opened 2023-10-13
Add web animations api feature

captainbrosset:
This a new feature proposal for the Web Animations API.
It corresponds to https://caniuse.com/web-animation


foolip (approved):

feature-group-definitions/web-animations.yml
```diff
@@ -0,0 +1,57 @@
+spec: https://drafts.csswg.org/web-animations/
```
foolip, line 1: Tests are failing because this URL isn't in browser-specs. This will fix it:

```suggestion
spec: https://drafts.csswg.org/web-animations-1/
```

PR #398, opened 2023-10-19
New css gradient feature group

captainbrosset:
This a new feature proposal for CSS gradients.
It corresponds to https://caniuse.com/css-gradients and its two sub features: https://caniuse.com/css-repeating-gradients and https://caniuse.com/css-conic-gradients.
I covered all of the BCD entries I could find.

This PR also introduces a schema change to allow multiple caniuse entries to be specified. Does that make sense? Or is this where we should have a notion of feature hierarchy instead?

I also wasn't sure whether to name this `gradients` or `css-gradients`. I know the docs favor the former, but the latter might be better knowing that gradients also exist in SVG and in Canvas.


feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+caniuse:
+  - css-gradients
+  - css-repeating-gradients
+  - css-conic-gradients
+compat_features:
+  # Background-image prop
+  - css.properties.background-image.gradients
+  - css.properties.background-image.multiple_backgrounds
+  # CSS gradient image type
+  - css.types.image.gradient
```
ddbeck, line 11: So far, we haven't rolled up everything required to use a feature, only the distinctive things that make up that feature. So I think we can omit `background-image` and the `gradient` type.

See also: https://github.com/web-platform-dx/web-features/issues/217

```suggestion
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  # Linear gradients
+  - css.types.image.gradient.linear-gradient
+  - css.types.image.gradient.linear-gradient.doubleposition
+  - css.types.image.gradient.linear-gradient.hue_interpolation_method
+  - css.types.image.gradient.linear-gradient.interpolation_color_space
```
ddbeck, line 16: Firefox hasn't implemented these yet. Since they seem to be unique in the bunch for being unsupported, I think we ought to omit them as well. (Commenting them out to show that this is an intentional choice, though I suppose we could also just delete those lines.)

```suggestion
  # - css.types.image.gradient.linear-gradient.hue_interpolation_method
  # - css.types.image.gradient.linear-gradient.interpolation_color_space
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  # Radial gradients
+  - css.types.image.gradient.radial-gradient
+  - css.types.image.gradient.radial-gradient.at
+  - css.types.image.gradient.radial-gradient.doubleposition
+  - css.types.image.gradient.radial-gradient.hue_interpolation_method
+  - css.types.image.gradient.radial-gradient.interpolation_color_space
+  - css.types.image.gradient.radial-gradient.interpolation_hints
+  # Conic gradients
+  - css.types.image.gradient.conic-gradient
+  - css.types.image.gradient.conic-gradient.doubleposition
```
ddbeck, line 29:
```suggestion
  # - css.types.image.gradient.conic-gradient
  # - css.types.image.gradient.conic-gradient.doubleposition
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  - css.types.image.gradient.linear-gradient.interpolation_color_space
+  - css.types.image.gradient.linear-gradient.interpolation_hints
+  - css.types.image.gradient.linear-gradient.to
+  - css.types.image.gradient.linear-gradient.unitless_0_angle
+  # Radial gradients
+  - css.types.image.gradient.radial-gradient
+  - css.types.image.gradient.radial-gradient.at
+  - css.types.image.gradient.radial-gradient.doubleposition
+  - css.types.image.gradient.radial-gradient.hue_interpolation_method
+  - css.types.image.gradient.radial-gradient.interpolation_color_space
```
ddbeck, line 25:
```suggestion
  # - css.types.image.gradient.radial-gradient.hue_interpolation_method
  # - css.types.image.gradient.radial-gradient.interpolation_color_space
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  # Conic gradients
+  - css.types.image.gradient.conic-gradient
+  - css.types.image.gradient.conic-gradient.doubleposition
+  - css.types.image.gradient.conic-gradient.hue_interpolation_method
+  - css.types.image.gradient.conic-gradient.interpolation_color_space
+  # Repeating linear gradients
+  - css.types.image.gradient.repeating-linear-gradient
+  - css.types.image.gradient.repeating-linear-gradient.doubleposition
+  - css.types.image.gradient.repeating-linear-gradient.hue_interpolation_method
+  - css.types.image.gradient.repeating-linear-gradient.interpolation_color_space
```
ddbeck, line 36:
```suggestion
  # - css.types.image.gradient.repeating-linear-gradient.hue_interpolation_method
  # - css.types.image.gradient.repeating-linear-gradient.interpolation_color_space
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  - css.types.image.gradient.repeating-radial-gradient
+  - css.types.image.gradient.repeating-radial-gradient.at
+  - css.types.image.gradient.repeating-radial-gradient.doubleposition
+  - css.types.image.gradient.repeating-radial-gradient.hue_interpolation_method
+  - css.types.image.gradient.repeating-radial-gradient.interpolation_color_space
+  - css.types.image.gradient.repeating-radial-gradient.interpolation_hints
+  # Repeating conic gradients
+  - css.types.image.gradient.repeating-conic-gradient
+  - css.types.image.gradient.repeating-conic-gradient.hue_interpolation_method
+  - css.types.image.gradient.repeating-conic-gradient.interpolation_color_space
```
ddbeck, line 50:
```suggestion
  # - css.types.image.gradient.repeating-conic-gradient.hue_interpolation_method
  # - css.types.image.gradient.repeating-conic-gradient.interpolation_color_space
```

feature-group-definitions/gradients.yml
```diff
@@ -0,0 +1,50 @@
...
+  - css.types.image.gradient.repeating-linear-gradient.interpolation_color_space
+  - css.types.image.gradient.repeating-linear-gradient.interpolation_hints
+  - css.types.image.gradient.repeating-linear-gradient.to
+  - css.types.image.gradient.repeating-linear-gradient.unitless_0_angle
+  # Repeating radial gradients
+  - css.types.image.gradient.repeating-radial-gradient
+  - css.types.image.gradient.repeating-radial-gradient.at
+  - css.types.image.gradient.repeating-radial-gradient.doubleposition
+  - css.types.image.gradient.repeating-radial-gradient.hue_interpolation_method
+  - css.types.image.gradient.repeating-radial-gradient.interpolation_color_space
```
ddbeck, line 45:
```suggestion
  # - css.types.image.gradient.repeating-radial-gradient.hue_interpolation_method
  # - css.types.image.gradient.repeating-radial-gradient.interpolation_color_space
```

ddbeck (changes requested):
> This PR also introduces a schema change to allow multiple caniuse entries to be specified. Does that make sense? Or is this where we should have a notion of feature hierarchy instead?

I think you have this right. I imagine several features will get folded up like this (I'm imagining all the subfeatures of something huge like `es6`). I did confirm that this folding would only affect older features, however (i.e., these are well beyond the "wider support" timelines that have been proposed).

cc: @foolip

feature-group-definitions/gradients.yml
captainbrosset: Ok, I think that makes sense, but I can see cases when this could be a problem, in theory.

Imagine a new CSS value `foo`. Let's say this value can be used by `bar` and `baz`.
What if one browser supports both `bar:foo` and `baz:foo` but another browser only supports `bar:foo`.

Maybe `baz:foo` is a cool feature that we want to document. If we don't add `baz` to the list of compat features, then we don't know if browser really support it.

ddbeck: > Imagine a new CSS value `foo`. Let's say this value can be used by `bar` and `baz`. What if one browser supports both `bar:foo` and `baz:foo` but another browser only supports `bar:foo`.
> 
> Maybe `baz:foo` is a cool feature that we want to document. If we don't add `baz` to the list of compat features, then we don't know if browser really support it.

Yes, I agree that this is a real concern: declaring support for a type, abstractly, doesn't say much about the practical outcome of using it with a specific property. That said, do you have a feeling about what properties should and should not qualify for this group?

I have a concrete example here: `background-image: <gradient>` works and so it doesn't alter the overall support story here. But what about `content: <gradient>`? It's specified as accepting a `<gradient>` value, but [the support story is not so good](https://developer.mozilla.org/en-US/docs/Web/CSS/content#browser_compatibility). Should we say that the overall CSS gradients feature isn't supported until that lands too? If not, what's the criteria for choosing some properties and not others?

I genuinely do not know the answer to these questions, so I'm more than open to ideas here.

---

Some optional background about BCD:

CSS types are probably BCD's weakest data. I've never felt good about what it covers. I believe the (undocumented) convention is to report a type as supported when the type is first supported as the value of _any_ property. That's probably good enough for the case of "is this type supported in any way?" but probably not so much for "can I use this type with property `baz`?

In a better world, BCD would also (or only) have keys along the lines of `css.properties.background-image.gradient.{conic-gradient,linear-gradient,…}`. Then we could say with some certainty whether you could _use_ a given CSS type across the properties where it's an allowed value.

ddbeck: Hi @captainbrosset, did you have any thoughts on this question?

> what properties should and should not qualify for this group?

(otherwise we can plow ahead with this and wait for the question to come up again)

captainbrosset: > I have a concrete example here: background-image: <gradient> works and so it doesn't alter the overall support story here. But what about content: <gradient>? It's specified as accepting a <gradient> value, but [the support story is not so good](https://developer.mozilla.org/en-US/docs/Web/CSS/content#browser_compatibility). Should we say that the overall CSS gradients feature isn't supported until that lands too? If not, what's the criteria for choosing some properties and not others?

We probably shouldn't gate CSS gradients on the fact that it's suported for `content`. So I'm fine with removing the lines you suggested.

It does, however, make it harder to reason about what goes into a group and what doesn't.

In that particular example, where would we signify that gradients aren't support in `content` yet? Would we make another group (maybe named `full-gradient`, or `gradient-v2-final`) that contains this group + the support for `content`?

It seems this project should be able to say if support for gradient in `content` is here or not. Do you agree?
If so, then I think people would naturally go to the `gradient` group to find this information.

Maybe that means that support for a group isn't only the "sum of its parts". Maybe some of the listed parts shouldn't count towards the group's overall suppor, but still be here for decision making. In this case, gradient for `content` would be listed somehow, maybe as a note, but wouldn't make the group become red.

ddbeck: So I just searched for "css gradients" in the hopes of getting a better idea of what web developers are hearing about when they hear about "gradients." I looked at the first several search results and also checked the docs for Tailwind and Bootstrap. With the major exception of MDN, everybody talks about _background_ gradients.

<details>
<summary>Some sample reading</summary>

- [CSS Gradient — Generator, Maker, and Background](https://cssgradient.io/) - Top result on Google, generates `background` properties. Several other generators in the top results do the same and caniuse links to [Ultimate CSS Gradient Generator - ColorZilla.com](https://www.colorzilla.com/gradient-editor/).
- [Gradient Color Stops - Tailwind CSS](https://tailwindcss.com/docs/gradient-color-stops) - In the _Backgrounds_ section
- [CSS Gradients - W3Schools](https://www.w3schools.com/css/css3_gradients.asp) - Examples all `background-image`
- [CSS Gradients | CSS-Tricks - CSS-Tricks](https://css-tricks.com/css3-gradients/) - Lede: "Just as you can declare the background of an element to be a solid color in CSS, you can also declare that background to be a gradient."
- [Make Beautiful Gradients in CSS, with linear-gradient, radial-gradient, or conic-gradient.](https://www.joshwcomeau.com/css/make-beautiful-gradients/) - Josh W. Comeau doesn't say it explicitly, but the examples use `background` and his generator uses `background-image`
- [Background · Bootstrap v5.3](https://getbootstrap.com/docs/5.3/utilities/background/#background-gradient) - Again `background-image` here

</details>

To me this suggests a fairly straightforward way out: we rename this feature to be `background-gradients`. This is consistent with commonly-used terminology and makes the scope of the feature explicit. Should we decide to cover `content` gradients, then we have a natural follow-up: `content-gradients`. We also leave the door open to some kind omnibus group, like `gradients` (containing both `content-gradients` and `background-gradients`).

(As a bonus, it also makes it less likely for us to collide with SVG gradients.)

ddbeck:
OK, I thought it only fair that I resolve the merge conflict since I created it in the first place. After a couple of false starts, that's done now.

I also reinstated the `css.properties.background-image.gradients` feature since we re-scoped the feature (see https://github.com/web-platform-dx/web-features/pull/398#discussion_r1383217978).

@foolip It'd be nice to get a review from someone who hasn't touched this a few times.

feature-group-definitions/background-gradients.yml
```diff
@@ -0,0 +1,46 @@
+spec: https://drafts.csswg.org/css-images-4/#gradients
+caniuse:
+  - css-gradients
+  - css-repeating-gradients
+  - css-conic-gradients
```
foolip, line 5: Including conic gradients makes the whole feature newer, but it's all still pre-2020 AFAICT. So let's try this.

feature-group-definitions/background-gradients.yml
```diff
@@ -0,0 +1,46 @@
...
+  - css.types.image.gradient.radial-gradient.at
+  - css.types.image.gradient.radial-gradient.doubleposition
+  # - css.types.image.gradient.radial-gradient.hue_interpolation_method
+  # - css.types.image.gradient.radial-gradient.interpolation_color_space
+  - css.types.image.gradient.radial-gradient.interpolation_hints
+  # Conic gradients
+  # - css.types.image.gradient.conic-gradient
+  # - css.types.image.gradient.conic-gradient.doubleposition
+  - css.types.image.gradient.conic-gradient.hue_interpolation_method
+  - css.types.image.gradient.conic-gradient.interpolation_color_space
```
foolip, line 27: Why not include css.types.image.gradient.conic-gradient but include some of its subentries? The precise opposite choices are more similar to linear and radial gradients above:

```suggestion
  # Conic gradients
  - css.types.image.gradient.conic-gradient
  - css.types.image.gradient.conic-gradient.doubleposition
  # - css.types.image.gradient.conic-gradient.hue_interpolation_method
  # - css.types.image.gradient.conic-gradient.interpolation_color_space
```

ddbeck: Yes, this is the right thing to do.

feature-group-definitions/background-gradients.yml
ddbeck: So this ends up being interoperable on 2020-11-17. Long enough for the wider-support status, though not quite pre-2020.

captainbrosset: Agreed, let's try it like this. I think this should be ready to merge.

foolip (approved):

PR #399, opened 2023-10-20
Add feature for AVIF

foolip:


foolip:
This would be the first image format. These aren't tracked in BCD, but I think that this is undeniably of interest to web developers, and is tracked by caniuse.com for good reasons.

If this is accepted I can also send PRs for WebP and JPEG XL.

ddbeck:
This seems like a good thing to cover and I'm happy to merge it as is.

But I'll note that https://github.com/mdn/browser-compat-data/pull/4193 is interesting reading: it's where we _didn't_ add formats/codecs/containers to BCD. The summary is that the support story for a format is approximately multiplicative (e.g., video element support × container support × codec support determines whether or not a particular video file will play in a browser). It makes things a lot easier if we're _not_ committed to comprehensiveness (e.g., we select some typical use case as our measure for support for a format).

ddbeck (approved):

foolip:
@ddbeck I've seen that issue before and I think that when the only option was to hang features like "opus_media_codec" onto other granular BCD entries, it does become unwieldy, with no entry representing "Opus works in all contexts". I think a single high-level entry for the format avoids some of those problems. I'll send more PRs :)

PR #407, opened 2023-11-02
Subgrid is now baseline

estelle:
make subgrid baseline
https://caniuse.com/css-subgrid


ddbeck (approved):
Thanks @estelle. The definition of Baseline is likely to change soon/finally and this will get modified as a consequence, but I appreciate the consistency for now.

PR #411, opened 2023-11-06
Reject hidden caniuse features

ddbeck:
~~The build will fail on this PR because `@types/caniuse-lite`'s type definitions are out of date. See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67330.~~ I've temporarily overridden the type so we can move on with this.

This PR checks that features' `caniuse` values point only to features that appear on caniuse.com. It disallows caniuse ID where the caniuse `shown` property is `false`.

This impacts the caniuse burndown list. I've chosen to mark them as complete, but also to note that they're hidden features.

This PR closes #275.

### DIFF
--- a/scripts/caniuse.ts
+++ b/scripts/caniuse.ts
@@ -20,7 +20,12 @@ const mapping = new Map<string, string | null>(
     Object.keys(lite.features).sort().map(id => [id, null])
 );
 
-const hiddenCaniuseItems = new Set(Object.entries(lite.features).flatMap(([id, data]) => !lite.feature(data).shown ? [id] : []));
+const hiddenCaniuseItems = new Set<string>();
+for (const [id, data] of Object.entries(lite.features)) {
+    if (!lite.feature(data).shown) {
+        hiddenCaniuseItems.add(id);
+    }
+}
 
 for (const [id, data] of Object.entries(features)) {
     if (!('caniuse' in data)) {
--- a/scripts/caniuse.ts
+++ b/scripts/caniuse.ts
@@ -20,6 +20,8 @@ const mapping = new Map<string, string | null>(
     Object.keys(lite.features).sort().map(id => [id, null])
 );
 
+const hiddenCaniuseItems = Object.entries(lite.features).flatMap(([id, data]) => !lite.feature(data).shown ? [id] : []);
+
 for (const [id, data] of Object.entries(features)) {
     if (!('caniuse' in data)) {
         continue;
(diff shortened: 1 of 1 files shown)


scripts/caniuse.ts
```diff
@@ -20,6 +20,8 @@ const mapping = new Map<string, string | null>(
     Object.keys(lite.features).sort().map(id => [id, null])
 );
 
+const hiddenCaniuseItems = Object.entries(lite.features).flatMap(([id, data]) => !lite.feature(data).shown ? [id] : []);
```
foolip, line 23:
```suggestion
const hiddenCaniuseItems = new Set(Object.entries(lite.features).flatMap(([id, data]) => !lite.feature(data).shown ? [id] : []));
```

scripts/caniuse.ts
```diff
@@ -28,22 +30,34 @@ for (const [id, data] of Object.entries(features)) {
     if (!mapping.has(caniuseId)) {
         throw new Error(`Invalid caniuse ID used for ${id}: ${caniuseId}`);
     }
+    if (hiddenCaniuseItems.includes(caniuseId)) {
```
foolip, line 33:
```suggestion
    if (hiddenCaniuseItems.has(caniuseId)) {
```

scripts/caniuse.ts
```diff
@@ -28,22 +30,34 @@ for (const [id, data] of Object.entries(features)) {
...
 
 for (const [caniuseId, id] of mapping.entries()) {
-    let checkbox = '[ ]';
-    let details = '';
-    if (id) {
-        checkbox = '[x]';
-        if (id !== caniuseId) {
-            details = ` (as ${id})`;
-        }
+    const isHidden = hiddenCaniuseItems.includes(caniuseId);
```
foolip, line 43:
```suggestion
    const isHidden = hiddenCaniuseItems.has(caniuseId);
```

scripts/caniuse.ts
foolip: But also, this one-liner takes a bit of work to decode. I would understand this faster:

```js
const hiddenCaniuseItems = new Set();
for (const [id, data] of Object.entries(lite.features)) {
  if (!lite.feature(data).shown) {
    hiddenCaniuseItems.add(id);
  }
}
```

foolip (approved):

scripts/caniuse.ts
ddbeck: I always forget `Set` is a thing. I much prefer this approach (and the unrolling of the flatmap, which was a bit too clever to begin with).

PR #434, opened 2023-11-27
Update Baseline schema for new definition of Baseline (and refresh some statuses)

ddbeck:
(detailed breakdown of support forthcoming, but a few features were left out because of incorrect keys or other issues with the definition itself)

### DIFF
--- a/feature-group-definitions/canvas-context-lost.yml
+++ b/feature-group-definitions/canvas-context-lost.yml
@@ -4,6 +4,7 @@ status:
   is_baseline: false
   support:
     chrome: "99"
+    chrome_android: "99"
     edge: "99"
 compat_features:
   - api.CanvasRenderingContext2D.isContextLost
--- a/index.ts
+++ b/index.ts
@@ -20,13 +20,17 @@ export interface FeatureData {
     usage_stats?: usage_stats_url | [usage_stats_url, usage_stats_url, ...usage_stats_url[]];  // A single URL or an array of two or more
 }
 
-type browserIdentifier = "chrome" | "edge" | "firefox" | "safari";
+type browserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
 
 interface SupportStatus {
-    /** Whether the feature achieved baseline status */
+    /** Whether the feature achieved Baseline status */
     is_baseline: boolean;
-    /** Date the feature achieved baseline status */
+    /** Whether the feature is widely supported (high) or newly supported (low) */
+    substatus?: "high" | "low";
+    /** Date the feature achieved Baseline status (legacy) */
     since?: string;
+    /** Date the feature achieved Baseline low status */
+    keystone_date?: string;
     /** Browser versions that most-recently introduced the feature */
     support?: {[K in browserIdentifier]?: string};
 }
--- a/package.json
+++ b/package.json
@@ -39,5 +39,8 @@
     "winston": "^3.11.0",
     "yaml": "^2.3.4",
     "yargs": "^17.7.2"
+  },
+  "peerDependencies": {
+    "picomatch": "^3.0.1"
   }
 }
--- a/feature-group-definitions/border-image.yml
+++ b/feature-group-definitions/border-image.yml
@@ -3,12 +3,16 @@ caniuse: border-image
 usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/43
 status:
   is_baseline: true
-  since: "2017-03-09"
+  substatus: high
+  keystone_date: 2017-02-01
   support:
     chrome: "56"
+    chrome_android: "56"
     edge: "12"
     firefox: "50"
+    firefox_android: "50"
     safari: "9.1"
+    safari_ios: "9.3"
 compat_features:
   - css.properties.border-image
   - css.properties.border-image.fill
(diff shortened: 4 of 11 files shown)


index.ts
```diff
@@ -20,13 +20,17 @@ export interface FeatureData {
...
-type browserIdentifier = "chrome" | "edge" | "firefox" | "safari";
+type browserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
 
 interface SupportStatus {
-    /** Whether the feature achieved baseline status */
+    /** Whether the feature achieved Baseline status */
     is_baseline: boolean;
-    /** Date the feature achieved baseline status */
+    /** Whether the feature is widely supported (high) or newly supported (low) */
+    substatus?: "high" | "low";
```
foolip, line 29: `substatus` should be required if `is_baseline` is true, and forbidden otherwise. Rather than expressing this in the schema, how about simplifying to an enum of possible states, like this: `baseline: false | "high" | "low"`

A mix of booleans and strings is similar to BCD's `version_added`, and means it's possible to still use `baseline` like a boolean.

index.ts
```diff
@@ -20,13 +20,17 @@ export interface FeatureData {
...
+    /** Date the feature achieved Baseline status (legacy) */
```
foolip, line 30: Why not just keep `since`?

index.ts
```diff
@@ -20,13 +20,17 @@ export interface FeatureData {
...
     since?: string;
+    /** Date the feature achieved Baseline low status */
+    keystone_date?: string;
```
foolip, line 33: If we can't just keep `since` above, I think `baseline_date` or `baseline_since` would be better here than `keystone_date`. We use that terminology in the definition to define what is Baseline, but from the outside in, it makes just as much or more sense to talk about a Baseline date, I think.

package.json
```diff
@@ -39,5 +39,8 @@
     "winston": "^3.11.0",
     "yaml": "^2.3.4",
     "yargs": "^17.7.2"
+  },
+  "peerDependencies": {
+    "picomatch": "^3.0.1"
```
foolip, line 44: Not used?

index.ts
LeoMcA: I suppose `since` + high gets ambiguous. Is that the date since baseline low? Or the date since baseline high?

index.ts
ddbeck: OK, I've switched to a `baseline` field with one of the strings or `false`.

package.json
ddbeck: I ran into an error about `picomatch` when I ran one of the `npm run …` scripts, but now I can't reproduce this. Rolling it back.

feature-group-definitions/subgrid.yml
```diff
@@ -2,13 +2,14 @@ spec: https://drafts.csswg.org/css-grid-2/#subgrids
 caniuse: css-subgrid
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4680
 status:
-  is_baseline: true
-  since: "2023-09-15" # since Edge 117
+  baseline: false
```
LeoMcA, line 5: Shouldn't this be baseline low?

ddbeck: This is "correct" because BCD says that the [`css.properties.grid-template-rows.subgrid` feature is unsupported on Firefox for Android](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows#browser_compatibility). This is "wrong" because the data is likely to be wrong (BCD says `css.properties.grid-template-columns.subgrid` _is_ supported on Firefox for Android).

Which is to say, the Baseline status is wrong for the right reason.

feature-group-definitions/canvas-context-lost.yml
```diff
@@ -4,6 +4,7 @@ status:
   is_baseline: false
```
LeoMcA, line 4:
```suggestion
  baseline: false
```

feature-group-definitions/subgrid.yml
LeoMcA: I see https://github.com/mdn/browser-compat-data/pull/21388 should fix this - nice :+1:

index.ts
```diff
@@ -20,13 +20,19 @@ export interface FeatureData {
...
+type BaselineHighLow = "high" | "low";
 
 interface SupportStatus {
-    /** Whether the feature achieved baseline status */
-    is_baseline: boolean;
-    /** Date the feature achieved baseline status */
+    /** Whether the feature is Baseline (low substatus), Baseline (high substatus), or not (false) */
+    baseline?: BaselineHighLow | false;
+    /** Whether the feature is Baseline (legacy) */
+    is_baseline?: true | false;
```
LeoMcA, line 31: Should this be removed from the schema? Or are we keeping it for backwards compatibility? (No need to do so for MDN)

index.ts
```diff
@@ -20,13 +20,19 @@ export interface FeatureData {
...
+    /** Date the feature achieved Baseline status (legacy) */
     since?: string;
```
LeoMcA, line 33: Similarly - no need to keep this in the schema, if MDN as consumer is the reason for it.

LeoMcA (approved):
Couple more schema comments (in terms of backwards compatibility) but otherwise lgtm

index.ts
ddbeck: I can't remove all of the old statuses yet—they need actual reviews for the feature lists. I'll do that in a follow up PR.

feature-group-definitions/subgrid.yml
ddbeck: And I've pushed a change to accept the latest value now. 👍

index.ts
foolip: Ah, I didn't consider that.

PR #458, opened 2023-12-01
`input-event`: set Baseline status

ddbeck:
I'm somewhat skeptical of this one. Safari has two issues:

- `InputEvent()` is listed as partial, due to missing some options on the constructor. I wonder if anyone's checked on this recently.
- `isComposing` was added later. This looks right.

Without the former, this is a newly available feature; without the latter, it's widely supported.

## input-event

| Key                                                                                                                             |    Baseline    | Low since date | Versions                                                                                                                    |
| :------------------------------------------------------------------------------------------------------------------------------ | :------------: | :------------- | --------------------------------------------------------------------------------------------------------------------------- |
| **input-event**                                                                                                                 | ❌ Not Baseline |                | Chrome 60<br>Chrome Android 60<br>Edge 79<br>Firefox 87<br>Firefox for Android 87<br>Safari ❌<br>Safari on iOS ❌            |
| [`api.HTMLElement.input_event`](https://developer.mozilla.org/docs/Web/API/HTMLElement/input_event#browser_compatibility)       |     ✅ High     | 2020-01-15     | Chrome 1<br>Chrome Android 18<br>Edge 79 🔑💎<br>Firefox 6<br>Firefox for Android 6<br>Safari 3.1<br>Safari on iOS 2        |
| [`api.InputEvent`](https://developer.mozilla.org/docs/Web/API/InputEvent#browser_compatibility)                                 |     ✅ High     | 2020-01-15     | Chrome 60<br>Chrome Android 60<br>Edge 79 🔑💎<br>Firefox 31<br>Firefox for Android 31<br>Safari 10.1<br>Safari on iOS 10.3 |
| [`api.InputEvent.InputEvent`](https://developer.mozilla.org/docs/Web/API/InputEvent/InputEvent#browser_compatibility)           | ❌ Not Baseline |                | Chrome 60<br>Chrome Android 60<br>Edge 79<br>Firefox 31<br>Firefox for Android 31<br>Safari ❌<br>Safari on iOS ❌            |
| [`api.InputEvent.data`](https://developer.mozilla.org/docs/Web/API/InputEvent/data#browser_compatibility)                       |     ✅ High     | 2020-01-15     | Chrome 60<br>Chrome Android 60<br>Edge 79 🔑💎<br>Firefox 67<br>Firefox for Android 67<br>Safari 10.1<br>Safari on iOS 10.3 |
| [`api.InputEvent.dataTransfer`](https://developer.mozilla.org/docs/Web/API/InputEvent/dataTransfer#browser_compatibility)       |     ✅ High     | 2020-01-15     | Chrome 60<br>Chrome Android 60<br>Edge 79 🔑💎<br>Firefox 67<br>Firefox for Android 67<br>Safari 10.1<br>Safari on iOS 10.3 |
| [`api.InputEvent.getTargetRanges`](https://developer.mozilla.org/docs/Web/API/InputEvent/getTargetRanges#browser_compatibility) |     ✅ High     | 2021-03-23     | Chrome 60<br>Chrome Android 60<br>Edge 79<br>Firefox 87 🔑💎<br>Firefox for Android 87<br>Safari 10.1<br>Safari on iOS 10.3 |
| [`api.InputEvent.inputType`](https://developer.mozilla.org/docs/Web/API/InputEvent/inputType#browser_compatibility)             |     ✅ High     | 2020-01-15     | Chrome 60<br>Chrome Android 60<br>Edge 79 🔑💎<br>Firefox 66<br>Firefox for Android 66<br>Safari 10.1<br>Safari on iOS 10.3 |
| [`api.InputEvent.isComposing`](https://developer.mozilla.org/docs/Web/API/InputEvent/isComposing#browser_compatibility)         |     🆕 Low     | 2023-03-27     | Chrome 60<br>Chrome Android 60<br>Edge 79<br>Firefox 31<br>Firefox for Android 31<br>Safari 16.4 🔑💎<br>Safari on iOS 16.4 |

<small>🔑💎 indicates a release that determines the Baseline low date.</small>

### DIFF
--- a/feature-group-definitions/input-event.yml
+++ b/feature-group-definitions/input-event.yml
@@ -1,5 +1,13 @@
 spec: https://w3c.github.io/uievents/#event-type-input
 caniuse: input-event
+status:
+  baseline: false
+  support:
+    chrome: "60"
+    chrome_android: "60"
+    edge: "79"
+    firefox: "87"
+    firefox_android: "87"
 compat_features:
   - api.HTMLElement.input_event
   - api.InputEvent
--- a/feature-group-definitions/input-event.yml
+++ b/feature-group-definitions/input-event.yml
@@ -1,13 +1,16 @@
 spec: https://w3c.github.io/uievents/#event-type-input
 caniuse: input-event
 status:
-  baseline: false
+  baseline: low
+  baseline_low_date: 2023-03-27
   support:
     chrome: "60"
     chrome_android: "60"
     edge: "79"
     firefox: "87"
     firefox_android: "87"
+    safari: "16.4"
+    safari_ios: "16.4"
 compat_features:
   - api.HTMLElement.input_event
   - api.InputEvent


tidoust:
> InputEvent() is listed as partial, due to missing some options on the constructor. I wonder if anyone's checked on this recently.

I see a relatively recent (<2 months old) pull request to add support for missing options in the constructor to WebKit, so that still seems current: https://github.com/WebKit/WebKit/pull/19346

captainbrosset:
I think this is correct. Missing a constructor option seems like a big deal to me, and I certainly wouldn't call a feature that doesn't have that kind of thing baseline.

This feature is typically the type of things I had in mind when I commented on how Baseline should help _close the loop_ with developers. Here, we have an incomplete feature because there's one bug left to fix in one browser, for it to be fully usable everywhere. Baseline has a unique opportunity to say "No, it's not baseline, and here's why: ...".

BCD contains the right info:
image

Wouldn't it be cool if, when you expand the baseline banner, you saw "The inputEventInit parameter is not yet supported on Safari. See bug 170416".

captainbrosset (approved):

foolip:
I think we should sometimes treat constructors as completely optional parts of a feature, which don't affect Baseline at all. I'd argue this for all subclasses of Event, because those constructors are just nice-to-haves for creating and firing events yourself. The feature that matters is "the browser creates and fires `InputEvent`s at the right time", not "I can create an `InputEvent` and fire it myself".

In other cases, like `AudioContext` the constructor is the entry point to the API and essential.

captainbrosset:
That makes sense to me, as long as we can make this clear in the feature. If someone was, in fact, looking for baseline data about creating and firing events, which feature would they look at? And how would they know that one is about creating events yourself, and the other about the browser firing events?

ddbeck:
Can we generalize this as a guideline that says something along the lines of "don't include event instructors in a feature, unless we have evidence that web developers consider it to be an essential part of a feature"?

BCD and MDN do distinguish events that fire from the event interfaces themselves. If you look through BCD for keys ending in `_event`, you'll find tons of 'em. We should make sure to include those (and we do so here, as `api.HTMLElement.input_event`).

ddbeck:
Actually, on second thought, it appears that the partial implementation for the constructor is no recorded. Maybe we don't need to generalize after all? 😅 

https://github.com/mdn/browser-compat-data/issues/12163
https://github.com/mdn/browser-compat-data/pull/21452

It does seem that WebKit lacks some of the members of the constructor options parameter, but the overall constructor works and BCD doesn't report the individual parameters' support. So we can refresh the status here and move on.

foolip:
> BCD and MDN do distinguish events that fire from the event interfaces themselves.

BCD does, the event being fired by the browser is the `api.HTMLElement.input_event` entry, and the event interface is `api.InputEvent`.

And the rarely-used constructor is `api.InputEvent.InputEvent`.

foolip (approved):

foolip:
I've filed https://github.com/web-platform-dx/web-features/issues/504 about a guideline for event constructors.

tidoust:
> It does seem that WebKit lacks some of the members of the constructor options parameter, but the overall constructor works and BCD doesn't report the individual parameters' support. So we can refresh the status here and move on.

Not trying to block anything, @foolip's rationale that "[event] constructors are just nice-to-haves" and not core to the feature under consideration seems good to me!

I just note that the rationale that "the overall constructor works and BCD doesn't report the individual parameters' support. So [...] we can move on" feels shallow: if a developer passes in some parameters to the constructor and gets an event whose properties are not setup according to these parameters, that seems problematic (and a bug that may not be easy to track down because everything would seem to work just fine). I wish BCD would have updated the implementation note rather than dropped it!

PR #474, opened 2023-12-04
Revise package README for new schema

ddbeck:


packages/web-features/README.md
```diff
@@ -13,19 +13,20 @@ import webFeatures from 'web-features';
...
 
-For baseline iconography, follow this procedure for each feature:
+For Baseline iconography, follow this procedure for each feature:
 
-1. If `status.is_baseline` is `true`, then show the baseline (🟢 affirmative, green check) badge.
-1. If `status.is_baseline` is `false`, then show the non-baseline (🟡 caution, yellow X) badge.
-1. If `status.is_baseline` is `undefined`, then **do not** show any baseline or non-baseline badge.
+1. If `status.baseline` is `"high"`, then show an affirmative "widely available" icon.
+1. If `status.baseline` is `"low"`, then show an affirmative "newly available" icon.
+1. If `status.baseline` is `false`, then show a "limited availability" non-Baseline icon.
```
tidoust, line 24: Commenting here while glancing at updates you made on open PRs. I can live with it of course, but it feels weird to me to have a mix of types string + boolean + undefined values for the property (especially since the update changes the name of the property, so no strong need to be backwards compatible). Developer's life seems complex enough with string + undefined ;)

ddbeck: For a bit of background on this choice, see: https://github.com/web-platform-dx/web-features/pull/434#discussion_r1407617325

I had a similar reaction as you did, though the case was made that you could do something like `Boolean(baseline)` and get a sensible result with `"high" | "low" | false`.

We still have undefined specifically as a problem. I think there's a mitigation for that: get Baseline statuses for every feature, then forbid undefined as an option. 

When it comes to mixing types here, I was convinced, but until we hit 1.0.0 I think we're free to change our minds on this. I'd welcome an issue or PR, if you've got an idea for how to deal with the mix of types (and name the corresponding fields).

tidoust: Thanks for the background. The comment mentions `version_added` in BCD, which follows a similar pattern. If that works for BCD...

PR #423, opened 2023-11-14
Add Baseline definition document

ddbeck:
On Tuesday, November 7, 2023, the web-features governance team agreed to a new definition of Baseline, incorporating feedback from web developers, WebDX community group participants, and commenters on the web-features repository (mostly summarized by #374).

PR #479, opened 2023-12-05
Move Baseline status definition review date text

ddbeck:
This PR moves the "due for review" date to the "Ownership and maintenance" section of the Baseline doc.

As @tidoust noted in https://github.com/web-platform-dx/web-features/pull/423#discussion_r1410847301, the existing phrasing is somewhat over determined. It suggests that the duration and only the duration would come up for review. This conflicts with the governance text that says that reviewing and revising the definition (in general) is on the agenda annually. This PR attempts to reconcile the competing texts.

### DIFF
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -118,6 +118,8 @@ This is but one of several possible stories to help keep in mind the needs and c
 The WebDX community group, through the [web-platform-dx/web-features-set owners group](../GOVERNANCE.md), maintains this document.
 Based on WebDX community group research, the web-features owners group decides matters such as the core browser set, releases, editorial overrides, and so on.
 
+The status definition is due for review by the governance group on 7 November 2024.
+
 ## Status definition
 
 The status has two substatuses: an interoperable low status and a wider-support high status.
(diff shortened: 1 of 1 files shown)


ddbeck:
(Since this touches the Baseline definition, it needs review from the owners group.)

atopal:
LGTM!

docs/baseline.md
```diff
@@ -118,6 +118,8 @@ This is but one of several possible stories to help keep in mind the needs and c
 The WebDX community group, through the [web-platform-dx/web-features-set owners group](../GOVERNANCE.md), maintains this document.
 Based on WebDX community group research, the web-features owners group decides matters such as the core browser set, releases, editorial overrides, and so on.
 
+The status definition is due for review by the governance group on 7 November 2024.
```
tidoust, line 121: The section and governance document use "owners group", not "governance group".

Also, it suddenly occurs to me that putting an explicit due date in the definition itself means the definition will have to be updated every year no matter what, even if the governance group determines that no updates are warranted. Is that intended? And does the date actually matter?

```suggestion
The owners group reviews the status definition at least once per calendar year.
```

foolip: I've filed https://github.com/web-platform-dx/web-features/issues/502 about "governance group" vs. "owners groups".

Beyond that this PR is just moving the date. I think it's fine to have the date in this document and being required to update it once a year. Not a big burden for the owners I think :)

foolip (approved):

docs/baseline.md
bden77294-creator: > The section and governance document use "owners group", not "governance group".
> 
> Also, it suddenly occurs to me that putting an explicit due date in the definition itself means the definition will have to be updated every year no matter what, even if the governance group determines that no updates are warranted. Is that intended? And does the date actually matter?

https://github.com/web-platform-dx/web-features/pull/479#discussion_r1415700081

bden77294-creator:
> you

bden77294-creator:
LICENSE.txt

PR #489, opened 2023-12-07
Remove compat_features that were migrated to BCD

foolip:
Migrated to tags in BCD:

https://github.com/mdn/browser-compat-data/pull/21457
https://github.com/mdn/browser-compat-data/pull/21519


foolip:
@ddbeck let me know when you'd like to remove `compat_features`. It will require running the migration script again, which I can do, but I'd like to do it only once more, not ongoing.

ddbeck:
Also linking this here because it's a connected thing: https://github.com/web-platform-dx/web-features/pull/110

(To make Baseline calculations traceable, it seems to me that we'd still want the option to generate some "dist" YAML files that show what BCD keys were used to generate a status. And probably to spare consumers walking the entire BCD tree to find out what the constituent features are, though I'm less certain on that point.)

PR #499, opened 2023-12-15
Add Trusted Types feature

foolip:


feature-group-definitions/trusted-types.yml
```diff
@@ -0,0 +1,37 @@
...
+usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/2722
+status:
+  baseline: false
+  support:
+    chrome: "83"
+    chrome_android: "83"
+    edge: "83"
+compat_features:
+  - api.TrustedHTML
+  # - api.TrustedHTML.toJSON
```
tidoust, line 12: What is the rationale for excluding the `toJSON` methods?

Support in Chromium and likely the method itself have been added a bit later on, but then that wouldn't change its baseline status in any case (more than 30 months since version 90 was shipped, and feature is not supported in other browser engines).

foolip: It's just that Trusted Types can be used since Chrome 83. The version number isn't used on MDN, but if caniuse imports web-features, version numbers would matter. So I don't want to use the wrong version number just because it doesn't change the Baseline status.

What we could do is to include the `toJSON` and treat the version numbers as an editorial override, not matching what we'd infer from BCD. At some point we'll need a "part of the feature but not required to use" annotation I think.

foolip:
@ddbeck can you review this?

ddbeck (approved):
So this is fine, but I think this is really attractive:

> What we could do is to include the `toJSON` and treat the version numbers as an editorial override, not matching what we'd infer from BCD. At some point we'll need a "part of the feature but not required to use" annotation I think.

It might be fine to take the computation we have above, uncomment all the comments, and put in a comment saying why we've overriden the value. Later we could set up some list of "calculation exempt" features, in a structured way that's exposed publicly.

foolip:
Alright, I'll uncomment the `toJSON` bits and add a comment to remind us of the editorial override, so it doesn't get lost in the move from `compat_features` to BCD tags.

PR #461, opened 2023-12-03
Features L-N: set Baseline statuses

ddbeck:
## motion-path

| Key                                                                                                                  | Baseline | Low since date | Versions                                                                                                                    |
| :------------------------------------------------------------------------------------------------------------------- | :------: | :------------- | --------------------------------------------------------------------------------------------------------------------------- |
| **motion-path**                                                                                                      |  🆕 Low  | 2023-08-21     | Chrome 116<br>Chrome Android 116<br>Edge 116 🔑💎<br>Firefox 72<br>Firefox for Android 79<br>Safari 16<br>Safari on iOS 16  |
| [`css.properties.offset-anchor`](https://developer.mozilla.org/docs/Web/CSS/offset-anchor#browser_compatibility)     |  🆕 Low  | 2023-08-21     | Chrome 116<br>Chrome Android 116<br>Edge 116 🔑💎<br>Firefox 72<br>Firefox for Android 79<br>Safari 16<br>Safari on iOS 16  |
| [`css.properties.offset-distance`](https://developer.mozilla.org/docs/Web/CSS/offset-distance#browser_compatibility) |  🆕 Low  | 2022-09-12     | Chrome 55<br>Chrome Android 55<br>Edge 79<br>Firefox 72<br>Firefox for Android 79<br>Safari 16 🔑💎<br>Safari on iOS 16     |
| [`css.properties.offset-path`](https://developer.mozilla.org/docs/Web/CSS/offset-path#browser_compatibility)         |  🆕 Low  | 2022-03-15     | Chrome 55<br>Chrome Android 55<br>Edge 79<br>Firefox 72<br>Firefox for Android 79<br>Safari 15.4 🔑💎<br>Safari on iOS 15.4 |
| `css.properties.offset-path.path`                                                                                    |  🆕 Low  | 2022-09-12     | Chrome 64<br>Chrome Android 64<br>Edge 79<br>Firefox 72<br>Firefox for Android 79<br>Safari 16 🔑💎<br>Safari on iOS 16     |
| [`css.properties.offset-rotate`](https://developer.mozilla.org/docs/Web/CSS/offset-rotate#browser_compatibility)     |  🆕 Low  | 2022-09-12     | Chrome 56<br>Chrome Android 56<br>Edge 79<br>Firefox 72<br>Firefox for Android 79<br>Safari 16 🔑💎<br>Safari on iOS 16     |
[...]

PR #501, opened 2023-12-18
`media-capture`: set Baseline status

ddbeck:
We probably need a note field now, to explain why this will (never?) be non-Baseline.

> media-capture is an interesting feature: it's implemented in, and arguably more targeted at, mobile browsers, and using the HTML `capture` attribute on desktop should not break anything (using the IDL attribute may though, since it's going to be undefined).
>
> In a way, [`html.elements.input.capture`](https://developer.mozilla.org/docs/Web/HTML/Attributes/capture#browser_compatibility) could be seen as having a high baseline status. I don't suggest doing that. That just seems like a good candidate feature for statuses beyond baseline that might be worth calling out to developers: you can use the attribute everywhere, just be aware that it won't do anything on some platforms.

_Originally posted by @tidoust in https://github.com/web-platform-dx/web-features/issues/461#issuecomment-1839195349_

### DIFF
--- a/feature-group-definitions/media-capture.yml
+++ b/feature-group-definitions/media-capture.yml
@@ -1,5 +1,11 @@
 spec: https://w3c.github.io/html-media-capture/
 caniuse: html-media-capture
+status:
+  baseline: false
+  support:
+    chrome_android: "25"
+    firefox_android: "79"
+    safari_ios: "10"
 compat_features:
   - api.HTMLInputElement.capture
   - html.elements.input.capture


foolip:
@ddbeck can you add in the notes you think we should have here? We can just add it to the schema, and a `notes` field which is either a string or an array of strings (like BCD) seems sensible to me.

foolip:
https://github.com/web-platform-dx/web-features/pull/825 will take care of this.

PR #503, opened 2023-12-18
`line-clamp`: set Baseline status

ddbeck:
This probably also needs a note, to explain the somewhat weird story:

1. `-webkit-line-clamp` is standardized and interoperable.
2. `line-clamp`, unprefixed, will likely exist in the future but is _not_ available today (even on a limited basis, AFAICT).


foolip:
@ddbeck should we go ahead and add a notes feature?

foolip:
Closing in favor of https://github.com/web-platform-dx/web-features/pull/977.

PR #506, opened 2023-12-18
Add speech recognition and synthesis features

foolip:


foolip:
Ping @ddbeck

feature-group-definitions/speech-recognition.yml
```diff
@@ -0,0 +1,4 @@
+name: Speech Recognition
```
ddbeck, line 1:
```suggestion
name: Speech recognition
```

feature-group-definitions/speech-synthesis.yml
```diff
@@ -0,0 +1,4 @@
+name: Speech Synthesis
```
ddbeck, line 1:
```suggestion
name: Speech synthesis
```

ddbeck (approved):
A couple nits, but otherwise good.

PR #510, opened 2023-12-19
Mark WebTransport as supported in Firefox

foolip:
Parts of the feature aren't fully supported, but it makes sense to say it's supported since Firefox 114: https://developer.mozilla.org/en-US/docs/Web/API/WebTransport#browser_compatibility
https://www.mozilla.org/en-US/firefox/114.0/releasenotes/


foolip:
This solves the problem that the Baseline banner at the top of https://developer.mozilla.org/en-US/docs/Web/API/WebTransport is wrong.

ddbeck (approved):
Yeah, going against a prominent releases not entry seems… off.

Issue #498, opened 2023-12-14
Is accessibility support included?


aardrian:
What steps is Baseline taking to ensure that when it affirms "this feature works across the latest devices and browser versions" a developer can confidently take the feature and deploy it without the risk of creating barriers for users (or legal risk, depending on jurisdiction)?

## Background

I see nothing in the Baseline documentation [nor goals](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#goals) about how to test, track, nor report whether or not something is (at least) [accessibility supported](https://www.w3.org/TR/WCAG22/#dfn-accessibility-supported) (nor what level of support across versions and assistive technology / browser combinations).

This is absent from all the other platforms Baseline [intends to supplement](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#non-goals) / borrow from, so the data will generally not come from them.

The only mention of "accessibility" I found in any existing open or closed issues was from a [comment in May](https://github.com/web-platform-dx/web-features/issues/174#issuecomment-1544444392) and not thoroughly addressed in followup conversation. A pull request [comment in February](https://github.com/web-platform-dx/web-features/pull/60#issuecomment-1424197145) lays out the concern as well, but the [follow-up issue](https://github.com/web-platform-dx/web-features/issues/66) does not address it.

For example, I am struggling to find guidance on:
* What features have any accessibility vetting;
* What versions of assistive technology (AT) are considered for support (2 major releases?);
* What platform accessibility features are supported with these features;
* Bugs that trigger something to be inaccessible despite claims (see years of Safari breaking tables with display properties);
* etc.

Without accessibility as a pillar of the support charts, including testing details, this cannot be relied on by developers. This feels counter to improving the developer experience.

If I am simply terrible at finding it in the repo, please point me to it.

tidoust:
> This is absent from all the other platforms Baseline [intends to supplement](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#non-goals) / borrow from, so the data will generally not come from them.

The web-features project agglomerates support data from other sources. For support in core web browsers, web-features currently builds on [BCD](https://github.com/mdn/browser-compat-data), which in turn builds on sources such as browser release notes and more generally speaking has close ties with core browser vendors. The [core browser set](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#core-browser-set) in the baseline definition is also a pragmatic one, as in "browsers for which we can find support data".

[...]

PR #519, opened 2024-01-08
Clarify that Baseline does not cover assistive technology

tidoust:
This adds an item to non-goals to clarify that support in assistive tools that may be used along a browser is not (yet?) covered by Baseline.

See discussion in https://github.com/web-platform-dx/web-features/issues/498

### DIFF
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -60,7 +60,7 @@ Baseline status cannot or will not satisfy the following non-goals:
   If a web developer needs to support a globally uncommon or discontinued browser (e.g., Internet Explorer 11), then the developer needs to know the specific limitations of that browser, not a broad overview of developers’ most commonly required browsers.
   Baseline can’t be both.
 * **Identify support in assistive technology.**
-  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that browsers do not expose to developers.
+  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that is not built into browsers.
   See also: [Future considerations](#future-considerations).
 * **Identify support in non-web environments.**
   Developers use web technologies in non-web settings, such as JavaScript in Node.js, Web APIs in Deno, or HTML and CSS in Electron-based applications.
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -179,6 +179,6 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
 * Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
-* Not exposed to platform accessibility APIs (AAPIs) as specified
+* Support in assistive technology that is not built into browsers.
 * Obsolete/deprecated/legacy/etc. (i.e., flagged as such in the specification or dropped from newer versions of the specification)
 * Having high-quality polyfills available
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -60,7 +60,7 @@ Baseline status cannot or will not satisfy the following non-goals:
   If a web developer needs to support a globally uncommon or discontinued browser (e.g., Internet Explorer 11), then the developer needs to know the specific limitations of that browser, not a broad overview of developers’ most commonly required browsers.
   Baseline can’t be both.
 * **Identify support in assistive technology.**
-  Baseline should cover core accessibility support in browsers but it does not cover support in assistive tools (e.g., screen magnifiers, screen readers, alternative keyboards) that may be used along a browser to meet the requirements of people with disabilities.
+  Baseline does not cover support in assistive technology (AT), which includes screen readers, screen magnifiers, voice control, and so on.
   See also: [Future considerations](#future-considerations).
 * **Identify support in non-web environments.**
   Developers use web technologies in non-web settings, such as JavaScript in Node.js, Web APIs in Deno, or HTML and CSS in Electron-based applications.
@@ -178,7 +178,6 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Upcoming (e.g., not yet shipped in all browsers)
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
-* Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
-* Buggy or unsupported in assistive tools
+* Buggy (e.g., supported but in ways that are surprising and semi-interoperable, not exposed to platform accessibility APIs (AAPIs) as specified)
 * Obsolete/deprecated/legacy/etc. (i.e., flagged as such in the specification or dropped from newer versions of the specification)
 * Having high-quality polyfills available
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -178,6 +178,7 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Upcoming (e.g., not yet shipped in all browsers)
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
-* Buggy (e.g., supported but in ways that are surprising and semi-interoperable, not exposed to platform accessibility APIs (AAPIs) as specified)
+* Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
+* Not exposed to platform accessibility APIs (AAPIs) as specified
 * Obsolete/deprecated/legacy/etc. (i.e., flagged as such in the specification or dropped from newer versions of the specification)
 * Having high-quality polyfills available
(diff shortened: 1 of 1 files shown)


docs/baseline.md
```diff
@@ -59,6 +59,9 @@ Baseline status cannot or will not satisfy the following non-goals:
   Many Baseline features will not achieve 100% user reach soon or perhaps ever.
   If a web developer needs to support a globally uncommon or discontinued browser (e.g., Internet Explorer 11), then the developer needs to know the specific limitations of that browser, not a broad overview of developers’ most commonly required browsers.
   Baseline can’t be both.
+* **Identify support in assistive technology.**
+  Baseline should cover core accessibility support in browsers but it does not cover support in assistive tools (e.g., screen magnifiers, screen readers, alternative keyboards) that may be used along a browser to meet the requirements of people with disabilities.
```
aardrian, line 63: Suggested re-write since this is a non-goal (and an aspirational _should_ seems weird here):

"Baseline will not cover support in assistive technology (AT), which includes screen readers, screen magnifiers, voice control, and so on."

docs/baseline.md
```diff
@@ -176,5 +179,6 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
 * Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
+* Buggy or unsupported in assistive tools
```
aardrian, line 182: Because this is a future consideration and Baseline is not going to add AT support, I suggest something a bit more straightforward as a future goal (since it keeps the scope strictly with browsers and does not require AT):

"Confirm if features are exposed to platform accessibility APIs (AAPIs) as specified."

tidoust:
This touches the baseline definition so needs to be approved by current owners, flagged as reviewers accordingly.

docs/baseline.md
```diff
@@ -175,6 +178,6 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Upcoming (e.g., not yet shipped in all browsers)
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
-* Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
+* Buggy (e.g., supported but in ways that are surprising and semi-interoperable, not exposed to platform accessibility APIs (AAPIs) as specified)
```
jgraham, line 181: I think I'd make accessible a separate bullet point, since something could be fully interoperable for non-AT users and fully unsupported in AT (I hope we don't have features in that state, but I still think it makes sense to make "accessible" a top level item rather than a sub-case of generally buggy).

tidoust: Done. To be clear, I don't necessarily think that Baseline should have a dedicated state for that. I was more using the list to make explicit that exposure to platform accessibility APIs is one of the areas that should be taken into account once we have enough support data.

LeoMcA (approved):

docs/baseline.md
```diff
@@ -59,6 +59,9 @@ Baseline status cannot or will not satisfy the following non-goals:
   Many Baseline features will not achieve 100% user reach soon or perhaps ever.
   If a web developer needs to support a globally uncommon or discontinued browser (e.g., Internet Explorer 11), then the developer needs to know the specific limitations of that browser, not a broad overview of developers’ most commonly required browsers.
   Baseline can’t be both.
+* **Identify support in assistive technology.**
+  Baseline does not cover support in assistive technology (AT), which includes screen readers, screen magnifiers, voice control, and so on.
+  See also: [Future considerations](#future-considerations).
```
ddbeck, line 64: I find myself agreeing with @foolip's https://github.com/web-platform-dx/web-features/pull/519#pullrequestreview-1883062568, but also reading this text as in conflict with that understanding—or at very least, I don't know what I ought to do or not do with this non-goal.

Perhaps we can be explicit about the way in which this non-goal relates to developers' role in accessibility?

```suggestion
* **Identify support in assistive technology.**
  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that browsers do not expose to developers and their applications.
  See also: [Future considerations](#future-considerations).
```

(I've also omitted the initialism, since we don't reuse this term elsewhere in the document.)

tidoust: Ah, I don't mind changing the text but I read *your* suggested update as in conflict with @foolip's comment ;)

Much of the "accessibility tree" that @foolip mentions as *in scope* is internal to browsers I believe, exposed to AT for sure, but not exposed to web developers and their applications. At least not until the [Accessibility Object Model](https://wicg.github.io/aom/spec/) specs make further progress. With that in mind, I'm reading your suggestion as saying that if the browser does not expose the correct information in the accessibility tree, then that's *out of scope* for Baseline.

ddbeck: Haha OK, I guess I don't know quite how to capture this line of thinking:

Baseline should care about the consistency of the core browser set to the extent that developers are offered a facility to control/manage/influence _browser_ behavior.

Things that ought to be out-of-scope:

- Plug-in, operating system, hardware, or non-core browser set user agent limitations (e.g., a bug in JAWS or VoiceOver, lack of OS support for a given Braille display, etc.)
- Inaccessible browser behavior required or permitted by specification, provided that behavior is consistent with other browsers (e.g., if browsers, as a group, are doing things that are _wrong_ but correct and consistent, then we ought to be able to say that developers can rely on that behavior, even if we hope that developers will make a judgment to not use those things).
- Subjective variation between browsers in terms of the user interaction model or quality (we'll see this in other areas, like variations on when and how browsers prompt users for permissions).

(I think the accessibility tree is a slightly special case here in that browsers—rather consistently—show developers the tree in developer tools, which makes a sort of implicit agreement with them about the way pages ought to alter the tree but I also think that's a specific conversation we could have, whether or not we adopt a change to the definition.)

aardrian: @tidoust
> Much of the "accessibility tree" that @foolip mentions as in scope is internal to browsers I believe, exposed to AT for sure, but not exposed to web developers and their applications

The accessibility tree (which is itself an abstraction) is exposed to authors in dev tools and individual properties are exposed in assorted dev tools views. AOM is not directly related to this for today.

tidoust: Right @aardrian, but when I see "exposed to developer *and their applications*", I read interfaces, IDL, not dev tools views. That's what I'm reacting to here.

@ddbeck, perhaps dropping "and their applications" is enough to make things clear, or at least clear enough.

foolip: What @ddbeck wrote matches my thinking. I mentioned the accessibility tree, but we don't need to focus too much on that. If  keyboard navigation doesn't work because of a browser bug, that also counts. Whereas an OS bug is out of scope for Baseline, but could of course be relevant to mention in BCD notes, and in web-features notes if we ever add that.

ddbeck: I don't think there's any loss of clarity (and certainly no gain) if we drop the "and their applications" bit:

```suggestion
* **Identify support in assistive technology.**
  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that browsers do not expose to developers.
  See also: [Future considerations](#future-considerations).
```

docs/baseline.md
```diff
@@ -60,7 +60,7 @@ Baseline status cannot or will not satisfy the following non-goals:
   If a web developer needs to support a globally uncommon or discontinued browser (e.g., Internet Explorer 11), then the developer needs to know the specific limitations of that browser, not a broad overview of developers’ most commonly required browsers.
   Baseline can’t be both.
 * **Identify support in assistive technology.**
-  Baseline does not cover support in assistive technology (AT), which includes screen readers, screen magnifiers, voice control, and so on.
+  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that browsers do not expose to developers.
```
foolip, line 63: I might clarify it like this:

```suggestion
  Baseline does not cover support for screen readers, screen magnifiers, voice control, and other assistive technology that is not built into browsers.
```

I would also make it "Identify support in assistive technology outside of browsers" above.

captainbrosset: +1 to the suggested change.

atopal (approved):
LGTM

captainbrosset (approved):

tidoust:
@ddbeck This seems good to go. Anything blocking merge?

docs/baseline.md
```diff
@@ -176,5 +179,6 @@ Here are some areas for future consideration and not currently in-scope for Base
 * Progressive enhancement safe (i.e., limited penalties for support failures)
 * Developer feedback requested
 * Buggy (e.g., supported but in ways that are surprising and semi-interoperable)
+* Not exposed to platform accessibility APIs (AAPIs) as specified
```
foolip, line 182: To use the exact wording from above, now that it's gotten a lot of polish:

```suggestion
* Support in assistive technology that is not built into browsers.
```

ddbeck (approved):
I think @foolip's suggestion on the future considerations text is a good one. Happy to see this merge if it's applied. 👍

tidoust:
Applied and merged!

PR #523, opened 2024-01-12
Link to standards positions for View Transitions

foolip:
This isn't in the schema yet, this is to illustrate how I think we should do this.

### DIFF
--- a/feature-group-definitions/view-transitions.yml
+++ b/feature-group-definitions/view-transitions.yml
@@ -1,6 +1,9 @@
 spec: https://drafts.csswg.org/css-view-transitions-1/
 caniuse: view-transitions
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4383
+standards_positions:
+ - https://github.com/mozilla/standards-positions/issues/677
+ - https://github.com/WebKit/standards-positions/issues/48
 status:
   baseline: false
   support:


atopal:
Would it make sense to have a key with the vendor name?

foolip:
@atopal we can infer the vendor name easily from the URL, so I thought just a list of URLs would be easier to manage.

foolip:
@zcorpan I'm curious what you think about organizing Mozilla standards positions by feature. The approach I've taken here isn't especially great because it's then up to web-features maintainers to triage and categorize RFPs. Do you think this organization would be useful in the first place, and if so what would be a better way to maintain it?

FWIW, I have gone over ~5 years worth of chromestatus.com entries and tried to map those to web-features, and think the best approach for us would be to make web-features a field on chromestatus entries.

zcorpan:
I think this mapping seems potentially useful. cc @bgrins

Not sure yet about maintenance.

foolip:
I closed this since it's only an idea. I might file an issue if I want to spend time on it.

captainbrosset:
I would welcome a discussion around maintaining a collection of links per feature. Either by making other sources of data (chrome-status, standards position, bug trackers, ...) point to our web-features IDs, or by just adding the links in the web-features YAML files.

foolip:
@captainbrosset I've filed https://github.com/web-platform-dx/web-features/issues/687 about links from within web-features. I'd also welcome other data sources using web-features of course.

PR #526, opened 2024-01-16
Include container query units in the CQ feature

foolip:
This has the same support status as the existing entries.

### DIFF
--- a/feature-group-definitions/container-queries.yml
+++ b/feature-group-definitions/container-queries.yml
@@ -20,3 +20,4 @@ compat_features:
   - css.properties.container
   - css.properties.container-name
   - css.properties.container-type
+  - css.types.length.container_query_length_units


tidoust (approved):
That looks good to me.

Now, this made me wonder about the [`CSSContainerRule`](https://developer.mozilla.org/en-US/docs/Web/API/CSSContainerRule) interface. I dug the history and see it [got removed as non essential](https://github.com/web-platform-dx/web-features/pull/157/files/479723f9184c2e1a38a2392d30715b015df0da78#diff-4bc3c60f6bc12e7675fbb91e2c59548c344df788fab15a587777162cb65b2ff9). So that's another example of BCD keys left out on purpose, with some rationale somewhat hidden in pull request comments :)

It's also a good example of keys that should probably become part of the group down the road, especially since core browsers now support the interface and its members. Adding them now would change the date at which the feature group would be considered widely available, but adding them in a year or so would no longer change anything.

For what it's worth, adding just `api.CSSContainerRule` wouldn't change the support data actually. It's just if you add `api.CSSContainerRule.containerName` and `api.CSSContainerRule.containerQuery` that the versions are no longer the same.

This makes me wonder whether BCD data for `api.CSSContainerRule` is correct? Or perhaps the interface members just changed, I don't know the history there.

foolip:
@tidoust `containerName` and `containerQuery` were added after CQ shipped, tracked in https://chromestatus.com/feature/5159369837117440 for Chrome. The `CSSContainerRule` interface itself shipped together with the initial CQ feature however.

I think their omission is a very minor issue since the typical way of using any CSS is via the CSS parser, not CSSOM.

I'll file an issue about where they should go, however.

PR #540, opened 2024-01-25
Publish types in `web-features` package

ddbeck:
Including TypeScript type definitions makes it somewhat easier to consume `web-features` in TypeScript code (which I'm now doing in the process of generating burndown lists for web-features to BCD, caniuse, and MDN pages). Also, this was requested by @LeoMcA.

I'll self-review to highlight or explain some bits.

You can test this out by doing these steps:

1. Check out this branch.
2. Run `npm install`, `npm run build`, then `cd packages/web-features && npm pack`. This generates a zip file containing the built package, as if it were on npm.
3. Take that zip file somewhere else, run `npm install <zip>` and try using `import webFeatures from "web-features"`. Voilà: types.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -3,45 +3,7 @@ import path from 'path';
 
 import { fdir } from 'fdir';
 import YAML from 'yaml';
-
-/** Web platform feature */
-export interface FeatureData {
-    /** Alias identifier */
-    alias?: string | [string, string, ...string[]];
-    /** Specification */
-    spec: specification_url | [specification_url, specification_url, ...specification_url[]];
-    /** caniuse.com identifier */
-    caniuse?: string | [string, string, ...string[]];
-    /** Whether a feature is considered a "baseline" web platform feature and when it achieved that status */
-    status?: SupportStatus;
-    /** Sources of support data for this feature */
-    compat_features?: string[];
-    /** Usage stats */
-    usage_stats?: usage_stats_url | [usage_stats_url, usage_stats_url, ...usage_stats_url[]];  // A single URL or an array of two or more
-}
-
-type browserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
-
-type BaselineHighLow = "high" | "low";
-
-interface SupportStatus {
-    /** Whether the feature is Baseline (low substatus), Baseline (high substatus), or not (false) */
-    baseline?: BaselineHighLow | false;
-    /** Date the feature achieved Baseline low status */
-    baseline_low_date?: string;
-    /** Browser versions that most-recently introduced the feature */
-    support?: {[K in browserIdentifier]?: string};
-}
-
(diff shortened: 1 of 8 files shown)


index.ts
```diff
@@ -3,45 +3,7 @@ import path from 'path';
...
-/** Specification URL
- * @format uri
-*/
-type specification_url = string;
-
-/** Usage stats URL
- * @format uri
-*/
-type usage_stats_url = string;
+import { FeatureData } from './types';
```
ddbeck, line 6: I isolate the publicly-interesting types here, so I can drop them into the `web-packages` directory without having to awkwardly slice this source file up.

packages/web-features/package.json
```diff
@@ -13,14 +13,17 @@
...
   },
+  "types": "./index.d.ts",
   "files": [
+    "index.d.ts",
     "index.js",
     "index.json"
   ],
   "scripts": {
-    "prepare": "tsc"
+    "prepare": "tsc && rm types.js && tsup ./index.ts --dts-only --format=esm --out-dir=."
```
ddbeck, line 23: Ironically, TypeScript does not give us very good tools to generate type definitions. This might change in the future (there's a `tsc` option on the roadmap for TS 5.4), but for now I found a halfway decent tool — `tsup` — that does what I need: builds one `index.d.ts` file containing everything we need.

scripts/build.ts
```diff
@@ -3,16 +3,23 @@ import fs from 'fs';
...
     const json = stringify(features);
     // TODO: Validate the resulting JSON against a schema.
     const path = new URL('index.json', packageDir);
     fs.writeFileSync(path, json);
-    fs.copyFileSync(new URL("LICENSE.txt", rootDir), new URL("LICENSE.txt", packageDir));
+    for (const file of filesToCopy) {
+        fs.copyFileSync(new URL(file, rootDir), new URL(file, packageDir));
+    }
+    execSync("npm install", { cwd: "./packages/web-features"});
+    execSync("npm run prepare", { cwd: "./packages/web-features"});
```
ddbeck, line 22: Here I add the steps to generate the types. When this is done, all of the files needed for `npm pack` or `npm publish` to work are present in `packages/web-features`

packages/web-features/package.json
romainmenke: I've been using [`api-extractor`](https://api-extractor.com) for this.
I also wanted to have a single `.d.ts` file and found that `api-extractor` works well and also has other interesting features.

`tsup` is obviously fine aswel :)

foolip (approved):

packages/web-features/package.json
ddbeck: Thanks for highlighting this, @romainmenke. In the interest of shipping, I'll merge this as is, but `api-extractor` (and its docs) seem more approachable than `tsup`. If I ever have to make changes to how this works, I'll revisit swapping tools then.

Issue #490, opened 2023-12-07
Feature Request: Human Friendly Name for Feature Definitions


jcscottiii:
Requirement(s):
1. While processing a given [feature-group-definition](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions),
  it would be nice to have a human friendly name for the feature.

Current problem:
- Some feature files use kebab-case like `idle-detection`. And I could convert the kebab casing to title case (e.g. Idle Detection). But not all features do that.
- Some feature file names may be too vague. (e.g. grid.yml). And the desired name may not be derivable from the filename (e.g. would want to use `CSS Grid`)
- Some features have a caniuse field that uses kebab case. But not all features have that. And some may not convert correctly if applying kebab to title casing conversions. (e.g. `caniuse: css-grid` -> `Css Grid` is not desirable)

Instead of having multiple ways to derive the name, I think having a field like `full_name` would be helpful.

ddbeck:
Yes, I very much like this idea. 

I suspect we'll want to preserve the option of more than one reader-friendly field, along the lines of:

- Short name: "Grid", "Temporal", "`<dialog>`"
- Medium name: "CSS Grid", "Temporal API", "The `<dialog>` element"
- [Long name left as an exercise to the reader]
- Short description: "Two-dimensional page layout made up of rows and columns…", "A date and time API and successor to `Date`…", "An element that embeds an image in the page…"

We won't need these all at once, but I suspect we'll eventually want them all eventually (e.g., for sending new features to caniuse, for MDN to generate links between features, etc.).

Based on @jcscottiii's suggestion here, a medium-length name has an immediate application and is the place we should start?

foolip:
I think we should start with descriptions that disambiguate all features from each other, and would look sensible if presented as an alphabetical list of all features.

Like for identifiers, I think we should avoid prefixes and suffixes in most cases. A lot of time it would just be the identifier capitalized.

foolip:
I'm starting with short names in https://github.com/web-platform-dx/web-features/pull/545.

PR #545, opened 2024-01-26
Add a short name to the schema (required)

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/490.


feature-group-definitions/document-picture-in-picture.yml
```diff
@@ -1,2 +1,3 @@
+name: "Document Picture in Picture"
```
foolip, line 1: To match the spec:

```suggestion
name: "Document Picture-in-Picture"
```

feature-group-definitions/edit-context.yml
```diff
@@ -1 +1,2 @@
+name: "Edit Context"
```
foolip, line 1: To match the API name:

```suggestion
name: "EditContext"
```

feature-group-definitions/flexbox-gap.yml
```diff
@@ -1,2 +1,3 @@
+name: "Flexbox gap"
```
foolip, line 1: https://caniuse.com/flexbox-gap calls this "gap property for Flexbox" but https://web.dev/articles/bridging-the-gap#flexbox_gap is an example of "Flexbox gap" as a shorter form.

feature-group-definitions/constraint-validation.yml
```diff
@@ -1,3 +1,4 @@
+name: "Constraint Validation API"
```
foolip, line 1: Includes "API" because there's also constraint validation that happens automatically, not triggered by an API.

feature-group-definitions/fullscreen.yml
```diff
@@ -1,3 +1,4 @@
+name: "Fullscreen API"
```
foolip, line 1: Includes "API" because browsers can go fullscreen without the API too.

feature-group-definitions/navigation.yml
```diff
@@ -1 +1,2 @@
+name: "Navigation API"
```
foolip, line 1: Includes "API" because navigation is a more general concept.

feature-group-definitions/sanitizer.yml
```diff
@@ -1 +1,2 @@
+name: "Sanitizer API"
```
foolip, line 1: Hmm, I don't think this needs "API". The API entrypoint is the `Sanitizer` constructor.

```suggestion
name: "Sanitizer"
```

feature-group-definitions/scheduler.yml
```diff
@@ -1,3 +1,4 @@
+name: "Scheduler API"
```
foolip, line 1: API because scheduling is a browser-internal thing as well. Similar argument to Navigation API.

feature-group-definitions/svg2-script-html-equivalence.yml
```diff
@@ -1,3 +1,4 @@
+name: "SVG <script>: async & defer"
```
foolip, line 1: Actually no `:` I think:

```suggestion
name: "SVG <script> async & defer"
```

feature-group-definitions/webcodecs.yml
```diff
@@ -1,3 +1,4 @@
+name: "Web Codecs"
```
foolip, line 1: Aligning with https://developer.mozilla.org/en-US/docs/Web/API/WebCodecs_API:

```suggestion
name: "WebCodecs"
```

foolip (commented):
Self-review to point out some interesting bits.

feature-group-definitions/picture-in-picture.yml
```diff
@@ -1,3 +1,4 @@
+name: "Picture in Picture"
```
ddbeck, line 1:
```suggestion
name: "Picture-in-Picture"
```

feature-group-definitions/scheduler.yml
ddbeck: > browser-internal thing

Do we have any evidence that web developers know or care about this fact?

feature-group-definitions/scrollintoview.yml
```diff
@@ -1,3 +1,4 @@
+name: "scrollIntoView()"
```
ddbeck, line 1: If we wanted to give a context clue here…

```suggestion
name: "scrollIntoView() DOM method"
```

ddbeck (approved):
I think all of this is _fine_, so I'm approving it.

That said, I've made some suggestions, thinking about how we might want to use this field. I'm not strongly committed to these suggestions (or couldn't even settle one one, in some cases). So perhaps an approach here is to adopt the changes that you like and open issues for anything that we ought to document as a naming guideline/convention?

feature-group-definitions/anchor-positioning.yml
```diff
@@ -1 +1,2 @@
+name: "Anchor Positioning"
```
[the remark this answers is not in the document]
foolip: The inconsistent capitalization bugged me too, but I wasn't sure which way to go. Let me try using sentence capitalization consistently and see if any case looks wrong to me.

feature-group-definitions/array-at.yml
```diff
@@ -1,3 +1,4 @@
+name: "Array.prototype.at"
```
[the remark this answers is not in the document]
foolip: Yeah, I think that's a good rule of thumb, and am eager for a path that doesn't lead to `.prototype.` everywhere.

I also like using `<code>` and then we should also generate a plaintext variant.

feature-group-definitions/broadcast-channel.yml
```diff
@@ -1,3 +1,4 @@
+name: "Broadcast Channel"
```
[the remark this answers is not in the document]
foolip: Right? I searched for it too 🙃

feature-group-definitions/canvas-context-lost.yml
```diff
@@ -1,3 +1,4 @@
+name: "contextlost and contextrestored events"
```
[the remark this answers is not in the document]
foolip: In that case, is just "contextlost" not a good name? For CSS properties I haven't named thing "x property" after all.

feature-group-definitions/content-visibility.yml
```diff
@@ -1,3 +1,4 @@
+name: "content-visibility"
```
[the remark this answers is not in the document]
foolip: I think we should use the bare name as the shortest name where it's unambiguous. Consider a case like `text-wrap: balance`, I don't think we should call it "text-wrap: balance property value". Is that taking consistency too far? For some very short CSS property names it's an issue I think, like for `overflow`.

feature-group-definitions/media-capture.yml
```diff
@@ -1,3 +1,4 @@
+name: "Media Capture"
```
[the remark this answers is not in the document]
foolip: I think we should change the id of this feature too, I did have it confused with Media Capture and Streams!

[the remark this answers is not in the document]
foolip: Let's go with your first suggestion.

feature-group-definitions/scheduler.yml
foolip: No, I don't think most web developers think about how work is scheduled unless they specialize in performance. But it's like "paint", "layout" or "fetch". Browsers do these things, but there are also APIs for them.

feature-group-definitions/scrollintoview.yml
foolip: More in line with suggestions around Array might be:

```suggestion
name: "Element scrollIntoView() method"
```

But I wonder if we shouldn't save the more uniform names for a longer name field?

feature-group-definitions/default.yml
```diff
@@ -1,3 +1,4 @@
+name: ":default"
```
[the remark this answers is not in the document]
foolip: This becomes an issue with `:has()` I think. ":has()" is a fine name, but ":has() functional pseudo-class" is a bit much. Just the word "default" would have been ambiguous I think, but pseudo-classes are the only things that start with a `:` so I think it's OK for a shortest-ish name.

feature-group-definitions/array-at.yml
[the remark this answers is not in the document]
foolip: I think we'll need to revisit this, but I think the shortest that makes sense to me is:

```suggestion
name: "Array at()"
```

feature-group-definitions/array-flat.yml
```diff
@@ -1,3 +1,4 @@
+name: "Array.prototype.flat and Array.prototype.flatMap"
```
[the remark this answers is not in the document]
foolip: Following the pattern of "Array at()":

```suggestion
name: "Array flat() and flatMap()"
```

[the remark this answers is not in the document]
ddbeck: I like this last pattern. A template for single JS and Web API bits like this might be "Interface methodname/propertyname" in the absence of more widely used names.

(Apologies for the weekend commenting—not expecting a quick turnaround here. I've had a messy week, personally.)

feature-group-definitions/canvas-context-lost.yml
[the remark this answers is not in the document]
ddbeck: Yes, I think "contextlost and contextrestored" works for me here.

feature-group-definitions/content-visibility.yml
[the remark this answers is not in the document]
ddbeck: Yes, this works for me as a convention too. I'm going to write up a bunch of these in a recap comment.

feature-group-definitions/scrollintoview.yml
ddbeck: Yes, a longer name field is better for this. A StackOverflow search makes "scrollIntoView()" alone look sufficient to me.

ddbeck:
To recap, I think these are some initial guidelines to emerge from the various edits here:

## Short name guidelines

Above all else, observe these two principles:
- Prefer names that web developers would actually use. Ask yourself: could you imagine hearing this a meeting or written in an issue tracker?
- Be open to revising in response to developer use and feedback. Descriptive names are cheap to change, as compared to identifiers.

In the absence of more definitive names names, start from these guidelines:
- Prefer sentence case, unless you must match an interface name, specification title, or conventional casing and punctuation.
- Prefer ordinary descriptive nouns ("Idle detection") over interface names (~~"IdleDetector"~~), specification titles (~~"Idle Detection API"~~), or metonymy (~~"IdleDetector.start()"~~). This is a judgment call; use your head (and a StackOverflow and GitHub issue search).
- If in doubt, choose the shortest plausible phrasing, but be open to revising it when someone is confused or complains. Someday we'll add a long name field, where we can be pedantic and consistent.
- Adhere to long-standing conventions, even when they're silly (e.g., "referer" or "pseudo-class").
- Specific issues:
    - CSS: Let syntax and casing—such as `:` (for pseudos), `@` (for at-rules), and `kebab-case`—do the talking (e.g., ":indeterminate" not ~~":indeterminate pseudo-class"~~).
    - HTML: Tags can stand in for the concept of an element (e.g., "\" not ~~"\ element"~~ or ~~"Image element"~~).
    - JavaScript and Web APIs: If you must name a specific method, property, or other part of an interface, start with this template: `<interface> <property/method/etc.>` (e.g., "Array at()" not ~~"Array at() method"~~).
    - WebAPIs: Append "API" only if it's needed to avoid an ambiguity (e.g., "Navigation API" not ~~"Navigation"~~).

foolip:
Thank @ddbeck, that encapsulates well the decisions we landed on so far. A few thoughts:

> Prefer sentence case, unless you must match an interface name, specification title, or conventional casing and punctuation.

This is the hardest guideline to follow I think. Both specification title and conventional casing would lead us to View Transitions and Idle Detections, but I stuck with sentence case. The hard part is the inconsistency we'd have between these and a case like "Fetch priority", and how to draw the line.

I think we should list only the exceptions we've used, which so far is only matching API names, which are also written together, e.g., BroadcastChannel.

> Adhere to long-standing conventions, even when they're silly (e.g., "referer" or "pseudo-class").

Neither of these cases are at play in the names we've used. Do you have a case in mind where we'll need to use this part of the guideline?

feature-group-definitions/sticky-positioning.yml
```diff
@@ -1,3 +1,4 @@
+name: "Sticky positioning"
```
foolip, line 1: Need to check if `position: sticky` might be more common.

feature-group-definitions/webcodecs.yml
```diff
@@ -1,3 +1,4 @@
+name: "WebCodecs"
```
foolip, line 1: Oh right, this is the spec name, it's not code.

tidoust:
Side note: all entries in browser-specs have a [`shortTitle`](https://github.com/w3c/browser-specs#shorttitle) property. That could be a useful place to look at when a specification title is needed.

Most of the time, the property value is derived from the spec's title automatically but it is sometimes set manually, and we'd be happy to improve the values as needed.

The short title is typically used by Bikeshed when a spec references a section in another spec. For instance, the AVIF's spec full title is "AOMedia Film Grain Synthesis 1 (AFGS1) specification". Its short title is "AVIF", and `[[av1-avif#layer-selector-property]]` would thus render as "[AVIF § 2.3.2.2 Layer Selector Property](https://aomediacodec.github.io/av1-avif/#layer-selector-property)" in a spec.

foolip:
Thanks @tidoust, I didn't know about `shortTitle`. Looking over them, it looks like you're keeping "API" in more cases than we're leaning towards here. Do you have guidelines for which to use?

foolip:
I'm going to merge this now and filing a new issue for the guidlines.

tidoust:
> Thanks @tidoust, I didn't know about `shortTitle`. Looking over them, it looks like you're keeping "API" in more cases than we're leaning towards here. Do you have guidelines for which to use?

I don't think (or don't recall ;)) that we gave much thoughts about keeping or dropping "API". The `shortTitle` property was mainly introduced for CSS specs. It seems fine to me to favor dropping API when possible. I created https://github.com/w3c/browser-specs/issues/1177 to track this from a browser-specs perspective.

captainbrosset:
>   * CSS: Let syntax and casing—such as `:` (for pseudos), `@` (for at-rules), and `kebab-case`—do the talking (e.g., ":indeterminate" not ~":indeterminate pseudo-class"~).

I'm on the fence about this one because I always think about beginners. It feels much more beginner-friendly, from my opinion, to say things like "has pseudo-element" rather than ":has", or "dialog html element" rather than "<dialog>".

Note that it could be a mix of both: ":has pseudo" and "<dialog> element". But, asking people to recognize a tiny syntax character to know what we're talking about feels like an unnecessary barrier to me.

>   * HTML: Tags can stand in for the concept of an element (e.g., "" not ~" element"~ or ~"Image element"~).

What if we have an SVG element feature at some point. How would someone know if "<foo>" is SVG or HTML?

PR #550, opened 2024-01-29
Adjust names of some features

foolip:
Omit the "event" suffix and use "and" instead of "&".

Use sentence case.


ddbeck (approved):
All good changes. Thank you, @foolip!

PR #553, opened 2024-01-29
Upgrade to Node v18

ddbeck:
As of 2023-09-11, v16 is end-of-life. See https://github.com/nodejs/release#end-of-life-releases.

This fixes the deprecation warning in GitHub Actions and the `EBADENGINE Unsupported engine` warning from `tsup`.

One weird thing here: after upgrading Node.js, I could not figure out how to get `ts-node` to _not_ give me an error about unknown file extensions. Something about module resolution changed between Node versions and I am not smart or patient enough to figure out a way past that (while refusing to change all the file extensions). `tsx` ought to work exactly the same, maybe even slightly faster.


scripts/build.ts
```diff
@@ -18,8 +18,8 @@ function build() {
     for (const file of filesToCopy) {
         fs.copyFileSync(new URL(file, rootDir), new URL(file, packageDir));
     }
-    execSync("npm install", { cwd: "./packages/web-features"});
-    execSync("npm run prepare", { cwd: "./packages/web-features"});
+    execSync("npm install", { cwd: "./packages/web-features",  encoding: "utf-8"});
+    execSync("npm run prepare", { cwd: "./packages/web-features", encoding: "utf-8"});
```
ddbeck, line 22: This is so that errors thrown are readable. I threw a lot of errors getting this to work.

package.json
```diff
@@ -13,28 +13,28 @@
     "node": ">=16.18.0"
```
foolip, line 13: Was this intentionally left at 16? Is 16 really the oldest Node the published package works on?

ddbeck: No, this is an oversight. We should bump this to match `@types/node`.

```suggestion
    "node": ">=18.19.0"
```

foolip (approved):
This doesn't affect https://github.com/web-platform-dx/web-features/blob/main/packages/web-features/package.json, so looks good.

Issue #546, opened 2024-01-27
Map api.CSSContainerRule.containerName and containerQuery


foolip:
@tidoust raised this in https://github.com/web-platform-dx/web-features/pull/526.

In https://github.com/web-platform-dx/web-features/pull/157 we decided to omit `containerName` and `containerQuery` from the `container-queries` features, since it would change the Baseline date and they're not really important.

However, they are a part of Container Queries, they're not a standalone feature or a meaningful "v2" of Container Queries. We need to find a way to include them in `container-queries` without changing the Baseline date of the feature.

captainbrosset:
If we had a way to add notes to features, they could at least be _mentioned_ there. The note would say something similar to what you said: "containerName and containerQuery are part of this feature but not important enough to impact the baseline status of this feature. They weren't used in the computation of the baseline status".

We could also have notes that apply to only single BCD keys:

```yaml
- api.CSSContainerRule.containerName
  include: false
  note: "containerName is not included in the baseline computation ..."
```

foolip:
I've sent https://github.com/web-platform-dx/web-features/issues/546 to fix this.

I agree we'll need notes at some point and have filed https://github.com/web-platform-dx/web-features/issues/557. Nevertheless, I don't think we need a note for this, because it's unlikely to matter to the vast majority of developers. And explaining that only the CSSOM reflection doesn't work while the actual syntax works is not easy.

PR #556, opened 2024-01-31
Include CSSContainerRule in container queries but override support data

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/546.


ddbeck (approved):

ddbeck:
This feels much more straightforward than splitting the feature. Great!

PR #563, opened 2024-01-31
Add various new features

foolip:


feature-group-definitions/details-name.yml
```diff
@@ -0,0 +1,5 @@
+name: Mutually exclusive <details> elements
```
ddbeck, line 1: Huh, I didn't know about this until today.

ddbeck (approved):

feature-group-definitions/details-name.yml
foolip: It's new!

PR #565, opened 2024-02-01
Add feature for individual transform properties

foolip:
The shortest plausible name for this is "translate, scale, and rotate",
but these are concepts that appear also in DOMMatrix and 2D canvas, so
to disambigutate include "CSS" as a prefix. Also include "properties" as
a suffix to make it clear that it's no about the ability to
translate/scale/rotate, which the `transform` property has, but the
indivudal properties.

### DIFF
--- a/feature-group-definitions/individual-transforms.yml
+++ b/feature-group-definitions/individual-transforms.yml
@@ -1,0 +1,10 @@
+name: CSS translate, scale, and rotate properties
+spec: https://drafts.csswg.org/css-transforms-2/#individual-transforms
+usage_stats:
+  - https://chromestatus.com/metrics/css/timeline/popularity/504 # translate
+  - https://chromestatus.com/metrics/css/timeline/popularity/505 # rotate
+  - https://chromestatus.com/metrics/css/timeline/popularity/506 # scale
+compat_features:
+  - css.properties.rotate
+  - css.properties.scale
+  - css.properties.translate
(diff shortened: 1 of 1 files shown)


feature-group-definitions/individual-transforms.yml
```diff
@@ -0,0 +1,10 @@
+name: CSS translate, scale, and rotate properties
```
captainbrosset, line 1: These seem to be pretty well documented as "individual transform properties". How about using this as the name?

foolip: I agree, that's a common naming of these that I've seen. My only concern is that if you know about the `scale` property, you'd have no way of finding it in a list of feature names where it's called "individual transform properties".

Ultimately I think this is an argument for both a structured/normalized name which would probably be "CSS translate, scale, and rotate properties" in this case, and a description which can spell it out.

@ddbeck what name do you think we should pick here?

ddbeck: If this is well documented as "individual transform properties" then I think that wins. If we need the disambiguation or search-ability, perhaps we could append it in parentheses: "Individual transform properties (translate, scale, and rotate)"

But maybe we ought to get on with adding a description field, just for our own auto documenting practice.

```yaml
name: Individual transform properties
description: Transform elements with separate `translate`, `rotate`, and `scale` CSS properties.
```

foolip: Looks good to me, I've updated with your suggestion verbatim.

ddbeck (approved):

PR #566, opened 2024-02-01
Add CSS 2D/3D transform features

foolip:
Because transforms exist in multiple other parts of the platform
(DOMMatrix, canvas, SVG) prefix the names with "CSS".

### DIFF
--- a/feature-group-definitions/transforms2d.yml
+++ b/feature-group-definitions/transforms2d.yml
@@ -1,4 +1,5 @@
-name: CSS 2D transforms
+name: 2D transforms
+description: The `transform` CSS property and its 2D transform functions allow rotating, scaling, skewing, and translating an element. Arbitrary 2D transforms are also possible using a transformation matrix.
 caniuse: transforms2d
 spec: https://drafts.csswg.org/css-transforms-1/
 compat_features:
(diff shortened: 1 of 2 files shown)


feature-group-definitions/transforms2d.yml
```diff
@@ -0,0 +1,19 @@
+name: CSS 2D transforms
```
ddbeck, line 1: Given our recent changes to [the guidelines](https://github.com/web-platform-dx/web-features/tree/main/docs#feature-guidelines), what do you think of this approach instead?

```suggestion
name: 2D transforms (CSS)
```

feature-group-definitions/transforms3d.yml
```diff
@@ -0,0 +1,17 @@
+name: CSS 3D transforms
```
ddbeck, line 1:
```suggestion
name: 3D transforms (CSS)
```

feature-group-definitions/transforms2d.yml
foolip: Maybe we just put CSS in the description?

ddbeck: Yes, that would be good too. I almost suggested it, but I figured that mean signing myself up to write them… which looked hard. 😅

foolip:
@ddbeck I've added some descriptions, and it was hard. They could probably be improved.

Also I know we'll have some trouble when trying to compute a baseline status for 3D transforms, because parts of it like `backface-visibility` are quite new. Let's figure out splitting when we get to that.

ddbeck (approved):
I think these descriptions are pretty good! Let's go with it.

PR #567, opened 2024-02-01
Add new text-wrap balance/pretty features

foolip:


captainbrosset:
How are you thinking about the granularity of these features? I was under the impression that we'd pretty much never create a feature that points to just one BCD key. Would a single "text wrapping" feature make sense, instead of these two?
Additionally, if we're going to publish the algorithm for consumers to compute api/property-level baseline status, I wonder if that means those small groups are not needed. What do you think?

foolip:
If these had shipped at the same time I would probably make a single group. However, if [BCD](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap#browser_compatibility) is correct here, Chrome shipped `balance` first, Firefox has only shipped `balance`, and STP has `balance` but not `pretty` enabled. So these won't have the same Baseline status.

We shouldn't fall back to wrapping a single BCD entry as the default solution to every problem of this sort, but when the individual features are identifiable and make sense on their own, I think it's the right path. https://developer.chrome.com/docs/css-ui/css-text-wrap-balance points towards this feature standing well enough on its own.

ddbeck (approved):
Yeah, this seems fine to me.

I think single BCD entry features is a possible smell, but not necessarily a sign of triviality. Especially in CSS, where BCD's least rigorous about reflecting the "API" exposed to developers.

(If I had my way, BCD would have entries for each property and value, a bit like Web API interfaces and properties do. This would probably reduce the likelihood of single-entry features here.)

PR #570, opened 2024-02-01
Add text-spacing shorthand

foolip:


foolip:
This is a draft because most likely `text-spacing-trim` and `text-autospace` will not ship at the same time. That would mean that we can't treat them as a group together with `text-spacing`.

foolip:
Chrome is going to ship just `text-spacing-trim` first:
https://groups.google.com/a/chromium.org/g/blink-dev/c/jVUR2ebE3e0/m/xTYzlOkvBAAJ

A group for the shorthand won't make sense, at least not initially, so I'll rename this.

ddbeck (approved):

PR #575, opened 2024-02-02
Use new import attribute in `README`

tomayac:
https://v8.dev/features/import-attributes

### DIFF
--- a/packages/web-features/README.md
+++ b/packages/web-features/README.md
@@ -10,7 +10,7 @@ npm install web-features
 
 ```js
 import webFeatures from 'web-features';
-// import webFeatures from 'web-features/index.json' assert { type: 'json' }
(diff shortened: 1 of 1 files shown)


foolip:
Do you known what release of Node.js started to support the new syntax? If it's very recent we may need to put in a comment or something to help people figure out the error they might see.

tomayac:
According to the [Node.js docs](https://nodejs.org/api/esm.html#import-attributes), it's experimental and was added in v21.0.

foolip:
@tomayac I see. I just tested in Node.js 18, and just `import webFeatures from 'web-features'` works so I think we should just delete the commented-out code.

foolip:
https://github.com/web-platform-dx/web-features/pull/599

foolip:
Fixed by removing the code. Thanks for pointing it out @tomayac!

PR #577, opened 2024-02-02
Add style container queries

foolip:


foolip:
I found and linked https://caniuse.com/css-container-queries-style and see that both caniuse.com and the spec calls this container style query/queries, so I've matched that.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/22213 about the general `style()` support.

ddbeck (approved):

PR #580, opened 2024-02-03
Add JPEG XL feature

foolip:
Trust https://caniuse.com/jpegxl for the Safari versions.

### DIFF
--- a/feature-group-definitions/jpegxl.yml
+++ b/feature-group-definitions/jpegxl.yml
@@ -1,0 +1,8 @@
+name: JPEG XL
+spec: https://jpeg.org/jpegxl/workplan.html
+caniuse: jpegxl
+status:
+  baseline: false
+  support:
+    safari: "17"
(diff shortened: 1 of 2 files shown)


feature-group-definitions/jpegxl.yml
```diff
@@ -0,0 +1,8 @@
+name: JPEG XL
+spec: https://jpeg.org/jpegxl/workplan.html
```
tidoust, line 2: I'm looking into adding an entry for JPEG XL in browser-specs. That URL does not strike me as a good *spec* URL though, since it targets a page that lists several specs. The [file format spec](https://www.iso.org/standard/85253.html) seems a better entry point from a browser perspective: supporting the file format likely means you also support the underlying JPEG XL core coding system spec.

```suggestion
spec: https://www.iso.org/standard/85253.html
```

ddbeck (approved):

PR #581, opened 2024-02-03
Add <selectlist> draft feature

foolip:


foolip:
@ddbeck this is an interesting case. https://caniuse.com/selectlist already exists and there's excitement around this feature, but it doesn't have a spec yet, and I don't think we can be 100% sure the eventually shipping feature will even be a `<selectlist>` element. This is more bleeding edge than we've dealt with in web-features before. What to do?

captainbrosset:
> I don't think we can be 100% sure the eventually shipping feature will even be a `<selectlist>` element

Correct, there's been discussions about changing this to extend the `<select>` element instead. This is still very much in flux. If anything, this should probably be named something like "Stylable select".

feature-group-definitions/selectlist.yml
```diff
@@ -0,0 +1,3 @@
+name: "<selectlist>"
```
captainbrosset, line 1: The explainer has now been changed to `<select>` instead of `<selectlist>`: https://open-ui.org/components/selectlist/

ddbeck:
I can see that we might want to include not-yet implemented features, but I worry about maintenance and data quality issues.

BCD's policy is to not include features without implementations. One benefit of that approach is that proposals that go nowhere (or implementations that are removed) don't live on as zombie features. Instead, lack of compat data is a clear signal to either add it (in the case of real features) or dispense with content (in the case of unimplemented features).

If we're going to do something different, then I'd like to see an obvious mechanism to dispense with unmaintained features (e.g., an editorial field with an expiration date could allow us to automatically generate removal PRs, if there's no change in the support status).

I'd also like to ensure that we have some kind of _there_ there. I'd prefer specs, but we might have an allowlist for issue trackers to qualify such primordial features (e.g., WHATWG, tc39/proposals, etc.)

foolip:
I have turned this into a draft feature (and no longer a draft PR, yeah, that might get confusing).

ddbeck (approved):

PR #588, opened 2024-02-07
Bump web-specs from 2.79.0 to 3.0.0 and adjust code accordingly

tidoust:
New version of web-specs contains the ISO spec for JPEG XL (see #580), but adding that spec, which does not have a public version, required making a breaking change in web-specs, to allow entries without a `nightly` property.

This update adjusts the code that processes web-specs entries accordingly, and removes the JPEG XL exception as it is no longer needed.


foolip (approved):
Thanks @tidoust!

Issue #522, opened 2024-01-10
Question: will there be a baseline_high_date field?


jcscottiii:
I was looking at recent updates to the [schema](https://github.com/web-platform-dx/web-features/blob/main/schemas/defs.schema.json) file. Upon examination, I had a quick question: will there be a baseline_high_date? 

For example, indeterminate is already marked as [high](https://github.com/web-platform-dx/web-features/blob/dfcd560dd00299d338ae3cdbf013b84d892056f2/feature-group-definitions/indeterminate.yml#L5C1-L8).

```yaml
status:
  baseline: high
  baseline_low_date: 2020-01-15
  support:
....
```

But as a consumer, I would be curious when it transitioned to `high`.  

cc: @foolip @atopal

foolip:
MDN didn't need this initially, but I think we should provide it. @ddbeck WDYT?

atopal:
I think we should add that, otherwise consumers have to compute that form the definition, but we explicitly want to have flexibility to adjust the the widely available time frame, since 30 months are currently just a proxy for wide availability. If we don't provide that, we might end up with different representations depending on what definition consumers have implemented.

ddbeck:
The Baseline definition specifies the low date, but the high date exists only by implication. Just to be clear here: we're talking about a `baseline_high_date` that is (given our current definition) the Baseline low date + 30 months (as opposed to, for example, the date on which we set `baseline: "high"`).

foolip:
It would indeed be computed. I think the benefit is people not having to figure out how to add 30 months when the day of the month is 29-31... (we should probably do whatever Temporal does)

PR #593, opened 2024-02-12
Compute Baseline high date from low date

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/522.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -6,6 +6,10 @@ import YAML from 'yaml';
 import { FeatureData } from './types';
 import { Temporal } from '@js-temporal/polyfill';
 
+// Number of months after Baseline low that Baseline high happens. Keep in sync with definition:
+// https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#wider-support-high-status
+const monthsFromBaselineLowToHigh = 30;
+
 // Some FeatureData keys aren't (and may never) be ready for publishing.
 // They're not part of the public schema (yet).
 // They'll be removed.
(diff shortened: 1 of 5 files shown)


ddbeck (approved):

index.ts
```diff
@@ -32,7 +33,18 @@ for (const fp of filePaths) {
...
-    const data = YAML.parse(src);
+    const data = YAML.parse(src) as FeatureData;
+
+    // Compute Baseline high date from low date.
+    if (data.status?.baseline_high_date) {
+        throw new Error(`baseline_high_date is computed and should not be used in source YAML. Remove it from ${key}.yml.`);
+    }
+    if (data.status?.baseline === 'high') {
+        const lowDate = Temporal.PlainDate.from(data.status.baseline_low_date);
+        const highDate = lowDate.add({ months: 30 });
```
foolip, line 44: The behavior is to clamp to the end of the month if the day of the month in `lowDate` is higher than the number of days in the resulting months. This is different from what this code would do:

```js
d=new Date('2020-08-31')
d.setMonth(d.getMonth()+30)
```

That results in 2023-03-03, where Temporal will give us 2023-02-28. Doesn't matter much, but it's a nice property that all things in one month will map to the same future month.

captainbrosset, line 44: Should we put `30` in a const near the top of the file to avoid it being a magic number here?

`const NUMBER_OF_MONTHS_FROM_LOW_TO_HIGH = 30;`

foolip: Good idea, I added a `monthsFromBaselineLowToHigh` (matching the style of the other constant up there)

Issue #544, opened 2024-01-26
Publish code to compute Baseline status for BCD


foolip:
There's code in https://github.com/ddbeck/strict-browser-compat-data which we should publish as an NPM package for the benefit of @LeoMcA and other consumers.

We'll maintain it in this repo but publish it as a separate NPM package called `compute-baseline`.

It should export one method with a signature a bit like this:

```js
function computeBaselineB(bcdKey, bcdItself) {}
```

foolip:
We should return something like:

```typescript
{
  baseline: "low" | "high" | false,
  baseline_low_date: "string",
  baseline_high_date: "string",
  support: object /* TBD, depends on what the code does */
}
```

captainbrosset:
At our Baseline WebDX CG call [yesterday](https://docs.google.com/document/d/1ree75ImLZjf60lTZ3BhCaLHygxgywr7SBXp-q0xPs8A/edit#heading=h.o2eqe43z028w), we quickly talked about this, and one point that came up is the fact that consumers will, over time, likely have a wide range of programming language needs. Forcing people to use a JS function might not always work. One way would be to make the algorithm clear, and document it.

Another way could be, and this is me thinking out loud, to just generate all individual feature baseline statuses in our package. After all, this would just be a build step, not something we maintain. Our package would contain the baseline status for all of the logical web feature groups we create and baseline status for all of the indivudal bcd keys.

foolip:
Simply publishing a bunch of JSON was the first thing that @ddbeck and I considered, and we'll want to do that eventually, but publishing the code has some benefits. In particular, it avoids this package from having to be co-versioned with BCD. If you already depend on BCD (for example in https://github.com/mdn/yari) then you don't need to worry about contradictions between BCD and Baseline data derived from BCD.

tidoust:
Closing. Done and release as part of the [`compute-baseline` package](https://www.npmjs.com/package/compute-baseline)

PR #597, opened 2024-02-12
Migrate Baseline calculation code

ddbeck:
This moves the core BCD and Baseline bits from https://github.com/ddbeck/strict-browser-compat-data/ into web-features.

I **do not** expect this to land as written. Parts of this are messy (tests), there's lots of TODOs (everywhere), and there's a few bits (especially `supportStatements.ts`) that are too clever by half. And I gave this package an absolutely terrible name in the hopes of forcing a better one.

Rather, I expect this to be something we scavenge from, perhaps through incremental PRs, until there's nothing left to merge.

Toward https://github.com/web-platform-dx/web-features/issues/544


packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,228 @@
...
+interface SupportStatus {
+  compatKey?: string;
+  baseline: "low" | "high" | false;
+  baseline_low_date: string | null;
+  baseline_high_date: string | null;
+  support: Map<Browser, Release | undefined>;
+  toJSON: () => string;
+}
+
+export function computeBaseline(
```
foolip, line 26: In order to write tests for this we need a way to pass in the BCD data to use.

packages/web-features-figuring/src/browser-compat-data/browser.test.ts
```diff
@@ -0,0 +1,38 @@
+import assert from "node:assert/strict";
+
+import { browser } from "./browser";
```
foolip, line 3: This object will change as we upgrade BCD. In mdn-bcd-collector we use mock BCD data for tests. That does have downsides however, it means the mock data could get out of sync with the BCD schema.

A middle ground could be if we had a way to dump the data we're testing so that we would notice if it changed, and can review the changes. Most of the time it won't invalidate the tests.

https://jestjs.io/docs/snapshot-testing is the style of testing I have in mind, but I don't know how to to it with Mocha or get tooling support for updating such tests.

A more basic approach is to prepare a BCD object from the real BCD by deleting all the parts we don't want, and then having a single `assert.deepStrictEqual()` test where we show what the resulting BCD tree looks like. The messy part is updating that test, it would involve logging the tree in the failing test and the pasting it into the code. That's the part that proper tooling support handles.

Thoughts?

packages/web-features-figuring/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "web-features-figuring",
```
foolip, line 2: To avoid accidentally publishing this, which could well happen since it sits next to a package we will publish.

```suggestion
{
  "private": true,
  "name": "web-features-figuring",
```

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,228 @@
...
+      .pop();
+
+    assert(keystoneRelease instanceof Release);
+    baseline_low_date = keystoneRelease.date().toISOString().slice(0, 10);
+
+    if (isBaselineHigh) {
+      const date = keystoneRelease.date();
+      // TODO: Replace with Temporal
+      date.setMonth(date.getMonth() + 30);
+      baseline_high_date = date.toISOString().slice(0, 10);
```
foolip, line 93: I guess we should only compute `baseline_high_date` in one place. But if we're going to compute the main Baseline status as well, that really sugests that this package is where most of the computation should be, and not in https://github.com/web-platform-dx/web-features/blob/main/index.ts. We also want to be able to review the resulting `baseline_high_date` which we currently can't.

I think the model is that web-features-figuring is really our build script, which produces some artifacts (logs) that we commit, and some (the complete resulting JSON) that we don't.

ddbeck:
cc: @Elchi3 @LeoMcA this might be interesting to both of you too.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,228 @@
...
+      return jsonify(this);
+    },
+  };
+}
+
+function calculate(compatKey: string): SupportStatus {
+  const f = feature(compatKey);
+  const s = support(f, browsers);
+
+  const allReleases = f.supportedBy();
```
ddbeck, line 67: This should use the `only` option, so we don't let non-real values in non-Baseline browsers render the feature unresolvable.

See also: https://github.com/ddbeck/strict-browser-compat-data/pull/1

packages/web-features-figuring/src/baseline/core-browser-set.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function highReleases(compat: Compat) {
+  return browsers(compat)
+    .flatMap((b) => b.releases())
+    .filter(isBaselineHighRelease);
+}
+
+function isBaselineHighRelease(release: Release) {
+  const baselineHighCutoff = new Date();
+  baselineHighCutoff.setMonth(baselineHighCutoff.getMonth() - 30);
```
captainbrosset, line 30: nit: consider defining `30` as a shared const somewhere else, so that we can reuse it where needed, and document it in the code, somewhat like:

```js
// This is the number of months that Baseline considers enough for a feature to go from low to high.
const BASELINE_LOW_TO_HIGH_TIME = 30;
```

ddbeck: Fixed with 3cb4e4e7090993a830a36e1031d3b3e319051bb3 👍

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -9,6 +9,11 @@ import { support } from "./support";
 import { Browser } from "../browser-compat-data/browser";
 import { Compat, defaultCompat } from "../browser-compat-data/compat";
 
+export const BASELINE_HIGH_DURATION = Temporal.Duration.from({ months: 30 });
```
foolip, line 12: We have this constant in the top-level index.ts too. Can it be shared?

ddbeck: Yes, done with a01258d. Though I imagine needing to do so will be temporary, until we can call `computeBaseline` from `index.ts` instead.

packages/web-features-figuring/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { Temporal } from "@js-temporal/polyfill";
+
+// Number of months after Baseline low that Baseline high happens. Keep in sync with definition:
+// https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#wider-support-high-status
+export const BASELINE_LOW_TO_HIGH_DURATION = Temporal.Duration.from({
+  months: 30,
+});
+
+export const VERY_FAR_FUTURE_DATE = Temporal.Now.plainDateISO().add({
+  years: 100,
```
foolip, line 10: Oh no, a Y2124 problem in the making! :)

I think this is fine, but will look if any uses of `VERY_FAR_FUTURE_DATE` can be avoided.

foolip: And a comment for what this constant (not a constant?)  is used for would be good.

packages/web-features-figuring/src/baseline/core-browser-set.ts
```diff
@@ -0,0 +1,44 @@
...
+function isBaselineHighRelease(release: Release) {
+  const baselineHighCutoff = Temporal.Now.plainDateISO().subtract(
+    BASELINE_LOW_TO_HIGH_DURATION,
+  );
+  return (
+    Temporal.PlainDate.compare(
+      release.date() ?? VERY_FAR_FUTURE_DATE,
+      baselineHighCutoff,
+    ) < 1
+  );
```
foolip, line 43: Would this be equivalent?

```suggestion
  return release.date() && Temporal.PlainDate.compare(release.date(), baselineHighCutoff) <= 0;
```

Or better still, add `if (!release.date()) return false;` at the top of the function and simplify more.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  let baseline_low_date = null;
+  let baseline_high_date = null;
+  if (isBaselineLow) {
+    const initialReleases = [...s.entries()].map(([, r]) => r);
+    const keystoneRelease = initialReleases
+      .sort((a, b) => {
+        assert(a !== undefined);
+        assert(b !== undefined);
+        return Temporal.PlainDate.compare(
+          a.date() ?? VERY_FAR_FUTURE_DATE,
```
foolip, line 92: Rather than dealing with dateless releases in many places, can we just filter out those releases in a central place? They can never have an effect on baseline calculation and just get in the way. Same with "preview" and perhaps some other releases.

packages/web-features-figuring/src/baseline/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const result = computeBaseline(
+      {
+        compatKeys: ["css.properties.border-color"],
+        checkAncestors: false,
+      },
+      compat,
+    );
+    assert.equal(result.baseline, "high");
+  });
```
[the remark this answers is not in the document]
foolip: One of the snapshots is 120k lines, that's a lot, it won't even load here on GitHub. That also means we won't really review changes to it.

It's very useful to know if a change will affect BCD in some way, but perhaps we can test that in some other way? Just rely on the planned log files and the change or lack of change we see in the computed statuses we see?

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+
+function calculate(compatKey: string, compat: Compat): SupportStatus {
+  const f = feature(compatKey);
+  const s = support(f, browsers(compat), compat);
+
+  const allReleases = f.supportedBy({ only: browsers(compat), compat });
+  const isBaselineLow = lowReleases(compat).every((r) =>
+    allReleases.includes(r),
+  );
+  const isBaselineHigh = highReleases(compat).every((r) =>
```
foolip, line 76: Instead of keeping track of which releases contribute to Baseline high, could we instead use the `baseline_low_date`, add `BASELINE_LOW_TO_HIGH_DURATION` and say that it's Baseline high if that date is not in the future?

We don't seem to use `highReleases` for anything other than this, and I _think_ this would mean less code. Ignore me if it would actually be more code.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      return jsonify(this);
+    },
+  };
+}
+
+function calculate(compatKey: string, compat: Compat): SupportStatus {
+  const f = feature(compatKey);
+  const s = support(f, browsers(compat), compat);
+
+  const allReleases = f.supportedBy({ only: browsers(compat), compat });
```
foolip, line 72: IIUC, this returns all of the releases that support the feature. Then we check if the latest browser releases (`lowReleases`) are a subset of this `allReleases`. Some comments to explain the rules would be helpful, since there are quite a few equivalent ways of crunching this data.

packages/web-features-figuring/src/baseline/support.ts
```diff
@@ -0,0 +1,23 @@
+import { Browser } from "../browser-compat-data/browser";
+import { Compat } from "../browser-compat-data/compat";
+import { Feature } from "../browser-compat-data/feature";
+import { Release } from "../browser-compat-data/release";
+
+type Support = Map<Browser, Release | undefined>;
+
+export function support(feature: Feature, browsers: Browser[], compat: Compat) {
```
foolip, line 8: Maybe TS infers the type, but as a reader I wasn't so sure. This would help me:

```suggestion
export function support(feature: Feature, browsers: Browser[], compat: Compat): Support {
```

packages/web-features-figuring/src/baseline/support.ts
```diff
@@ -0,0 +1,23 @@
...
+  const support: Support = new Map();
+  for (const b of browsers) {
+    support.set(b, undefined);
+  }
+
+  for (const rel of feature.supportedBy({ only: browsers, compat })) {
+    const currRel = support.get(rel.browser);
+
+    if (!currRel || currRel.compare(rel) >= 0) {
```
foolip, line 17: What is `currRel` and `rel`? The absence of `compareVersions` anywhere makes me suspect that a bug is on this line, if these are strings.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+    allReleases.includes(r),
+  );
+
+  const baseline =
+    (isBaselineHigh ? "high" : false) || (isBaselineLow ? "low" : false);
+
+  let baseline_low_date = null;
+  let baseline_high_date = null;
+  if (isBaselineLow) {
+    const initialReleases = [...s.entries()].map(([, r]) => r);
```
foolip, line 86: Untested, but I think `s` is a Set and this would do the same?

```suggestion
    const initialReleases = [...s.values()];
```

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+    const keystoneRelease = initialReleases
+      .sort((a, b) => {
+        assert(a !== undefined);
+        assert(b !== undefined);
+        return Temporal.PlainDate.compare(
+          a.date() ?? VERY_FAR_FUTURE_DATE,
+          b.date() ?? VERY_FAR_FUTURE_DATE,
+        );
+      })
+      .pop();
```
foolip, line 96: The same without modifying the `initialReleases`:

```suggestion
      .at(-1);
```

packages/web-features-figuring/src/browser-compat-data/release.ts
```diff
@@ -0,0 +1,54 @@
...
+  date(): Temporal.PlainDate | null {
+    const { release_date } = this.data;
+    if (release_date === undefined) {
+      return null;
+    }
+
+    return Temporal.PlainDate.from(release_date);
+  }
+
+  compare(otherRelease: Release) {
```
foolip, line 30: Hmm, does this make comparison operators work in TypeScript?

packages/web-features-figuring/src/browser-compat-data/browser.ts
```diff
@@ -0,0 +1,84 @@
...
+        // This shouldn't happe in practice, but we must appease the type checker
+        throw Error(`${this} doesn't have releases data. That's weird.`);
+      }
+
+      this._releases = [];
+      for (const [key, value] of Object.entries(this.data.releases)) {
+        this._releases.push(new Release(this, key, value));
+      }
+
+      this._releases.sort((a, b) =>
```
foolip, line 50: Does this do the same as sorting the version strings with `compareVersions`? Does BCD not record any patch release that came after a major release? Looking for corner cases 😈

packages/web-features-figuring/src/browser-compat-data/release.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const index = this.browser.releases().indexOf(this);
+    const otherIndex = this.browser.releases().indexOf(otherRelease);
+
+    if (index < otherIndex) {
+      return -1;
+    } else if (index > otherIndex) {
+      return 1;
+    }
+    return 0;
```
foolip, line 45:
```suggestion
    return index - otherIndex;
```

packages/web-features-figuring/src/browser-compat-data/release.ts
```diff
@@ -0,0 +1,54 @@
+import { Temporal } from "@js-temporal/polyfill";
+import { ReleaseStatement } from "@mdn/browser-compat-data";
+
+import { Browser } from "./browser";
+
+export class Release {
```
foolip, line 6: I like this helper! A release as a tuple of browser and version string is very convenient, and I've written some pretty inconvenient code passing around the two as a pair.

foolip:
I've done some reviewing, but will have to keep going later. This is going to be great once it's churning all of the data for us.

foolip:
Motivation to get this done: https://twitter.com/wesbos/status/1771239902782398709

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  support: Map<Browser, Release | undefined>;
+  toJSON: () => string;
+}
+
+export function computeBaseline(
+  featureSelector: FeatureSelector,
+  compat: Compat = defaultCompat,
+): SupportStatus {
+  const { compatKeys } = featureSelector;
+  const keys = featureSelector.checkAncestors
```
foolip, line 35: AFAICT there are no tests for when `checkAncestors` is true. I'm not sure when it would change the result either since BCD has lints for subfeatures not predating their parent features. Perhaps with flags that aren't cascaded the way I'd like them to be?

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+        assert(a !== undefined);
+        assert(b !== undefined);
+        return Temporal.PlainDate.compare(
+          a.date() ?? VERY_FAR_FUTURE_DATE,
+          b.date() ?? VERY_FAR_FUTURE_DATE,
+        );
+      })
+      .pop();
+
+    assert(keystoneRelease instanceof Release);
```
foolip, line 98: If we ever take the `VERY_FAR_FUTURE_DATE` case below things have gone wrong. Can we guard against it by simplifying this assert?

```suggestion
    assert(keystoneRelease?.date());
```

Or just `keystoneRelease.date()` if TS doesn't think `keystoneRelease` could be null.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  for (const { baseline } of statuses) {
+    if (baseline === false) {
+      return false;
+    }
+    if (baseline === "low") {
+      bestOf = "low";
+    }
+    if (baseline === "high" && bestOf === "high") {
+      bestOf = "high";
+    }
```
foolip, line 152: This can be omitted since `bestOf` is initially "high".

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+    const data = compat.query(current);
+    if (typeof data === "object" && data !== null && "__compat" in data) {
+      ancestors.push(current);
+    }
+  }
+  return ancestors;
+}
+
+function minStatus(statuses: SupportStatus[]): "high" | "low" | false {
+  let bestOf: "high" | "low" = "high";
```
foolip, line 142: Here, let's `assert(statuses.length)` since otherwise we have to deal with the case of an empty array.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      bestOf = "low";
+    }
+    if (baseline === "high" && bestOf === "high") {
+      bestOf = "high";
+    }
+  }
+  return bestOf;
+}
+
+function minSupport(
```
foolip, line 157: Can you document the intended behavior of this helper? It looks like it will find the earliest supported version even in ship-remove-ship situations. Some code might need that, but not when computing Baseline, right?

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  let baseline_low_date = null;
+  if (baseline !== false) {
+    baseline_low_date = latestDate(
+      statuses.map((status) => status.baseline_low_date),
+    );
+  }
+
+  let baseline_high_date = null;
+  if (baseline === "high") {
+    baseline_high_date = latestDate(
```
foolip, line 52: Can't we compute the aggregate high date from the aggregate low date? That seems simpler to me, less things to debug if something isn't working as expected.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  const sorted = dates
+    .map((date) => {
+      assert(date !== null);
+      return Temporal.PlainDate.from(date);
+    })
+    .sort(Temporal.PlainDate.compare);
+
+  assert(sorted.length >= 1);
+
+  return sorted.at(-1)?.toString() ?? null;
```
foolip, line 198: This works, and I'm sure it's fast enough, but it can be both simpler and faster to just keep track of a min/max in one loop, O(n) instead of O(n log(n)). Since dates are strings of the same format, it actually works fine to compare then as strings too. What I'd suggest:

```suggestion
  let latest = null;
  for (const date of dates) {
    if (date === null) {
      return null;
    }
    assert(date.length === 10);
    if (!latest || date > latest) {
      latest = date;
    }
  }
  return latest;
```

packages/web-features-figuring/src/browser-compat-data/supportStatements.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  get version_removed() {
+    return super.version_removed as string | false;
+  }
+
+  // TODO: `supportedBy()` ought to be (partially) implemented on non-real value
+  // support statements. For example, `"version_added": true` should allow for
+  // returning `[this.browser.current()]` at least.
+  supportedBy() {
```
foolip, line 151: Should this and `hasCaveats` move into the baseline directory, to keep this directly as an unopinionated wrapper around BCD?

packages/web-features-figuring/src/browser-compat-data/typeUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+  Browsers,
+  BrowserStatement,
+  CompatData,
+  CompatStatement,
+  Identifier,
+  MetaBlock,
+} from "@mdn/browser-compat-data";
+
+export function isIndexable(o: unknown): o is Record<string, unknown> {
+  if (typeof o === "object" && o !== null && Object.keys(o).length > 0) {
```
foolip, line 11: Why is the `Object.keys(o).length` check needed here? Shouldn't `{}` be considered indexable even though it has no keys?

I thought this might be to handle `Map`, but `Object.keys()` for a map returns an empty array.

foolip:
@ddbeck I think I've looked at the most important bits here now, sorry it trickled in over many days and many comments. A bunch of very minor questions and suggestions.

The main thing I think we'll need to improve soon (not now) is the handling the "caveats" category that requires human review and judgment. Those are now logged, but not in a form where we could output them to the kind of per-feature compute log file we've been discussing. We'll need this to efficiently review a computed status and know when to dig deeper.

The approach of expanding support statements and then checking release-by-release has some pros and cons. It could make it easier to understand why something isn't Baseline if it's just due to a single missing release that's not obvious from looking at the statements array. It's also closer to the caniuse model which might help us in the future with release-associated notes. The main con is the transformation back and forth feels a bit wasteful, but if it works it works!

Let's get this landed and then wired up to generate a status for some feature :)

packages/web-features-figuring/src/baseline/core-browser-set.ts
ddbeck: Two good points. Getting the type checking to play nice required some small adjustments. Done in 62113328b75a3827fa29856be799d7513374c98c.

packages/web-features-figuring/src/browser-compat-data/browser.ts
foolip: Great issue, thank you! Let's fix it here by using `compare-versions` since that is what mdn-bcd-collector does, so a lot of the data is built around this assumption. If we find that it doesn't match the date order in some case, we'll need to sort that out, but not in this PR.

packages/web-features-figuring/src/browser-compat-data/typeUtils.ts
ddbeck: Good catch. Removed the length check to no ill effect.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    : compatKeys;
+
+  const statuses = keys.map((key) => calculate(key, compat));
+  const support = collateSupport(statuses.map((status) => status.support));
+
+  const keystoneDate = findKeystoneDate(
+    statuses.flatMap((s) => [...s.support.values()]),
+  );
+  const { baseline, baseline_low_date, baseline_high_date } =
+    keystoneDateToStatus(keystoneDate);
```
ddbeck, line 50: This is the biggest, most important change: I now follow the Baseline definition more literally. First we find the ostensible keystone date, then use that to determine if it's low, high, or false.

This is nice because it works the same way for the individual compat keys as it does for the final summation.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,241 @@
...
+ * Compute the Baseline support ("high", "low" or false, dates, and releases)
+ * for a single compat key.
+ */
+function calculate(compatKey: string, compat: Compat): SupportStatus {
+  const f = feature(compatKey);
+  const s = support(f, browsers(compat));
+  const keystoneDate = findKeystoneDate([...s.values()]);
+
+  const { baseline, baseline_low_date, baseline_high_date } =
+    keystoneDateToStatus(keystoneDate);
```
ddbeck, line 73: Reusing the same logic as above, to find the keystone date and status for a single compat key.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+  return ancestors;
+}
+
+/**
+ * Collate several support summaries, taking the most-recent release for each browser across all of the summaries.
+ */
+function collateSupport(
+  supports: Map<Browser, Release | undefined>[],
+): Map<Browser, Release | undefined> {
```
ddbeck, line 114: This was formerly `minSupport`, now renamed (and commented) to more clearly describe what's going on here.

packages/web-features-figuring/src/baseline/index.ts
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * Given several dates, find the most-recent date and determine the
+ * corresponding Baseline status and high and low dates.
+ */
+function keystoneDateToStatus(date: Temporal.PlainDate | null): {
+  baseline: BaselineStatus;
+  baseline_low_date: BaselineDate;
+  baseline_high_date: BaselineDate;
+} {
```
ddbeck, line 147: A brand new method to find the status, which sidesteps the Set difference stuff.

packages/web-features-figuring/src/baseline/support.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const unqualifiedReleases = [];
+    const qualifiedReleases = [];
+    for (const { release, qualifications } of releases) {
+      if (qualifications === undefined) {
+        unqualifiedReleases.push(release);
+      } else {
+        qualifiedReleases.push({ release, qualifications });
+      }
+    }
```
ddbeck, line 26: It's not _super_ obvious here, but now `supportedBy()` returns `{ release, qualifications? }[]`, where qualifications are things like prefixes and partials. If a feature is supported in two support statements, then you'll get two `{ release, qualifications? }`. So, for example, if there's data showing support with both a prefix and without, then you'll get both `{ release, qualifications }` and `{ release }`. Tidy!

(I've just rejected the releases that have qualifications. We could get more clever with this, if we like.)

packages/web-features-figuring/src/baseline/support.ts
```diff
@@ -0,0 +1,89 @@
...
+    const qualifications: string[] = Array.from(
+      new Set(
+        releases.flatMap(({ release, qualifications }) =>
+          Object.keys(qualifications),
+        ),
+      ),
+    );
+    console.warn(
+      `${feature}: ${browser} has ${releases.length} releases deemed unsupported due to ${qualifications.join(", ")}. See underlying data for details.`,
+    );
```
ddbeck, line 87: I'm still just printing warnings—not sure where we want the warnings to actually go just yet.

ddbeck (commented):
@foolip Thank you for your comments. I think I've managed to address everything* you've mentioned and there's now much less code! 🎉  I've done a self-review to highlight the most important changes.

\*except the `withAncestors` test. I don't know of a feature where that would make a difference just yet.

packages/web-features-figuring/src/browser-compat-data/supportStatements.ts
```diff
@@ -168,13 +162,34 @@ export class RealSupportStatement extends SupportStatement {
...
 
+    let releases;
     if (this.version_removed === false) {
-      return this.browser.releases().filter((rel) => rel.compare(start) >= 0); // Release is on or after start
+      releases = this.browser.releases.filter((rel) => rel.compare(start) >= 0); // Release is on or after start
+    } else {
+      const end: Release = this.browser.version(this.version_removed);
+      releases = this.browser.releases.filter(
+        (rel) => rel.compare(start) >= 0 && rel.compare(end) < 0,
+      ); // Release is on after start and before the end
```
foolip, line 172:
```suggestion
      ); // Release is on or after start and before the end
```

packages/web-features-figuring/src/browser-compat-data/supportStatements.ts
```diff
@@ -168,13 +162,34 @@ export class RealSupportStatement extends SupportStatement {
...
+    }
+    if (this.data.alternative_name) {
+      qualifications.alternative_name = this.data.alternative_name;
+    }
+    if (this.flags.length) {
+      qualifications.flags = this.flags;
+    }
+    if (this.partial_implementation) {
+      qualifications.partial_implementation = this.partial_implementation;
+    }
```
foolip, line 187: This amounts to a filter of `this.data` for specific keys. But that's OK, it works.

foolip:
Also, please rename the package to `compute-baseline` as discussed on chat. Then 🚀

packages/web-features-figuring/src/browser-compat-data/browser.ts
```diff
@@ -20,11 +23,24 @@ export function browser(id: string, compat = defaultCompat): Browser {
...
   constructor(id: BrowserName, data: BrowserStatement) {
     this.id = id;
     this.data = data;
+
+    const sortedReleaseData: [version: string, data: ReleaseStatement][] =
+      Object.entries(data.releases).sort(([a], [b]) => compareVersions(a, b));
+
+    const releases = sortedReleaseData.map(
+      ([version, data], index) => new Release(this, version, data, index),
+    );
```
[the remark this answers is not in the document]
ddbeck: I'm going to leave this one for now. This is one of those places where TS makes the "simpler" thing more complex, because it can't follow unwinding of `version` from `data.releases`.

ddbeck:
I'm merging this now. Thanks to everyone who looked at any part of this.

Next week, I'll wire this up to write YAML. Onward!

PR #598, opened 2024-02-12
Add optional descriptions as internal (for now) documentation

foolip:


feature-group-definitions/input-event.yml
```diff
@@ -1,4 +1,5 @@
 name: input event
+description: The `input` event is fired when a form control (e.g., `<input>`, `<select>`, or `<textarea>`) or any element with the `contenteditable` attribute is updated.
```
ddbeck, line 2: Maybe putting this in an imperative would help us be more clear about what developers are going to do with it?

> Capture changes to form controls and `contenteditable` elements.

Or with a lighter touch, using the active voice:

> The `input` event fires when a form control changes or an element with the `contenteditable` attribute changes.

feature-group-definitions/webvtt.yml
```diff
@@ -1,4 +1,5 @@
 name: WebVTT
+description: File format for text tracks (captions or subtitles) for video. Added using the `<track>` element.
```
ddbeck, line 2: As an imperative:

> Provide captions or subtitles for video content with the WebVTT file format and `<track>` elements.

Active voice:

> WebVTT is a captions or subtitles file format for `<track>` elements.

ddbeck (approved):
Out of band, you mentioned wanting to speak more closely to developers in these descriptions. Left some ideas for you, but completely optional.

feature-group-definitions/input-event.yml
foolip: I like both of these better than what I wrote myself. The imperative from the developer's point of view makes a lot of sense to me, but I worry that it's a bit novel and will turn out to be messy.

There's a lot of precedent for the active voice in HTML's green "domintro" boxes which are meant for web developers. Example from https://html.spec.whatwg.org/multipage/media.html#seeking:

> Seeks to near the given time as fast as possible, trading precision for speed.

That works well in the imperative form too:

> Seek to near the given time as fast as possible, trading precision for speed.

However, I think we'll have some cases where naming the API will make it clearer, and in those cases I think the active voice works better:

> The `fastSeek()` method seeks to near the given time as fast as possible, trading precision for speed.

These possible imperative forms are awkward:

> Use the `fastSeek()` method to seek to near the given time as fast as possible, trading precision for speed.

> Seeks to near the given time as fast as possible using the `fastSeek()` method, trading precision for speed.

When we decide to publish descriptions, perhaps our guideline will turn out to be sometimes imperative and sometimes active voice, but for now I'll go with your suggestion for active voice:

```suggestion
description: The `input` event fires when a form control changes or an element with the `contenteditable` attribute changes.
```

feature-group-definitions/webvtt.yml
foolip: Let's go with:

```suggestion
description: WebVTT is a captions and subtitles format. WebVTT files are loaded using the `<track>` element, and the `VTTCue` API can be used to create or update cues dynamically.
```

PR #601, opened 2024-02-15
Add comments for compression streams use counters

foolip:


feature-group-definitions/compression-streams.yml
```diff
@@ -1,8 +1,8 @@
 name: Compression streams
 spec: https://wicg.github.io/compression/
 usage_stats:
-  - https://chromestatus.com/metrics/feature/timeline/popularity/3060
-  - https://chromestatus.com/metrics/feature/timeline/popularity/3061
+  - https://chromestatus.com/metrics/feature/timeline/popularity/3060 # CompressionStream
```
ddbeck, line 4: A non-blocking question: does Chrome Platform Status offer a way to get identifiers for these URLs, without scraping the page? If I'm going to generate "dist" versions of the YAML for review, maybe this is something we can pull in automatically.

ddbeck (approved):

feature-group-definitions/compression-streams.yml
foolip: I'm not sure, but the names are often idiosyncratic and style varies. If we wanted labels for use counters we'd either need to just accept the odd names, or add our own.

PR #602, opened 2024-02-15
Add aspect-ratio feature

foolip:


foolip:
A reviewer might wonder why we should bund the aspect-ratio-from-width-and-height feature with the `aspect-ratio` CSS property, if predates the CSS property? I think it makes sense because now that feature is defined as a "presentation hint" which does use the `aspect-ratio` property. Minimal demo in https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=12379.

PR #603, opened 2024-02-15
Add COLRv1 color font format

foolip:


foolip:
@drott thanks for https://github.com/mdn/browser-compat-data/pull/22231, I've merged that and will update the comment for the spec exception here.

ddbeck (approved):

PR #604, opened 2024-02-15
Add ic and lh units

foolip:


foolip:
I suggest reviewing this together with https://github.com/web-platform-dx/web-features/pull/631. Any concerns would apply to both.

feature-group-definitions/lh.yml
```diff
@@ -0,0 +1,7 @@
+name: lh unit
+description: The CSS `lh` unit corresponds to the requested line height, i.e., the computed value of the `line-height` property. Some lines may be higher than this based on their content.
```
ddbeck, line 2:
```suggestion
description: The CSS `lh` unit corresponds to the requested line height, the computed value of the `line-height` property. Some lines may be higher than this based on their content.
```

Optional style nit, since haven't actually adopted a style guide: I like and (over)use Latin abbreviations, but I try to avoid them in public-facing texts and style guides [often](https://developers.google.com/style/abbreviations) [forbid](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/ie) [them](https://support.apple.com/guide/applestyleguide/a-apsg3acde405/1.0/web/1.0). I don't think there's any loss of meaning to omit it here.

foolip: Very well! I will now try harder to use Latin abbreviations in our discussions, but fewer in descriptions.

PR #608, opened 2024-02-15
Add CSS math functions

foolip:


feature-group-definitions/calc-constants.yml
```diff
@@ -0,0 +1,8 @@
+name: CSS calc() constants e, pi, infinity, and NaN
```
ddbeck, line 1:
```suggestion
name: calc() CSS function constants e, pi, infinity, and NaN
```

feature-group-definitions/calc.yml
```diff
@@ -0,0 +1,8 @@
+name: CSS calc()
```
ddbeck, line 1:
```suggestion
name: calc() CSS function
```

feature-group-definitions/css-comparison.yml
```diff
@@ -0,0 +1,6 @@
+name: CSS min(), max(), and clamp()
```
ddbeck, line 1:
```suggestion
name: min(), max(), and clamp() CSS comparison functions
```

ddbeck, line 1: Maybe the ID is off too? `comparison-functions`?

feature-group-definitions/css-exponential.yml
```diff
@@ -0,0 +1,8 @@
+name: CSS pow(), sqrt(), hypot(), log(), and exp()
```
ddbeck, line 1:
```suggestion
name: pow(), sqrt(), hypot(), log(), and exp() CSS exponent functions
```

feature-group-definitions/css-round.yml
```diff
@@ -0,0 +1,6 @@
+name: CSS round(), mod(), and rem()
```
ddbeck, line 1:
```suggestion
name: round(), mod(), and rem() CSS functions
```

feature-group-definitions/css-sign.yml
```diff
@@ -0,0 +1,5 @@
+name: CSS abs() and sign()
```
ddbeck, line 1:
```suggestion
name: abs() and sign() CSS functions
```

feature-group-definitions/css-trig.yml
```diff
@@ -0,0 +1,10 @@
+name: CSS sin(), cos(), tan(), asin(), acos(), atan(), and atan2()
```
ddbeck, line 1:
```suggestion
name: sin(), cos(), tan(), asin(), acos(), atan(), and atan2() CSS trigonometry functions
```

ddbeck (changes requested):
Ideas for putting the unique things first in the names here

feature-group-definitions/calc.yml
foolip: I'm not sure why I didn't call this just `calc()`.

I think the description should be "The `calc()` CSS function does thing."

foolip:
@ddbeck do you want me to attempt descriptions for these to simplify the names? Or do you think "CSS" and "function" should be in the names too?

ddbeck:
@foolip I'd be happy with simplified names with descriptions, yes. Or even just simplified names. I assumed there was some kind of special distinguishing going on here with the names (e.g., maybe you wanted to avoid confusion with `Math.max()` or something), so I tried to maintain the intent of keeping "CSS" in the names. But if we can drop them too, then let's do it.

feature-group-definitions/calc-constants.yml
```diff
@@ -0,0 +1,9 @@
+name: calc() constants
```
foolip, line 1: @ddbeck can you review this name and description before I change them all? I found it hard to write the description in an active voice, since a constant is pretty darn passive, it is used but does not have behavior. How do you think we should phrase that?

ddbeck: Yeah, this is a place where the passive voice is genuinely OK.

> The `e`, `pi`, `infinity`, and `NaN` constants are accepted by CSS `calc()` expressions.

By comparison, the active version puts the wrong information first:

> `calc()` CSS function expressions accept the constants `e`, `pi`, `infinity`, and `NaN`.

Functions are easier to write actively, but also work passively:

> The `min()`, `max()`, and `clamp()` functions compare CSS values.
> `sin()`, `cos()`, … `atan2()` are trigonometry functions for CSS values.

feature-group-definitions/css-comparison.yml
foolip: Maybe just min-max-clamp since that's even shorter?

ddbeck (approved):
Very nice. Thank you @foolip!

PR #611, opened 2024-02-15
Add animation-composition

foolip:


feature-group-definitions/animation-composition.yml
```diff
@@ -0,0 +1,5 @@
+name: animation-composition
```
ddbeck, line 1:
```suggestion
name: animation-composition
description: The `animation-composition` CSS property chooses which composite operation to use when more than one animation applies to a property.
```

ddbeck (approved):
Approved, plus my bonus attempt at trying to describe this

feature-group-definitions/animation-composition.yml
foolip: Should that end with "applies to an element"?

ddbeck: Implicitly it's an element too, but if I understand correctly, `animation-composition` only comes into play when two animations apply to an element that affect the same property. Or to turn that around, if you apply two animations to an element but one changes `background-color` and the other changes `transform`, then `animation-composition` doesn't matter.

Though now that I've written this, we could tighten it up a bit:

```suggestion
name: animation-composition
description: The `animation-composition` CSS property chooses how to combine animations that affect the same property.
```

foolip: Ah, yes, the thing being animated is a property, which applies to an element. Let's go with your shorter description.

PR #612, opened 2024-02-15
Add :autofill

foolip:


feature-group-definitions/autofill.yml
```diff
@@ -0,0 +1,4 @@
+name: ":autofill"
```
captainbrosset, line 1: What did we land on with regards to a `description` field?
Features like this tend to benefit from a longer description like: `description: "CSS :autofill pseudo-class"` I think.

foolip: We don't have a guideline yet and descriptions aren't published, but I've added the description I think makes sense here.

captainbrosset (approved):

PR #613, opened 2024-02-15
Add blocking="render"

foolip:


feature-group-definitions/blocking-render.yml
```diff
@@ -0,0 +1,11 @@
+name: 'blocking="render"'
+details: "The `blocking="render"` attribute for `<link>`, `<script>`, and `<style>` elements blocks rendering until the external script or stylesheet has been loaded."
```
foolip, line 2: @domenic given that you reviewed https://github.com/whatwg/html/pull/7474 you must understand this feature deeply. I made an attempt at describing what this features does. I know it must be wrong because it's [potentially render-blocking](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#potentially-render-blocking) and not "definitely render-blocking", but I don't know when it doesn't actually block, or whether developers need to know about this from a short description.

Advice appreciated :)

domenic: I think your summary seems reasonable. The caveats are mostly detailed in https://html.spec.whatwg.org/#render-blocking-mechanism . Things like:

- This only works for elements in the `<head>`
- There's an implementation-defined timeout
- Only works for text/html documents

... that sort of thing.

@noamr is probably our current expert on this?

noamr: - Works only in the head
- Doesn't work on non-async classic scripts
- For links, only wokrs on `<link rel=stylesheet>` and `<link rel=expect`>. For the latter it would block until the element indicated by `href` is seen and is fully parsed (end tag seen or added programmatically).

feature-group-definitions/blocking-render.yml
```diff
@@ -0,0 +1,11 @@
+name: 'blocking="render"'
+details: 'The `blocking="render"` attribute for `<link>`, `<script>`, and `<style>` elements blocks rendering until the external script or stylesheet has been loaded.'
```
ddbeck, line 2: This'll fix the test

```suggestion
description: 'The `blocking="render"` attribute for `<link>`, `<script>`, and `<style>` elements blocks rendering until the external script or stylesheet has been loaded.'
```

feature-group-definitions/blocking-render.yml
foolip: Thanks @domenic and @noamr for the feedback. Of those caveats, I think the implementation-defined timeout is the most relevant to understand, and what some (evil QA people) might wonder about when reading about the feature. Is the timeout the same in all browsers, and what is it?

I'll add some comments to the YAML for things you've listed but I'm not sure should go in the description, so it's easier to find for future web-features maintainers.

noamr: > Thanks @domenic and @noamr for the feedback. Of those caveats, I think the implementation-defined timeout is the most relevant to understand, and what some (evil QA people) might wonder about when reading about the feature. Is the timeout the same in all browsers, and what is it?
> 

Chrome currently doesn't implement  such timeout. Since render blocking always relies on resource loading (js, css, html), the usual network timeouts apply.
> 
> 
> I'll add some comments to the YAML for things you've listed but I'm not sure should go in the description, so it's easier to find for future web-features maintainers.

feature-group-definitions/blocking-render.yml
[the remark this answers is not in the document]
noamr: Keep in mind that `<link rel=expect>` only ships in Chrome M124, so it wasn't hiding there for long.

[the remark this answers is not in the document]
foolip: It was hard to figure out what this feature is, so I created https://github.com/mdn/mdn/issues/534

PR #614, opened 2024-02-15
Add contain-intrinsic-size

foolip:


feature-group-definitions/contain-intrinsic-size.yml
```diff
@@ -0,0 +1,10 @@
+name: contain-intrinsic-size
+spec: https://drafts.csswg.org/css-sizing-4/#intrinsic-size-override
+usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/651 # TODO
```
ddbeck, line 3: What's this `TODO` about?

foolip: The name of the use counter is just "intrinsic-size" so I probably wanted to double check. I'll make it a comment.

ddbeck (approved):
Thank you!

PR #615, opened 2024-02-15
Add field-sizing

foolip:


feature-group-definitions/field-sizing.yml
```diff
@@ -0,0 +1,2 @@
+name: field-sizing
+spec: https://drafts.csswg.org/css-ui-4/#field-sizing
```
ddbeck, line 2: This was new to me and I didn't know what it was. In the interest of making sense of this for others:

```suggestion
spec: https://drafts.csswg.org/css-ui-4/#field-sizing
description: The `field-sizing` CSS property sets whether an element may expand based on its contents. The property applies only to elements with a default preferred size, such as a `<textarea>`, which has a fixed size regardless of its content.
```

ddbeck (approved):

PR #618, opened 2024-02-15
Add font-palette and animated font-palette

foolip:


foolip:
@ddbeck I think this is an interesting question of splitting vs merging. In the fullness of time this is definitely one feature, but right now there's a difference in when the different parts were implemented which maps cleanly to both distinct spec concepts, BCD entries and use counters. That's why I decided to split these.

cc @drott @tursunova

feature-group-definitions/font-palette-animation.yml
```diff
@@ -0,0 +1,5 @@
+name: font-palette animation
+spec: https://drafts.csswg.org/css-fonts-4/#font-palette-prop
+usage_stats: https://chromestatus.com/metrics/css/timeline/animated/709
+compat_features:
+  - css.types.color.palette-mix
```
ddbeck, line 5: This doesn't exist and probably should. Maybe file an issue on BCD?

(I'm not stressed about missing keys getting in here and there, if we're committed to tagging in BCD.)

feature-group-definitions/font-palette.yml
```diff
@@ -0,0 +1,11 @@
...
+spec:
+  - https://drafts.csswg.org/css-fonts-4/#font-palette-prop
+  - https://drafts.csswg.org/css-fonts-4/#font-palette-values
+usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/709
+compat_features:
+  - css.at-rules.font-palette-values
+  - css.at-rules.font-palette-values.base-palette
+  - css.at-rules.font-palette-values.font-family
+  - css.at-rules.font-palette-values.override-colors
+  - css.properties.font-palette
```
ddbeck, line 11: Perhaps this should have `CSSFontPaletteValuesRule` and descendants too?

feature-group-definitions/font-palette-animation.yml
```diff
@@ -0,0 +1,5 @@
+name: font-palette animation
+spec: https://drafts.csswg.org/css-fonts-4/#font-palette-prop
+usage_stats: https://chromestatus.com/metrics/css/timeline/animated/709
```
ddbeck, line 3: It's not clear to me that this usage counter belongs here, duplicating the other feature. Can you elaborate on this?

ddbeck (commented):
This split seems fine and not terribly ambiguous. Maybe BCD could track the animated-ness of the animation feature? 🤷‍♂️

feature-group-definitions/font-palette-animation.yml
foolip: Oh, I added this with the intention to add it to BCD but then forgot. I'll do that before merging this PR.

feature-group-definitions/font-palette-animation.yml
foolip: It's not the same URL, this is for when `font-palette` is animated.

feature-group-definitions/font-palette-animation.yml
ddbeck: Ah, this was a copy-and-paste failure on my part (I opened "both). Sorry!

feature-group-definitions/font-palette-animation.yml
ddbeck: Ah, this was a copy-and-paste failure on my part (I opened "both). Sorry!

feature-group-definitions/font-palette.yml
foolip: Done above.

feature-group-definitions/font-palette-animation.yml
foolip: Added in https://github.com/mdn/browser-compat-data/pull/22242 at a different path than I originally though. I'll just merge this and we'll sort out any mismatch when we migrate `compat_features` fully.

PR #620, opened 2024-02-15
Add IntersectionObserver v1

foolip:


foolip:
@szager if you have any thoughts on how/whether to split IntersectionObserver as high-level features, please let me know.

foolip:
I've split v2 into https://github.com/web-platform-dx/web-features/pull/628 so that v1 can be merged without figuring out the spec URL questions for v2.

feature-group-definitions/intersection-observer.yml
```diff
@@ -0,0 +1,22 @@
+name: IntersectionObserver
```
ddbeck, line 1: I think there's a case to be made for either here, but the spec uses the phrase "intersection observer" (not just the interface name) pretty routinely. I think it's more readable with the space and we should do the same. But caniuse does use the interface name, so this is not a blocking suggestion

```suggestion
name: Intersection observer
```

ddbeck (approved):

feature-group-definitions/intersection-observer.yml
foolip: I think we should use the phrase form when there's no ambiguity and there's some precedent for it, so I'll accept your suggestion.

PR #622, opened 2024-02-15
Add <slot>

foolip:


foolip:
Question for reviewers: Should this be combined with `<template>`? It's technically possible to use just `<slot>`, but do web developers care?

ddbeck:
I think no to combining, but perhaps later yes. 😅 

HTML elements are going to have these dependencies. The _reductio ad absurdum_ for this would be something like web developers never use `<slot>` without `<html>` and `<body>` so maybe we should combine them too?

Elements are routinely documented as a distinct things (e.g., MDN has separate reference pages for `<details>` and `<summary>` even though neither page has an example of the tag used without the other), even when they're closely tied to each other.

But if we represent them separately, we should still find a way to show a close relationship between them. Though they are not the same thing, it would break developer expectations if only one of `<slot>` or `<template>` worked. In the near term, maybe that means putting some or all of the `<template>` compat keys into `<slot>` as well (though I don't think the reverse applies), if it makes a difference to support calculations.

Perhaps later, we can show some relationship elsewhere in the schema (e.g., an abstract `slot-and-template` feature made up the concrete element features, with no compat keys of its own), or through some broader group (e.g., through a declarative Shadow DOM group).

ddbeck (approved):
(I think this is ready to merge as written, considering changes stemming from the `<template>` relationship question as optional.)

foolip:
Let me try to define the feature for `<template>` as well. I'm not sure they'll have the same Baseline date.

I think grouping the two as a "HTML temples" group (not feature!) might make sense.

foolip:
I've added `<template>` in https://github.com/web-platform-dx/web-features/pull/626 and found really no reason to group these.

PR #626, opened 2024-02-16
Add <template>

foolip:


feature-group-definitions/template.yml
```diff
@@ -0,0 +1,9 @@
+name: <template>
+spec: https://html.spec.whatwg.org/multipage/scripting.html#the-template-element
+usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/2769
+compat_features:
+  - api.HTMLTemplateElement
+  - api.HTMLTemplateElement.content
+  - api.HTMLTemplateElement.shadowRootMode
```
foolip, line 7: Oops, this belongs to Declarative Shadow DOM.

ddbeck: Are you going to edit this? I think it's the only thing holding it back.

ddbeck (approved):

ddbeck:
(approved assuming that one key is dropped)

PR #631, opened 2024-02-16
Add cap unit

foolip:


foolip:
I suggest reviewing this together with https://github.com/web-platform-dx/web-features/pull/604. Any concerns would apply to both.

ddbeck (approved):

PR #632, opened 2024-02-16
Add declarative shadow DOM (<template shadowrootmode>)

foolip:


feature-group-definitions/declarative-shadow-dom.yml
```diff
@@ -0,0 +1,8 @@
+name: Declarative shadow DOM
+description: The `shadowrootmode` attribute on `<template>` creates a shadow DOM without the use of JavaScript. It is a declarative alternative to the `attachShadow()` method.
```
foolip, line 2: @mfreed7 is this description accurate? The goal here is to explain the feature in one or two sentences to web developers. It doesn't need to be extremely precise, but shouldn't be misleading of course.

foolip: For comparison, caniuse describes this more in terms of the use case:

> Proposal to allow rendering elements with shadow dom (aka web components) using server-side rendering.

mfreed7: I think the description as-is is fairly accurate. And I like it better than caniuse's description, because there are other use cases also. This description describes what the feature *does*, not how it can be used.

One nit: use`creates a shadow **root**` instead of a `shadow DOM`.

foolip: Great, thanks for the review! I'll update the wording.

ddbeck (approved):

PR #633, opened 2024-02-16
Add constructed stylesheets

foolip:


foolip:
@andruud do you think it makes sense to treat the `CSSStyleSheet` constructor and the `adoptedStyleSheets` arrays as a single feature for web developers? They happened around the same time and I couldn't figure out a way to use one without the other.

andruud:
Yes, that makes sense. The methods `replace` and `replaceSync` also belong under the same umbrella.

foolip:
Thanks @andruud! I've included the replacement methods too now.

ddbeck (approved):

PR #634, opened 2024-02-16
Add CSS appearance property

foolip:


feature-group-definitions/appearance.yml
```diff
@@ -0,0 +1,11 @@
+name: appearance
+description: "The `appearance` CSS property controls the appearance of form controls. Using `appearance: none` disables any default native appearance and allows the elements to be styled with CSS like any other element."
```
foolip, line 2: @zcorpan is this a reasonable description of `appearance` for web developers? I'm focusing on the main use case and leaving aside some of the bizarre things that can be done.

zcorpan: Generally reasonable but "like any other element" is not accurate because some form controls still have weird behavior, e.g. buttons.
```suggestion
description: "The `appearance` CSS property controls the appearance of form controls. Using `appearance: none` disables any default native appearance and allows the elements to be styled with CSS."
```

zcorpan (approved):

captainbrosset (approved):

foolip:
Thanks @zcorpan and @captainbrosset!

PR #643, opened 2024-02-23
Add HTTP/1.1, HTTP/2, and HTTP/3.

foolip:
The HTTP/1.1 list of BCD keys is based on skimming the ToC of the
original https://datatracker.ietf.org/doc/html/rfc2616 alongside the
output of `npm run traverse http` in BCD. Oversights are possible.

HTTP/2 and HTTP/3 don't appear to have any representation in BCD but do
exist in caniuse, so add those entries without `compat_features`.


feature-group-definitions/http11.yml
```diff
@@ -0,0 +1,76 @@
...
+  - http.status.307
+  - http.status.401
+  - http.status.403
+  - http.status.404
+  - http.status.406
+  - http.status.407
+  - http.status.409
+  - http.status.410
+  - http.status.412
+  - http.status.416
```
ddbeck, line 71:
```suggestion
  - http.status.416
  - http.status.418
```

This is a haha only serious suggestion. There are a few other >416 statuses. Should they be here?

ddbeck (approved):
This is fine, but I did have a question in my spot checking.

feature-group-definitions/http11.yml
foolip: https://datatracker.ietf.org/doc/html/rfc2616 only lists 417 Expectation Failed, but https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417 doesn't have a compat table. If there was a BCD entry we should map it to HTTP/1.1. The statuses 418, 425, and 451 aren't in https://datatracker.ietf.org/doc/html/rfc2616.

Issue #548, opened 2024-01-29
Guidelines for short names


foolip:
Short names were added in https://github.com/web-platform-dx/web-features/pull/545 and @ddbeck suggested guidelines at https://github.com/web-platform-dx/web-features/pull/545#issuecomment-1913286828.

foolip:
It's worth thinking more about sentence case vs. Capitalizing Every Word. https://github.com/web-platform-dx/web-features/pull/545 was merged with sentence case. I've sent https://github.com/web-platform-dx/web-features/pull/549 to show what would change if we capitalize.

These cases are not great with sentence case:

- Cascade layers and Container queries, which I've seen capitalized many, many times.
- Constraint validation API, where the trailing "API" makes it look like like a typo.
- Web animations, which seems more generic and might cover CSS animations as well.

And these cases are not great when capitalized:

- "Flexbox Gap", but we could make it "Flexbox <code>gap</code>"
- "input Event", although that would be less strange as "<code>input</code> Event"
- More generally, any name that ends with "attribute", "event", "method", "property" or similar, which we'll have more of in a future long name. Consider "CSS <code>aspect-ratio</code> Property", where `aspect-ratio` feels squashed in between those Big Words.

@tidoust also pointed out that https://github.com/w3c/browser-specs has a `shortTitle` property, and that capitalizes every word.

I'm still leaning towards keeping sentence case, but am not at all committed to it.

tidoust:
Regarding `shortTitle` property values in browser-specs, the [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/#Title) recommends to "Capitalize title words following U.S. usage". Most if not all specification titles follow that convention.

The same manual also lists some [commonly misspelled terms](https://www.w3.org/Guide/manual-of-style/#Terms) with the example of "real-time communications", where capitalization depends on whether the term is used in titles and headings (and references to titles and headings) or to talk about the generic idea.

[...]

PR #655, opened 2024-02-29
Refactor docs to cover both names and identifiers

foolip:
Add more examples and document the additional shortening from names to
identifiers that we've done in many cases.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,39 +2,58 @@
 
 ## Feature guidelines
 
+### Names
+
+Feature authors should (in descending order of priority):
+
+- Prefer names to known to be in widespread use by web developers.
+  Favor describing things as they are most-widely known, even if it's not the most technically correct option.
+
+  - 👍 Recommended: JavaScript
+  - 👎 Not recommended: ECMAScript
+  - 👍 Recommended: Declarative shadow DOM
+  - 👎 Not recommended: `shadowrootmode` attribute
+
+- Avoid prefixes that mark a feature as specific to a technology, such as CSS, HTML, or JavaScript.
+  Features can and do cross such boundaries.
+
+  - 👍 Recommended: Container queries
+  - 👎 Not recommended: CSS container queries
+  - 👍 Recommended: `<dialog>`
+  - 👎 Not recommended: HTML `<dialog>`
+
+- Avoid frequently-used abbreviations and nouns, such as API and Web.
+
+  - 👍 Recommended: Async clipboard
+  - 👎 Not recommended: Async clipboard API
+  - 👍 Recommended: Workers
+  - 👎 Not recommended: Web workers
+
+- Prefer common, descriptive noun phrases over abbreviations, metonymy, and syntax.
+
+  - 👍 Recommended: Offscreen canvas
+  - 👎 Not recommended: `OffscreenCanvas`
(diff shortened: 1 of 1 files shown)


foolip:
While writing this I encountered a few issues:

- `structuredClone()` has "structured-clone" as the identifier and `checkVisibility()` as "check-visibility", but `scrollIntoView()` has "scrollintoview". Should we change it to "scroll-into-view" or add "match caniuse when shorter" to the guideline?
- should "array-copywithin", "array-findlast", "array-fromasync", and "array-isarray" be separated into words even though that makes the identifiers longer than the names? What about "bigint"?
- The identifiers "viewport-relative-units" and "viewport-relative-unit-variants" are not derived from the names at all
- What about features names for case-insensitive attribute names or values like "modulepreload" and "tabindex"? Do we need to explain why we use those names?

Dear reviewers, please don't block this PR on figuring out the answers to all of these issues, but keep them in mind :)

ddbeck:
>  * `structuredClone()` has "structured-clone" as the identifier and `checkVisibility()` as "check-visibility", but `scrollIntoView()` has "scrollintoview". Should we change it to "scroll-into-view" or add "match caniuse when shorter" to the guideline?

I think `structured-clone` is special being a global and that "structured cloning" is something said by documentation and such. But perhaps `check-visibility` should be renamed to something like `element-checkvisibility` and `elements-scrollintoview`, keeping in the spirit of the `Array` method identifiers. (see also https://github.com/web-platform-dx/web-features/pull/656#discussion_r1514509663).

> * should "array-copywithin", "array-findlast", "array-fromasync", and "array-isarray" be separated into words even though that makes the identifiers longer than the names? What about "bigint"?

Continuing from my previous point, perhaps we need a guideline for single interface method features. Those feel like "easy" cases where we might just use the method name and move on.

In the case of `bigint`, I think developers just say "bigint" and "bigints" and less commonly "big ints", so we have that to go on here.

>  * The identifiers "viewport-relative-units" and "viewport-relative-unit-variants" are not derived from the names at all

I think I'm responsible for these and I tried to rely phrases "known to be in widespread use by web developers." I tried to validate this by reading blog posts and Stack Overflow questions about their use.  I don't think it'd be outrageous to use the unit names directly, but the description field fills that need too.

> * What about features names for case-insensitive attribute names or values like "modulepreload" and "tabindex"? Do we need to explain why we use those names?

It wouldn't hurt to make that explicit for names, though I think the "lowercase alphanumeric" bit handles the identifiers already.

(Also, I learned today that the values are case insensitive! I had long assumed that the sequence of bytes given by `modulepreload` was significant, not the sequence of letters. Fascinating!)

ddbeck (approved):

ddbeck:
> But perhaps `check-visibility` should be renamed to something like…

Over in https://github.com/web-platform-dx/web-features/pull/655#issuecomment-1980936576, Philip makes a good case for preferring brevity over consistency across the whole set of features. (That said, we _should_ prefer consistency when we add text to distinguish things for ambiguity, as in the case of arrays and typed arrays.)

foolip:
Thanks for the review @ddbeck! I'll merge this and we should keep going in https://github.com/web-platform-dx/web-features/issues/548 with the guidelines.

PR #656, opened 2024-02-29
Add audio/video fastSeek() method

foolip:


feature-group-definitions/fast-seek.yml
```diff
@@ -0,0 +1,12 @@
+name: fastSeek()
```
ddbeck, line 1: Perhaps this should be something like:

```suggestion
name: HTMLMediaElement fastSeek() method
```

To make the context more obvious and to be more consistent with the way we've done things for `Array` methods?

(I don't know if this also implies that we should rename this to `htmlmediaelement-fastseek` too.)

foolip: Consistency you say? How about `checkVisibility()` and `scrollIntoView()`?

For Array, in many cases the prefix is needed to distinguish arrays and types arrays, and the `includes()` method also exists on `String` and `IDBKeyRange`.

Also consider `requestVideoFrameCallback()` which is certainly long enough to stand on its own, and is on `HTMLVideoElement`, not `HTMLMediaElement`.

I think the guideline I'm imagining is along the lines of "use the shortest plausible name except when ambiguous and when there's a group of related features, make the same choice for all of them for consistency within that group."

ddbeck: > I think the guideline I'm imagining is along the lines of "use the shortest plausible name except when ambiguous and when there's a group of related features, make the same choice for all of them for consistency within that group."

OK, yes, I like the idea of using the shortest without creating ambiguity. I'll backtrack a little on https://github.com/web-platform-dx/web-features/pull/655#issuecomment-1980936576 too.

ddbeck: Yes, I'm satisfied now. I'll approve this PR, but I'll leave it to you to decide to merge or wait for Patrick's review.

PR #660, opened 2024-03-01
Add snapshots as YAML files to track name/spec

foolip:
This also ensures that only snapshots which are defined can be used by a
feature.

Snapshots are still not published, and this does not affect the built
JSON at all.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -17,28 +17,43 @@ const omittables = [
     "snapshot"
 ]
 
-function scrub(data: FeatureData) {
+function scrub(data: any) {
     for (const key of omittables) {
         delete data[key];
     }
-    return data;
+    return data as FeatureData;
 }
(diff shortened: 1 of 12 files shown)


ddbeck (approved):

Elchi3 (approved):
Maybe a single yml file as the snapshots allow-list would be enough, but this works, too.

foolip:
> Maybe a single yml file as the snapshots allow-list would be enough, but this works, too.

I feel the pull of this argument, and went back and forth, but ultimately favored similarity of structure between features, groups and snapshots. Out of the three, snapshots look the silliest as separate YAML files. If we end up not adding anything more to the schema, perhaps we should squash it into a single file again.

But I'm glad you think this works, because then I can merge :)

Issue #532, opened 2024-01-18
Feature group guidelines


foolip:
We should have detailed guidelines and principles for how to create feature groups in https://github.com/web-platform-dx/web-features/tree/main/docs.

Known discussions:

- https://github.com/web-platform-dx/web-features/issues/24
- https://github.com/web-platform-dx/web-features/issues/495
- https://github.com/web-platform-dx/web-features/issues/504

foolip:
The main questions to resolve from https://github.com/web-platform-dx/web-features/issues/495:

- How do we handle "non-critical behavioral sub feature"?
- How do we handle features that arrive in bits in pieces, possibly many individual launches over time? Should we somehow mark the critical (constitutional) parts, or the non-critical parts?
- What's the role of subgroups in the above?

captainbrosset:
At the last WebDX CG meeting, I volunteered to create a shared doc to list the current discussions and questions to resolve. The idea is that, over time, this document would morph into a series of guidelines for how to create groups.

captainbrosset:
I started unearthing relevant discussions from GitHub and moving them to this shared document: https://docs.google.com/document/d/1Y2hTj9IqgLxOXmlePKttA0muB47DCI5tkeIIevmc18M/edit?usp=sharing

Please add to it if you know of others.
Please also comment so we can move this document forward and help it become a set of guidelines we can use going forward.

tidoust:
Additional questions and conventions that could perhaps be worth addressing in the guidelines:

- What makes a good feature group? The [README](https://github.com/web-platform-dx/web-features#web-features-exploring-the-set-of-interoperable-features-of-the-web-platform) says that "`web-features` identifies, defines, and categorizes capabilities of the web platform that web developers care about". Can we expand on that a bit to create a slightly more actionable list of checkpoints when we consider a feature group for inclusion?
- What are the recommended fields in the YAML file when the feature group is created? The schema only requires the `spec` field. Should the `caniuse` and/or `compat_features` fields be set upon creation, when possible? The `usage_stats` field? The human-friendly name, when it exists (see #490)?
- What should the `spec` field contain? The code currently asserts that the URL maps to a [nightly URL](https://github.com/w3c/browser-specs/#nightlyurl) in `web-specs`. More generally, human-friendly documentation about the YAML schema would ease contributions and reviews. That could be done in a separate document, linked from a guidelines doc.

foolip:
[...]

PR #673, opened 2024-03-07
Use (initial support) as the suffix for "basic" features

foolip:
This puts the suffix in the short name as it will be needed to
disambiguate the feature in a list of short names. This made it
necessary to write new descriptions, and an attempt was made.

### DIFF
--- a/feature-group-definitions/array.yml
+++ b/feature-group-definitions/array.yml
@@ -1,5 +1,5 @@
-name: Arrays
-description: Basic JavaScript arrays
+name: Arrays (initial support)
+description: Arrays are ordered lists of JavaScript values.
 spec:
   - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array-objects
   - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array-constructor
--- a/feature-group-definitions/typed-arrays.yml
+++ b/feature-group-definitions/typed-arrays.yml
@@ -1,5 +1,5 @@
-name: Typed arrays
-description: Basic typed arrays
+name: Typed arrays (initial support)
+description: "Typed arrays are ordered lists of JavaScript values, where all values are of the same numerical type, for example 8-bit integers or 32-bit floating point numbers."
 spec:
   - https://tc39.es/ecma262/multipage/indexed-collections.html#table-49
   - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray-constructors


feature-group-definitions/array.yml
```diff
@@ -1,5 +1,5 @@
-name: Arrays
-description: Basic JavaScript arrays
+name: Arrays (initial support)
+description: Arrays are ordered lists of JavaScript values.
```
captainbrosset, line 2:
```suggestion
description: Ordered lists of JavaScript values.
```

Suggestion to avoid repeating the name of the feature in the description. The description should just _describe_ the feature.

feature-group-definitions/typed-arrays.yml
```diff
@@ -1,5 +1,5 @@
-name: Typed arrays
-description: Basic typed arrays
+name: Typed arrays (initial support)
+description: "Typed arrays are ordered lists of JavaScript values, where all values are of the same numerical type, for example 8-bit integers or 32-bit floating point numbers."
```
captainbrosset, line 2:
```suggestion
description: "Ordered lists of JavaScript values, where all values are of the same numerical type, for example 8-bit integers or 32-bit floating point numbers."
```

captainbrosset (approved):

Elchi3 (approved):
LGTM, no strong feeling on Patrick's point. I will note that in MDN short descriptions we do the repetition. (certain dominating search engines apparently enjoy that)

feature-group-definitions/array.yml
foolip: I've made a breakdown of descriptions written so far in https://github.com/web-platform-dx/web-features/issues/532#issuecomment-1986073057 for discussion. I think that when the verb is "is" or "represent" that perhaps the active voice isn't doing any good, and perhaps we should have two styles, with some guide for which to prefer.

foolip:
I'll leave this in draft to allow for some discussion about the description style over in https://github.com/web-platform-dx/web-features/issues/532.

feature-group-definitions/array.yml
captainbrosset: Let's close this without changes. I actually think my suggestion wasn't a good one and dropped a [comment](https://github.com/web-platform-dx/web-features/issues/532#issuecomment-1996892324) on 532 about why. Let's keep the long form.

foolip:
Thanks for review @Elchi3 and @captainbrosset! Merging now.

foolip:
I've sent https://github.com/web-platform-dx/web-features/pull/689 to document this.

PR #677, opened 2024-03-08
Rename the SVG <script> async/defer feature

foolip:
The type="module" support makes more sense as part of JS modules:
https://github.com/web-platform-dx/web-features/pull/676

### DIFF
--- a/feature-group-definitions/svg-async-defer.yml
+++ b/feature-group-definitions/svg-async-defer.yml
@@ -1,4 +1,5 @@
 name: SVG <script> async and defer
+alias: svg2-script-html-equivalence
 spec: https://svgwg.org/svg2-draft/interact.html#ScriptElement
 status:
   baseline: false


ddbeck (approved):
https://github.com/web-platform-dx/web-features/pull/676/files#r1517665897 says you're removing `svg.elements.script.type.module` from this feature. I think it's fine to remain (so the alias still makes sense), so I'm approving as is, but if you think it shouldn't be in the feature after all, I'd like to hear a little more about why.

Elchi3 (commented):
If you leave it in, you are creating situation 4 for MDN: https://github.com/web-platform-dx/web-features/issues/654#issuecomment-1985837520. MDN doesn't have its own page for this BCD key, so it might not be a problem.

ddbeck:
> The feature "SVG `<script>` works like HTML `<script>`" is very contingent on how browsers evolved. It feels much more contingent-y than most features, and eventually won't be considered a feature at all. Note how I dropped the modules part from the name of this feature to make it shorter, but that didn't really make sense...

Hmmm. I feel that something like this should be allowed to (co-)exist? Spitballing here, but maybe we'd be better served by a few (partly overlapping) groups:

- SVG script equivalency
- SVG script async
- SVG script defer
- JS modules

The first one would probably be short-lived (we'd acquire some irrelevance guideline, like BCD), while the others would persist for a long time.

feature-group-definitions/svg-async-defer.yml
```diff
@@ -1,4 +1,5 @@
 name: SVG <script> async and defer
+alias: svg2-script-html-equivalence
 spec: https://svgwg.org/svg2-draft/interact.html#ScriptElement
 status:
   baseline: false
```
niloferm1984, line 5: Baseline true 
The cat look through the glass window 
My head turn so yellow

niloferm1984 (approved):

foolip:
The JS modules part of this features was extracted in https://github.com/web-platform-dx/web-features/pull/676.

The identifier for this will continue to bug me, in particular the number 2, so I've updated this to `svg-async-defer`. But should we even have this feature, or should we put back JS modules and rename it to `svg-script-parity` or something?

ddbeck:
@foolip can you show example YAML for that? I'm not quite sure what you mean.

niloferm1984:
Thank u

niloferm1984:
Thank u 


Sent from Yahoo Mail for iPhone


On Saturday, March 16, 2024, 4:25 pm, Daniel D. Beck ***@***.***> wrote:




@foolip can you show example YAML for that? I'm not quite sure what you mean.

foolip:
@ddbeck I mean like https://github.com/web-platform-dx/web-features/pull/700.

feature-group-definitions/svg-async-defer.yml
niloferm1984: Correctly all of u ever

ddbeck:
Thanks, @foolip. Very helpful to understand the options here. I guess looking back on what I said in https://github.com/web-platform-dx/web-features/pull/677#issuecomment-1986178750, I think my preferences on this are now:

- Split this PR in two, as in `svg-script-async` and `svg-script-defer`. (Reasoning here that web developers won't be looking for _both_ `async` and `defer`, they'll just be wondering why the one they tried to use isn't working.)
- Drop #700 and the "equivalency" angle entirely. I see now that "equivalency with HTML" is _not_ a feature. It's _why_ async and defer should work in SVGs, but I have a hard time seeing it as a thing one could _use_. As if you'd want the option to insert a script tag into either HTML or SVG dynamically or something? It's too weird.

ddbeck:
And on a meta issue: I've sent a message to François to look into spam controls on the repo—I don't have the necessary privileges in the web-platform-dx org.

niloferm1984:
Ok 


Sent from Yahoo Mail for iPhone


On Saturday, March 16, 2024, 5:01 pm, Daniel D. Beck ***@***.***> wrote:




And on a meta issue: I've sent a message to François to look into spam controls on the repo—I don't have the necessary privileges in the web-platform-dx org.

foolip:
@ddbeck sounds like we're on the same page, I don't think this amounts to a coherent feature.

niloferm1984:
Ok thanks 


Sent from Yahoo Mail for iPhone


On Saturday, March 16, 2024, 8:58 pm, Philip Jägenstedt ***@***.***> wrote:




@ddbeck sounds like we're on the same page, I don't think this amounts to a coherent feature.

foolip:
Reopening just to serve as a reminder to split into smaller features.

foolip:
I'll close this and open an issue instead.

PR #678, opened 2024-03-08
Add color() and relative color syntax features

foolip:
Support for RCS isn't computed by hand, waiting for the scripts...

### DIFF
--- a/feature-group-definitions/relative-color.yml
+++ b/feature-group-definitions/relative-color.yml
@@ -1,5 +1,5 @@
-name: Relative color syntax
-description: "The `from` keyword for color functions (`color()`, `hsl()`, `oklch()`, etc.) creates a new color based on a given color by modifying the values of the input color."
+name: Relative colors
+description: "The `from` keyword for color functions (`color()`, `hsl()`, `oklch()`, etc.) creates a new color based on a given color by modifying the values of the input color. Also known as relative color syntax (RCS)."
 spec: https://drafts.csswg.org/css-color-5/#relative-color-function
 caniuse: css-relative-colors
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/4632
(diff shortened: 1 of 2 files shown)


feature-group-definitions/relative-color.yml
```diff
@@ -0,0 +1,14 @@
+name: Relative color syntax
```
foolip, line 1: @ddbeck this might interest you in light of your comment in https://github.com/web-platform-dx/web-features/pull/628#discussion_r1517867790

I considered naming this something like "Relative colors" and putting "Also known as relative color syntax" in the description, but didn't. I can't make up my mind if I should or not. It's "commonly known as" vs. "shortest plausible name" doing battle.

romainmenke: I really like "relative colors".
This also aligns with the documentation work that is currently happening for MDN :
- https://github.com/mdn/content/pull/32004
- https://pr32004.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/CSS_colors/Relative_colors

------

Relative colors also depends on `mixed_type_params` in our data : https://github.com/csstools/cssdb/blob/c5ebff87d042438d63285cf3291f62d3affd734b/cssdb.settings.json#L1424-L1441

Currently that isn't relevant as no implementation has a complete and bug free version. But might become relevant if WebKit fixes relative colors without also fixing `mixed_type_params`.

foolip: I've changed the name to "Relative colors", happy to match MDN!

What does `mixed_type_parameters` mean? It's not linked to any browser bugs, but just from the name I suppose that something like `oklch(50% 0.1 40)` originally didn't work? Does it also have to do with the `deg` unit?

romainmenke: Since commenting that I changed my mind.
For web features, relative color syntax isn't related to `mixed_type_params`.
Even if you obviously cannot mix percentages and numbers in a relative color value unless that sub feature is also supported.

-------

> What does mixed_type_parameters mean?

Before that addition these were valid:
- `rgb(255 255 255)`
- `rgb(100% 100% 100%)`

But this was not: `rgb(100% 255 100%)`

With the addition of the new color functions it quickly became clear that no author was going to remember the exact incantation for each.

When taking `var()` into account it also becomes unknown what the actual type is when writing : `rgb(100% var(--g) 100%)`.

So it was relaxed to always allow any combination of suitable value types.

https://drafts.csswg.org/css-color-4/#color-syntax-modern

ddbeck (approved):

PR #680, opened 2024-03-11
Use correct spec link for long-animation-frames

noamr:


feature-group-definitions/long-animation-frame-timing.yaml
```diff
@@ -1,2 +1,2 @@
 name: Long animation frame timing
-spec: https://w3c.github.io/longtasks/#sec-loaf-timing
+spec: https://w3c.github.io/long-animation-frames/
```
foolip, line 2: @noamr this URL is a 404, oops?

noamr: Yea, I asked that the short name be changed and the long name was also changed :( fixed now

foolip: Should we also rename this feature to "Long animation frames" then? Or "Long animation frame detection"? From a web developer point of view, what is it?

noamr: Perhaps "Long Animation Frame Timing" (like resource timing)

foolip: OK, so keep the current name. I suspect you may not like the way we capitalize, and if so please file an issue because we want to hear from people if it doesn't make sense in some cases. I'll go ahead and merge this though.

PR #684, opened 2024-03-13
Add Lab and LCH color spaces

foolip:


feature-group-definitions/lab.yml
```diff
@@ -0,0 +1,25 @@
+name: Lab and LCH
+description: "The CIE Lab color space expresses colors in terms of perceived lightness and how red/green and blue/yellow a color is. LCH is a variant of Lab with polar coordinates. These color spaces can be used with the CSS `color()`, `lab()`, and `lch()` functions. Also as CIELAB and CIELCH."
```
foolip, line 2: When do we include "CIE" and not? Not in the identifier because that matches the syntax of `lab()`. Not in the short name because the shortest plausible name doesn't need CIE, it's often known as just "Lab". But yes here early in the description, so that it's possible to figure out that "CIE" is a thing related to this, and finally Googling CIELAB at the end would find https://en.wikipedia.org/wiki/CIELAB_color_space.

romainmenke: The order/wording is maybe a bit confusing?
I read it as "Can also be used as CIELAB and CIELCH", while the intention is "Also known as CIELAB and CIELCH".

---

Rest looks good!

foolip: Oh, I'm missing a word!

ddbeck (approved):

PR #695, opened 2024-03-15
Ensure description is not too long

foolip:
As suggested by @ddbeck in https://github.com/web-platform-dx/web-features/issues/532#issuecomment-1999483461


captainbrosset (approved):

ddbeck (approved):
Thank you!

PR #702, opened 2024-03-18
Simplify accent-color description

foolip:


ddbeck (approved):
Yes, this is better. I let some spec-speak leak in. Thank you!

PR #703, opened 2024-03-18
Simplify field-sizing description

foolip:


feature-group-definitions/field-sizing.yml
```diff
@@ -1,3 +1,3 @@
 name: field-sizing
 spec: https://drafts.csswg.org/css-ui-4/#field-sizing
-description: The `field-sizing` CSS property sets whether an element may expand based on its content. The property applies only to elements with a default preferred size, such as a `<textarea>`, which has a fixed size regardless of its content.
+description: The `field-sizing` CSS property allows form controls such as `<textarea>` to be sized based on their content.
```
foolip, line 3: @ddbeck my main goal here was to say what `field-sizing: content` does, since `field-sizing: fixed` is the pre-existing default. It's sort of like how `box-sizing: content-box` isn't the feature, `box-sizing: border-box` is.

I couldn't find a good way to fit `field-sizing: content` into a phrase however. "The `field-sizing: content` CSS property and value ..." is too much...

ddbeck: This is good! I think it's slightly over-fitting to say that `field-sizing: content` is _the_ feature (e.g., you might not want this, while the framework that you're using does). I like how you've thread the needle here.

ddbeck (approved):

PR #704, opened 2024-03-18
Update descriptions and their max length to fix build

foolip:


foolip:
Self-review and merging since the build is broken.

PR #707, opened 2024-03-19
[Masonry] Adding a description and removing compat features

captainbrosset:
The BCD entries for this feature have already been added to the BCD as tags, no need to repeat them here (see this [search query](https://github.com/search?q=repo%3Amdn%2Fbrowser-compat-data%20%22web-features%3Amasonry%22&type=code)).
This PR also adds a short description of the feature.

### DIFF
--- a/feature-group-definitions/masonry.yml
+++ b/feature-group-definitions/masonry.yml
@@ -1,11 +1,6 @@
 name: Masonry
+description: Masonry is a type of grid layout where the items on one of the axes are tightly packed together, instead of leaving gaps to align across the other axis.
 spec: https://drafts.csswg.org/css-grid-3/
 status:
   baseline: false
   support: {}
-compat_features:
-  - css.properties.align-tracks
-  - css.properties.grid-template-columns.masonry
-  - css.properties.grid-template-rows.masonry
-  - css.properties.justify-tracks
-  - css.properties.masonry-auto-flow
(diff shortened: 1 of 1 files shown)


feature-group-definitions/masonry.yml
```diff
@@ -1,11 +1,6 @@
 name: Masonry
+description: Masonry is a type of grid layout where the items on one of the axes are tightly packed together, instead of leaving gaps to align across the other axis.
```
ddbeck, line 2: A couple of ideas for this one:

- Note the domain here (styles)
- Explain the metaphor

```suggestion
description: Masonry is a type of CSS grid layout where the items on one of the axes are tightly packed together, like brickwork, instead of leaving gaps to align across the other axis.
```

feature-group-definitions/masonry.yml
```diff
@@ -1,11 +1,6 @@
...
 spec: https://drafts.csswg.org/css-grid-3/
 status:
   baseline: false
   support: {}
-compat_features:
-  - css.properties.align-tracks
-  - css.properties.grid-template-columns.masonry
-  - css.properties.grid-template-rows.masonry
-  - css.properties.justify-tracks
-  - css.properties.masonry-auto-flow
```
ddbeck, line 11: This is OK since were making other changes, but don't take it as a license to seek out every such key just yet. 😄 

I think it'll be a lot easier to clean these up when we have the tools to easily determine whether there's a mismatch between web-features and BCD.

ddbeck (approved):
Have some (optional) ideas for the description for you. Thanks!

feature-group-definitions/masonry.yml
captainbrosset: Makes sense to me, thanks. Especially for cases where we specifically don't want the web-features entry to map 1:1 to what's in BCD.

ddbeck (approved):

PR #713, opened 2024-03-20
Add anchor positioning caniuse and description fields

captainbrosset:
This PR adds the `caniuse` and `description` fields to the anchor positioning feature.

### DIFF
--- a/feature-group-definitions/anchor-positioning.yml
+++ b/feature-group-definitions/anchor-positioning.yml
@@ -1,2 +1,4 @@
 name: Anchor positioning
+description: Anchor positioning allows authors to position an element based on the position of another element, for example to place a tooltip next to the content it references.
 spec: https://drafts.csswg.org/css-anchor-position-1/#anchoring
+caniuse: css-anchor-positioning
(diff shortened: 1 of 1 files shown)


feature-group-definitions/anchor-positioning.yml
```diff
@@ -1,2 +1,4 @@
 name: Anchor positioning
+description: Anchor positioning allows authors to position an element based on the position of another element, for example to place a tooltip next to the content it references.
```
ddbeck, line 2: Optional style suggestion, to tighten this up:

```suggestion
description: Anchor positioning places an element based on the position of another element. For example, to place a tooltip next to the content it references.
```

As a general rule, I think we might want to avoid talking about "authors" in the abstract, since they're the audience for this. I'd prefer to talk directly to them. So we might even consider something like:

```suggestion
description: Anchor positioning places an element based on the position of another element. For example, you can place a tooltip next to the content it references.
```

ddbeck (approved):
Optional style suggestion for you to consider. Thank you!

ddbeck (approved):
Thanks, @captainbrosset!

PR #716, opened 2024-03-20
Add `counter-style`

ddbeck:
Note that when we set a status for this one, it might get a little weird: BCD says `speak-as` is not widely supported, but I'm not sure that's correct (e.g., I couldn't find a webkit bug about it). The rest of this seems boring and unobjectionable.


feature-group-definitions/counter-style.yml
```diff
@@ -0,0 +1,28 @@
+name: "@counter-style"
+description: The `@counter-style` CSS at-rule defines custom counter styles for list items. For example, you can use a sequence of specific symbols instead of numbers for an ordered list.
```
foolip, line 2: `@font-face` and `@scope` don't have descriptions yet, but I think "CSS at-rule" will work in general 👍

foolip (approved):

PR #720, opened 2024-03-22
Add background-clip as two features

foolip:


feature-group-definitions/background-clip-text.yml
```diff
@@ -0,0 +1,5 @@
+name: "background-clip: text"
+description: "The `background-clip: text` CSS property value allows the background to be clipped to the text in the element."
```
foolip, line 2: Attention reviewers! This is the first use of the phrase "CSS property value" referring to a CSS property and a specific value. We'll need it for `text-wrap: balance` too. I don't love it, but it's some kind of tradeoff between the slightly too wrong "CSS property" and the too pedantic "CSS property and property value".

foolip:
I think this is the first instance of two features that very clearly have a parent-child relationship, where eventually we'll want to fold `background-clip: text` into `background-clip`. This relationship isn't recorded in any way.

feature-group-definitions/background-clip-text.yml
ddbeck: I think this is a good compromise. Others I could think of:

- "The `background-clip: text` CSS property and value allows…" (this one's OK—preserves a little bit of the pedantry at the cost of an "and")
- "The `text` value of the `background-clip` CSS property allows…" (wordy and buries the context a little, but not outrageous)
- "The `background-clip` CSS property's value `text` allows…"  (starting to get ugly, lots nouns piled up front here)
- "The `background-clip: text` CSS property-value pair allows…" (pedantic with the smell of "you should already know what I'm talking about")

On the specific description here:

```suggestion
description: "The `background-clip: text` CSS property value draws the background underneath only the text in the element."
```

I'm not sure "clipping" is recognizable and it felt like it might be defining the word by using it. But maybe this is too aggressive or specific about what's going on.

feature-group-definitions/background-clip.yml
```diff
@@ -0,0 +1,8 @@
+name: background-clip
+description: The `background-clip` CSS property allows the background to be clipped to the padding box or content box, in addition to the default of the whole border box.
```
ddbeck, line 2: Like above, what do you think of this, to dodge the "allows" bit. (A variation, to see an alternative to "drawing" here.)

```suggestion
description: The `background-clip` CSS property sets the extent of the background: the padding box, the content box, or the default border box.
```

ddbeck (approved):
This is fine as is, but some optional style talk in line comments. Also:

> a parent-child relationship […] This relationship isn't recorded in any way.

Maybe a `TODO` comment on each, pointing to the other, would be a good move? Again, optional.

feature-group-definitions/background-clip.yml
foolip: Thanks, I like this rewrite!

feature-group-definitions/background-clip-text.yml
foolip: I think that clipping is fairly established and understood in any kind of graphics context, but that doesn't mean it's the simplest way to describe this.

Your wording does provide a clue to the question "why can I still only see the text" which is "you need to make the text transparent or translucent to see the background."

foolip:
@ddbeck I've added some notes now, WDYT? (Not TODOs, because there is nothing to do right now.)

ddbeck (approved):
That works for me. Thank you!

PR #721, opened 2024-03-22
Add baseline-source

foolip:


feature-group-definitions/baseline-source.yml
```diff
@@ -0,0 +1,8 @@
+name: baseline-source
+description: The `baseline-source` CSS property controls how inline-level boxes are aligned within a line of text.
```
foolip, line 2: @bfgeek is this a reasonable description for web developers of what this feature does? I did some testing and I think it's about the case where the inline-level box has multiple lines of text and therefore multiple baselines. The wording I've used here makes it sounds very similar to `vertical-align` but what's the simplest way of explaining it? Maybe "how inline-level boxes with multiple lines of text are aligned with the surrounding text"?

bfgeek: It fixes the issue that the default of which baseline is selected for flex/grid/block are different. (flex/grid first, block - last).

Allows developers for an inline-flex to select the last baseline instead of the first. (Also fixes the broken table in block behaviour).

foolip: How would you describe the behavior itself? Is it "how inline-level boxes with multiple lines of text are aligned with the surrounding text"?

Perhaps we then add on "By default, which baseline is used depends on the `display` value." By not saying how, it's a hint that using `baseline-source` lets you decide.

foolip: I'll apply this change:

```suggestion
description: The `baseline-source` CSS property controls how inline-level boxes with multiple lines of text are aligned with the surrounding text. By default, which baseline is used depends on the `display` property value.
```

feature-group-definitions/baseline-source.yml
```diff
@@ -0,0 +1,8 @@
+name: baseline-source
+description: The `baseline-source` CSS property controls how inline-level boxes with multiple lines of text are aligned with the surrounding text. By default, which baseline is used depends on the `display` property value.
```
ddbeck, line 2: Mostly me just trying to work in the word "typographic" to give this a little more context.

```suggestion
description: The `baseline-source` CSS property controls how inline-level boxes with multiple lines of text are aligned with the surrounding text. By default, the `display` property value sets the typographic baseline.
```

ddbeck (approved):

feature-group-definitions/baseline-source.yml
foolip: I think working in "typographic" is good, but "sets" feels wrong here. The choice depends on `display`, but that's not at all obvious and more of an accident than by design. It's good if the unfortunateness of this comes through in the description as much as possible. Minimally:

```suggestion
description: The `baseline-source` CSS property controls how inline-level boxes with multiple lines of text are aligned with the surrounding text. By default, which typographic baseline is used depends on the `display` property value.
```

ddbeck: OK, cool—I do not want to imply intention that does not exist. Your version works for me.

PR #725, opened 2024-03-24
Add a comment about what spec URLs are allowed

foolip:


foolip:
Note that this is different from BCD:

https://github.com/mdn/browser-compat-data/blob/fd1cb702c59712fc3ef2056ed3a846292e3ae8c3/test/linter/test-spec-urls.ts#L41-L53

ddbeck (approved):

PR #726, opened 2024-03-24
Only allow specs in "good" standing

foolip:
https://github.com/w3c/browser-specs/blob/main/README.md#standing

There are no exceptions to this rule now, but exceptions could still be
made via defaultAllowlist.

BCD also has this rule:
https://github.com/mdn/browser-compat-data/blob/fd1cb702c59712fc3ef2056ed3a846292e3ae8c3/test/linter/test-spec-urls.ts#L43

### DIFF
--- a/scripts/specs.ts
+++ b/scripts/specs.ts
@@ -4,16 +4,20 @@ import webSpecs from 'web-specs' assert { type: 'json' };
 
 import features from '../index.js';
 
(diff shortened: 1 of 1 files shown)


foolip:
This will conflict with https://github.com/web-platform-dx/web-features/pull/725, I'll rebase after one of the two is merged.

ddbeck (approved):
I merged the other one first, but I'd be happy to see this merge after the rebase. 👍

ddbeck (approved):

PR #727, opened 2024-03-24
Add prefers-color-scheme / color-scheme / light-dark()

foolip:
These three features are all related to dark mode.

### DIFF
--- a/feature-group-definitions/color-scheme.yml
+++ b/feature-group-definitions/color-scheme.yml
@@ -1,5 +1,5 @@
 name: color-scheme
-description: The `color-scheme` CSS property defines which color schemes (light or dark) an element can be rendered in and prevents automatic dark mode adjustments by the browser.
+description: The `color-scheme` CSS property sets which color schemes (light or dark) an element uses and may prevent automatic dark mode adjustments by the browser.
 spec: https://drafts.csswg.org/css-color-adjust-1/#color-scheme-prop
 usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/637
 compat_features:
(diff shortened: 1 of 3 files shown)


feature-group-definitions/color-scheme.yml
```diff
@@ -0,0 +1,8 @@
+name: color-scheme
+description: The `color-scheme` CSS property defines which color schemes (light or dark) an element can be rendered in and prevents automatic dark mode adjustments by the browser.
```
ddbeck, line 2: A few changes for brevity and possibly clarity. Specifically:

- I initially read "rendered in" as meaning "(don't) render at all"
- The `only` bit is optional

```suggestion
description: The `color-scheme` CSS property sets which color schemes (light or dark) an element uses and may prevent automatic dark mode adjustments by the browser.
```

Also, this reminded me of something: I once wrote [a little style guide for writing short descriptions](https://github.com/mdn/short-descriptions/blob/master/writing-short-descriptions.md)! It was for CSS only (at least to start), so it's a little over-specific for us to adopt in full, but it does suggest to me some changes here, specifically:

> Short descriptions should avoid the use of words like "specifies", "defines", or "determines" when the word "sets" is just as good and results in no loss of meaning.

feature-group-definitions/prefers-color-scheme.yml
```diff
@@ -0,0 +1,5 @@
+name: prefers-color-scheme
+description: The `prefers-color-scheme` CSS media query is used to detect the requested color scheme, light or dark.
```
ddbeck, line 2: style nit: brevity

```suggestion
description: The `prefers-color-scheme` CSS media query detects the requested color scheme, light or dark.
```

feature-group-definitions/color-scheme.yml
foolip: I like these edits, applying without further debate (for once) 😄

ddbeck (approved):

PR #728, opened 2024-03-24
Update `spec` link for Relative colors

romainmenke:
Section `5.1` is only about `color(from ...)`.

While section `4` is a better general introduction to relative colors.

`https://drafts.csswg.org/css-color-5/#relative-colors` links to section `4`


foolip (approved):
Thank you!

PR #738, opened 2024-03-25
Add descriptions for a few features

foolip:


feature-group-definitions/aborting.yml
```diff
@@ -1,4 +1,5 @@
 name: AbortController and AbortSignal
+description: The `AbortController` and `AbortSignal` interfaces allow aborting an ongoing operation, such as a `fetch()` request.
```
ddbeck, line 2: style nit: avoiding "allow" (though I'll grant that "handle" is not _great_)

```suggestion
description: The `AbortController` and `AbortSignal` interfaces handle cancelling an ongoing operation, such as a `fetch()` request.
```

feature-group-definitions/broadcast-channel.yml
```diff
@@ -1,4 +1,5 @@
 name: BroadcastChannel
+description: The `BroadcastChannel` interface allows sending messages between same-origin browsing contexts, for example between the same page loaded in multiple tabs.
```
ddbeck, line 2: style nits:

- avoiding "allows"
- maybe this is idiosyncratic but I'm not sure "for example" is formally correct as a coordinating conjunction (brought to you because TIL that the equivalent in Dutch _is_)

```suggestion
description: The `BroadcastChannel` interface sends messages between same-origin browsing contexts, such as between the same page loaded in multiple tabs.
```

feature-group-definitions/canvas-context-lost.yml
```diff
@@ -1,4 +1,5 @@
 name: contextlost and contextrestored
+description: The `contextlost` event for `<canvas>` fires when the canvas backing storage is lost, while the `contextrestored` is fired when it is recreated.
```
ddbeck, line 2: style nit: consistency

```suggestion
description: The `contextlost` event for `<canvas>` fires when the canvas backing storage is lost, while the `contextrestored` event fires when it is recreated.
```

ddbeck (commented):
Some ideas to make these a little nicer. Thank you!

feature-group-definitions/aborting.yml
foolip: I think "handle" sounds like these interfaces do something automatically, as oppose to enabling developers to do something. To write this as "The X aborts Y" is hard, since it's the `abort()` method on an `AbortController` instance `controller`, when `controller.signal` (an `AbortSignal` instance) has been given to the API that supports aborting. Too much into the weeds IMHO.

We'll need some pattern for cases like this common case where a complex API is passive, but allows a developer to act. Ideas:

- allows
- enables
- supports
- ???

feature-group-definitions/broadcast-channel.yml
foolip: I like such as. I'll update two existing cases for "for example" as conjunction.

foolip: https://github.com/web-platform-dx/web-features/pull/742

foolip: And I'll apply this part of the suggestion:

```suggestion
description: The `BroadcastChannel` interface allows sending messages between same-origin browsing contexts, such as between the same page loaded in multiple tabs.
```

feature-group-definitions/aborting.yml
foolip: That's a first for "you" but it's not uncommon to see in docs for developers. Sure, why not. I'll also go with cancel.

PR #741, opened 2024-03-25
Add groups as YAML files similar to snapshots

foolip:


Elchi3 (approved):
Looks good to me and seems okay to merge given this isn't published yet, right?

I wondered if we want to allow multiple parents. I think we want, but maybe we can extend this when we've hit the first case.

PR #742, opened 2024-03-25
Avoid use of "for example" as a conjunction

foolip:


docs/baseline.md
```diff
@@ -29,7 +29,7 @@ See [Audience illustration for Baseline status](#audience-illustration-for-basel
 
 Goal 2: **Change with the web and web developers’ needs.**
 Features can and should join and leave Baseline status as their respective interoperability, reliability, commitment, and availability profiles change.
-Likewise, Baseline statuses should change as developers’ needs change (for example, developers come to rely on a new browser or an established browser fades from use).
+Likewise, Baseline statuses should change as developers’ needs change. (For example, if developers come to rely on a new browser or an established browser fades from use.)
```
ddbeck, line 32: Since this stands alone, we can probably lose the parentheses too.

```suggestion
Likewise, Baseline statuses should change as developers’ needs change. For example, if developers come to rely on a new browser or an established browser fades from use.
```

ddbeck (approved):
One minor addition to this, but otherwise looks great. Thank you!

PR #744, opened 2024-03-26
✨ add safari for js groupBy

ctcpip:


foolip (approved):
Thanks @ctcpip! I also updated the Baseline status and date.

Support confirmed in BCD:
https://github.com/mdn/browser-compat-data/blob/91029dd9c22f9148c35d6a080cbd4c1e4ec1b7b0/javascript/builtins/Object.json#L835
https://github.com/mdn/browser-compat-data/blob/91029dd9c22f9148c35d6a080cbd4c1e4ec1b7b0/javascript/builtins/Map.json#L452

PR #748, opened 2024-03-27
Add font-size-adjust

foolip:


feature-group-definitions/font-size-adjust.yml
```diff
@@ -0,0 +1,9 @@
+name: font-size-adjust
+description: The `font-size-adjust` CSS property sets the size of lowercase letters relative to the size of uppercase letters.
```
foolip, line 2: @drott is this a reasonable description of this feature for web developers?

drott: No, this is inaccurate, the [MDN text](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust#:~:text=The%20font%2Dsize%2Dadjust%20CSS%20property%20provides%20a%20way%20to%20modify%20the%20size%20of%20lowercase%20letters%20relative%20to%20the%20size%20of%20uppercase%20letters%2C%20which%20defines%20the%20overall%20font%2Dsize.%20This%20property%20is%20useful%20for%20situations%20where%20font%20fallback%20can%20occur.) is also too specific to lowercase letters and incorrect, as the property allows more than that.

perhaps:
> The [font-size-adjust](https://drafts.csswg.org/css-fonts-5/#propdef-font-size-adjust) property provides a way to preserve apparent size of text across fonts in the fallback chain by adjusting the used font size so that the specified metric is scaled to the same size regardless of the font used.

foolip: I see, so it allows matching of all of the font metrics listed in the spec and MDN? That's ex-height, cap-height, ch-width, ic-width, and ic-height. Do all implementations support all of these, or will we need to break this feature up into varying levels of support?

drott: Yes, I expect all impls to support all of them. We discussed whether ic-height is needed or going to be removed from the spec, but that discussion has shifted to what we do when a [metric is unavailable in the font](https://github.com/w3c/csswg-drafts/issues/6384).

foolip: Thanks @drott! I took your suggested text and tried to simplify it just a little bit, but hopefully not to the point where it's wrong now. @ddbeck can you review?

feature-group-definitions/font-size-adjust.yml
```diff
@@ -0,0 +1,9 @@
+name: font-size-adjust
+description: The `font-size-adjust` CSS property preserves the apparent text size by adjusting the font size so that the specified metric is scaled to the same size regardless of the font used. This is useful when fallback fonts may be used.
```
ddbeck, line 2: I like this new description—I feel like I was able to figure out what is going on, which is a first for this property for me. 😅 

I tried a somewhat aggressive revision here, to make it easier to understand the (several!) relationships described.

```suggestion
description: The `font-size-adjust` CSS property preserves apparent text size, regardless of the font used, by scaling fonts to the same size with respect to a specific metric, such as x-height. This can help make fallback fonts look the same size.
```

foolip: Looks good to me, thanks @ddbeck!

PR #750, opened 2024-03-28
Adding missing feature descriptions - Batch 1

captainbrosset:


feature-group-definitions/background-fetch.yml
```diff
@@ -1,4 +1,5 @@
 name: Background fetch
+description: The Background Fetch API allows service workers to delegate downloading large amounts of data to the system, even when the web page is closed.
```
foolip, line 2:
```suggestion
description: Background fetch allows service workers to download data in the background even when the web page is closed.
```

feature-group-definitions/background-gradients.yml
```diff
@@ -1,4 +1,5 @@
 name: Background gradients
+description: The CSS gradient functions create background images consisting of progressive transitions between two or more colors.
```
foolip, line 2:
```suggestion
description: The `linear-gradient()`, `radial-gradient()`, and `conic-gradient()` CSS functions create backgrounds consisting of progressive transitions between two or more colors.
```

feature-group-definitions/border-image.yml
```diff
@@ -1,4 +1,5 @@
 name: Border images
+description: The `border-image` CSS property draws an image around an element's border.
```
foolip, line 2: Hmm, doesn't it replace the painting of the border, and not painting around it?

feature-group-definitions/bigint.yml
```diff
@@ -1,4 +1,5 @@
 name: BigInt
+description: BigInt is a built-in JavaScript type that represents numeric values which are too large to be represented by the Number type.
```
foolip, line 2: Trying to fit this into the "The $syntax $technology $typeofthing" pattern like "The `aspect-ratio` CSS property ..." How about this:

```suggestion
description: The `BigInt` JavaScript type represents integers of any size, including integers too large for the primitive `Number` type.
```

feature-group-definitions/cascade-layers.yml
```diff
@@ -1,4 +1,5 @@
 name: Cascade layers
+description: CSS Cascade Layers solve CSS rule specificity conflicts by providing priority levels for different groups of CSS rules, such as low-priority styles like resets, and high-priority styles like UI components.
```
foolip, line 2:
```suggestion
description: The `@layer` CSS at-rule avoids specificity conflicts by providing priority levels for different groups of CSS rules, such as low-priority styles like resets, and high-priority styles like UI components.
```

foolip (commented):
Thank you @captainbrosset! I left (too many) suggestions based on the kinds of patterns I've been following, but I'm not a tech writer so I might have bad taste. I would defer to @ddbeck if there's anything we have different preferences on.

feature-group-definitions/async-await.yml
```diff
@@ -1,4 +1,5 @@
 name: Async functions
+description: The `async` and `await` keywords enable the asynchronous, promise-based behavior of a function to be written without using promise chains.
```
ddbeck, line 2: What do you think of saying what this means in a more explicit style, like this?

```suggestion
description: The `async` and `await` keywords allow you to write asynchronous functions without using using the `Promise()` constructor and chains of `then()` methods.
```

feature-group-definitions/async-clipboard.yml
```diff
@@ -1,4 +1,5 @@
 name: Async clipboard
+description: The Clipboard API provides asynchronous read and write access to the content of the system clipboard, by requesting permission from the user.
```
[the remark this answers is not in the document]
ddbeck: Riffing on @foolip here:

```suggestion
description: The `navigator.clipboard` API asynchronously reads and writes to the system clipboard.
```

feature-group-definitions/background-fetch.yml
ddbeck: Even shorter:

```suggestion
description: Background fetch downloads data in the background even when the web page is closed.
```

feature-group-definitions/background-gradients.yml
ddbeck: Since "transitions" are something else, how about this:

```suggestion
description: The `linear-gradient()`, `radial-gradient()`, and `conic-gradient()` CSS functions create backgrounds that progress between two or more colors.
```

feature-group-definitions/border-image.yml
ddbeck: How about this?

```suggestion
description: The `border-image` CSS property draws an image around an element.
```

ddbeck (commented):
I have some ideas!

feature-group-definitions/async-await.yml
foolip: I think this is perhaps a bit too explicit. Async/await is also useful when you use other APIs that return promises, even if there's just one of them. So it's not necessarily about the constructor or a chain.

feature-group-definitions/async-clipboard.yml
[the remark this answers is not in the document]
captainbrosset: > I think we should avoid the spec name unless it is also the best name for the feature.

I agree that for this particular case, "Clipboard API" is probably not the best name for the feature. At least I've not really heard developers talk about it this way. But otherwise, I think it's pretty common for devs to refer to features by their spec names rather than by the name of the main interface/method/property of the feature. 
Guess it will have to be on a case by case basis.

feature-group-definitions/async-await.yml
ddbeck: OK, in that case, rewriting solely to avoid "enables".

```suggestion
description: The `async` and `await` keywords allow you to use the asynchronous, promise-based behavior of a function without using promise chains.
```

ddbeck (approved):
I'm approving this, though I'll leave it un-merged for a little while, in case Philip wants to review again. Thank you, Patrick!

feature-group-definitions/async-await.yml
```diff
@@ -1,4 +1,5 @@
 name: Async functions
+description: The `async` and `await` keywords allow you to use the asynchronous, promise-based behavior of a function without using promise chains.
```
foolip, line 2: As I noted before it's possible to pick holes in the "using promise chains" bit here, but I think this is close enough to the motivation in practice.

foolip (approved):

PR #751, opened 2024-03-28
Adding missing feature descriptions - Batch 2

captainbrosset:


feature-group-definitions/check-visibility.yml
```diff
@@ -1,4 +1,5 @@
 name: checkVisibility()
+description: The `Element.checkVisibility()` method checks if an element is visible.
```
ddbeck, line 2: I wasn't sure how this compared to something like intersection observer. My attempt to clarify:

```suggestion
description: The `Element.checkVisibility()` method checks if an element is visible. For example, it checks whether an element has the style `display: none`.
```

feature-group-definitions/class-syntax.yml
```diff
@@ -1,4 +1,5 @@
 name: Classes
+description: The `class` JavaScript keyword defines a new class, which is a type of function that can be instantiated multiple times by using the `new` keyword.
```
ddbeck, line 2: I'm having a hard time with this one. The description isn't wrong, but I don't think it says much about what it does for developers. Trying something completely different here, just to get thinking about this:

```suggestion
description: Classes are an object-oriented syntax for JavaScript prototypes.
```

Maybe @Elchi3 has an idea here, as well.

feature-group-definitions/colrv1.yml
```diff
@@ -1,4 +1,5 @@
 name: COLRv1
+description: The COLRv1 font format is used by color fonts to support multi-color glyphs.
```
ddbeck, line 2: For brevity:

```suggestion
description: COLRv1 is a font format that supports multi-color glyphs.
```

(That said, I'm a little nervous about using "supports" in these descriptions, but I couldn't think of an alternative here.)

feature-group-definitions/compression-streams.yml
```diff
@@ -1,4 +1,5 @@
 name: Compression streams
+description: The JavaScript Compression Streams API compresses and decompresses streams of data using the gzip or deflate formats.
```
ddbeck, line 2: A couple of ideas here:

- I think this is a web API only, so I don't think it helps to mention JS.
- We can name the interfaces, since there aren't too many? (Less certain about the utility here.)

```suggestion
description: The `CompressionStream` and `DecompressionStream` interfaces compress and decompress data using the gzip or deflate formats.
```

ddbeck: I've run out of time, so this is a partial review—I haven't looked at any of the ones below.

ddbeck (commented):
This is a partial review—sorry!—but I thought I'd offer what I can now.

feature-group-definitions/check-visibility.yml
```diff
@@ -1,4 +1,5 @@
 name: checkVisibility()
+description: "The `Element.checkVisibility()` method checks if an element is visible. For example, it checks whether an element has the style `display: none`."
```
foolip, line 2: I think we should avoid the `Interface.method()` notation unless it's about a static method like `CSS.supports()`, since it looks like code but won't work. Sometimes I think it's OK to use a variable name that is very likely to be used, like `element. checkVisibility()` in this case, but let's see if it can be avoided:

```suggestion
description: "The `checkVisibility()` method checks if an element is visible. For example, it checks whether an element has the style `display: none`."
```

In cases where we have to spell out the interface name I think one of these styles would work:

```suggestion
description: "`Element`'s `checkVisibility()` method checks if an element is visible. For example, it checks whether an element has the style `display: none`."
```

```suggestion
description: "The `checkVisibility()` method for `Element` checks if an element is visible. For example, it checks whether an element has the style `display: none`."
```

feature-group-definitions/constraint-validation.yml
```diff
@@ -1,4 +1,5 @@
 name: Constraint validation API
+description: The Constraint Validation API provides a programmatic way to validate form controls.
```
foolip, line 2: I think this is just around the edge of how much is reasonable to spell out, but I think it works:

```suggestion
description: The `checkValidity()`, `reportValidity()` and `setCustomValidity()` methods are a programmatic way to validate form controls.
```

feature-group-definitions/contain-intrinsic-size.yml
```diff
@@ -1,4 +1,5 @@
 name: contain-intrinsic-size
+description: The `contain-intrinsic-size` CSS property sets the size of an element that the browser will use for layout calculations when the element is subject to size containment, within a CSS query container.
```
foolip, line 2: @vmpstr can you help review this description?

feature-group-definitions/container-style-queries.yml
```diff
@@ -1,4 +1,5 @@
 name: Container style queries
+description: CSS style container queries apply styles to an element based on the computed styles of its container.
```
foolip, line 2:
```suggestion
description: CSS container style queries apply styles to an element based on the computed styles of its container.
```

feature-group-definitions/content-visibility.yml
```diff
@@ -1,4 +1,5 @@
 name: content-visibility
+description: The `content-visibility` CSS property controls whether the browser should skip an element's rendering, including layout and painting, until it is needed.
```
foolip, line 2: Somewhat simpler:

```suggestion
description: The `content-visibility` CSS property skips an element's rendering, including layout and painting, until it is needed.
```

feature-group-definitions/check-visibility.yml
```diff
@@ -1,4 +1,5 @@
 name: checkVisibility()
+description: "The `checkVisibility()` method checks if an element is visible. For example, it checks whether an element has the style `display: none`."
```
bkardell, line 2: The name even was contentious because this is hard to summarize, but maybe should quote the spec "provides a set of simple checks for whether an element is potentially 'visible', with options to define which precise notion of 'visibility' is desired."

feature-group-definitions/constraint-validation.yml
```diff
@@ -1,4 +1,5 @@
 name: Constraint validation API
+description: The `checkValidity()`, `reportValidity()` and `setCustomValidity()` methods are a programmatic way to validate form controls.
```
bkardell, line 2: I wonder if this should be less specific- it's not impossible that this could be added to, right? Maybe it could be just something like "An API with methods providing programatic access to HTML's form field constraint validations, including providing custom validations."?

feature-group-definitions/check-visibility.yml
foolip: Noting that there are multiple kinds of visibility sounds good. Trying to match the guidelines in https://github.com/web-platform-dx/web-features/pull/789:

```suggestion
description: "The `checkVisibility()` method checks if an element is potentially visible, with optional parameters for the kinds of visibility to check. For example, it checks whether the element has the style `display: none`, but can also check for `visibility: hidden`."
```

I reduced it to just one example from the default checks and the optional checks to resist the temptation of making it complete.

feature-group-definitions/constraint-validation.yml
captainbrosset: Yeah, I think I'd prefer something like that too. @foolip ?

foolip: Sounds good to me!

feature-group-definitions/container-style-queries.yml
```diff
@@ -1,4 +1,5 @@
 name: Container style queries
+description: CSS container style queries apply styles to an element based on the computed styles of its container.
```
ddbeck, line 2: One-off suggestion here, based on https://github.com/web-platform-dx/web-features/pull/801/files#r1554637063.

(Also avoiding starting the description with CSS. Not sure if that's too contorted.)

```suggestion
description: Container style queries with the `@container` at-rule apply styles to an element based on the values of custom properties of its container.
```

foolip:
The test failures are because of the "needs to be updated" lines in the output. I've filed https://github.com/web-platform-dx/web-features/issues/827 about this being buried in other output.

foolip:
@captainbrosset in order to update all dist files, you can run `npm run dist feature-group-definitions/*.dist.yml`. That should resolve whatever the problems are, and is safe to run even if no changes are needed.

foolip:
> safe to run even if no changes are needed

With one exception! If you accidentally edit .dist.yml instead of the source file and try to get them in sync with the dist script, you will instead overwrite your changes.

feature-group-definitions/constraint-validation.yml
ddbeck: Maybe we can preserve the search-ability here with something like this?

```suggestion
description: Methods that validate form controls before submission, such as `checkValidity()`, `reportValidity()` and `setCustomValidity()`.
```

captainbrosset:
With the exception of contain-intrinsic-size, which @foolip would like @vmpstr to review, this PR should be about ready to go.

feature-group-definitions/contain-intrinsic-size.yml
foolip: This description seems good to me, so I'll just go ahead and merge this PR, and ask for review after the fact via a chat message.

foolip (approved):

PR #752, opened 2024-03-28
Adding missing feature descriptions - Batch 3

captainbrosset:


feature-group-definitions/device-posture.yml
```diff
@@ -1,4 +1,5 @@
 name: Device posture
+description: The Device Posture API provides information about the physical posture of a device, such as whether a foldable device is folded or unfolded.
```
foolip, line 2: I've filed https://github.com/web-platform-dx/web-features/issues/822 about the style of names and descriptions for APIs. For now, I think the least silly suggestion I can make that keeps this consistent is just a capitalization change:

```suggestion
description: The device posture API provides information about the physical posture of a device, such as whether a foldable device is folded or unfolded.
```

feature-group-definitions/document-picture-in-picture.yml
```diff
@@ -1,3 +1,4 @@
 name: Document picture-in-picture
+description: The Document Picture-in-Picture API makes it possible to open an always-on-top window that can be populated with arbitrary HTML content.
```
foolip, line 2:
```suggestion
description: The document picture-in-picture API makes it possible to open an always-on-top window that can be populated with arbitrary HTML content.
```

feature-group-definitions/edit-context.yml
```diff
@@ -1,2 +1,3 @@
 name: EditContext
+description: The EditContext API is used to build rich text editors that support advanced text input experiences, such as Input Method Editor (IME) composition, emoji picker, or any other platform-specific editing-related UI surfaces.
```
foolip, line 2:
```suggestion
description: The `EditContext` API allows you to build rich text editors that support advanced text input experiences, such as Input Method Editor (IME) composition, emoji picker, or any other platform-specific editing-related UI surfaces.
```

foolip (approved):

feature-group-definitions/custom-properties.yml
```diff
@@ -1,4 +1,5 @@
 name: Custom properties
+description: Custom properties (also known as CSS variables) are CSS properties prefixed with `--`, and whose values can be reused throughout CSS rules by using the `var()` function.
```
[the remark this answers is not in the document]
captainbrosset: I'm usually a fan of `using` or `by using` instead of `with`. According to Microsoft Style Guide, `with` is ambiguous, and `using` eliminates the ambiguity: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/u/using-by-using-with

feature-group-definitions/details.yml
```diff
@@ -1,4 +1,5 @@
 name: <details>
+description: The `<details>` element creates a disclosure widget which content is visible only when the widget is toggled open, such as when using a nested `<summary>` element.
```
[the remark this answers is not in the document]
captainbrosset: I'm not 100% sure about the "toggled open" part. How about "... which can be **expanded** to reveal ..."

Also, do you think we should mention the fact that multiple details with the same name work as a mutually exclusive accordion?

feature-group-definitions/custom-properties.yml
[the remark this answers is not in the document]
foolip: OK, a smaller edit to avoid that change:

```suggestion
description: Custom properties are CSS properties prefixed with `--`. Their values can be reused throughout CSS rules by using the `var()` function. Also known as CSS variables.
```

feature-group-definitions/details.yml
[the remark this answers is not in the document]
foolip: Saying "expanded" sounds good to me. I think it's best not to mention the `name` attribute since it's a separate feature. If we had some way to link other features in descriptions we could say "see also", but we don't correctly.

captainbrosset:
@ddbeck this already got reviewed byu @foolip once. Let me know if you have any comments about those.

feature-group-definitions/custom-properties.yml
```diff
@@ -1,4 +1,5 @@
 name: Custom properties
+description: Custom properties are CSS properties prefixed with `--`. Their values can be reused throughout CSS rules by using the `var()` function. Also known as CSS variables.
```
ddbeck, line 2: I think an example would go a long ways here:

```suggestion
description: Custom properties are CSS properties prefixed with `--` that set values you can reuse with the `var()` function. For example, you can set a `--key-color` property to reuse as `border-color: var(--key-color)`. Also known as CSS variables.
```

feature-group-definitions/details.yml
```diff
@@ -1,4 +1,5 @@
 name: <details>
+description: The `<details>` element is a disclosure widget which can be expanded to reveal additional details. When closed, only the nested `<summary>` element is visible.
```
ddbeck, line 2: style nit: circularity

```suggestion
description: The `<details>` element is a disclosure widget which can be expanded to reveal additional content. When closed, only the nested `<summary>` element is visible.
```

feature-group-definitions/dialog.yml
```diff
@@ -1,4 +1,5 @@
 name: "<dialog>"
+description: The `<dialog>` HTML element represents a modal or non-modal dialog box.
```
ddbeck, line 2: How about an example or two?

```suggestion
description: The `<dialog>` HTML element represents a modal or non-modal dialog box, such as a confirmation prompt or inspector widget.
```

feature-group-definitions/document-picture-in-picture.yml
```diff
@@ -1,3 +1,4 @@
 name: Document picture-in-picture
+description: The document picture-in-picture API makes it possible to open an always-on-top window that can be populated with arbitrary HTML content.
```
ddbeck, line 2: style nit, brevity

```suggestion
description: The document picture-in-picture API makes an always-on-top window from arbitrary HTML content.
```

feature-group-definitions/fetch-priority.yml
```diff
@@ -1,4 +1,5 @@
 name: Fetch priority
+description: The optional `priority` parameter of the `fetch()` method, and the `fetchPriority` HTML attribute specifies the priority of the fetch request, relative to other requests of the same type.
```
ddbeck, line 2: I know I usually say to put the unique thing first, but I had a really hard time parsing this. What do you think of this?

```suggestion
description: The `fetch()` `priority` option and the `fetchPriorty` HTML attribute give hints to the browser about which requests to do before other requests of the same type.
```

feature-group-definitions/flexbox-gap.yml
```diff
@@ -1,3 +1,4 @@
 name: Flexbox gap
+description: The `gap` CSS property in a flexbox layout specifies the size of the gap between flex items and between flex lines.
```
ddbeck, line 2: "flex lines" feels a bit obscure to me, what do you think of this?

```suggestion
description: The `gap` CSS property in a flexbox layout sets the size of the space between flex items on the main axis (columns) and cross axis (rows).
```

feature-group-definitions/dialog.yml
captainbrosset: Not sure what an "inspector widget" is. Maybe:

```suggestion
description: The `<dialog>` HTML element represents a modal or non-modal dialog box, such as a confirmation prompt or a subwindow used to enter data.
```

feature-group-definitions/document-picture-in-picture.yml
captainbrosset: I like it. How about "creates" instead of "makes":

```suggestion
description: The document picture-in-picture API creates an always-on-top window from arbitrary HTML content.
```

feature-group-definitions/flexbox-gap.yml
captainbrosset: I agree it's obscure, but using "items" for cross axis is incorrect. Flex items are things that are placed along the main axis. And, when the container is set to wrap, the things that are placed along the cross axis are flex lines.
So it's a choice of either being more technically correct, or easier to understand.
That being said, I think it ends up being equivalent. Flex lines aren't something you can see in the DOM or style with CSS, so any gap between flex lines will end up being the exact same gap between flex items.

Maybe something simpler?

```suggestion
description: The `gap` CSS property in a flexbox layout sets the size of the space between items.
```

This way we don't mention the 2 dimensions, which might actually be better because over in `flexbox.yaml` we say that it's a "one-dimensional layout system".

feature-group-definitions/dialog.yml
ddbeck: Not a Mac user, eh? 😄 But yes, this works!

feature-group-definitions/document-picture-in-picture.yml
ddbeck: That's fine with me.

feature-group-definitions/flexbox-gap.yml
ddbeck: Yes, this shorter version is much better!

foolip (approved):

captainbrosset:
@ddbeck I've addressed the review feedback and re-generated the dist files. This should be ready to merge. Giving you a chance to re-review if you want to.

Quick question: do you usually squash PR commits? (this one has silly looking "fix typo" commits in it, and squashing nicely removes those).

foolip:
> Quick question: do you usually squash PR commits? (this one has silly looking "fix typo" commits in it, and squashing nicely removes those).

I squash and rewrite the commit messages, because GitHub ruins them by default. Most of the time I trim it back to the original commit message, if it's still accurate.

ddbeck (approved):

ddbeck:
@captainbrosset I was talking to @foolip earlier and how it must be a little rough to be on the receiving end of all these suggestions. I just wanted to say that I'm very grateful you've written these and that you've been so responsive on the edits. There have been a lot of unstated assumptions about descriptions and through this mass effort we're finding good patterns going forward, and very quickly. Thank you!

foolip:
Yes indeed, thank you @captainbrosset, web-features descriptions will forever be better thanks to you. And they'll be published sooner too!

captainbrosset:
Thank you, both, for the kind words. I completely understand the need for these discussions, and they're part of the work. Coming up with the initial descriptions is, in a way, the easy part. But it's a vital part nonetheless, because it acts as a forcing function to get the ball rolling, and assigns someone to see these PRs through.

PR #753, opened 2024-03-28
Adding missing feature descriptions - Batch 4

captainbrosset:


feature-group-definitions/focus-visible.yml
```diff
@@ -1,4 +1,5 @@
 name: ":focus-visible"
+description: The `:focus-visible` CSS pseudo-class selects elements that match the `:focus` pseudo-class and are determined by the browser to be focused using the keyboard or a similar input device.
```
foolip, line 2: @bkardell can you review this?

feature-group-definitions/font-optical-sizing.yml
```diff
@@ -1,4 +1,5 @@
 name: font-optical-sizing
+description: The `font-optical-sizing` CSS property sets whether text rendering is optimized for viewing at different sizes.
```
foolip, line 2: I guess that `font-optical-sizing: none` is the new feature here. But not calling that out seems OK.

feature-group-definitions/font-palette.yml
```diff
@@ -1,4 +1,5 @@
 name: font-palette and @font-palette-values
+description: The `font-palette` CSS property specifies one of the many palettes contained in a color font that browser should use for that font.
```
foolip, line 2:
```suggestion
description: The `font-palette` CSS property selects a color palette from the font or set a custom palette.
```

feature-group-definitions/font-synthesis.yml
```diff
@@ -1,4 +1,5 @@
 name: font-synthesis
+description: The `font-synthesis` CSS shorthand property specifies whether or not the browser should synthesize bold, italic, small-caps, and other typefaces when they're missing from the specified font-family.
```
foolip, line 2:
```suggestion
description: The `font-synthesis` CSS property sets whether or not the browser should synthesize bold, italic, small-caps, and other typefaces when they're missing from the font.
```

feature-group-definitions/fonts.yml
```diff
@@ -1,4 +1,5 @@
 name: "@font-face"
+description: The `@font-face` CSS at-rule specifies a custom font with which to display text, which can be loaded from either a remote server or a locally-installed font on the device.
```
foolip, line 2:
```suggestion
description: The `@font-face` CSS at-rule specifies a custom font with which to display text, which can be either loaded from a server or a font installed on the device.
```

feature-group-definitions/fullscreen.yml
```diff
@@ -1,4 +1,5 @@
 name: Fullscreen API
+description: The Fullscreen API is used to present a specific element in fullscreen mode.
```
foolip, line 2:
```suggestion
description: The fullscreen API is used to present a specific element in fullscreen mode.
```

feature-group-definitions/grid-animation.yml
```diff
@@ -1,4 +1,5 @@
 name: Grid animation
+description: Grid animation allows to animate the `grid-template-columns` and `grid-template-rows` CSS properties.
```
foolip, line 2:
```suggestion
description: Grid animation allows you to animate the `grid-template-columns` and `grid-template-rows` CSS properties.
```

foolip (approved):

feature-group-definitions/font-palette.yml
captainbrosset: I'm a bit puzzled b y the "set a custom palette", unsure how it differs from "selects a color palette".

Here's another attempt:

```suggestion
description: The `font-palette` CSS property selects a color palette from the font or creates a new palette by blending together two color palettes from the font.
```

feature-group-definitions/fonts.yml
captainbrosset:
```suggestion
description: The `@font-face` CSS at-rule specifies a custom font with which to display text, which can be either loaded from a server or installed on the device.
```

I think we can get rid of the repetition.

feature-group-definitions/font-palette.yml
foolip: I haven't tried writing demos for this, but my understand is that you can do at least 3 things:

- Use a palette from the font without changes by using `@font-palette-values` with `base-palette`
- Override some or all colors using `override-colors`
- Mix any two palettes using `palette-mix()`

@tursunova is that right?

feature-group-definitions/fonts.yml
foolip: The problem I tried to avoid in my first suggestions is that it condenses down to "custom font can be loaded from a locally-installed font", and that "a remote server" and "locally-installed font" is an odd pair.

With the new wording it sounds like the rule has the effect of either loading a font or installing a font, which is also not right.

I'm not sure how to fix this. @ddbeck to the rescue?

feature-group-definitions/font-variant-alternates.yml
```diff
@@ -1,4 +1,5 @@
 name: font-variant-alternates
+description: The `font-variant-alternates` CSS property controls the usage of alternate glyphs for characters, as defined in `@font-feature-values` at-rules.
```
ddbeck, line 2: opened https://github.com/web-platform-dx/web-features/pull/847

```suggestion
description: The `font-variant-alternates` CSS property selects which alternate glyphs a font should use, as defined by `@font-feature-values` at-rules.
```

feature-group-definitions/fullscreen.yml
```diff
@@ -1,4 +1,5 @@
 name: Fullscreen API
+description: The fullscreen API is used to present a specific element in fullscreen mode.
```
ddbeck, line 2: Avoiding a circular description

```suggestion
description: The fullscreen API makes a specific element fill the whole screen and hides most browser UI.
```

ddbeck (commented):
A few more suggestions. Thank you!

feature-group-definitions/fonts.yml
foolip: Very good, let's have that one!

feature-group-definitions/font-palette-animation.yml
```diff
@@ -1,4 +1,5 @@
 name: font-palette animation
+description: The `palette-mix()` CSS function creates a new `font-palette` value by blending together two `font-palette` values contained in a color font.
```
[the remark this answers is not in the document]
foolip: The palette can come from the font, and then some or all of the colors can be replaced.

feature-group-definitions/font-palette.yml
tursunova: Yes, that's right. I'm not sure about the phrasing, but maybe add something like: 

"...selects a color palette from the font or creates a new palette **by overriding colors from the palette or** by blending together two color palettes from the font"

foolip: Thank you @tursunova!

Since a palette from the font is always the base, perhaps we can describe this as a palette from the font with optional overrides. Bringing it all together:
```suggestion
description: The `font-palette` CSS property selects a color palette from the font, optionally overriding individual colors in the `@font-palette-values` at-rule.
```

foolip: There's also the question of `palette-mix()`. It shouldn't be mentioned in this feature because it was introduced for animations, and even though it can be used as `font-palette: palette-mix(--a, --b)` without animation, I think that's a niche thing best not mentioned.

feature-group-definitions/font-palette-animation.yml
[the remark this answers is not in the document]
foolip: After confirming my understanding with @tursunova, I think this description isn't right for animating `font-palette`. `palette-mix()` is part of the feature because it represents the intermediate states, but it's not really how you use the features. Rather, you'd just animate from `font-palette: --a` to `font-palette: --b` and the mixing is done by the browser.

So the description here should be similar to https://github.com/web-platform-dx/web-features/pull/855, somehow expressing "ability to animate". @ddbeck can you weigh in?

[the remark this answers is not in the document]
ddbeck: Oh, I completely misread the feature—going on the compat features and missing the name. I thought this was about `palette-mix()` existing, in general, and not its ability to be animated.

Building on my previous comment in https://github.com/web-platform-dx/web-features/pull/855#discussion_r1562324143

```suggestion
description: You can animate color fonts between two `font-palette` values.
```

If we want to make the connection to `palette-mix` explicit:

```suggestion
description: You can animate color fonts between two `font-palette` values, using the `palette-mix()` function to control the color mix.
```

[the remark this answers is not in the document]
foolip: I think the first is better, because using `palette-mix()` explicitly is likely to be rare, perhaps only when you want to pick the interpolation color space, which is complexity I don't think most people will want to think about. The default should be good.

feature-group-definitions/flexbox.yml
```diff
@@ -1,4 +1,5 @@
 name: Flexbox
+description: CSS Flexbox Layout is a one-dimensional layout system, which lays out content either horizontally or vertically, with optional wrapping.
```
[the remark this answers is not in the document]
foolip: For Grid (in another PR) I reduced it to just one "layout"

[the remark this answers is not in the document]
captainbrosset: So, either this to match the grid description:

```suggestion
description: CSS Flexbox is a one-dimensional layout system, which places content either horizontally or vertically, with optional wrapping.
```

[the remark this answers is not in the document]
captainbrosset: Or this, to start directly with the word Flexbox:

```suggestion
description: Flexbox is a one-dimensional layout system, which places content either horizontally or vertically, with optional wrapping.
```

captainbrosset:
One comment left to resolve, but then the PR should be about ready to merge.

feature-group-definitions/flexbox.yml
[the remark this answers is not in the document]
foolip: I think it's perfectly fine to not say "CSS" here. If we really want it, we could say "CSS layout system".

ddbeck (approved):
Once the dists are refreshed, this should be ready to go. Thank you!

PR #754, opened 2024-03-28
Adding missing feature descriptions - Batch 5

captainbrosset:


feature-group-definitions/grid.yml
```diff
@@ -1,4 +1,5 @@
 name: Grid
+description: CSS Grid Layout is a two-dimensional layout system, which lays content out in rows and columns.
```
foolip, line 2:
```suggestion
description: CSS Grid is a two-dimensional layout system, which lays content out in rows and columns.
```

feature-group-definitions/has.yml
```diff
@@ -1,4 +1,5 @@
 name: ":has()"
+description: The `:has()` CSS pseudo-class selects an element if any of the selectors passed as parameters would match at least one element.
```
foolip, line 2: Can you move this into a separate PR together with `:is()` and `:where()`? I have some thoughts that I think should apply to all, and hopefully that'll be more effective done as one.

feature-group-definitions/html-media-capture.yml
```diff
@@ -1,4 +1,5 @@
 name: HTML media capture
+description: The `capture` HTML attribute for `<input type="file">` elements allows the user to, optionally, capture media using the device's camera, microphone, or other media-capture capabilities.
```
foolip, line 2: Trying to simplify a bit to the concrete:

```suggestion
description: The `capture` HTML attribute for `<input type="file">` elements allows the user to capture media using the device's camera or microphone.
```

If there's something beyond camera or microphone implemented in some browser, we can name that directly.

feature-group-definitions/http11.yml
```diff
@@ -1,4 +1,5 @@
 name: HTTP/1.1
+description: HTTP/1.1 is a revision of the HTTP network protocol used by the World Wide Web.
```
foolip, line 2: Hmm, this feature might be hard to explain. There was both HTTP 0.9 and 1.0, so this isn't the earliest version of the protocol, but it is the oldest worth describing I think.

Maybe...

```suggestion
description: HTTP/1.1 is a network protocol used by browsers and servers. It has been superseded by HTTP/2 and HTTP/3.
```

feature-group-definitions/http2.yml
```diff
@@ -1,3 +1,4 @@
 name: HTTP/2
+description: The HTTP/2 protocol is a major revision of the HTTP network protocol, providing improved performance and efficiency by using a single TCP connection to send multiple streams of data at once.
```
foolip, line 2: This and HTTP/3 looks good to me!

feature-group-definitions/idle-detection.yml
```diff
@@ -1,4 +1,5 @@
 name: Idle detection
+description: The Idle Detection API is used to notify a webpage of the user's idle, active, and locked state.
```
foolip, line 2:
```suggestion
description: The `IdleDetector` API is used to notify a webpage of the user's idle, active, and locked state.
```

feature-group-definitions/image-set.yml
```diff
@@ -1,3 +1,4 @@
 name: image-set()
+description: The `image-set()` CSS function defines a set of images for at different resolutions or pixel densities, which the browser can pick from, depending on the device capabilities.
```
foolip, line 2:
```suggestion
description: The `image-set()` CSS function provides a set of images at different resolutions or pixel densities, which the browser can pick from, depending on the device capabilities.
```

feature-group-definitions/import-maps.yml
```diff
@@ -1,4 +1,5 @@
 name: Import maps
+description: The `type="importmap"` attribute for the `<script>` HTML element indicates that the body of the element contains an import map. An import map is a JSON object that controls how the browser should resolve module specifiers when importing JavaScript modules.
```
foolip, line 2: I would avoid the word "body" which I associate with `<body>` or a fetch request/response body, while the children of an elements are its contents, or `textContent` property. How about:

```suggestion
description: A `<script type="importmap">` HTML element provides an import map as a JSON string. An import map controls how the browser should resolve module specifiers when importing JavaScript modules.
```

feature-group-definitions/has.yml
captainbrosset: See #862.

foolip (approved):

foolip:
In the interest of getting https://github.com/web-platform-dx/web-features/pull/761 done ASAP I'm going to merge this and make a list of things that @ddbeck might want to review post facto and send tweaks for.

PR #756, opened 2024-03-28
Adding missing feature descriptions - Batch 7

captainbrosset:


feature-group-definitions/popover.yml
```diff
@@ -1,4 +1,5 @@
 name: Popover
+description: The Popover API is used to display content on top of other page content, and is controlled either declaratively using HTML attributes, or via JavaScript.
```
foolip, line 2:
```suggestion
description: The `popover` HTML attribute creates an overlay to display content on top of other page content. Popovers can be shown declaratively using only HTML, or using the `showPopover()` method. Also known as a popup.
```

feature-group-definitions/read-write-pseudos.yml
```diff
@@ -1,4 +1,5 @@
-name: ":read and :write"
+name: ":read-only and :read-write"
+description: The `:read-only` and `:read-write` CSS pseudo-classes match elements that are read-only or read-write, respectively, such as when `<input>` or `<textarea>` elements have the `readonly` attribute set or not set, respectively.
```
foolip, line 2: The sentence is a bit long. Trying this:

```suggestion
description: The `:read-only` and `:read-write` CSS pseudo-classes match elements that are read-only or read-write, respectively. For example, `:read-only` matches `<input>` and `<textarea>` elements with the `readonly` attribute.
```

feature-group-definitions/registered-custom-properties.yml
```diff
@@ -1,4 +1,5 @@
 name: Registered custom properties
+description: The `CSS.registerProperty()` static method, and the `@property` CSS at-rule register custom properties which types and behaviors can be defined.
```
foolip, line 2:
```suggestion
description: The `CSS.registerProperty()` static method and the `@property` CSS at-rule register custom properties for which types and behaviors can be defined.
```

feature-group-definitions/sanitizer.yml
```diff
@@ -1,2 +1,3 @@
 name: Sanitizer
+description: The Sanitizer API sanitizes untrusted strings of HTML, Document and DocumentFragment objects. After sanitization, unwanted elements or attributes are removed, and the returned objects can safely be inserted into a document's DOM.
```
foolip, line 2: This will merge into HTML and may no longer be called Sanitizer API after that. Can it be described in terms of `setHTML()` and `parseHTML()`?

feature-group-definitions/scheduler.yml
```diff
@@ -1,4 +1,5 @@
 name: Scheduler API
+description: The Prioritized Task Scheduling API provides a way to prioritize all tasks belonging to an application.
```
foolip, line 2:
```suggestion
description: The `scheduler` API provides a way to prioritize all tasks belonging to an application.
```

feature-group-definitions/scope.yml
```diff
@@ -1,3 +1,4 @@
 name: "@scope"
+description: The `@scope` CSS at-rule defines the scope for a group of rules.
```
foolip, line 2:
```suggestion
description: The `@scope` CSS at-rule sets the scope for a group of rules.
```

feature-group-definitions/scroll-driven-animations.yml
```diff
@@ -1,4 +1,5 @@
 name: Scroll-driven animations
+description: CSS scroll-driven animations are a type of CSS animations that don't run over time, but instead are driven by the user's scroll position through the page.
```
foolip, line 2: I think it applies to all scroll containers, so dropping "through the page":

```suggestion
description: CSS scroll-driven animations are a type of CSS animations that don't run over time, but are instead driven by the user's scroll position.
```

feature-group-definitions/scroll-into-view.yml
```diff
@@ -1,4 +1,5 @@
 name: scrollIntoView()
+description: An element's `scrollIntoView()` method scrolls the element's ancestor containers such that the element is visible to the user.
```
foolip, line 2:
```suggestion
description: The `scrollIntoView()` method scrolls an element's ancestor containers such that the element is visible to the user.
```

foolip (approved):

foolip (approved):

PR #757, opened 2024-03-28
Adding missing feature descriptions - Batch 8

captainbrosset:


feature-group-definitions/scroll-snap.yml
```diff
@@ -1,4 +1,5 @@
 name: Scroll snap
+description: CSS scroll snap properties control the panning and scrolling behavior within a scroll container.
```
foolip, line 2: Since the properties aren't named:

```suggestion
description: CSS scroll snap controls the panning and scrolling behavior within a scroll container.
```

Also paging @flackr for review of this description.

feature-group-definitions/search-input-type.yml
```diff
@@ -1,4 +1,5 @@
 name: '<input type="search">'
+description: The `<input>` HTML element with the `type="search"` attribute represents a text field that's used for users to enter search queries, and that might be styled differently by the browser.
```
foolip, line 2:
```suggestion
description: The `<input>` HTML element with the `type="search"` attribute represents a text field for search queries, which might be styled differently by the browser.
```

feature-group-definitions/search.yml
```diff
@@ -1,4 +1,5 @@
 name: <search>
+description: The `<search>` HTML element is a container that represents the parts of the web page with search functionality.
```
foolip, line 2:
```suggestion
description: The `<search>` HTML element is a container representing the parts of the web page with search functionality.
```

feature-group-definitions/server-timing.yml
```diff
@@ -1,4 +1,5 @@
 name: Server timing
+description: Server timing is part of the Performance API and allows to annotate network requests with server timing information.
```
foolip, line 2: How would we name https://w3c.github.io/resource-timing/#dom-performanceresourcetiming if it was added as a feature? I think we should use that name here, and later we might even be able to cross-link features.

I think "performance resource timing API" and "`PerformanceResourceTiming` API" would both work. Picking one:

```suggestion
description: The `serverTiming` property of the `PerformanceResourceTiming` API contains server timing information about network requests.
```

feature-group-definitions/set-methods.yml
```diff
@@ -1,4 +1,5 @@
 name: Set methods
+description: "`difference()`, `intersection()`, `isDisjointFrom()`, `isSubsetOf()`, `isSupersetOf()`, `symmetricDifference()`, and `union()` methods of the JavaScript `Set` object."
```
foolip, line 2:
```suggestion
description: "The `difference()`, `intersection()`, `isDisjointFrom()`, `isSubsetOf()`, `isSupersetOf()`, `symmetricDifference()`, and `union()` methods of the JavaScript `Set` object performs operations between two sets."
```

feature-group-definitions/shadow-dom.yml
```diff
@@ -1,4 +1,5 @@
 name: Shadow DOM
+description: Shadow DOM is a part of Web Components. It allows for attaching encapsulated "shadow" DOM trees to the DOM. A shadow DOM tree is rendered separately from the main DOM, avoiding the component's scripts and styles to collide with other parts of the document.
```
foolip, line 2: This is one where I think a "X does Y" description isn't practical. I like this approach, just tweaking:

```suggestion
description: Shadow DOM allows you to attach encapsulated "shadow" DOM trees to elements. A shadow DOM tree is a separate component, isolated from the scripts and styles in other parts of the document. This is a part of Web Components.
```

feature-group-definitions/slot.yml
```diff
@@ -1,4 +1,5 @@
 name: <slot>
+description: The `<slot>` HTML element, part of Web Components, is a placeholder inside a web component where consumers of the component can insert their own markup.
```
foolip, line 2:
```suggestion
description: The `<slot>` HTML element is a placeholder inside a web component where consumers of the component can insert their own markup.
```

Later we can link "web component" to that group or feature, I think.

feature-group-definitions/search.yml
captainbrosset: I will not die on that hill, but I prefer my original phrasing. It feels simpler to understand. I don't think all languages use the "ing" form, which means that it might take a bit longer for some people to parse that sentence.

feature-group-definitions/scroll-snap.yml
captainbrosset: Looks good to me, but leaving this one open for @flackr to comment.

feature-group-definitions/search.yml
foolip: I suspect that @ddbeck would agree with you. In any even I'm happy with your original wording, I guess I went too far trying to make it shorter.

feature-group-definitions/scroll-snap.yml
foolip: I'll ping @flackr, but let's not block on that, I'd love to get this merged sooner rather than later, when @ddbeck has reviewed.

flackr: This text looks good to me!

feature-group-definitions/server-timing.yml
foolip: @captainbrosset did you mean to leave this suggestion out?

foolip (approved):

feature-group-definitions/server-timing.dist.yml
```diff
@@ -2,6 +2,7 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 name: Server timing
+description: Server timing is part of the Performance API and allows to annotate network requests with server timing information.
```
foolip, line 5:
```suggestion
description: The `serverTiming` property of the `PerformanceResourceTiming` API contains server timing information about network requests.
```

PR #758, opened 2024-03-28
Adding missing feature descriptions - Batch 9

captainbrosset:


feature-group-definitions/structured-clone.yml
```diff
@@ -1,4 +1,5 @@
 name: structuredClone()
+description: The global `structuredClone()` JavaScript method creates a deep clone of an object by using the structured clone algorithm. The method also allows for transferring values from the original object to the cloned object.
```
foolip, line 2: Explaining tranferables is a bit tricky. It's not an optional use case you might have, it's something you have to do because `structuredClone()` will throw if you don't. Trying:

```suggestion
description: The `structuredClone()` global method creates a deep copy of an object. Values that cannot be cloned can instead be transferred, making the original value no longer usable.
```

I dropped the algorithm reference because it's a spec concept that most devs won't know. If we could link it that might be useful, but we can't now.

feature-group-definitions/svg2-script-html-equivalence.yml
```diff
@@ -1,4 +1,5 @@
 name: SVG <script> async and defer
+description: The `async` and `defer` attributes of the `<script>` SVG element allow to control the load and execution of scripts in SVG documents.
```
foolip, line 2:
```suggestion
description: The `async` and `defer` attributes of the `<script>` SVG element control the loading and execution of scripts in SVG documents.
```

feature-group-definitions/tabindex.yml
```diff
@@ -1,4 +1,5 @@
 name: tabindex
+description: The `tabindex` HTML attribute makes elements focusable, and defines the elements' relative ordering for sequential focus navigation.
```
foolip, line 2:
```suggestion
description: The `tabindex` HTML attribute makes elements focusable, and defines the element's relative ordering for sequential focus navigation.
```

feature-group-definitions/template.yml
```diff
@@ -1,4 +1,5 @@
 name: <template>
+description: The `<template>` HTML element holds HTML fragments, which can be cloned and inserted into the document by JavaScript.
```
foolip, line 2:
```suggestion
description: The `<template>` HTML element holds HTML fragments which you can clone and insert into the document using JavaScript.
```

feature-group-definitions/speech-recognition.yml
```diff
@@ -1,4 +1,5 @@
 name: Speech recognition
+description: The `SpeechRecognition` interface of the Web Speech API controls the speech recognition service that might be available in the browser.
```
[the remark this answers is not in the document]
captainbrosset: If we're going to greate a Web Speech group, should we leave out the last sentence?

```suggestion
description: The `SpeechRecognition` API converts human speech from a microphone to text.
```

feature-group-definitions/speech-synthesis.yml
```diff
@@ -1,4 +1,5 @@
 name: Speech synthesis
+description: The `SpeechSynthesis` interface of the Web Speech API controls the speech synthesis service that might be available in the browser.
```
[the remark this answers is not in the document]
captainbrosset:
```suggestion
The `SpeechSynthesis` API converts text into audio using a synthetic voice.
```

feature-group-definitions/tabindex.yml
captainbrosset: My original spelling: `elements'` relied on the fact that elements was plural (hence adding a trailing apostrophe). If we keep the first part of the sentence plural, but then switch to singular for the second part, it might be harder to parse.

In this suggestion, I'm making everything singular:

```suggestion
description: The `tabindex` HTML attribute make an element focusable, and defines the element's relative ordering for sequential focus navigation.
```

feature-group-definitions/speech-recognition.yml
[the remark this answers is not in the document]
foolip: Yeah, that's just as well.

feature-group-definitions/tabindex.yml
foolip: Oh, I took it to be singular even though a plural "elements" preceded it. I think your new suggestion work, let's go with that.

foolip (approved):

PR #759, opened 2024-03-28
Adding missing feature descriptions - Batch 10

captainbrosset:


feature-group-definitions/text-wrap-balance.yml
```diff
@@ -1,4 +1,5 @@
 name: "text-wrap: balance"
+description: "The `text-wrap: balance` CSS declaration wraps text content across lines, while using an algorithm that balances that number of characters on each line, enhancing legibility."
```
foolip, line 2:
```suggestion
description: "The `text-wrap: balance` CSS property value balances the length of each line when text is broken into multiple lines. Also known as headline balancing."
```

(I don't think "CSS property value" is great, but we have it already, and I'd like to change them all together in a single PR.)

feature-group-definitions/text-wrap-pretty.yml
```diff
@@ -1,4 +1,5 @@
 name: "text-wrap: pretty"
+description: "The `text-wrap: pretty` CSS declaration wraps text content across lines, while using an algorithm that favors better layout over speed."
```
foolip, line 2:
```suggestion
description: "The `text-wrap: pretty` CSS property value prioritizes better layout over speed when text is broken into multiple lines."
```

feature-group-definitions/transition-behavior.yml
```diff
@@ -1,4 +1,5 @@
 name: transition-behavior
+description: The `transition-behavior` CSS property specifies whether transitions will be started for properties whose animation behavior is discrete. Properties with a discrete animation type can't be interpolated and swap from their start value to the end value at 50%.
```
foolip, line 2: Since `transition-behavior: allow-discrete` is the only value other than the default, we can spell it out:

```suggestion
description: The `transition-behavior: allow-discrete` CSS property value allows transitions for properties whose animation behavior is discrete. Such properties can't be interpolated and swap from their start value to the end value at 50%.
```

feature-group-definitions/trusted-types.yml
```diff
@@ -1,4 +1,5 @@
 name: Trusted types
+description: The Trusted Types API is used to lock down insecure parts of the DOM API, to prevent client-side cross-site scripting (XSS) attacks.
```
foolip, line 2: Trusted types feels too embedded into lots of parts of the platform to be called an API on its own.

```suggestion
description: Trusted types allow you to lock down insecure parts of the DOM API and prevent client-side cross-site scripting (XSS) attacks.
```

feature-group-definitions/view-transitions.yml
```diff
@@ -1,4 +1,5 @@
 name: View transitions
+description: The View Transitions API transitions between different states of a document, or between different documents.
```
foolip, line 2: I think we'll need another feature for MPA view transitions, which is the "or between different documents" bit here:

```suggestion
description: View transitions allow you to create animated visual transitions between different states of a document.
```

foolip: Actually, ignore this suggestion. The SPA shipped first in Chrome:
https://chromiumdash.appspot.com/commit/c03f83e180726ba898a2191f616aa7df9c581510

But perhaps we will treat that as a partial implementation later, rather than having two features.

Leaving only the good parts of my suggestion:

```suggestion
description: View transitions allow you to create animated visual transitions between different states of a document, or between different documents.
```

feature-group-definitions/temporal.yml
```diff
@@ -1,3 +1,4 @@
 name: Temporal
+description: The Temporal API provides a way to work with dates and times in JavaScript.
```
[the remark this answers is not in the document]
captainbrosset: I like the comparison with `Date`, since someone might be confused about why there are two ways to deal with Dates. However, I think I prefer the `The X thing does Y` than `X is a thing for Y`. It feels more consistent with the other descriptions we have.

```suggestion
description: The `Temporal` API allows you to work with dates, times, time zones, and durations. It is more powerful than the `Date` API.
```

feature-group-definitions/text-wrap-balance.yml
captainbrosset: Usually, when I talk about these things, I say that `text-wrap` is the property, that `balance` is the value, and that `text-wrap: balance` is the declaration. I don't really like the phrase "property value" because I always someone would be confused by it and think we're only talking about `balance`, or only about `text-wrap`.

That being said, I agree that it's better to be consistent for now. As you said, we can always bulk-change if we think we have a better proposal in the future.

feature-group-definitions/temporal.yml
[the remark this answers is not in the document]
foolip: That's a better description all around, let's go with that!

feature-group-definitions/transition-behavior.yml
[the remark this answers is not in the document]
foolip: I think using the CSS property name is a good default. Some features are named for a property + value pair, and in this case that would be `transition-behavior-allow-discrete.yml`. But we don't know for sure that there will ever be another interesting value, so I think sticking with just `transition-behavior.yml` until that time works.

foolip (approved):

PR #762, opened 2024-03-29
New CSS Custom Highlight API feature

captainbrosset:
A few resources about the feature:

* https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API
* https://css-tricks.com/css-custom-highlight-api-early-look/
* https://www.youtube.com/watch?v=1qldqyT324o

I'll open a BCD PR to add the corresponding tags there.


feature-group-definitions/custom-highlight
```diff
@@ -0,0 +1,3 @@
+name: Custom Highlight
```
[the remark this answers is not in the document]
captainbrosset: The spec, and MDN, call this CSS Custom Highlight API, singular. I guess Custom highlights does make sense when you think about what the feature does. But I don't think anyone talks about this API in those terms.

feature-group-definitions/custom-highlight
```diff
@@ -0,0 +1,3 @@
...
+description: The CSS Custom Highlight API allows styling arbitrary text ranges on a document, without the need to add extra elements to the DOM.
```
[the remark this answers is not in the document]
captainbrosset: I like this, but again, I'm slightly troubled by the fact that "Custom highlights" isn't a thing that exists. We're inventing a new term here, which might very well not map how people have been calling this so far, or how people will later call it.

feature-group-definitions/custom-highlight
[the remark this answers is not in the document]
captainbrosset: Whoops, sorry about this. Fixing now.

foolip:
In https://github.com/web-platform-dx/web-features/pull/771 I also suggest a name for a feature that nobody else has used, by following the style that we have established. I even go so far as to claim it's "also known as" Compute Pressure API even though I think that's the main name it is known by.

My confidence that we should almost always omit "API" is decreasing. @ddbeck in the interest of making progress, perhaps we should collect some names that we're not so sure about in an issue, but keep adding features even when in doubt about the name?

feature-group-definitions/custom-highlight.yml
```diff
@@ -0,0 +1,3 @@
+name: Custom highlights
```
foolip, line 1: Commenting on the first line for lack of a better place. I think the shortest identifier we can get away with here is just `highlight` (`highlight.yml`) since that matches both the `Highlight` constructor and the `::highlight()` pseudo-element.

captainbrosset: Very surprising to me, because `highlight` is kind of a generic name, but it indeed looks like nothing else on the web platform uses that name yet. So that might work.

captainbrosset:
> My confidence that we should almost always omit "API" is decreasing. @ddbeck in the interest of making progress, perhaps we should collect some names that we're not so sure about in an issue, but keep adding features even when in doubt about the name?

I'm also not super confident with our current naming strategy. Keeping names as short as possible, and omitting common terms feels like unnecessary rules at this point, creating more problems than they're solving.

feature-group-definitions/custom-highlight.yml
foolip: I take it you're not a fan of some of our shortest identifiers like `search`, `set`, or `where`. Perhaps it was a mistake to try to make identifiers as short as possible and use a bit of syntax where possible. But I think `highlight` fits with the current approach, so let's go with that for now.

foolip:
> > My confidence that we should almost always omit "API" is decreasing. @ddbeck in the interest of making progress, perhaps we should collect some names that we're not so sure about in an issue, but keep adding features even when in doubt about the name?
> 
> I'm also not super confident with our current naming strategy. Keeping names as short as possible, and omitting common terms feels like unnecessary rules at this point, creating more problems than they're solving.

Can you comment on https://github.com/web-platform-dx/web-features/issues/548 with what you think we should do for "API" specifically?

I do think that the current rules solve at least one real issue, which is when we should include "CSS" in the shortname. The current answer is "never" and easy to apply consistently.

feature-group-definitions/custom-highlight.yml
foolip: @captainbrosset are you OK with renaming this to `highlight.yml`? I think that will stand the test of time even if we change the name and description.

captainbrosset:
@ddbeck @foolip are we ok to merge this one as is?

captainbrosset:
By the way, the corresponding BCD PR is https://github.com/mdn/browser-compat-data/pull/22731

foolip (approved):

foolip:
Thanks @captainbrosset! Let's keep figuring out naming and descriptions in https://github.com/web-platform-dx/web-features/issues/822. I'm not happy with all of the decisions we've made so far, but it's also quite hard to articulate guidelines that don't have any drawbacks...

PR #770, opened 2024-04-02
`class-syntax`: Add `super` operator

ddbeck:
I _think_ this belongs with the rest of the class syntax stuff (excluding the private class properties stuff).

Inspired by looking at this feature for https://github.com/web-platform-dx/web-features/pull/751/files#r1547468185.

### DIFF
--- a/feature-group-definitions/class-syntax.yml
+++ b/feature-group-definitions/class-syntax.yml
@@ -18,4 +18,5 @@ compat_features:
   - javascript.classes.constructor
   - javascript.classes.extends
   - javascript.classes.static
+  - javascript.operators.super
   - javascript.statements.class


foolip:
It looks like support for `super` came slightly before `class` in both Chrome and Safari:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/class#browser_compatibility

But in Edge and Firefox they came at the same time. I think that treating it as part of classes makes sense.

Elchi3 (approved):
LGTM, thank you! 👍

PR #774, opened 2024-04-02
Fix name of :read-only and :read-write feature

foolip:


captainbrosset:
For info, this change also exists in https://github.com/web-platform-dx/web-features/pull/756/files#diff-ede10c87bb77559a8bbc80777b753de0cce10d13215d5c1d17299032ff16f844

PR #777, opened 2024-04-03
Rename fonts to font-face

foolip:
This matches the syntax, and "fonts" is very generic.


foolip:
I'm sending this because I had trouble remembering what it is from just the identifier, which isn't true of any other feature.

ddbeck (approved):

PR #784, opened 2024-04-04
`compute-baseline`: fix incorrect status for future-dated "high" statuses

ddbeck:
Presently, an actual Baseline `"low"` status can result in an erroneous `"high"` status with a future-dated `baseline_high_date`. For example, given the keystone date of 2024-04-04 (today), you'd get this back:

```js
{
  baseline: "high",
  baseline_low_date: "2024-04-04",
  baseline_high_date: "2026-10-04"
}
```

This PR fixes the faulty check for future `baseline_high_date` values and puts in some tests to prevent it from happening again.

### DIFF
--- a/packages/compute-baseline/src/baseline/index.test.ts
+++ b/packages/compute-baseline/src/baseline/index.test.ts
@@ -77,7 +77,7 @@ describe("computeBaseline", function () {
 });
 
 describe("keystoneDateToStatus()", function () {
-  it('returns `"low"` for recent dates', function () {
+  it('returns "low" for recent dates', function () {
     const status = keystoneDateToStatus(
       Temporal.Now.plainDateISO().subtract({ days: 7 }),
     );
--- a/packages/compute-baseline/src/baseline/index.test.ts
+++ b/packages/compute-baseline/src/baseline/index.test.ts
@@ -3,7 +3,8 @@ import assert from "node:assert/strict";
 import * as chai from "chai";
 import chaiJestSnapshot from "chai-jest-snapshot";
 
-import { computeBaseline } from ".";
+import { computeBaseline, keystoneDateToStatus } from ".";
+import { Temporal } from "@js-temporal/polyfill";
 
 chai.use(chaiJestSnapshot);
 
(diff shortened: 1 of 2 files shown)


packages/compute-baseline/src/baseline/index.test.ts
```diff
@@ -74,3 +75,37 @@ describe("computeBaseline", function () {
...
+  it('returns `"low"` for recent dates', function () {
+    const status = keystoneDateToStatus(
+      Temporal.Now.plainDateISO().subtract({ days: 7 }),
+    );
+    assert.equal(status.baseline, "low");
+    assert.equal(typeof status.baseline_low_date, "string");
+    assert.equal(status.baseline_high_date, null);
+  });
+
+  it('returns `"high"` for long past dates', function () {
```
foolip, line 89: Since these strings will never be treated as markdown, I like this better:

```suggestion
  it('returns "high" for long past dates', function () {
```

packages/compute-baseline/src/baseline/index.test.ts
```diff
@@ -74,3 +75,37 @@ describe("computeBaseline", function () {
...
+    const status = keystoneDateToStatus(Temporal.PlainDate.from("2020-01-01"));
+    assert.equal(status.baseline, "high");
+    assert.equal(typeof status.baseline_low_date, "string");
+    assert.equal(typeof status.baseline_high_date, "string");
+  });
+
+  it("returns `false` for future dates", function () {
```
foolip, line 96:
```suggestion
  it("returns false for future dates", function () {
```

packages/compute-baseline/src/baseline/index.test.ts
```diff
@@ -74,3 +75,37 @@ describe("computeBaseline", function () {
...
+    const status = keystoneDateToStatus(
+      Temporal.Now.plainDateISO().add({ days: 90 }),
+    );
+    assert.equal(status.baseline, false);
+    assert.equal(status.baseline_low_date, null);
+    assert.equal(status.baseline_high_date, null);
+  });
+
+  it("returns `false` for null dates", function () {
```
foolip, line 105:
```suggestion
  it("returns false for null dates", function () {
```

foolip (approved):

packages/compute-baseline/src/baseline/index.test.ts
```diff
@@ -74,3 +75,37 @@ describe("computeBaseline", function () {
     assert.equal(result.baseline_high_date, "2018-01-29"); // 30 months later
   });
 });
+
+describe("keystoneDateToStatus()", function () {
+  it('returns `"low"` for recent dates', function () {
```
ddbeck, line 80:
```suggestion
  it('returns "low" for recent dates', function () {
```

PR #785, opened 2024-04-04
Add utilities for finding BCD tag data

ddbeck:
This PR adds two new helpers to compute-baseline:

- A `walk()` generator on `Compat`, to offer something a bit like BCD's `traverse` command. This code (mostly) already exists, just not attached to the `Compat` interface.
- A `tags` getter on feature objects, to have checking for an array of tags in a BCD `__compat` object in one place.

This is the last of the table setting for generating statuses from authored YAML files.

### DIFF
--- a/packages/compute-baseline/src/browser-compat-data/feature.ts
+++ b/packages/compute-baseline/src/browser-compat-data/feature.ts
@@ -38,6 +38,15 @@ export class Feature {
     return `[Feature ${this.id}]`;
   }
 
+  /**
+   * Return the feature's tags as an array (wthether there are any tags or not).
+   */
+  get tags(): string[] {
+    return Array.isArray(this.data.__compat?.tags)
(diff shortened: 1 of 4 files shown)


packages/compute-baseline/src/browser-compat-data/feature.ts
```diff
@@ -38,6 +38,15 @@ export class Feature {
     return `[Feature ${this.id}]`;
   }
 
+  /**
+   * Return the feature's tags as an array (wthether there are any tags or not).
+   */
+  get tags(): string[] {
+    return Array.isArray(this.data.__compat?.tags)
+      ? this.data.__compat?.tags
+      : [];
```
ddbeck, line 47: This is probably less literal-minded:

```suggestion
    return this.data.__compat?.tags ?? [];
```

PR #787, opened 2024-04-04
Add "dist" YAML files for generated statuses and compat features

ddbeck:
This PR demonstrates an approach to generating statuses and consuming BCD data, first described in https://github.com/web-platform-dx/web-features/pull/110#issuecomment-1516080168

Here's the intended workflow:

1. If desired, generate a `dist.yml` file with `npx tsx ./dist.ts feature-group-definitions/<some-target-id>.yml`.
  
   (I didn't generate for _everything_ because the PR would be huge. We can opt-in to this, one file at a time.)

2. Review `<some-target-id>.dist.yml`. If anything looks off, override the results by editing `<some-target-id>.yml` (NOT the dist file). Then regenerate `<some-target-id>.dist.yml`.

3. When satisfied, merge the results.

From then on, the package will use `<some-target-id>.dist.yml` as the source of data about the feature.

I'll self review to highlight some interesting pieces.

Closes #110.


dist.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Generate a dist YAML document from a feature definition YAML file path.
+ *
+ * This passes through most of the contents of a feature definition. If
+ * possible, it fills in `compat_features` from @mdn/browser-compat-data and, if
+ * successful, generates a `status` block.
+ */
+function toDist(fp: string): YAML.Document {
+  const yaml = YAML.parseDocument(fs.readFileSync(fp, { encoding: "utf-8" }));
```
ddbeck, line 53: If this function looks a bit complicated, it is! This works the way it does to carry forward the order and comments from the authored YAML into the generated YAML.

We could probably simplify this a bit if we wrote a separate formatter to run on _both_ authored and generated files (e.g., to ensure a specific key order, line lengths, etc.)—but that's bit of a separate task and not on the critical path to generating statuses.

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+          .toString(),
+      );
+
+      // …then fix the order.
+      assert(YAML.isMap(statusNode));
+      const highDateNode = statusNode.items.pop();
+      const targetIndex = statusNode.items.findIndex(
+        (item) => item.key === "baseline_low_date",
+      );
+      statusNode.items.splice(targetIndex, 0, highDateNode);
```
ddbeck, line 129: This will replace `index.ts`'s date calculations (eventually, when all files are dist-able).

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+  return features;
+}
+
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
+  for (const y of process.argv.slice(2)) {
+    updateDistFile(y);
+  }
+}
```
ddbeck, line 159: This ought to use the yargs machinery we use elsewhere, for good help messages and such. But I wanted to confirm that we're happy with the overall approach before putting in the work.

feature-group-definitions/aborting.dist.yml
```diff
@@ -0,0 +1,28 @@
...
+  baseline_low_date: 2019-03-25
+  baseline_high_date: 2021-09-25
+  support:
+    chrome: "66"
+    chrome_android: "66"
+    edge: "16"
+    firefox: "57"
+    firefox_android: "57"
+    safari: "12.1"
+    safari_ios: "12.2"
```
ddbeck, line 20: This file demonstrates consuming a web-features `compat_features` list to generate a brand-new status. ✨

feature-group-definitions/anchor-positioning.dist.yml
```diff
@@ -0,0 +1,8 @@
+# Generated from: anchor-positioning.yml
```
ddbeck, line 1: This file demonstrates dist-ing a feature where there's no status, `compat_features`, or tags in BCD.

feature-group-definitions/array-at.dist.yml
```diff
@@ -0,0 +1,25 @@
...
+    chrome: "92"
+    chrome_android: "92"
+    edge: "92"
+    firefox: "90"
+    firefox_android: "90"
+    safari: "15.4"
+    safari_ios: "15.4"
+compat_features:
+  - javascript.builtins.Array.at
+  - javascript.builtins.TypedArray.at
```
ddbeck, line 25: This file demonstrates dist-ing a feature that already has both a status and `compat_features` list. We don't do anything here.

feature-group-definitions/cap.dist.yml
```diff
@@ -0,0 +1,18 @@
+# Generated from: cap.yml
+# Do not edit this file by hand. Edit the source file instead!
+# Last modified: 2024-04-04T09:35:23.998323918+00:00[UTC]
+
+name: cap unit
+description: The CSS `cap` unit corresponds to the height of Latin capital letters.
+spec: https://drafts.csswg.org/css-values-4/#cap
```
ddbeck, line 7: This file demonstrates the real dream. Only these lines were authored! The status and compat features came from `compute-baseline` and BCD.

feature-group-definitions/webdriver-bidi.dist.yml
```diff
@@ -0,0 +1,19 @@
...
+# Last modified: 2024-04-04T09:37:18.561438534+00:00[UTC]
+
+name: WebDriver BiDi
+spec: https://w3c.github.io/webdriver-bidi/
+# WebDriver BiDi is not in BCD, see https://github.com/mdn/browser-compat-data/issues/20207
+# The browser releases listed here are from:
+# https://developer.chrome.com/blog/webdriver-bidi-2023
+# https://bugs.chromium.org/p/chromedriver/issues/detail?id=4016
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1753997
+status:
```
ddbeck, line 12: This file demonstrates preserving comments in the original file.

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  const currentDist = fs.existsSync(distPath)
+    ? YAML.parseDocument(fs.readFileSync(distPath, { encoding: "utf-8" }))
+    : undefined;
+  const newDist = toDist(fp);
+
+  currentDist.commentBefore = undefined; // Suppress the generated comment for comparisons
+
+  // Write a dist only if it's new or modified
+  if (!currentDist || stringify(currentDist) !== stringify(newDist)) {
```
foolip, line 32: Can you save `stringify(newDist)` in a variable to not to it twice?

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+    newDist.commentBefore = [
+      `Generated from: ${id}.yml`,
+      `Do not edit this file by hand. Edit the source file instead!`,
+      `Last modified: ${Temporal.Now.zonedDateTimeISO("UTC")}`,
```
foolip, line 36: This date is going to lead to merge conflicts any time there are two reviews in flight. Let's use `git log` to find the last modified date instead, when we need it.

feature-group-definitions/webdriver-bidi.dist.yml
foolip: I thought this would be hard/impossible, but here it is, it works!

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+    yaml.contents.items.splice(
+      targetIndex,
+      0,
+      statusNode as YAML.Pair<any, any>,
+    );
+  } else {
+    // If there is a support block, then we set its baseline_high_date, if
+    // applicable.
+
+    // TODO: consider checking if a generated status would be the same, and warn
```
foolip, line 98: This would be nice, yes.

dist.ts
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Get all of the features with a given @mdn/browser-compat-data tag.
+ *
+ * @param {string} tag A tag (excluding the `web-features:` prefix).
+ */
+function lookupTag(tag: string): Feature[] {
+  const namespaceTag = `web-features:${tag}`;
+
+  const features = [];
+  for (const feature of compat.walk()) {
```
foolip, line 147: This definitely works, but how about walking BCD just once and creating a reverse mapping from web-features identifier to `Feature`? Lookup would be very quick.

dist.ts
foolip: I am very happy.

feature-group-definitions/aborting.dist.yml
```diff
@@ -0,0 +1,28 @@
...
+caniuse: abortcontroller
+status:
+  baseline: high
+  baseline_low_date: 2019-03-25
+  baseline_high_date: 2021-09-25
+  support:
+    chrome: "66"
+    chrome_android: "66"
+    edge: "16"
+    firefox: "57"
```
foolip, line 17: Shouldn't this be 53? https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility

feature-group-definitions/anchor-positioning.dist.yml
foolip: It's nice to see this working, but can you drop it from this PR? It would be nice to treat the absence of a `.dist.yml` as a burndown list for statuses being available.

feature-group-definitions/array-at.dist.yml
```diff
@@ -0,0 +1,25 @@
...
+description: Array at() and typed array at() methods
+spec: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.at
+snapshot: ecmascript-2022
+group:
+  - arrays
+  - typed-arrays
+status:
+  baseline: low
+  baseline_low_date: 2022-03-14
+  support:
```
foolip, line 15: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility agrees.

feature-group-definitions/cap.dist.yml
```diff
@@ -0,0 +1,18 @@
...
+status:
+  baseline: false
+  support:
+    chrome: "118"
```
foolip, line 11: https://developer.mozilla.org/en-US/docs/Web/CSS/length#browser_compatibility says 117

ddbeck: `api.CSS.cap_static` [says 118](https://developer.mozilla.org/en-US/docs/Web/API/CSS#browser_compatibility) and that seems to be the truth of it: https://github.com/mdn/browser-compat-data/pull/20984#pullrequestreview-1724941202

feature-group-definitions/aborting.dist.yml
ddbeck: Opera is 53. Already, the machines are reading the tables better than we are 😄

dist.ts
ddbeck: I fixed this alongside dropping the date from comments (which is why I was doing it in the first place) in 7b358fb3d246ede8c98ddd57b2ac96959697256c.

dist.ts
ddbeck: Fixed, see: https://github.com/web-platform-dx/web-features/pull/787/files#r1553521058.

feature-group-definitions/anchor-positioning.dist.yml
ddbeck: Done with 54a8a10.

dist.ts
ddbeck: Good idea. I've done this in da1d3dec6a9ee6d02e4279544e8edf74fcbe9ac5. Would you take another look? I'm not sure I like the style of initializing the map very much.

feature-group-definitions/cap.dist.yml
foolip: I guess that's a small enough difference that we just go with it, even though web developers will use `cap` as a unit much more often than `CSS.cap()`.

feature-group-definitions/aborting.dist.yml
foolip: Excellent!

dist.ts
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * @param {string} fp The path to the human-authored YAML file.
+ */
+function updateDistFile(fp: string): void {
+  const { dir, name: id } = path.parse(fp);
+  const distPath = path.join(dir, `${id}.dist.yml`);
+
+  const dist = toDist(fp);
+  fs.writeFileSync(distPath, dist.toString({ lineWidth: 0 }));
+}
```
ddbeck, line 25: Since the header comment is stable, there's no need to read ahead—we can just write out the dist now.

dist.ts
```diff
@@ -0,0 +1,196 @@
...
+    const status = generated.statusByCompatFeaturesOverride ?? generated.status;
+    if (status) {
+      insertStatus(yaml, status);
+    }
+  }
+
+  insertHeaderComments(yaml, id);
+
+  return yaml;
+}
```
ddbeck, line 89: @foolip in trying to emit a warning when status overrides are needless, I "accidentally" refactored this to separate the "find the highest source of truth" bit from the "edit YAML" bit.

This doesn't change the results of the writing dists (at least for the set of files that I've generated dists for—a nice benefit of this approach is that Git now tells me whether I've modified those files). But the code is quite different. If you want me to back this out and submit it as a separate PR rather than re-review, let me know.

dist.ts
foolip: Oh. Ignore one of my review comments.

dist.ts
foolip: I've reviewed it, looks good!

PR #794, opened 2024-04-05
Remove compat keys for line-clamp

foolip:
We will most likely need to handle this feature manually and decide if we want it to cover just the unprefixed `line-clamp` or also `-webkit-line-clamp`. To avoid computing anything automatically, remove the compat keys.


ddbeck (approved):
Should I retire https://github.com/web-platform-dx/web-features/pull/503 as well? Or hold that open as a testing ground for notes?

PR #795, opened 2024-04-05
Remove duplicate BCD paths

foolip:
BackgroundFetchManager is a plain duplicate.

There is a separate feature for `<slot>` that has some of the same compat
keys as Shadow DOM. Leave those only for `<slot>`.


foolip:
I found these while updating and running https://github.com/web-platform-dx/web-features/pull/473.

Elchi3 (approved):
I guess there is a question in here if we want compat_keys to be in several feature-definitions but for now I think it would be good to not duplicate.

(If everything is tagged in BCD, these duplications become more obvious at authoring time)

foolip:
I think we should generally avoid tagging one BCD entry with multiple web-features tags. There might be exceptions, but very few.

PR #797, opened 2024-04-06
Generate .dist.yml where compat_features are sourced from BCD

foolip:
These are cases where the source YAML doesn't include compat_features,
but the generated .dist.yml does, based on BCD tags.


feature-group-definitions/ic.dist.yml
```diff
@@ -0,0 +1,17 @@
+# Generated from: ic.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: ic unit
+description: The CSS `ic` unit corresponds to the width of CJK ideographic characters.
+spec: https://drafts.csswg.org/css-values-4/#ic
+status:
+  baseline: false
```
foolip, line 8: This isn't baseline because of the `CSS.ic()` static method not being implemented in Firefox:
https://developer.mozilla.org/en-US/docs/Web/API/CSS#browser_compatibility

But this isn't really preventing its use in a vast majority of cases. I think we should remove api.CSS.ic_static from this feature or use an editorial override.

feature-group-definitions/lh.dist.yml
```diff
@@ -0,0 +1,17 @@
+# Generated from: lh.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: lh unit
+description: The CSS `lh` unit corresponds to the requested line height, the computed value of the `line-height` property. Some lines may be higher than this based on their content.
+spec: https://drafts.csswg.org/css-values-4/#lh
+status:
+  baseline: false
```
foolip, line 8: This is also affected by missing statics in https://developer.mozilla.org/en-US/docs/Web/API/CSS#browser_compatibility

feature-group-definitions/masonry.dist.yml
```diff
@@ -0,0 +1,15 @@
+# Generated from: masonry.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Masonry
+description: Masonry is a type of CSS grid layout where the items on one of the axes are tightly packed together, like brickwork, instead of leaving gaps to align across the other axis.
+spec: https://drafts.csswg.org/css-grid-3/
+status:
+  baseline: false
+  support: {}
```
foolip, line 9: This is correct.

feature-group-definitions/server-timing.dist.yml
```diff
@@ -0,0 +1,25 @@
+# Generated from: server-timing.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Server timing
+spec: https://w3c.github.io/server-timing/
+caniuse: server-timing
+status:
+  baseline: low
+  baseline_low_date: 2023-03-27
+  support:
```
foolip, line 10: This is a straightforward case of everything being the same versions:
https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing#browser_compatibility

feature-group-definitions/storage-buckets.dist.yml
```diff
@@ -0,0 +1,28 @@
+# Generated from: storage-buckets.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Storage buckets
+spec: https://wicg.github.io/storage-buckets/
+status:
+  baseline: false
+  support:
+    chrome: "122"
```
foolip, line 9: This is correct:
https://chromestatus.com/feature/5739224579964928
https://chromiumdash.appspot.com/commit/834e1568e27cb1f088627f3c9f6c36c2803115f0

feature-group-definitions/temporal.dist.yml
```diff
@@ -0,0 +1,312 @@
+# Generated from: temporal.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Temporal
+spec: https://tc39.es/proposal-temporal/
+caniuse: temporal
+status:
+  baseline: false
+  support: {}
```
foolip, line 9: https://caniuse.com/temporal agrees.

feature-group-definitions/web-bluetooth.dist.yml
```diff
@@ -0,0 +1,69 @@
+# Generated from: web-bluetooth.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Web Bluetooth
+description: The Web Bluetooth API enables selecting and communicating with nearby Bluetooth devices.
+spec: https://webbluetoothcg.github.io/web-bluetooth/
+caniuse: web-bluetooth
+status:
+  baseline: false
+  support: {}
```
foolip, line 10: This is wrong, needs splitting out and investigating.

feature-group-definitions/web-midi.dist.yml
```diff
@@ -0,0 +1,59 @@
+# Generated from: web-midi.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Web MIDI
+description: The Web MIDI API enables selecting MIDI input and output devices and sending and receiving MIDI messages.
+spec: https://webaudio.github.io/web-midi-api/
+caniuse: midi
+status:
+  baseline: false
+  support: {}
```
foolip, line 10: This is incorrect.

foolip (commented):
Self-review to verify or reject the generated statuses.

feature-group-definitions/edit-context.dist.yml
```diff
@@ -0,0 +1,50 @@
+# Generated from: edit-context.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: EditContext
+spec: https://w3c.github.io/edit-context/
+status:
+  baseline: false
+  support:
+    chrome: "121"
```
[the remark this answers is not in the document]
foolip: Fixed in https://github.com/mdn/browser-compat-data/pull/22779.

feature-group-definitions/server-timing.dist.yml
ddbeck: Like finding a four-leaf clover 😆

ddbeck (approved):

ddbeck:
@foolip rebase needed here, otherwise ready to go. Thank you!

PR #800, opened 2024-04-06
Fix error message to show BCD path of feature

foolip:


packages/compute-baseline/src/browser-compat-data/feature.ts
```diff
@@ -73,7 +73,7 @@ export class Feature {
 
       if (!(s instanceof RealSupportStatement)) {
         throw Error(
-          `${feature} contains non-real values. Cannot expand support.`,
+          `${this.id} contains non-real values. Cannot expand support.`,
```
ddbeck, line 76: What do you think of making this even more specific?

```suggestion
          `${this.id} contains non-real values for `${browser.name}`. Cannot expand support.`,
```

ddbeck (approved):
One little idea here, but fine to land as is

PR #801, opened 2024-04-06
Generate lots more .dist.yml files

foolip:

PR #802, opened 2024-04-06
Add .dist.yml for non-trivial cases

foolip:
These are cases that were excluded from
https://github.com/web-platform-dx/web-features/pull/801 due to some
complication, small or large.


foolip:
I'm letting this sit as a burndown draft to rebase over time. There will be conflicts if any of these files are added, and then I can double check if the issues from https://github.com/web-platform-dx/web-features/pull/801 were addressed.

feature-group-definitions/calc-constants.dist.yml
```diff
@@ -0,0 +1,19 @@
+# Generated from: calc-constants.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: calc() constants
+description: The `e`, `pi`, `infinity`, and `NaN` constants are accepted in CSS math functions such as `calc()`.
+spec: https://drafts.csswg.org/css-values-4/#calc-keywords
+status:
+  baseline: false
```
foolip, line 8: This will be fixed by https://github.com/mdn/browser-compat-data/pull/22847

PR #805, opened 2024-04-06
Generate details.dist.yml

foolip:
Add the caniuse link and exclude the toggle event.


feature-group-definitions/details.yml
```diff
@@ -1,8 +1,11 @@
 name: <details>
 spec: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element
+caniuse: details
 compat_features:
   - api.HTMLDetailsElement
   - api.HTMLDetailsElement.open
-  - api.HTMLDetailsElement.toggle_event
+  # Exclude toggle event as it changes the Chrome/Safari dates by 3-5 years, and
+  # it's not a critical part of the feature.
+  # - api.HTMLDetailsElement.toggle_event
```
ddbeck, line 9: An idea: given that `HTMLDetailsElement`'s `toggle` event is a distinct thing from a popover's `ToggleEvent`, it might be a good idea to have it be a separate feature, even.

ddbeck (approved):

feature-group-definitions/details.yml
foolip: Look like the toggle event on `<details>` should also be a `ToggleEvent`:
https://html.spec.whatwg.org/multipage/interactive-elements.html#queue-a-details-toggle-event-task

I'm not sure what will make sense, let's come back to this when mapping the very very long tail of unmatched BCD keys.

PR #806, opened 2024-04-06
Generate backdrop-filter.dist.yml

foolip:


foolip:
This will be a simple case after https://github.com/mdn/browser-compat-data/pull/22788 lands and we update BCD.

foolip:
I think we should have a CI job that fails if we have listed `compat_features` that exactly match BCD. That would then fail when we roll a new version of BCD.

Anything that remains would be differences to BCD still in need of upstreaming.

If we have a better way to ignore specific BCD keys, we might be able to get rid of even more.

foolip:
For this PR, we can just wait until the next BCD release I think.

foolip:
We have a bunch of things in flight in BCD now, and can just make this another one of those.

ddbeck (approved):

PR #807, opened 2024-04-06
Generate .dist.yml for cap/ic/lh units

foolip:
After removing the CSS Typed OM bits, these are all Baseline low.


ddbeck:
Just to make sure I understand what's intended here (based on this and the BCD PR): you expect that `api.CSS.cap_static` and friends will eventually find their way into some sort of CSS typed OM feature(s)?

foolip:
I'm not sure yet. I don't think they make sense in the main CSS typed OM feature, because that's older than these units. And "CSS typed OM support for cap/ic/lh" is way too specific and not likely to help anyone. So I think these might actually belong to a kind of "intersection feature", meaning that if feature A+B are both supported, then there's also this additional API surface that we expect.

Alternatively, we find a way to say that certain BCD keys aren't expected to be used by web-features at all, and just add a note about why there.

ddbeck:
Hmm, OK, yes. That makes sense. Groups or snapshots might have answers as well.

ddbeck (approved):

PR #808, opened 2024-04-06
Add media source (MSE) and managed media source

foolip:


tidoust (approved):
That looks good to me. I suppose we want to align with Can I Use's scope for Media Source here. There are a couple of additional MSE features that are not part of that definition, and that we might want to add as well, possibly later on:

1. MSE in workers, which BCD currently flags as "experimental" (only supported in Chrome)
2. Support for text tracks in MSE (only supported in Safari, save iOS).

PR #809, opened 2024-04-06
Add WebRTC and WebRTC encoded transform

foolip:


feature-group-definitions/webrtc.yml
```diff
@@ -0,0 +1,10 @@
+name: WebRTC
+description: The WebRTC API establishes real-time communication channels directly between browsers. It is commonly used in video conferencing applications.
+spec: https://w3c.github.io/webrtc-pc/
+caniuse: rtcpeerconnection
+# The WebRTC API is huge and has been evolving for over a decade. These are
+# central interfaces that result in the right computed status (matching caniuse)
+# but this is <1% complete.
```
ddbeck, line 7: Seems like tagging this (and then setting sensible overrides, or establishing reasonable follow-on features) is going to be a lot of work. Want to file an issue to go with this?

ddbeck (approved):

PR #810, opened 2024-04-07
Add the media session and audio session APIs

foolip:


feature-group-definitions/audio-session.yml
```diff
@@ -0,0 +1,12 @@
+name: Audio session
+description: The `navigator.audioSession` API controls how audio playback interacts with other applications. For example, playing music can pause audio from other applications, while ambient audio can play at the same time.
```
ddbeck, line 2: We've used both "API" and "interface" in places like this. What do you think of making some kind of convention? I like the idea of:

- _interface_ to refer to a specific, named method, property, namespace, etc. (e.g., `SomeInterface` or `someMethod()`, etc.)
- _API_ to refer to an abstract collection of interfaces (e.g., The WebRTC API).

```suggestion
description: The `navigator.audioSession` interface controls how audio playback interacts with other applications. For example, playing music can pause audio from other applications, while ambient audio can play at the same time.
```

feature-group-definitions/media-session.yml
```diff
@@ -0,0 +1,25 @@
+name: Media session
+description: The `navigator.mediaSession` API integrates with platform UI for media playback. It can be used to set metadata such as title and artwork, and to handle user actions like playing, pausing, or seeking.
```
ddbeck, line 2: I'm not sure what "integrates with platform UI" means and the spec is not helping me here. Do you know more about this? What does platform UI refer to?

foolip: It's mainly mobile phone lock screen controls for audio playback which show the title and album art, and controls from play/pause/prev/next. Desktop operating systems can also have a thing like this, but I've not interacted with those myself.

feature-group-definitions/audio-session.yml
foolip: Another reason I'm reluctant about "interface" even for Web IDL `interface` is that it's just an historical accident that we call them interfaces instead of classes. I think we should refer to `URL` (defined with Web IDL) and `Map` (defined in ECMAScript) in similar terms. Both are created with `new` and are a lot like classes, just not called that in specs.

foolip: But we should do whatever MDN does...

feature-group-definitions/media-session.yml
ddbeck: How about this then?

```suggestion
description: The `navigator.mediaSession` API integrates with device UI for media playback. It can be used to set metadata, such as title and artwork, and to handle user actions, like playing, pausing, or seeking.
```

ddbeck: Implying, I guess that "device" would be our preferred term for "device or operating system"

feature-group-definitions/audio-session.yml
foolip: I think using API is fine for anything that refers to a constructor (`URL`) or object (`navigator.audioSession`). For properties and methods I think we'll normally want to call them that, but there may be exceptions.

I would not expect a single crisp definition of API if you ask web developers, or if you ask browser engineers :)

feature-group-definitions/media-session.yml
foolip: That WFM.

feature-group-definitions/audio-session.yml
foolip: So, is "The `navigator.audioSession` API" a good idea or not? I'm afraid it's a bit too novel and that web-features will end up looking quirky compared to MDN or web.dev. But I think the alternative is "The Audio Session API" and then we have to revisiting sentence case in short names I think, for consistency.

ddbeck: If we want to avoid novelty (and not re-use interface generally), then I suppose my recommendation is to say "The `navigator.audioSession` object controls…". Longer thoughts on this follow.

---

If we want to match MDN with references to specific _things_ like `navigator.something`, we've actually got a number of options (because MDN isn't really consistent internally—giving us choices too). Adapted a bit to our constraints, but here we go:

- "The `navigator.audioSession` **property** controls…" (puts emphasis on `navigator`: it's `navigator`'s property)
- "The `navigator.audioSession` **object** controls…" (puts emphasis on `audioSession`: implying the relationship to `AudioSession`), a highly condensed version of MDN's style of "The `audioSession` [read-only property of the `Navigator` interface returns a] `AudioSession` object"
- "The `navigator.audioSession` property **returns** a…" (puts emphasis on the property accessor—I hate this one, it's the worst—MDN shouldn't even be doing this.)

From these, I think option 2 is the "right" one: we care more about the object than we do the fact that `navigator` gives it to you or the syntax you do it with.

But there's also this "API" route. I think it's harder to do that consistently without doing something novel or without being vague. MDN mostly uses in these Capital Letter API situations, which tends to be in reference to whole specs-as-features. And even then it's not terribly consistent. For example, the Canvas API tutorial starts with text on "how to use the `<canvas>` element to draw 2D graphics"—which is much closer to our style here anyway.

ddbeck: Oh interestingly, the Google style guide explicitly [says **not** to do what I suggested earlier](https://developers.google.com/style/word-list#api):

> Don't use API when referring to a method or a class. For example, don't write This resource has one API to mean "This resource has one method." 

I'm guessing somebody thought a lot longer and harder about this than I did, so let's definitely not do that.

foolip: OK, let's go with "object" for now. It's still a bit novel and maybe "Audio Session API" would be more familiar to some people, but I guess that's an argument for changing the short name as well.

I guess it's time to start another big issue with names that I'm not totally happy with :)

foolip: I've filed https://github.com/web-platform-dx/web-features/issues/822 and don't feel very excited about "object" any longer. @ddbeck do you think it would be acceptable to revert back to "API" while we figure out the guidlines?

ddbeck: Yes, there's an issue now, so that's fine with me.

ddbeck (approved):
Approved, pending the revert of `object` to `API`.

PR #812, opened 2024-04-07
Add margin-trim

foolip:


feature-group-definitions/margin-trim.yml
```diff
@@ -0,0 +1,12 @@
+name: margin-trim
+description: The `margin-trim` CSS property controls margins at the start and end of a container, removing the margins of child elements when they meet the edges of the container.
```
ddbeck, line 2: style, for brevity: what do you think?

```suggestion
description: The `margin-trim` CSS property removes the margins of child elements when they meet the starting or ending edges of the container.
```

ddbeck (approved):
Optional description revision for you

feature-group-definitions/margin-trim.yml
foolip: "starting or ending edges" feels a bit abstract to me. In the typical use case it will be the top or bottom. Maybe simplify this even further?

```suggestion
description: The `margin-trim` CSS property removes the margins of child elements when they meet the edges of the container.
```

ddbeck: That works for me. 👍

PR #813, opened 2024-04-07
Add media element pseudo-classes

foolip:
https://drafts.csswg.org/selectors/#resource-pseudos calls these
Resource State Pseudo-classes, but that name is generic and any
pseudo-classes added for `` would probably also go in that section.

All of these are specifically about media elements, so use that as the
name.


feature-group-definitions/media-pseudos.yml
```diff
@@ -0,0 +1,13 @@
+name: Media element pseudo-classes
+description: The `:playing`, `:paused`, `:seeking`, `:buffering`, `:stalled`, `:muted`, and `:volume-locked` CSS pseudo-classes matches `<audio>` and `<video>` elements based on their state.
```
ddbeck, line 2:
```suggestion
description: The `:playing`, `:paused`, `:seeking`, `:buffering`, `:stalled`, `:muted`, and `:volume-locked` CSS pseudo-classes match `<audio>` and `<video>` elements based on their state.
```

ddbeck (approved):
Suggesting a typo fix, otherwise I'm happy with this.

PR #820, opened 2024-04-08
Test `computeBaseline`'s `withAncestors` option

ddbeck:
This PR tests the `withAncestors` option. There are two things in this PR:

1. A test that asserts that the results of the Baseline calculation `api.Notification.body` is different with and without ancestor features (i.e., the reported support for `api.Notification.body` is _greater_ than the reported supported for `api.Notification` and `api.Notification.body`).

   Why is this important? See the MDN page [Notification: body property](https://developer.mozilla.org/en-US/docs/Web/API/Notification/body#browser_compatibility) and compare it to what you see in [Notification](https://developer.mozilla.org/en-US/docs/Web/API/Notification#browser_compatibility).
   
   If you look at [this gist of such features](https://gist.github.com/ddbeck/e3e78201c60170152b020fc983bb163b) you'll see that there are dozens that are hypothetically like it—fewer than I expected! However, cases like `api.Notification.body`, where there's a standalone MDN page for the feature, represent the most-risky cases of this.
  
2. The script I used to find that feature (and several others like it): `scripts/find-troublesome-ancestors.ts`.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -11,4 +11,5 @@
 
 !/scripts/dist.ts
 !/scripts/feature-init.ts
+!/scripts/find-troublesome-ancestors.ts
 !/scripts/release.ts
--- a/packages/compute-baseline/src/baseline/index.test.ts
+++ b/packages/compute-baseline/src/baseline/index.test.ts
@@ -1,10 +1,10 @@
 import assert from "node:assert/strict";
 
+import { Temporal } from "@js-temporal/polyfill";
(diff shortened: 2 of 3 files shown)


ddbeck:
Also I'll note that the helper script only finds features where the Baseline _status_ for a feature's ancestors is less than the feature itself. It doesn't find cases where the Baseline low date is earlier for ancestor features than the feature itself.

foolip (approved):

PR #826, opened 2024-04-09
Add a BCD key to font-palette animation

foolip:
See https://github.com/web-platform-dx/web-features/pull/801/files#r1557484343


foolip:
Leaving as draft until after https://github.com/web-platform-dx/web-features/pull/801 so that I can update dist as well.

Issue #822, opened 2024-04-09
Guideline for API names and descriptions


foolip:
In https://github.com/web-platform-dx/web-features/pull/810 and https://github.com/web-platform-dx/web-features/pull/752 we have some names and descriptions that feel a bit quirky compared to how most resources talk about them.

Should the name be:
 
- Audio session or Audio Session API
- Media session or Media Session API
- Device posture or Device Posture API

Should the description start with:

- The `navigator.audioSession` object, the `navigator.audioSession` API, or the Audio Session API?
- The `navigator.mediaSession` object, the `navigator.mediaSession` API, or the Media Session API?
- The `navigator.devicePosture` object, the `navigator.devicePosture` API, or the Device Posture API?

Considerations:

- If we include "API", do we always do it or only when that's the spec name? https://w3c.github.io/device-posture/ includes it but the other two do not
- Calling something "object" is technically correct, but I worry that it overfocuses on the API entry point, and that the features aren't normally referred to be their entry point, especially things on `navigator`, since `navigator` is effectively a namespace to avoid putting stuff on `window`.

foolip:
https://github.com/web-platform-dx/web-features/pull/762 is another feature we might want to revisit.

foolip:
And https://github.com/web-platform-dx/web-features/pull/771, which I just merged with the description "The `PressureObserver` API ..."

ddbeck:
OK, I've been thinking over this general problem of "never say API".

For feature _names_, I think we made the right choice: we do not want to create features that have the appearance of having "completed" an API. Suppose the Media Session API acquires new and interesting capabilities. A "media session" feature is still usable, even if the whole of the Media Session API is not implemented. Naming "Media Session API" makes it harder to draw that distinction.

That said, I'm thinking that "Device Posture API" or "Media Session API" would be very good _groups_ (and will inevitably appear in snapshots).

That leaves descriptions: I'm more comfortable with descriptions being a moving target. So _initially_ any feature could start with "The Special API does …" or similar. When follow-on features appear, those descriptions can change to reflect that without losing the stability of the feature name.

foolip:
API in group names sounds reasonable, and groups would tend to more often match whole specs and their names. I've sent https://github.com/web-platform-dx/web-features/pull/861 for Web Speech API.

[...]

PR #828, opened 2024-04-09
Refer to AbortController and BroadcastChannel as APIs

foolip:
This is more in line with https://github.com/web-platform-dx/web-features/pull/810, although the discussion continues in https://github.com/web-platform-dx/web-features/issues/822.


captainbrosset (approved):
I know we haven't settled on the discussion yet, but I like this and we can change it over time.

PR #830, opened 2024-04-09
Add .dist.yml in no-op cases

foolip:
These are cases where running the dist scripts shows BOTH of these messages:

> status override matches generated status from compat_features override. Consider deleting this override.
> status override matches generated status from tags. Consider deleting this override.

The simplification of the source YAML is left for a separate commit to
leave no doubt that this is indeed a no-op.


ddbeck (approved):
I checked ~20 of these closely and found most of the others pretty familiar. Looks good—excited for the follow up.

feature-group-definitions/webusb.dist.yml
```diff
@@ -0,0 +1,106 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+name: WebUSB
+spec: https://wicg.github.io/webusb/
+caniuse: webusb
+usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/1519
+status:
+  baseline: false
+  support:
+    chrome: "101"
```
foolip, line 11: https://caniuse.com/webusb disagrees. I'll back this out.

foolip: No I won't. I'll fix it after merging this.

foolip:
The follow up is https://github.com/web-platform-dx/web-features/pull/831, but we'll need a new BCD release before we can remove more.

PR #831, opened 2024-04-10
Remove status and compat_features where this doesn't change .dist.yml

foolip:
This change was made by deleting the status and compat_features from
*.yml using `sed -i '' '/^status:$/,$d' *.yml`, regenerating all the
*.dist.yml, and keeping the changes where only the source changed.


ddbeck (approved):
This is very satisfying. Thank you!

PR #837, opened 2024-04-10
`device-orientation-events`: add intial feature

ddbeck:
This adds device orientation events `DeviceMotionEvent` and `DeviceOrientationEvent`. Despite the somewhat historical notes on https://caniuse.com/deviceorientation, it does seem like there's a usefully interoperable core here.

That said, the MDN docs suggest that the reference frames differ noticeably across platforms. That might be worthy of a note someday.


feature-group-definitions/device-orientation-events.yml
```diff
@@ -0,0 +1,16 @@
+name: Device orientation events (intial support)
```
foolip, line 1: Can you add a TODO here about the reference frame so that we can come back to it?

ddbeck: Done!

foolip (approved):

feature-group-definitions/device-orientation-events.dist.yml
```diff
@@ -0,0 +1,31 @@
...
+status:
+  baseline: low
+  baseline_low_date: 2023-09-18
+  support:
+    chrome: "31"
+    chrome_android: "31"
+    edge: "12"
+    firefox: "6"
+    firefox_android: "6"
+    safari: "17"
```
foolip, line 18: What was the final piece here? Your old markdown tables would be very useful to look at during review :D

ddbeck: All of it: `DeviceMotionEvent`, `DeviceOrientationEvent`, and the window events themselves.

And a lot of that scripting used the precursor to compute-baseline—I'll see if I can quickly reinstate that.

foolip: As an alternative, how about just adding a comment after each BCD key with its Baseline date in a comment?

PR #838, opened 2024-04-10
Add `autofocus` feature

ddbeck:
This is a Baseline 2023 feature, since all browsers now target all elements (not just specific form controls) equally.


feature-group-definitions/autofocus.dist.yml
```diff
@@ -0,0 +1,23 @@
...
+
+name: Autofocus
+description: The `autofocus` HTML attribute gives focus to an element on page load.
+spec: https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute
+caniuse: autofocus
+status:
+  baseline: low
+  baseline_low_date: 2023-02-14
+  support:
+    chrome: "109"
```
foolip, line 12: This is extremely far from the data in https://caniuse.com/autofocus. Trim until it matches?

ddbeck: Hmm, now I'm thinking that we should unhook this from caniuse. Caniuse describes it (emphasis mine):

> Allows a **form field** to be immediately focused on page load.

But what we have interoperability for is autofocus as a global HTML attribute, which was only adopted by the spec in 2019: https://github.com/whatwg/html/pull/4830. It'd be super weird to say that `HTMLElement.autofocus` was supported in 2016.

I suppose then there's case to be made to split this in two: one as "autofocus for form fields" (mapped to caniuse, a status, and no compat features) and one as "autofocus global attribute" (compat features as we have here, no link to the caniuse feature).

foolip: Would it make sense to treat this as a single feature, but retroactively call it partially supported for the very long period of time when it was just `<input type="text" autofocus>` and similar?

ddbeck: Yes, that's exactly what BCD does, even: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#browser_compatibility

Then the hard part is defining "partial support" for the status block 😅

foolip: Right, capturing the Baseline 2023 part of this isn't possible then. But the autofocus global attribute isn't a big feature anyway, so I think it's OK if this one is made invisible.

How about making this match caniuse for now, with a TODO to include the global attribute when we have partial impl. and notes?

ddbeck: Done with ab80235b9af17f07660f6acab84ee905e1b4c3b8

PR #840, opened 2024-04-10
`file-and-directory-entries`: add feature

ddbeck:
Another Baseline 2023 feature.


feature-group-definitions/file-and-directory-entries.yml
```diff
@@ -0,0 +1,38 @@
+name: File and directory entries
+description: File and directory entries represent a virtual file system for working with files and directories dragged and dropped into the browser.
+spec: https://wicg.github.io/entries-api/
+compat_features:
```
foolip, line 4: Can you add api.HTMLInputElement.webkitEntries and api.DataTransferItem.webkitGetAsEntry as long as that doesn't change support status? Those are the entry points to get a `FileSystemEntry` object.

feature-group-definitions/file-and-directory-entries.dist.yml
```diff
@@ -0,0 +1,52 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+name: File and directory entries
+description: File and directory entries represent a virtual file system for working with files and directories dragged and dropped into the browser.
+spec: https://wicg.github.io/entries-api/
+status:
+  baseline: low
+  baseline_low_date: 2023-03-27
+  support:
+    chrome: "102"
```
foolip, line 11: I think this feature should match these compat tables:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitEntries#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry#browser_compatibility

And the differences are small enough that I think we can accept Chrome 22 and Edge 79 as the versions instead of the earliest versions.

feature-group-definitions/file-and-directory-entries.yml
ddbeck: Done with 1598243.

feature-group-definitions/file-and-directory-entries.dist.yml
ddbeck: I don't think this matters much anymore. Firefox for Android doesn't support the entry points (I double checked this), so the status is now false. 😞

foolip: I think our browser versions should be correct so that we don't have to reevaluate everything in order for caniuse to be able to import features. Leaving TODOs would be fine though.

ddbeck: @foolip I removed an outlier, `createSyncAccessHandle` with e646c54. That pushes us back to Chrome 86, which is where a bunch of these compat features appear. What do you think?

foolip: @ddbeck let's go with that for now. If we had https://github.com/web-platform-dx/web-features/issues/852 or something like that it would be easier to review and see at a glance where the most reasonable cutoff point is. I'll comment on that issue that we should revisit this feature.

foolip (approved):

PR #843, opened 2024-04-10
Improve contain-intrinsic-size description

foolip:
As suggested by @vmpstr. Thanks!

### DIFF
--- a/feature-group-definitions/contain-intrinsic-size.yml
+++ b/feature-group-definitions/contain-intrinsic-size.yml
@@ -1,5 +1,5 @@
 name: contain-intrinsic-size
-description: The `contain-intrinsic-size` CSS property sets the size of an element that the browser will use for layout calculations when the element is subject to size containment, within a CSS query container.
+description: The `contain-intrinsic-size` CSS property sets the intrinsic size of an element. When using size containment, the browser will layout the element as if it had a single child of this size.
 spec: https://drafts.csswg.org/css-sizing-4/#intrinsic-size-override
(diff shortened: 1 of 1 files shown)


feature-group-definitions/contain-intrinsic-size.yml
```diff
@@ -1,5 +1,5 @@
 name: contain-intrinsic-size
-description: The `contain-intrinsic-size` CSS property sets the size of an element that the browser will use for layout calculations when the element is subject to size containment, within a CSS query container.
+description: The `contain-intrinsic-size` CSS property sets the intrinsic size of an element. When using size containment, the browser will layout the element as if it had a single child of this size.
```
foolip, line 2: @vmpstr and I had one point of uncertainty here, namely "intrinsic size". It's very hard to describe the concept succinctly. It is the "as if you had a single child of size size"-size, but most layout modes don't shrink wrap to that size, so it's not the resulting inner or outer size of the element, more like a constraint.

I think this wording is OK, but suggestions welcome.

ddbeck: FWIW, I think this is OK. It's hard to understand, but it's a difficult concept even after reading the docs. Since you've included a good phrases for search, either to find this feature or to go learn more ("size containment", "intrinsic size", and the name of the property), I think it's OK.

That said, I'm not in a position to judge the correctness of the "as if it had a single child of this size."

ddbeck (approved):

PR #845, opened 2024-04-10
Add an HTML group with subgroups

foolip:
This is mainly to group forms and media elements, but a few features make sense in the parent group too.

These features defined by HTML are not in any of these groups:

- `<link rel="modulepreload">`
- BroadcastChannel
- contextlost and contextrestored
- Fetch priority
- Import maps
- JavaScript modules
- Navigation API
- Offscreen canvas
- structuredClone()

`<link rel="modulepreload">` is a borderline case, maybe it's HTML, maybe it's Modules, or maybe it's both. This can be revisted.

### DIFF
--- a/groups/html.yml
+++ b/groups/html.yml
@@ -1,0 +1,8 @@
+# This group isn't intended for everything defined in HTML:
+# spec: https://html.spec.whatwg.org/multipage/
+#
+# It is for the markup language, and not for all APIs and infrastructure
+# defined in HTML. For example, structuredClone() should not be in this group.
+# APIs on HTML*Element and CSS features specific to an HTML element generally
+# do make sense to group under HTML, however.
+name: HTML
--- a/feature-group-definitions/audio-video-tracks.yml
+++ b/feature-group-definitions/audio-video-tracks.yml
@@ -1,5 +1,6 @@
 name: Audio and video tracks
 description: The `audioTracks` and `videoTracks` APIs for media elements switch audio and video tracks during playback.
+group: media-elements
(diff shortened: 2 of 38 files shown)


foolip:
@Elchi3 is this in line with how you think about groups? The forms group I'm adding here spans HTML and CSS specs but is about `<form>` and the many form control elements. Similarly, I think we might add "Modules" as a group spanning JavaScript and HTML.

groups/html.yml
```diff
@@ -0,0 +1,8 @@
+# This group isn't intended for everything defined in HTML:
+# spec: https://html.spec.whatwg.org/multipage/
+#
+# It is for the markup language, and not for all APIs and infrastructure
+# defined in HTML. For example, structuredClone() should not be in this group.
+# APIs on HTML*Element and CSS features specific to an HTML element generally
+# do make sense to group under HTML, however.
```
ddbeck, line 7: I'm slightly anxious about this. Do we contemplate a DOM group existing and how does that impact this?

ddbeck (approved):
I'm OK with this, but I do have a question.

groups/html.yml
foolip: I think DOM would be everything pertaining to the `Node` hierarchy except language-specific interfaces like everything `HTMLElement` and `SVGElement`. I don't expect it will be that messy, but always a little :)

foolip: However, sometimes people say "DOM APIs" for what I would call "APIs defined using Web IDL", a vastly broader group of things. I don't think that should be a group at all, it's just the collection of all BCD keys under api.*.

ddbeck: Ah ha, OK that's fine.

PR #846, opened 2024-04-10
Convert browser version overrides to compat_features overrides

foolip:


feature-group-definitions/js-modules.dist.yml
```diff
@@ -0,0 +1,34 @@
...
+  baseline: high
+  baseline_low_date: 2018-05-09
+  baseline_high_date: 2020-11-09
+  support:
+    chrome: "61"
+    chrome_android: "61"
+    edge: "16"
+    firefox: "60"
+    firefox_android: "60"
+    safari: "11"
```
foolip, line 24: Safari was 10.1 before.

feature-group-definitions/lab.dist.yml
```diff
@@ -0,0 +1,28 @@
...
+status:
+  baseline: low
+  baseline_low_date: 2023-05-09
+  support:
+    chrome: "111"
+    chrome_android: "111"
+    edge: "111"
+    firefox: "113"
+    firefox_android: "113"
+    safari: "15"
```
foolip, line 16: Safari was 15.4 before, but since it's lower now I'm just going to trust it. It doesn't change the Baseline low date.

feature-group-definitions/map.dist.yml
```diff
@@ -0,0 +1,55 @@
...
+group: maps
+status:
+  baseline: high
+  baseline_low_date: 2016-09-20
+  baseline_high_date: 2019-03-20
+  support:
+    chrome: "38"
+    chrome_android: "38"
+    edge: "12"
+    firefox: "42"
```
foolip, line 31: Firefox was 13 before, need to find the culprit key.

feature-group-definitions/map.dist.yml
```diff
@@ -0,0 +1,55 @@
...
+    firefox_android: "42"
+    safari: "10"
```
foolip, line 33: Safari was 8 before.

feature-group-definitions/set.dist.yml
```diff
@@ -0,0 +1,39 @@
...
+description: Set objects store unique values of any type.
+spec: https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set-objects
+snapshot: ecmascript-2015
+group: sets
+status:
+  baseline: high
+  baseline_low_date: 2015-11-03
+  baseline_high_date: 2018-05-03
+  support:
+    chrome: "43"
```
foolip, line 14: Chrome was 38 before.

feature-group-definitions/set.dist.yml
```diff
@@ -0,0 +1,39 @@
...
+    chrome_android: "43"
+    edge: "12"
+    firefox: "42"
```
foolip, line 17: Firefox was 13 before.

feature-group-definitions/set.dist.yml
```diff
@@ -0,0 +1,39 @@
...
+    firefox_android: "42"
+    safari: "9"
```
foolip, line 19: Safari was 8 before.

foolip (commented):
Self-review to identify changes in browser versions.

feature-group-definitions/js-modules.dist.yml
foolip: It's because of `nomodule` and I think it's actually correct to treat this as an essential part of the feature and bump the Safari version slightly here. AIUI, this attribute is the fallback mechanism and important to deploying JS modules in practice. I'll add TODO about this though.

feature-group-definitions/lab.dist.yml
foolip: Actually, I won't trust it. I'd expect `lab()` and `oklab()` to be shipped together, and `oklab()` stays at 15.4.

PR #849, opened 2024-04-11
Add GitHub Action to label changes to feature definitions

ddbeck:
This PR will fail (since there's no config on the target branch). I tested it on my own fork, however, and it _ought_ to work.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,5 @@
+feature-definition:
+  - changed-files:
+      - all-globs-to-any-file:
+          - "feature-group-definitions/*.yml"
+          - "!feature-group-definitions/*.dist.yml"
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,5 @@
+feature-definition:
```
foolip, line 1: Anticipating that we'll rename feature-group-definitions/ to just features/:

```suggestion
features:
```

foolip (approved):

.github/labeler.yml
ddbeck: This is the label name, not the path—I've put in cc1be44 to make sure we cover that from the get go (I'm anticipating a second label for dist-only changes, so I'd like to keep the label name as-is).

PR #850, opened 2024-04-11
Add ::spelling-error and ::grammar-error

foolip:


feature-group-definitions/spelling-grammar-error.yml
```diff
@@ -0,0 +1,11 @@
+name: "::spelling-error and ::grammar-error"
+description: The `::spelling-error` and `::grammar-error` CSS pseudo-elements match text that is highlighted as misspelled and grammatically incorrect, respectively.
```
ddbeck, line 2: "Highlighted" implies some visual styling to me. Maybe this?

```suggestion
description: The `::spelling-error` and `::grammar-error` CSS pseudo-elements match text that is identified as misspelled and grammatically incorrect, respectively.
```

ddbeck (approved):

feature-group-definitions/spelling-grammar-error.yml
foolip: There is visual styling, and I think the main point of these pseudo-elements is to change or remove that styling.

foolip: (The browser isn't looking for these errors except when they will be highlighted, so this is really in form controls and other text input.)

ddbeck: OK, it's fine to leave this as is then. I'll merge this.

PR #853, opened 2024-04-11
Add text-wrap: nowrap

foolip:
This is a Baseline 2024 feature.


feature-group-definitions/text-wrap-nowrap.yml
```diff
@@ -0,0 +1,6 @@
+name: "text-wrap: nowrap"
+description: "The `text-wrap: nowrap` CSS declaration prevents text being broken into multiple lines. Text that does not fit will instead overflow."
```
ddbeck, line 2: Previously, you used "value" where you've used "declaration" here (to refer to the pair of property and value). This is also fine with me—it's probably more technically correct—but is that a preference? I think a guideline for this would be nice to have.

Suggestions here to make the tenses consistent.

If you prefer _declaration_:

```suggestion
description: "The `text-wrap: nowrap` CSS declaration prevents breaking text into multiple lines. Text that does not fit overflows instead."
```

If you prefer _value_:


```suggestion
description: "The `text-wrap: nowrap` CSS value prevents breaking text into multiple lines. Text that does not fit overflows instead."
```

foolip: I've took "CSS declaration" from some of @captainbrosset's descriptions. It looks like none of those have landed yet and that background-clip-text.yml is the only thing to stay consistent with. That's "CSS property value" which I think is probably too long, but let's change them all at once. So:

```suggestion
description: "The `text-wrap: nowrap` CSS property value prevents text breaking into multiple lines. Text that doesn't fit overflows instead."
```

ddbeck (approved):

PR #855, opened 2024-04-11
Add feature for animatable display

foolip:


feature-group-definitions/display-animation.yml
```diff
@@ -0,0 +1,11 @@
+name: display animation
```
foolip, line 1: I've named this similarly to `font-palette-animation` for consistency.

feature-group-definitions/display-animation.yml
```diff
@@ -0,0 +1,11 @@
...
+description: "Animation and transitions between `display: none` and any other `display` value, as well as between `content-visibility: hidden` and any other `content-visibility` value."
```
foolip, line 2: This description is lacking something. The problem with saying "The `display` CSS property can be animated" is that it sounds like a claim _about_ a feature and not a description _of_ a feature.

foolip (commented):
Self-review to point out a thing.

feature-group-definitions/display-animation.yml
ddbeck: So we're trying to capture the idea of "has the property of being non-discretely animitable" and the way we've written most descriptions makes constructing such a thing kinda ghastly. I kinda hate this:

```suggestion
description: "Elements are animatable into or out of visibility, given by `display: none` to any other `display` value or by `content-visibility: hidden` to any other `content-visibility` value. This also applies to transitions."
```

What if we throw caution to the wind and write something rather different than we've done elsewhere:

```suggestion
description: "You can animate elements between `display: none` and any other `display` value or animate between `content-visibility: hidden` and any other `content-visibility` value. This also applies to transitions."
```

foolip: Let's go without your second option. Being different and standing out isn't a bad thing here, me might get feedback about it then, and if not it'll stand out as an implicit TODO every time we see it. To be clear, I think it's a pretty good description, and I can't improve on it.

PR #857, opened 2024-04-12
`aria-attribute-reflection`: add initial support feature

ddbeck:
I couldn't find any rhyme or reason for the ones that aren't interoperable, so I created this as an initial support feature.


feature-group-definitions/aria-attribute-reflection.dist.yml
```diff
@@ -0,0 +1,96 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+name: ARIA attribute reflection (initial support)
+description: WAI-ARIA attributes have corresponding properties on `Element` and `ElementInternals` objects.
+spec: https://w3c.github.io/aria/#accessibilityroleandproperties-correspondence
+status:
+  baseline: low
+  baseline_low_date: 2023-10-24
+  support:
+    chrome: "103"
```
foolip, line 11: https://chromiumdash.appspot.com/commit/dbef8d4838807211e2940f0929c8109cdf06b9d5 is the initial support for reflected ARIA properties in Chromium, and includes most of this stuff.

Can you exclude `ariaBrailleLabel`, `ariaBrailleRoleDescription`, `ariaDescription`, `ariaInvalid`, and `ariaSetSize` and see if that results in Chrome 81?

feature-group-definitions/aria-attribute-reflection.dist.yml
```diff
@@ -0,0 +1,96 @@
...
+    chrome_android: "103"
+    edge: "103"
+    firefox: "119"
```
foolip, line 14: Matches `role` in https://developer.mozilla.org/en-US/docs/Web/API/Element#browser_compatibility 👍

feature-group-definitions/aria-attribute-reflection.dist.yml
```diff
@@ -0,0 +1,96 @@
...
+    firefox_android: "119"
+    safari: "16.4"
```
foolip, line 16: `role` is 12.1 in https://developer.mozilla.org/en-US/docs/Web/API/Element#browser_compatibility. Can you check which parts of the feature are newer?

feature-group-definitions/aria-attribute-reflection.dist.yml
ddbeck: We talked about this more on our call the other week. The minimalist set consisting of `role` and `ariaLabel` is still Chrome 103, thanks to `role`. As a consequence, I've added back everything that gets us down to Safari 12.1 (which is basically ElementInternals, plausibly a separate feature). Done with 6ab45161c3852b74d0240ca2847ddf64cd10cf10.

foolip (approved):

PR #867, opened 2024-04-12
Add Web authentication (initial support)

foolip:


feature-group-definitions/webauthn.yml
```diff
@@ -0,0 +1,24 @@
+name: Web authentication (initial support)
+description: The web authentication API allows you to create public key-based credentials and use them for authentication, such as when signing in. Also known as WebAuthn.
```
foolip, line 2: @agl @nsatragno might you be able to help me review this description? It's for web developers and I'd like it to be accurate most also easy to understand. If WebAuthn is used for a specific thing 90% of the time, mentioning that would be helpful. I understand there's a connection to FIDO and perhaps to passkeys, but I don't know exactly how.

I there are subfeatures of WebAuthn that are important to cover separately, what are they? A lot has shipped since Chrome 67, but I don't know how to group it.

foolip: In the course of looking closer I realized there's a lot more to untangle here, and I sent https://github.com/web-platform-dx/web-features/pull/868. Feedback on how to structure all of these related technologies and API surfaces in web-developer-facing features would be much appreciated.

foolip:
I have folded this into https://github.com/web-platform-dx/web-features/pull/868.

PR #868, opened 2024-04-13
Add credential management features

foolip:


foolip:
This is draft because it's confusing and I need to do more research to figure this out. Big picture, `navigator.credentials` allows access to various kinds of credentials, and there are at these subclasses of `Credential` known by webref:

- [`PasswordCredential`](https://w3c.github.io/webappsec-credential-management/#passwords)
- [`FederatedCredential`](https://w3c.github.io/webappsec-credential-management/#federated) which is _not_ FedCM
- [`IdentityCredential`](https://fedidcg.github.io/FedCM/#browser-api-identity-credential-interface) a.k.a. FedCM
- [`DigitalCredential`](https://wicg.github.io/digital-identities/#the-digitalcredential-interface)
- [`OTPCredential`](https://wicg.github.io/web-otp/#OTPCredential)
- [`PublicKeyCredential`](https://w3c.github.io/webauthn/#sctn-api) a.k.a WebAuthn or Web Authentication API

[Credential Management](https://w3c.github.io/webappsec-credential-management/) has a lot of API surface, but it seems to all be in service of one of the specific kinds of credentials.

Then there are the more user-facing technologies like passkeys, yubikeys, "Login with GitHub", SMS authentication, that relate to the above in different ways, some perhaps not at all.

Parking this here to ask the pros for help :)

foolip:
`OTPCredential` should be added as WebOTP.

features/credential-management.yml.dist
```diff
@@ -0,0 +1,71 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: high
+  baseline_low_date: 2020-01-15
+  baseline_high_date: 2022-07-15
+  support:
+    chrome: "60"
+    chrome_android: "60"
+    edge: "79"
```
foolip, line 11: This hinges on `navigator.credentials.store`, and I've confirmed that is missing from Edge 18, BCD is correct. This does affect the Baseline low date, so I'd like a second opinion from a reviewer on whether this is right, or if we think that `store()` is a nice-to-have. One data point is that caniuse says Edge 79.

features/web-otp.yml.dist
```diff
@@ -0,0 +1,26 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome: "93"
+    chrome_android: "84"
+    edge: "93"
+compat_features:
+  # ⬇️ Same status as overall feature ⬇️
```
foolip, line 11: @ddbeck is there a compute-baseline bug here? Without `compute_from`, I'd expect the overall status to match the last block, or sometimes no block.

ddbeck: This was due to inconsistent ancestors checking, fixed with https://github.com/web-platform-dx/web-features/pull/1485. Merging main into this and refreshing the dists will fix it.

foolip: Thank you, sorted now!

features/credential-management.yml.dist
ddbeck: I had a look around and `store()` appears in lots of examples for this API (e.g., the first few blog posts I found via Google). Whether it's strictly essential to the feature or not, it does appear to be part of the popular understanding of what this thing is. Let's keep it.

features/fedcm.yml
```diff
@@ -0,0 +1,8 @@
+name: Federated credential management
+description: The `IdentityCredential` API. Also known as FedCM.
```
ddbeck, line 2:
```suggestion
description: The `IdentityCredential` API delegates authentication to a third-party identity provider directly, instead of depending on third-party cookie. Also known as FedCM.
```

features/federated-credentials.yml
```diff
@@ -0,0 +1,11 @@
+name: Federated credentials
+description: The `FederatedCredential` API... not to be confused with FedCM.
```
ddbeck, line 2: I have literally no idea how this is different from federated credential _management_. 🤷 

```suggestion
description: The `FederatedCredential` API delegates authentication to a third-party identity provider directly, instead of depending on third-party cookie. Not to be confused with federated credential management (FedCM).
```

features/web-otp.yml
```diff
@@ -0,0 +1,11 @@
+name: WebOTP
+description: The `OTPCredential` API TODO one-time-passwords (OTP) TODO.
```
ddbeck, line 2:
```suggestion
description: The `OTPCredential` API represents a one-time-password (OTP).
```

ddbeck (commented):
Some ideas here about descriptions. Thank you!

features/fedcm.yml
foolip: Just missing an "s" in "third-party cookies", which I'll add.

features/federated-credentials.yml
foolip: When I asked @nsatragno about the difference a few months ago she said:

> They're designed very differently and `FederatedCredential` has enough usage that I don't think it'll be deprecated any time soon. `FederatedCredential` only stores a set of usernames with their provider. FedCM is much more involved and has the browser mediate between the website and the identity provider. They solve the same problem in the same way a typewriter and an IDE solve the same problem of writing code...

But I'm not sure how to describe them in simple terms that gets at the difference. @nsatragno do you have any suggestions, starting from the description @ddbeck has suggested? Note that it's the same description as for FedCM in all substance.

nsatragno: Yeee this is very confusing and will not help developers u.u. But they are _very_ different things.

FederatedCredential is an API that relying parties (i.e. websites a user signs in to) can use to store and retrieve username and identity provider url pairs. This can be used to have the browser remember these details for the next time the user signs in. However, FedereatedCredential does not concern itself with how the federated sign-in works. It's a pretty basic storage, very similar to PasswordCredential.

The FedCM API instead defines a protocol used by identity providers (i.e. websites that can provide identity for a user, "sign in with...") to support sign-in flows mediated by the browser. The protocol defines endpoints that the identity provider must implement and will be accessed by the browser. The relying party calls the FedCM javascript API to trigger the browser showing federated authentication UI.

For more information on FedCM, you can read https://developers.google.com/privacy-sandbox/3pcd/fedcm-developer-guide.

--

I would describe FederatedCredential to developers as:

```suggestion
description: The `FederatedCredential` API allows websites to store the details required to authenticate a user using a third-party identity provider. Not to be confused with federated credential management (FedCM).
```

foolip: Thanks @nsatragno and @ddbeck! Applying the suggestion.

features/fedcm.yml
```diff
@@ -0,0 +1,8 @@
+name: Federated credential management
+description: The `IdentityCredential` API delegates authentication to a third-party identity provider directly, instead of depending on third-party cookies. Also known as FedCM.
```
foolip, line 2: @nsatragno before we land this, can you also help check if this is a reasonable description of FedCM? Not sure if you checked it already.

ddbeck: Let's get rid of it. It was meant to contrast with the "depending on" bit, but I can see why it would be an iffy choice. We don't lose anything in this form, I think:

```suggestion
description: The `IdentityCredential` API delegates authentication to a third-party identity provider, instead of using third-party cookies. Also known as FedCM.
```

ddbeck (approved):

PR #869, opened 2024-04-13
Add BCD as a dev dependency

foolip:
This should allow Dependabot to send dependency updates for BCD, and it
was stuck at 5.5.18 and 5.5.19 in the package-lock.json files.

compute-baseline needs it as a dependency to run tests, and
package-lock.json is needed for `npm ci` to work.

The top-level package.json needs it as a dependency since it's a peer
dependency of compute-baseline, and because `npm test:dist` needs it,
and that dependency should be pinned. starting-style.dist.yml needed
updating, and those updates should go with dependency bumps.

This is a bit convoluted and perhaps will confuse Dependabot, but it's
what would make sense if compute-baseline was in a separate repository
and we installed it from NPM.

### DIFF
--- a/feature-group-definitions/starting-style.dist.yml
+++ b/feature-group-definitions/starting-style.dist.yml
@@ -9,6 +9,8 @@ status:
     chrome: "117"
     chrome_android: "117"
     edge: "117"
+    safari: "17.5"
+    safari_ios: "17.5"
 compat_features:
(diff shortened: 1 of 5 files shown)


foolip:
Obsoleted by https://github.com/web-platform-dx/web-features/pull/891.

PR #872, opened 2024-04-13
Simplify Lab and Oklab editorial overrides

foolip:


feature-group-definitions/lab.dist.yml
```diff
@@ -0,0 +1,28 @@
...
+status:
+  baseline: low
+  baseline_low_date: 2023-05-09
+  support:
+    chrome: "111"
+    chrome_android: "111"
+    edge: "111"
+    firefox: "113"
+    firefox_android: "113"
+    safari: "15"
```
foolip, line 16: This is earlier than the manual 15.4 version before. It matches BCD, and there's no reason to doubt the source:
https://github.com/mdn/browser-compat-data/pull/10898

PR #874, opened 2024-04-13
Add status for IntersectionObserver v2

foolip:
This isn't in BCD, but this matches other sources:
https://chromestatus.com/feature/5878481493688320
https://caniuse.com/intersectionobserver-v2


feature-group-definitions/intersection-observer-v2.yml
```diff
@@ -3,6 +3,10 @@ description: The `trackVisibility` parameter for the `IntersectionObserver` cons
 spec: https://w3c.github.io/IntersectionObserver/v2/
 caniuse: intersectionobserver-v2
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/3051
-compat_features:
-  - api.IntersectionObserver.IntersectionObserver.options_trackVisibility_parameter
-  - api.IntersectionObserverEntry.isVisible
```
ddbeck, line 8: It's weird that we thought that this was in BCD, but it wasn't. Is this something we ought to file an issue against in BCD?

ddbeck (approved):

feature-group-definitions/intersection-observer-v2.yml
foolip: My memory of creating https://github.com/web-platform-dx/web-features/pull/628 is imperfect, but I think I just predicted BCD keys without checking if they were in BCD. I've filed https://github.com/mdn/browser-compat-data/issues/22862.

PR #875, opened 2024-04-13
Generate .dist.yml for features with no status or BCD keys

foolip:
This will ensure we notice if tags are added to BCD that result in a
compmuted status.


foolip:
Actually we shouldn't do this. Let's require status instead: https://github.com/web-platform-dx/web-features/pull/899

Issue #876, opened 2024-04-13
compute-baseline doesn't handle Edge 79 removal correctly 


foolip:
Generating `audio-video-tracks.dist.yml` currently produces this:

```yml
# Generated from: audio-video-tracks.yml
# Do not edit this file by hand. Edit the source file instead!

name: Audio and video tracks
description: The `audioTracks` and `videoTracks` APIs for media elements switch audio and video tracks during playback.
group: media-elements
spec: https://html.spec.whatwg.org/multipage/media.html#media-resources-with-multiple-media-tracks
caniuse:
  - audiotracks
  - videotracks
status:
  baseline: false
  support:
    edge: "12"
    safari: "8"
    safari_ios: "8"
compat_features:
  - api.AudioTrack
  - ...
```

Edge 12 is wrong, see here:
https://developer.mozilla.org/en-US/docs/Web/API/AudioTrack#browser_compatibility

PR #877, opened 2024-04-13
Add audio-video-tracks.dist.yml

foolip:
<s>This is to avoid the wrong status being generated by accident:
https://github.com/web-platform-dx/web-features/issues/876</s>


foolip:
The `npm run test:dist` bit fails and I can't reproduce it locally. My best guess is that https://github.com/web-platform-dx/web-features/pull/869 will resolve the problem.

ddbeck (approved):

PR #878, opened 2024-04-13
Add .dist.yml for image formats

foolip:
WebP data is from https://caniuse.com/webp. Like for AVIF, partial
support was not captured and left as a TODO.


feature-group-definitions/webp.yml
```diff
@@ -2,3 +2,15 @@ name: WebP
 spec: https://www.ietf.org/archive/id/draft-zern-webp-15.html
 caniuse: webp
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/3797
+# TODO: Add notes about animated WebP.
+status:
+  baseline: high
+  baseline_low_date: 2020-09-16
+  support:
+    chrome: "9"
```
ddbeck, line 10: Caniuse says this is partial through Chrome 22 (for transparent and lossless images) or 31 (for animated). Was it intended to take the earlier versions?

foolip: It was intentional yes, and I did the same for AVIF where the Safari version would otherwise change.

I think matching caniuse and treating animations as an essential part of support would be OK too, but my hunch is that still images are the overwhelming use case, and that support for animations would make more sense as notes that don't imply partial implementation.

ddbeck (approved):

feature-group-definitions/webp.yml
ddbeck: Maybe we could expand on the comments then? Something like…

```
# TODO: Add notes about animated <format> coming later in Chrome (but don't mark this as partially supported)
```

PR #879, opened 2024-04-13
Generate .dist.yml for color-*

foolip:
These are Baseline 2022/2023 features.


foolip:
This depends on https://github.com/web-platform-dx/web-features/pull/869 because some of the tags are recently migrated to BCD.

ddbeck (approved):

PR #880, opened 2024-04-13
Split <slot> into two features and add statuses

foolip:


feature-group-definitions/slot-assign.dist.yml
```diff
@@ -0,0 +1,23 @@
+# Generated from: slot-assign.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Imperative slot assignment
+description: The `assign()` method for `<slot>` elements does TODO. Also known as manual slot assignment.
```
foolip, line 5: @mfreed7 can you help me describe what this feature is? https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assign doesn't really say. I know it's the imperative alternative to the declarative method, but how to describe what it's for in simple terms?

mfreed7: Sure. I agree that the MDN description looks more or less like a copy of the spec. Not how a developer would understand it. Maybe this?

-----

The assign() method of the [HTMLSlotElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement) interface assigns the provided nodes to that slot, provided they are direct children of the shadow host. This allows the slotting behavior to be controlled via Javascript, rather than automatically by the browser based on matching up the `slot` and `name` attributes. For this reason, this feature is also known as "manual slot assignment".

In order for `assign()` to have any effect, the shadow root must have been created with the [`slotAssignment` parameter](`https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#slotassignment) set to `"manual"`.

-----

foolip: Thank you @mfreed7, that is very very helpful! I think we'll need to shorten it somewhat to fit our (self-imposed) length limit on descriptions, but I'll try to retain all the information.

feature-group-definitions/slot-assign.dist.yml
```diff
@@ -0,0 +1,23 @@
+# Generated from: slot-assign.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Imperative slot assignment
+description: The `assign()` method for `<slot>` elements assigns the given nodes to the slot. This allow you to control slotting using Javascript, as an alternative to using the `slot` and `name` HTML attributes. The nodes must be direct children of a shadow host and the shadow root must be created with the `slotAssignment `parameter set to "manual". Also known as manual slot assignment.
```
foolip, line 5: > Error: description in slot-assign.yml is too long, 377 characters. The maximum allowed length is 300.

Trying to jettison the least essential, this is 302 characters and would pass if we counted plaintext length:

```suggestion
description: The `assign()` method for `<slot>` elements assigns nodes to the slot. This is an alternative to using the `slot` and `name` HTML attributes. The nodes must be children of a shadow host and the shadow root must be created with the `slotAssignment` set to "manual". Also known as manual slot assignment.
```

ddbeck:
```suggestion
description: The `assign()` method for `<slot>` elements assigns nodes to the slot, as an alternative to using the `slot` and `name` HTML attributes. The nodes must be children of a shadow host and the shadow root must be created with the `slotAssignment` set to "manual". Also known as manual slot assignment.
```

NB: this is the dist file, you'll want to apply it to the authored YAML.

ddbeck (approved):

PR #881, opened 2024-04-14
Add WebAuthn easy public key access

foolip:


foolip:
Note that `getTransports()` was implemented in the same releases of Firefox and Safari, but it doesn't seem related. I guess it was easy to implement and was nearby in the spec.

ddbeck (approved):

PR #882, opened 2024-04-14
Convert descriptions to plaintext + HTML

foolip:


foolip:
Since we have both dist and build, we need to decide what should happen where. This could be done as a build step too.

captainbrosset:
I would really like there to be two flavors for the name and description fields, as you proposed in another issue: `name` + `description` and `nameHTML` + `descriptionHTML`. The former would be plain text, with markdown styling removed, and the latter would be markdown rendered as HTML.
I have no opinion on whether this should be generated during dist file generation or at built time.

foolip:
If you see a need for a plaintext variant for each, that sounds OK to me, and easy to provide. It would also make sense to use that as the description length limit, to not include markup and push people away from using `<code>` just to stay under the length limit.

foolip:
I guess the simplest way to generate plain text is to parse the HTML and get the `textContent`.

package.json
```diff
@@ -37,6 +37,7 @@
     "compute-baseline": "file:./packages/compute-baseline",
     "fast-json-stable-stringify": "^2.1.0",
     "fdir": "^6.1.1",
+    "marked": "^12.0.1",
```
ddbeck, line 40: `marked` is not very compliant with CommonMark. While we're especially unlikely to exercise any of Markdown's "fun" bits, I'd prefer to choose a good dependency _now_, in case we start using Markdown for other things. May I suggest `remark` or `micromark` instead?

foolip: I picked `marked` because it has no dependencies, but I've never used any of these libraries before so I'm not opinionated. Is there one that you like, and do you know if any of them make it easy to produce plain text without parsing the resulting HTML?

ddbeck: `remark` is the thing to use then. It won't _look_ very dependency-light (the "unified" project is made up of a bunch of tiny packages that transform markdown/html/etc., though they're mostly developed together), but it's nice enough to join up `remark` + [`remark-html`](https://unifiedjs.com/explore/package/remark-html/#use) or [`strip-markdown`](https://unifiedjs.com/explore/package/strip-markdown/#use) get different outputs from the same parsed Markdown.

(If that sounds like a hassle, I've done this sort of thing before and I can open a competing PR to demo that.)

captainbrosset (approved):
I'll let you and Daniel figure out the md2html mechanics, but wanted to approve this anyway.

package.json
foolip: I've pushed some changes.

foolip:
@ddbeck I've pushed some changes that "work" but I don't really love it. I'll leave comments about the parts I hope can be improved.

I feel like doing this as part of the dist step isn't going to be great, and perhaps we should instead make .dist.yml serve only as a lockfile for BCD keys? That would also avoid having to regenerate dist in many PRs when we're changing something other than the BCD keys. Descriptions would be part of the build step instead. WDYT?

feature-group-definitions/aborting.dist.yml
```diff
@@ -25,3 +25,4 @@ compat_features:
   - api.AbortSignal
   - api.AbortSignal.abort_event
   - api.AbortSignal.aborted
+descriptionHTML: The <strong>AbortController</strong> and <code>AbortSignal</code> APIs allow you to cancel an ongoing operation, such as a <code>fetch()</code> request.
```
foolip, line 28: I couldn't figure out how to insert `descriptionHTML` right after `description`. But it won't matter if we trim dist files into just the computed `status` and `compat_features`.

scripts/dist.ts
```diff
@@ -137,6 +146,32 @@ function toDist(sourcePath: string): YAML.Document {
...
 
+function convertMarkdown(yaml: YAML.Document) {
+  // Markdown is only used the descriptions right now, but could easily be
+  // supported in names as well.
+  const description = yaml.get("description");
+  if (typeof description !== "string") {
+    return;
+  }
+
+  const text = remark()
```
foolip, line 157: I'm not happy with using both `remark()` and `unified()` and parsing the Markdown twice. There must be a way to parse the Markdown string once, and turn the resulting document into both plaintext and HTML.

ddbeck: Sorry, I think I might've steered you wrong with my previous comment. I think something like this would work well (enough for now):

(hast-util-to-string does the same thing as the DOM's `textContent`)

```typescript
  import { toString as hastTreeToString } from "hast-util-to-string";
  import rehypeStringify from "rehype-stringify";
  import remarkParse from "remark-parse";
  import remarkRehype from "remark-rehype";
  import { unified } from "unified";

  // snip
  
  const mdTree = unified().use(remarkParse).parse(description);
  const htmlTree = unified().use(remarkRehype).runSync(mdTree);
  const text = hastTreeToString(htmlTree);
  let html = unified().use(rehypeStringify).stringify(htmlTree);
```

feature-group-definitions/aborting.dist.yml
ddbeck: It's a little fussy. You have to create a document node, then splice the document's items array to insert it near a target item. See `insertStatus` for an example: https://github.com/web-platform-dx/web-features/blob/a7e9e1c058519fda5986864cade005fc8a739aea/scripts/dist.ts#L173

ddbeck:
> I feel like doing this as part of the dist step isn't going to be great, and perhaps we should instead make .dist.yml serve only as a lockfile for BCD keys? That would also avoid having to regenerate dist in many PRs when we're changing something other than the BCD keys. Descriptions would be part of the build step instead. WDYT?

I think I'd be OK with that, yes. I like the _idea_ of having a single, authoritative YAML file. But the built HTML and plain text fields seem like a nuisance.

scripts/dist.ts
foolip: Getting the plaintext via HTML instead of directly from the Markdown tree adds a step between Markdown and plaintext that shouldn't be strictly needed, but in terms of dependencies and lines of code it's no worse. I'll try it.

foolip:
@ddbeck I've moved the markdown conversion to index.ts now. Do you think that's right place? It makes it harder to review the text and HTML descriptions, but maybe it's the right tradeoff? We can add some tests...

index.ts
```diff
@@ -87,8 +94,30 @@ function* identifiers(value) {
...
+    return { text, html };
+}
+
 const features: { [key: string]: FeatureData } = {};
 for (const [key, data] of yamlEntries('feature-group-definitions')) {
+    // Convert markdown to text+HTML.
+    if (data.description) {
+        const { text, html } = convertMarkdown(data.description);
+        data.description = text;
+        data.descriptionHTML = html;
```
ddbeck, line 118: For all our other fields we use `_` as a word separator. Maybe this instead?

```suggestion
        data.description_html = html;
```

ddbeck (approved):
One note on the field name.

I think this is the right thing to do, with the conversion in `index.ts`. I agree that it's harder to peek at the resulting HTML, but I think there's good reasons not to put it into `dist` files:

- We will never directly author HTML. Unlike lists of keys or spec URLs or whatever, the only time we'll want the HTML to be different is when the Markdown ought to be different too. There will never be a good reason to publish different texts depending on format.
- We're not working with large documents where Markdown hazards, such as indentation and long runs between opening and closing syntax, are likely to lead to display errors.
- If we find that we _do_ ship problems with the HTML, we could put in tests against the most likely errors (which, given the text we're working with here, are all but certain to be unclosed <code>`</code> or HTML tags).

PR #883, opened 2024-04-14
Add WebRTC SCTP information

foolip:


feature-group-definitions/webrtc-sctp.yml
```diff
@@ -0,0 +1,11 @@
+name: WebRTC SCTP information
+description: The `sctp` object on `RTCPeerConnection` provides information about the negotiated SCTP transport. SCTP is the protocol used by `RTCDataChannel`.
```
foolip, line 2: @ddbeck do you think https://github.com/web-platform-dx/web-features/tree/main/docs#provides applies here?

ddbeck: I don't think it does, though if you want to dodge it:

```suggestion
description: The `sctp` object on `RTCPeerConnection` represents the negotiated SCTP transport. SCTP (Stream Control Transmission Protocol) is the protocol that `RTCDataChannel` uses.
```

I suggest expanding the abbreviation here, _unless_ it's not widely known by its unabbreviated form (e.g., I probably wouldn't spell out API or TCP/IP).

ddbeck (approved):

feature-group-definitions/webrtc-sctp.yml
foolip: I think that "represents" is accurate, although the its read-only nature is more clear from "information about". But I like it better overall.

PR #884, opened 2024-04-14
Add old and new canvas text features

foolip:
Canvas text baselines is a Baseline 2023 feature.


feature-group-definitions/canvas-text.yml
```diff
@@ -0,0 +1,11 @@
+name: Canvas text metrics
+description: The `fillText()` and  `strokeText()` methods draw text to 2D canvas. The `measureText()` method measures the width and other metrics which can be used to position the text.
```
ddbeck, line 2: nit (could also be "to 2D canvas**es**")

```suggestion
description: The `fillText()` and  `strokeText()` methods draw text to a 2D canvas. The `measureText()` method measures the width and other metrics which can be used to position the text.
```

ddbeck (approved):

foolip:
I had to regenerate `broadcast-channel.dist.yml` too to make CI pass. Sneaking it into this PR so fix it ASAP. (It's broken on main.)

PR #895, opened 2024-04-15
Fix Baseline calculation when support was added then removed

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/876


packages/compute-baseline/src/baseline/index.test.ts.snap
```diff
@@ -1,5 +1,56 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`computeBaseline disregards support that's been removed 1`] = `
+"{
+  \\"baseline\\": false,
+  \\"support\\": {}
```
foolip, line 6: Why doesn't Safari 8 show up here? https://developer.mozilla.org/en-US/docs/Web/API/AudioTrack#browser_compatibility

packages/compute-baseline/src/baseline/support.ts
```diff
@@ -24,9 +24,14 @@ export function support(feature: Feature, browsers: Browser[]): Support {
       }
     }
 
-    logReleaseOmissions(feature, qualifiedReleases, unqualifiedReleases);
+    const currentlySupported = unqualifiedReleases.includes(b.current());
```
foolip, line 27: Ah, this is a nice and simple fix.

packages/compute-baseline/src/baseline/index.test.ts.snap
ddbeck: This is because the branch was behind main and the newest BCD. I've merged main into this branch and updated this snapshot.

feature-group-definitions/starting-style.dist.yml
```diff
@@ -9,8 +9,6 @@ status:
     chrome: "117"
     chrome_android: "117"
     edge: "117"
-    safari: "17.5"
-    safari_ios: "17.5"
```
ddbeck, line 13: @foolip This introduces a potentially undesirable oddity: we will no longer see version info for future-dated releases. I think this is correct, but it is a difference from before.

foolip: Ah, I didn't realize 17.5 was a future release. I agree that's the right change though.

foolip (approved):

Issue #890, opened 2024-04-15
Make status mandatory


foolip:
Let's guarantee that `status` exists for all features in the published package.

cc @ddbeck

foolip:
I also think we should generate `false` for browsers that don't have support, so that it's clear from a single entry what the core browser set is, and what's missing if it's not Baseline.

foolip:
I'm trying part of this in https://github.com/web-platform-dx/web-features/pull/899.

foolip:
Burndown list based on the errors in that PR:

- [ ] anchor-positioning
- [ ] appearance
- [ ] audio-video-tracks
- [ ] background-gradients
- [ ] baseline-source
- [ ] calc-constants
- [ ] calc
- [ ] check-visibility
- [ ] color-scheme
- [ ] compression-streams
- [ ] constructed-stylesheets
- [ ] contain-intrinsic-size
- [ ] counter-style
- [ ] document-picture-in-picture
- [ ] field-sizing
- [ ] font-size-adjust
- [ ] font-synthesis
- [ ] fullscreen
- [ ] highlight
- [ ] http11
- [ ] http2
- [ ] http3
- [ ] image-set
- [ ] intersection-observer-v2
- [ ] intersection-observer
- [ ] light-dark
- [ ] line-clamp
- [ ] managed-media-source
- [ ] mathml
- [ ] media-source
- [ ] min-max-clamp
- [ ] modulepreload
- [ ] navigation
- [ ] picture-in-picture
- [ ] pointer-lock
- [ ] relative-color
- [ ] sanitizer
- [ ] scope
- [ ] scroll-driven-animations
- [ ] scroll-snap
- [ ] scrollend
- [ ] search-input-type
- [ ] show-picker-input
- [ ] slot
- [ ] speech-recognition
- [ ] speech-synthesis
- [ ] text-box-trim
- [ ] text-fragments
- [ ] text-wrap-balance
- [ ] transforms2d
- [ ] transforms3d
- [ ] web-bluetooth
- [ ] web-midi
- [ ] webp

PR #899, opened 2024-04-16
Make Baseline status required

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/890.


foolip:
This depends on adding a lot of statuses, either manually or through .dist.yml.

ddbeck (approved):

PR #901, opened 2024-04-16
Add .dist.yml for 3 CSS features

foolip:
What these have in common is that they had `safari: preview` in
https://github.com/web-platform-dx/web-features/pull/802 but that bug
has now been fixed.


feature-group-definitions/light-dark.yml
```diff
@@ -1,5 +1,3 @@
 name: light-dark()
 description: The `light-dark()` CSS function accepts two colors and uses one depending on the current color scheme.
 spec: https://drafts.csswg.org/css-color-5/#light-dark
-compat_features:
-  - css.types.color.light-dark
```
captainbrosset, line 5: I'm not seeing the `light-dark` tag in BCD. How come we can get rid of this in the source file?

ddbeck: It's a little buried but it's there: https://github.com/mdn/browser-compat-data/blame/7752b0a038a04ee47f112e696706bd53a589b0c6/css/types/color.json#L807

PR #909, opened 2024-04-19
Add inert

foolip:


feature-group-definitions/inert.yml
```diff
@@ -0,0 +1,7 @@
+name: Inert subtrees
```
ddbeck, line 1:
```suggestion
name: Inert elements
```

"Subtrees" is a metaphor we're not going to explain and I don't know if it's self-evident. Since docs more often talk about parents and children, I'm suggesting to sidestep this (with a little help from the description).

Alternatively we could do like `autofocus` though this feels a little too terse to me:

```suggestion
name: Inert
```

feature-group-definitions/inert.yml
```diff
@@ -0,0 +1,7 @@
...
+description: TODO
```
ddbeck, line 2:
```suggestion
description: The `inert` HTML attribute marks an element and its descendants as non-interactive. Inert elements don't get focus or fire `click` events.
```

feature-group-definitions/inert.yml
foolip: I think "`inert` attribute" or "Inert elements" are both OK. https://wpt.fyi/interop-2023 did call this simply "Inert" but I agree it's unnecessarily terse, in that case I'd want it to be `inert` with code syntax to point at the attribute.

ddbeck (approved):

PR #911, opened 2024-04-19
Add messageerror

foolip:


feature-group-definitions/messageerror.dist.yml
```diff
@@ -0,0 +1,36 @@
...
+    firefox: "65"
+    firefox_android: "65"
+    safari: "16.4"
+    safari_ios: "16.4"
+compat_features:
+  # BroadcastChannel as a whole wasn't supported in Edge 18.
+  #- api.BroadcastChannel.messageerror_event
+  - api.DedicatedWorkerGlobalScope.messageerror_event
+  - api.MessagePort.messageerror_event
+  - api.ServiceWorkerContainer.messageerror_event
```
foolip, line 25: Including this increases the Chrome and Firefox versions, and it's not essential for _every_ use of the messageerror event. But I find it hard to argue the feature is complete without it. We could consider partial implementation before this point.

feature-group-definitions/messageerror.yml
```diff
@@ -0,0 +1,22 @@
+name: messageerror
+description: The `messageerror` event TODO.
```
ddbeck, line 2:
```suggestion
description: The `messageerror` event fires on a target, such as a window or worker, when an incoming message cannot be deserialized. This event can fire for many types of messages, such as broadcast channels or cross-document messaging.
```

feature-group-definitions/messageerror.yml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  - https://html.spec.whatwg.org/multipage/indices.html#event-messageerror
+  - https://w3c.github.io/ServiceWorker/#eventdef-serviceworkerglobalscope-messageerror
+compat_features:
+  # BroadcastChannel as a whole wasn't supported in Edge 18.
+  #- api.BroadcastChannel.messageerror_event
```
ddbeck, line 8: Is this exception needed? AFAICT, this key is younger than all the reported status versions, across the board.

foolip: Including it changes Edge 18 to Edge 79, and that doesn't seem right for what is the intersection between two features.

feature-group-definitions/messageerror.yml
```diff
@@ -0,0 +1,22 @@
+name: messageerror
+description: The `messageerror` event fires on a target, such as a window or worker, when an incoming message cannot be deserialized. This event can fire for many types of messages, such as broadcast channels or cross-document messaging.
```
foolip, line 2: @ddbeck what do you think of this further edit?

```suggestion
description: The `messageerror` event fires on a target, such as a window or worker, when an incoming message cannot be deserialized. This event can fire for many types of messages, such as cross-document messages or broadcast channel messages.
```

Trying to avoid the reading that a broadcast channel is itself a kind of message.

PR #912, opened 2024-04-19
Add push messages (Push API)

foolip:


feature-group-definitions/push.yml
```diff
@@ -0,0 +1,51 @@
+name: Push messages
+description: The Push API TODO.
```
ddbeck, line 2:
```suggestion
description: The Push API subscribes to and receives server-initiated messages. Subscribers receive pushed messages in the background, even after periods inactive or offline.
```

foolip:
I've excluded two more keys and added an elaborate TODO about aligning BCD, caniuse and web-features on whether this features is partially supported on iOS or not.

ddbeck (approved):

PR #913, opened 2024-04-19
Add screen orientation and screen orientation lock

foolip:


feature-group-definitions/screen-orientation.yml
```diff
@@ -0,0 +1,10 @@
+name: Screen orientation
+description: The `screen.orientation` API TODO.
```
ddbeck, line 2:
```suggestion
description: The `screen.orientation` API gets information about the orientation of the viewport, such as landscape or portrait. With this API, you can adapt an application's layout or behavior in response to changes in orientation.
```

I went back and forth a couple of times on whether to say `window.screen.orientation` or just `screen.orientation`. I'm assuming you give the latter for a reason, so I left it that way.

feature-group-definitions/screen-orientation-lock.yml
```diff
@@ -0,0 +1,6 @@
+name: Screen orientation lock
+description: The `screen.orientation.lock()` method TODO.
```
ddbeck, line 2:
```suggestion
description: The `screen.orientation.lock()` method prevents changes to the screen orientation, typically in fullscreen applications such as games. For example, while locked, rotating a phone to the side won't change the screen orientation from landscape to portrait.
```

feature-group-definitions/screen-orientation.yml
ddbeck: "Most customary" and using published examples to figure out what's most "most customary" seems like an _excellent_ way to approach this. Opened https://github.com/web-platform-dx/web-features/pull/917 Thank you!

ddbeck (approved):

PR #917, opened 2024-04-19
Description guideline: when to use global/`window` object references

ddbeck:
from https://github.com/web-platform-dx/web-features/pull/913#discussion_r1572601975

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -193,3 +193,10 @@ For example, instead of "The `input()` method enables the usage of…" write "Th
 ([#753](https://github.com/web-platform-dx/web-features/pull/753#discussion_r1560914869))
 
 It's OK to use "usage" only in the sense of "consumption" (as in "mobile data usage") or "customary practice" (as in "the style guide's usage recommendations").
+
+##### `window`
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -193,3 +193,10 @@ For example, instead of "The `input()` method enables the usage of…" write "Th
 ([#753](https://github.com/web-platform-dx/web-features/pull/753#discussion_r1560914869))
 
 It's OK to use "usage" only in the sense of "consumption" (as in "mobile data usage") or "customary practice" (as in "the style guide's usage recommendations").
+
+##### `window`
+
+Because `window` is both the global object and represents the browser window, there are cases where it makes sense to refer to either `window.<object>` or `<object>` (such as `window.fetch` and `fetch`).
```
foolip, line 199:
```suggestion
Because `window` is both the global object and represents the browser window, there are cases where it makes sense to refer to either `window.<name>` or `<name>` (such as `window.fetch()` or `fetch()`).
```

foolip (approved):

foolip (approved):

PR #927, opened 2024-04-22
Update popover.yml

rakleed:
Add information about Firefox support: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/125#apis


foolip:
This depends on https://github.com/mdn/browser-compat-data/pull/22848 being released in BCD, but since we have the status manually here we can just update it now.

foolip (approved):

foolip:
Thank you @rakleed!

PR #928, opened 2024-04-22
Add image-set.dist.yml

foolip:
The Baseline low date is from Safari 17.


feature-group-definitions/image-set.yml
```diff
@@ -2,3 +2,21 @@ name: image-set()
...
+  baseline: low
+  baseline_low_date: 2023-05-05
+  support:
+    chrome: "113"
+    chrome_android: "113"
+    edge: "113"
+    firefox: "89"
+    firefox_android: "89"
+    safari: "14"
+    safari_ios: "14"
```
ddbeck, line 18: caniuse says this is partial until 17. I see you did not reflect that change in the BCD PR. For posterity, want to make a note about what you're thinking here?

(marking approved because I'm presuming reasonableness and I mostly want something to look back on in case anyone asks)

ddbeck (approved):

feature-group-definitions/image-set.yml
foolip: Oh, looks like more BCD updates will be needed here to be consistent between Firefox and Safari regarding support for `type()`.

foolip: @ddbeck can you review https://github.com/mdn/browser-compat-data/pull/22902 and https://github.com/mdn/browser-compat-data/pull/22912? I'll need to build on those.

ddbeck: @foolip Approved and merged the Chrome/Safari PR, but I had questions on the Firefox one. Thank you!

foolip: I've sent https://github.com/mdn/browser-compat-data/pull/22939 and will change to Safari 17 here.

PR #929, opened 2024-04-22
Add media capture (getUserMedia() and enumerateDevices())

foolip:


feature-group-definitions/media-capture.dist.yml
```diff
@@ -0,0 +1,46 @@
...
+# BCD.
+status:
+  baseline: high
+  baseline_low_date: 2017-09-19
+  baseline_high_date: 2020-03-19
+  support:
+    chrome: "53"
+    chrome_android: "53"
+    edge: "12"
+    firefox: "39"
```
foolip, line 22: The Chrome, Edge and Safari versions in https://caniuse.com/stream match here, but Firefox doesn't, caniuse says 36. That comes down to me including `enumerateDevices()` in this feature, which I think is reasonable, since `enumerateDevices()` on its own doesn't amount to much, it's in support of either media capture or audio device output.

feature-group-definitions/media-capture.yml
```diff
@@ -0,0 +1,42 @@
+name: Media capture
+description: TODO
```
ddbeck, line 2: Not sure it's a good idea exactly, but I'm following caniuse's lead here by mentioning the `getUserMedia()` entry point.

```suggestion
description: The `navigator.mediaDevices.getUserMedia()` API requests access to devices that produce streams of data, such as microphones or video cameras.
```

feature-group-definitions/media-capture.yml
```diff
@@ -0,0 +1,42 @@
+name: Media capture
+description: The `navigator.mediaDevices.getUserMedia()` API requests access to devices that produce streams of data, such as microphones or video cameras.
```
foolip, line 2: @ddbeck WDYT about making it more concrete?

```suggestion
description: The `navigator.mediaDevices.getUserMedia()` API requests access to devices that produce audio or video streams, such as microphones or video cameras.
```

ddbeck: Yes, that's an good edit. 👍

ddbeck (approved):

PR #931, opened 2024-04-23
Add `wasm-simd` WebAssembly feature

ddbeck:


feature-group-definitions/simd.yml
```diff
@@ -0,0 +1,6 @@
+name: SIMD
+description: The 128-bit SIMD (Single Instruction Multiple Data) extension to WebAssembly performs one instruction on multiple units of data, when running on hardware that supports such instructions. Also known as vector instructions.
+spec: https://webassembly.github.io/spec/core/bikeshed/#vector-instructions%E2%91%A0
```
foolip, line 3: When there's this trailing stuff representing a circled number like ①, that's because there were multiple generated IDs that conflicted, that needed to be disambiguated. And happily, https://webassembly.github.io/spec/core/bikeshed/#vector-instructions is an anchor on the very same section:

```suggestion
spec: https://webassembly.github.io/spec/core/bikeshed/#vector-instructions
```

feature-group-definitions/simd.yml
```diff
@@ -0,0 +1,6 @@
+name: SIMD
```
foolip, line 1: Even though it's not minimal, I think we should follow caniuse in name and ID here:

```suggestion
name: WebAssembly SIMD
```

There's no feature called just SIMD that to disambiguate from, but SIMD is a concept outside of web technologies, and I think "this browser supports SIMD" sounds a bit off, a bit like "this browser supports GPUs".

ddbeck: Thank you! I initially included WebAssembly but couldn't articulate _why_, so I assumed it was wrong. You explained it perfectly.

ddbeck:
@foolip should we prefix the feature ID, too? e.g., `webassembly-simd` or `wasm-simd`?

foolip:
I think we should make it `wasm-simd`, with the expectation that we'll call WebAssembly `wasm`, just like https://caniuse.com/wasm.

foolip (approved):

PR #933, opened 2024-04-23
Limit the length of names to 80 characters

foolip:
Shorten two CSS function features, the longest of which was 56
characters long. The remaining names that list function names consists
of 3 functions or less.

The longest remaining name is 43 characters long:
ARIA attribute reflection (initial support)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -14,6 +14,9 @@ import { unified } from 'unified';
 
 import { BASELINE_LOW_TO_HIGH_DURATION } from 'compute-baseline';
 
+// The longest name allowed, to allow for compact display.
+const nameMaxLength = 80;
+
 // The longest description allowed, to avoid them growing into documentation.
 const descriptionMaxLength = 300;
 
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -8,6 +8,9 @@ import { Temporal } from '@js-temporal/polyfill';
 
 import { BASELINE_LOW_TO_HIGH_DURATION } from 'compute-baseline';
 
+// The longest name allowed, to allow for compact display.
+const nameMaxLength = 50;
```
captainbrosset, line 12: I like the fact that we enfore a max limit. I imagine this could be a useful guarantee to consumers of the data.
I don't know about picking 50 though. It feels a tad short and I'm guessing we might need to up that limit at some point in the future.
How about picking 100 instead? It feels more manageable on the long run, and would still play the role of communicating **a** limit to consumers.

captainbrosset: My vote goes to 80.

foolip: 80 it is! I'll also back out the name changes that aren't needed then.

index.ts
```diff
@@ -127,9 +130,12 @@ for (const [key, data] of yamlEntries('feature-group-definitions')) {
         data.status.baseline_high_date = String(highDate);
     }
 
-    // Ensure description is not too long.
+    // Ensure name and description are not too long.
+    if (data.name?.length > nameMaxLength) {
+        throw new Error(`name in ${key}.yml is too long, ${data.name.length} characters. The maximum allowed length is ${nameMaxLength}.`);
```
captainbrosset, line 135:
```suggestion
        throw new Error(`The name field in ${key}.yml is too long, ${data.name.length} characters. The maximum allowed length is ${nameMaxLength}.`);
```

index.ts
```diff
@@ -127,9 +130,12 @@ for (const [key, data] of yamlEntries('feature-group-definitions')) {
...
+    }
     if (data.description?.length > descriptionMaxLength) {
-        throw new Error(`description in ${key}.yml is too long, ${data.description.length} characters. The maximum allowed length is ${descriptionMaxLength}.`)
+        throw new Error(`description in ${key}.yml is too long, ${data.description.length} characters. The maximum allowed length is ${descriptionMaxLength}.`);
```
captainbrosset, line 138:
```suggestion
        throw new Error(`The description field in ${key}.yml is too long, ${data.description.length} characters. The maximum allowed length is ${descriptionMaxLength}.`);
```

captainbrosset (approved):

PR #934, opened 2024-04-23
`compute-baseline`: Add README and LICENSE files

ddbeck:


packages/compute-baseline/README.md
```diff
@@ -0,0 +1,57 @@
...
+`npm install --save @mdn/browser-compat-data@latest`
+
+## Usage
+
+```javascript
+import { computeBaseline } from "compute-baseline";
+
+computeBaseline({
+  compatKeys: ["css.properties.border-color"],
+  checkAncestors: false,
```
foolip, line 48: Is it necessary to specify `checkAncestors` if it's false?

And example of bring-your-own-BCD would be helpful here.

foolip (approved):

packages/compute-baseline/README.md
ddbeck: `withAncestors` is required. I put in https://github.com/web-platform-dx/web-features/pull/936 to remove this requirement. I'll push a change shortly illustrating the bring-your-own-BCD stuff.

packages/compute-baseline/README.md
```diff
@@ -0,0 +1,113 @@
...
+To install the package, run:
+
+`npm install --save compute-baseline`
+
+If you wish to specify which version of `@mdn/browser-compat-data` (or manage its upgrades explicitly, such as with Dependabot), then install the latest `@mdn/browser-compat-data` too.
+Run:
+
+`npm install --save @mdn/browser-compat-data@latest`
+
+## Usage
```
ddbeck, line 41: @foolip a few new examples for usage here. Want to take a look?

packages/compute-baseline/README.md
```diff
@@ -40,15 +40,71 @@ Run:
...
+  support: Map(7) { … }
+  toJSON: [Function: toJSON]
+}
+```
+
+Use the `toJSON()` method to get a `web-features`-like plain JSON representation of the status.
+
+### Check support for a single support key
+
+Sometimes it can be hepful to know if parent features have less support than the specific feature you'r checking (for example, the parent is behind a prefix or flag) when computing a status for a deeply-nested feature.
```
foolip, line 76:
```suggestion
Sometimes it can be helpful to know if parent features have less support than the specific feature you're checking (for example, the parent is behind a prefix or flag) when computing a status for a deeply-nested feature.
```

foolip (approved):

PR #935, opened 2024-04-23
Support Markdown in name to produce text+HTML names

foolip:
This doesn't change any published name field, it only adds `name_html`.


foolip:
I have some doubts about this.

Would consumers really want to show a mix of prose and code?

There are cases where it's not obvious what the best use of `<code>` is:

- Array at() or `Array` `at()`?
- Flexbox gap or Flexbox `gap`?
- Temporal or `Temporal`?
- WebTransport or `WebTransport`?

captainbrosset:
This is great because, right now, if you want to display the list of feature names in an HTML page, some feature names end up getting evaluated as HTML.

For example, search.yml defines: `name: <search>`. If I want to display this in an HTML page, I need to do my own escaping. If we changed the name field to be `` name: `<search>` ``, then we'd generate a `nameHTML` property in the final JSON that looks like `"name_html":"<code>&#x3C;search></code>`, which is safe to output in an HTML page.

foolip:
@captainbrosset are you using these names in https://microsoftedge.github.io/TopDeveloperNeeds/ and would you want code to be wrapped in `<code>` in the names? I have some concern it would look messy in a compact list of feature names, but with the presentation on that dashboard there's enough space between things to make that not a strong concern.

In other words, if you'd like to have this, let's add it as `name_html` and let consumers pick.

Note that if you decided you don't want the wrapping `<code>`, you'd still need to take the plaintext and escape it, as it will include `<`. But when used as a text node, that is actually the only character that needs escaping. (In attribute values you need more escaping.)

foolip:
Also @captainbrosset, any thoughts on the cases in https://github.com/web-platform-dx/web-features/pull/935#issuecomment-2076557044?

captainbrosset:
> @captainbrosset are you using these names in https://microsoftedge.github.io/TopDeveloperNeeds/ and would you want code to be wrapped in `<code>` in the names?

Yes, I am using these names for some of the features listed on the dashboard, but none of them include `<`. 

> I have some concern it would look messy in a compact list of feature names.

I tend to agree. While code snippets look good and are useful when used thoughtfully in text, they tend to hinder legibility in short titles. Compare this (current):

image

With this (what would happen if `backdrop-filter` was tagged as a code snippet):

image

The code highlighting doesn't feel meaningfully helpful in the title.

I think my only ask here, as a consumer, is to avoid `<` characters from breaking the HTML rendering.

foolip:
I see. So that request is rather for `name_escaped_as_html`, not the `name_with_extra_code_tags_as_html` I'm going after here.

Or do you mean we just shouldn't have `<` in names, at all?

captainbrosset:
> I see. So that request is rather for `name_escaped_as_html`, not the `name_with_extra_code_tags_as_html` I'm going after here.

Correct

> Or do you mean we just shouldn't have `<` in names, at all?

No, I think it's fine to have them in names.

ddbeck:
FWIW, I like the idea of having `_html` suffixed items (`name_html`, `description_html`, ) meaning "use this in an HTML context" not necessarily guaranteeing that they'll be marked up in precisely the same ways.

foolip:
Is it really worth an extra field just to escape HTML though?

PR #937, opened 2024-04-23
Update array and promise descriptions

foolip:


foolip:
@Elchi3 it would be good to update these before we publish all descriptions as part of the package. I've suggested the structure of the description but left TODOs where it should describe what the thing does.

This isn't high priority and I think we might publish descriptions without these changes too.

(This started out as just adding backticks, but grew into this PR.)

feature-group-definitions/array-at.yml
```diff
@@ -1,5 +1,5 @@
 name: Array at()
-description: Array at() and typed array at() methods
+description: The `at()` method for arrays and typed arrays TODO.
```
foolip, line 2: I'm wondering if we should instead say "Also supported for typed arrays" at the end of each description to keep the descriptions simpler?

feature-group-definitions/array-isarray.yml
```diff
@@ -1,5 +1,5 @@
 name: Array.isArray()
-description: Array.isArray() static method
+description: The `Array.isArray()` static method TODO.
```
ddbeck, line 2: This breaks the rules of circular definition but it feels unreasonable to do otherwise.

```suggestion
description: The `Array.isArray()` static method checks whether a value is an array.
```

feature-group-definitions/array-of.yml
```diff
@@ -1,5 +1,5 @@
 name: Array.of()
-description: Array.of() and typed array of() static methods
+description: The `Array.of()` and typed array `.of()` static methods TODO.
```
ddbeck, line 2:
```suggestion
description: The `Array.of()` and typed array `.of()` static methods create new arrays from the values of any number of arguments.
```

feature-group-definitions/array-splice.yml
```diff
@@ -1,5 +1,5 @@
 name: Array splice()
-description: Array splice() method
+description: The array `splice()` method TODO.
```
ddbeck, line 2:
```suggestion
description: The array `splice()` method changes an array in-place. You can use it to delete items, overwrite items, or insert items, starting from an index.
```

feature-group-definitions/promise-any.yml
```diff
@@ -1,5 +1,5 @@
 name: Promise.any()
-description: Promise.any() static method
+description: The `Promise.any()` static method TODO.
```
ddbeck, line 2:
```suggestion
description: The `Promise.any()` static method returns a promise that fulfills as soon as the first of an iterable of promises fulfills, with that promise's value. Otherwise, it rejects with an `AggregateError` when all of the promises have rejected.
```

feature-group-definitions/promise-finally.yml
```diff
@@ -1,5 +1,5 @@
 name: Promise finally()
-description: Promise finally() method
+description: The promise `finally()` method TODO.
```
ddbeck, line 2:
```suggestion
description: The promise `finally()` method executes a function when the promise settles (resolves or rejects).
```

feature-group-definitions/promise-withresolvers.yml
```diff
@@ -1,5 +1,5 @@
 name: Promise.withResolvers()
-description: Promise.withResolvers() static method
+description: The `Promise.withResolvers()` static method TODO.
```
ddbeck, line 2: I am too tired to make sense of this one today, sorry.

ddbeck (commented):
I proposed a bunch of new descriptions. I had no idea several of these existed (and wished I had, for a few), so some of these descriptions may be less than optimal.

feature-group-definitions/array-by-copy.yml
```diff
@@ -1,5 +1,5 @@
 name: Array by copy
-description: Array and typed array by copy methods toReserved(), toSorted(), toSpliced(), and with()
+description: The `toReserved()`, `toSorted()`, `toSpliced()`, and `with()` methods for arrays and typed arrays TODO.
```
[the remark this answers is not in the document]
foolip: Do you want "of" instead of "for" here too, or only in some cases?

feature-group-definitions/array-flat.yml
```diff
@@ -1,5 +1,5 @@
 name: Array flat() and flatMap()
-description: Array flat() and flatMap() methods
+description: The `flat()` and `flatMap()` methods for arrays and typed arrays TODO.
```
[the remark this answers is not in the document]
foolip: Promote leads me to think that the input is modified in some way, but this always creates a new array, right?

feature-group-definitions/promise-withresolvers.yml
foolip: Live to fight another day!

feature-group-definitions/array-by-copy.yml
```diff
@@ -1,5 +1,5 @@
 name: Array by copy
-description: Array and typed array by copy methods toReserved(), toSorted(), toSpliced(), and with()
+description: The `toReserved()`, `toSorted()`, `toSpliced()`, and `with()` methods for arrays and typed arrays return changed copies of arrays. They stand in contrast to methods such as `sort()` or `reverse()` that change arrays in place.
```
ddbeck, line 2:
```suggestion
description: The `toReserved()`, `toSorted()`, `toSpliced()`, and `with()` methods of arrays and typed arrays return changed copies of arrays. They stand in contrast to methods such as `sort()` or `reverse()` that change arrays in place.
```

feature-group-definitions/array-fill.yml
```diff
@@ -1,5 +1,5 @@
 name: Array fill()
-description: Array fill() and typed array fill() methods
+description: The `fill()` method for arrays and typed arrays sets all or some items of an array to a given a value.
```
ddbeck, line 2:
```suggestion
description: The `fill()` method of arrays and typed arrays sets all or some items of an array to a given a value.
```

feature-group-definitions/array-fromasync.yml
```diff
@@ -1,5 +1,5 @@
 name: Array.fromAsync()
-description: Array.fromAsync() static method
+description: The `Array.fromAsync()` static method copies items of an async iterable object to make a new array.
```
ddbeck, line 2:
```suggestion
description: The `Array.fromAsync()` static method copies items from an async iterable object to make a new array.
```

feature-group-definitions/array-includes.yml
```diff
@@ -1,5 +1,5 @@
 name: Array includes()
-description: Array includes() and typed array includes() methods
+description: The `includes()` method for arrays and typed arrays returns whether a given value appears in the array.
```
ddbeck, line 2:
```suggestion
description: The `includes()` method of arrays and typed arrays returns whether a given value appears in the array.
```

feature-group-definitions/array-flat.yml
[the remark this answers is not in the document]
ddbeck: Excellent point. Having another go at this:

```suggestion
description: The `flat()` and `flatMap()` methods for arrays and typed arrays returns a copy of an array such that the items of nested arrays become direct members of the previously enclosing array. In other words, the methods reduce the number of dimensions of a multidimensional array.
```

feature-group-definitions/promise-withresolvers.yml
ddbeck: Here we are, another day!

```suggestion
description: The `Promise.withResolvers()` static method is an alternative to the `Promise()` constructor that returns the the promise and the resolution functions. You can use this to access `resolve and `reject` outside the scope of the executor function.
```

foolip: Nice, I didn't know what this was, I might just use it some time.

ddbeck (approved):
Happy to see this merge once the dists are refreshed.

feature-group-definitions/array-flat.yml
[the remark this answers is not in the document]
foolip: I think that "become direct members of the previously enclosing array" still sounds like it's going to modify the input array. The "creates a new array"  in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat is key I think.

foolip:
@ddbeck I've applied all changes except the one for `flat()` and `flatMap()`. (I'll update dist all in one go later.)

feature-group-definitions/array-flat.yml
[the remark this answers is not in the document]
ddbeck: Round 3:

```suggestion
description: The `flat()` and `flatMap()` methods for arrays and typed arrays creates a new array such that each nested array item is concatenated into it.
```

(I didn't steal the MDN description wholesale because I think the recursive bit is more confusing than descriptive in the default case of depth 1.)

PR #946, opened 2024-04-24
Add status for HTTP/1.1

foolip:


foolip:
This depends on https://github.com/mdn/browser-compat-data/pull/22930 being released.

foolip:
To not be blocked on a BCD release I've added a status block.

PR #952, opened 2024-04-24
Generate dist file for anchor positioning

captainbrosset:
This adds the dist file for the anchor-positioning feature.
This was generated by the `dist` command, the tag exists in BCD.

Oddly, `support` is empty, even if Chromium-based browsers started supporting it starting with Chrome/Edge 125. I checked the version of BCD we use, and it does have up to date data. I'm also not seeing a BCD key that should be removed for support to data to appear. So I'm missing something.


feature-group-definitions/anchor-positioning.dist.yml
```diff
@@ -0,0 +1,170 @@
+# Generated from: anchor-positioning.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Anchor positioning
+description: Anchor positioning places an element based on the position of another element. For example, you can place a tooltip next to the content it references.
+spec: https://drafts.csswg.org/css-anchor-position-1/#anchoring
+caniuse: css-anchor-positioning
+status:
+  baseline: false
+  support: {}
```
foolip, line 10: I was expecting to see Chrome 125 here. Are one of the BCD keys not supported in Chrome 125?

foolip: Ah, of course, it's because it's in beta and not released yet.

foolip (approved):

feature-group-definitions/anchor-positioning.dist.yml
captainbrosset: Ah ok, that's why! I was surprised too, and looked at all of the keys one by one, failing to find the one that wasn't supported.

PR #953, opened 2024-04-24
Add font-size-adjust.dist.yml

foolip:
<s>Subfeatures are excluded because they're newer than the base feature:
https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust#browser_compatibility</s>


feature-group-definitions/font-size-adjust.dist.yml
```diff
@@ -0,0 +1,16 @@
...
+caniuse: font-size-adjust
+status:
+  baseline: false
+  support:
+    firefox: "3"
+    firefox_android: "4"
+    safari: "16.4"
+    safari_ios: "16.4"
+compat_features:
+  - css.properties.font-size-adjust
```
ddbeck, line 16: The Firefox notes on this are _weird_:

> Before Firefox 40, <code>font-size-adjust: 0</code> was incorrectly interpreted as <code>font-size-adjust: none</code> (<a href="https://bugzil.la/1144885">bug 1144885</a>).

Fixed in version 40, but support for `none` came later in Firefox 72? This doesn't make sense. This makes me wonder whether `none` really was a later addition (and I ought to come up with a test case), or if I'm misunderstanding the meaning of the note and linked bug.

foolip: Firefox 72 is because of https://github.com/mdn/browser-compat-data/issues/22931. We should treat all CSS property values in BCD as suspect until that is resolved.

foolip: I'm fixing that in https://github.com/mdn/browser-compat-data/pull/22936.

ddbeck: Oh yikes about the ranged values. Reviewed and merged https://github.com/mdn/browser-compat-data/pull/22936.

foolip: That hasn't been released yet. I'll add a TODO to unblock this.

PR #954, opened 2024-04-24
Add CSS Module Scripts feature

captainbrosset:
This PR adds a new feature to the repo for CSS Module Scripts.


feature-group-definitions/css-modules.yml
```diff
@@ -0,0 +1,9 @@
+name: CSS module scripts
```
captainbrosset, line 1: I know we've used the term "JavaScript module" for the JS counterpart, but I chose to use the "scripts" suffix here because there was a lot of discussion around this name in the feature proposal originally.

feature-group-definitions/css-modules.yml
```diff
@@ -0,0 +1,9 @@
...
+description: 'CSS module scripts allow CSS code to be organized into reusable units. Other modules use `import ... with {type: "css"}` to load CSS modules as constructable stylesheets.'
+spec:
+  - https://html.spec.whatwg.org/multipage/webappapis.html#css-module-script
+usage_stats:
+  - https://chromestatus.com/metrics/feature/timeline/popularity/3805
+compat_features:
+  - javascript.statements.import
+  - javascript.statements.import.import_attributes.type_css
```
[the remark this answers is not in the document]
foolip: I agree, and if it's deprecated in BCD you'll get a warning that the feature can never be Baseline.

feature-group-definitions/css-modules.dist.yml
```diff
@@ -0,0 +1,16 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+name: CSS module scripts
+description: 'CSS module scripts allow CSS code to be organized into reusable units. Other modules use `import ... with {type: "css"}` to load CSS modules as constructable stylesheets.'
+spec: https://html.spec.whatwg.org/multipage/webappapis.html#css-module-script
+usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/3805
+status:
+  baseline: false
+  support:
+    chrome: "123"
```
foolip, line 11: https://chromestatus.com/feature/5948572598009856 says Chrome 93, is that wrong?

captainbrosset: I believe that's when the old _assertion_ syntax (`assert`) got added. Since the new syntax relies on import _attributes_ instead (`with`), I think 123 is correct.

captainbrosset: See https://chromestatus.com/feature/5205869105250304

foolip: That makes sense. It's akin to the feature being supported with another name or with a prefix, which we don't count as supported for web-features.

foolip (approved):

PR #957, opened 2024-04-25
Remove background gradients feature

foolip:
This grouping of features turned out to not make sense, it's a mix of
very old and rather new stuff. Conic gradients was split into a separate
feature: https://github.com/web-platform-dx/web-features/pull/1017


feature-group-definitions/background-gradients.dist.yml
```diff
@@ -0,0 +1,45 @@
...
+caniuse:
+  - css-gradients
+  - css-repeating-gradients
+  - css-conic-gradients
+status:
+  baseline: high
+  baseline_low_date: 2020-11-17
+  baseline_high_date: 2023-05-17
+  support:
+    chrome: "72"
```
foolip, line 16: This feature is a bit of an outlier in our approach to capturing initial support, and doesn't fit with the direction I've suggested in https://github.com/web-platform-dx/web-features/issues/866.

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient#browser_compatibility is probably the oldest part of this feature, and it's very old!

Should we actually split this feature along the same lines as caniuse.com?

ddbeck: This is where we combined them in the first place: https://github.com/web-platform-dx/web-features/pull/398

Maybe we have another composition case. We seem to encounter some cases where it'd be nice to shorthand to say that feature Y is feature A+B. That has a bit of messiness for consumers, so I don't think we can do that _now_. But we can prepare for that future by splitting the features up and grouping them.

ddbeck: I suppose now we have https://github.com/web-platform-dx/web-features/issues/971.

I think it is something like #217 plus computed statuses based on child features. There's an open question of how consumers would see a composite feature. Would the composite feature contain a list of compat keys, the union of the subordinate features? Or would the composite feature contain references to the subordinate features only.

foolip: How about I create a group for gradients and split this into features like caniuse?

ddbeck: Yes, let's do that. (Sorry that I was unclear about suggesting this last week.)

foolip: I ended up sending https://github.com/web-platform-dx/web-features/pull/1017. The other two features are older, and while we should eventually have them, I don't see why we should spend time on it right now. I'll update this PR to delete instead.

ddbeck (approved):

PR #960, opened 2024-04-25
Add notifications (it's complicated)

foolip:


feature-group-definitions/notifications.yml
```diff
@@ -0,0 +1,38 @@
+name: Notifications
+description: The notifications API TODO.
```
ddbeck, line 2:
```suggestion
description: The notifications API sends system notifications to the user, often even when the page or browser is not the foreground application.
```

feature-group-definitions/notifications.yml
```diff
@@ -0,0 +1,38 @@
...
+# - showNotification() in service workers for persistent notifications.
+#
+# Finally, BCD and caniuse.com disagree about whether it was Safari 6 or 7 that
+# first supported notifications, and BCD originally agreed with caniuse.com:
+# https://github.com/mdn/browser-compat-data/pull/7949
+#
+# Due to this complexity, manually combine these 3 data points:
+# https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification#browser_compatibility
+# https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification#browser_compatibility
+# https://caniuse.com/notifications (excluding partial support on iOS)
```
ddbeck, line 29: After this merges, I'd like you to file an issue that just does the fancy preview permalink thing to this comment. Even after we tackle the various issues that prevent this from being handled normally, it'll still be a significant effort to unravel it for this feature specifically.

foolip: I have filed https://github.com/web-platform-dx/web-features/issues/973 and will update the link after this PR lands.

ddbeck (approved):

PR #962, opened 2024-04-25
Add locks

foolip:


feature-group-definitions/locks.yml
```diff
@@ -0,0 +1,13 @@
+name: Locks
```
foolip, line 1: Should we call it just locks, web locks, or something else?

That API is a bit like a mutex, allowing different bits of code to avoid using the same resource at the same time, but that's something you do with the API, the API itself doesn't "lock resources" or something like that.

ddbeck: Oh good the spec is "web locks." I like how this, said aloud, sounds like "web logs." Let's leave it as "locks" for now. If it's confusing, we can revisit it with developer feedback.

feature-group-definitions/locks.yml
```diff
@@ -0,0 +1,13 @@
...
+description: The `navigator.locks` API TODO.
```
ddbeck, line 2:
```suggestion
description: The `navigator.locks` API coordinates work with shared resources through mutually exclusive ownership of a resource's name. Also known as web locks.
```

foolip:
I've changed the identifier to `web-locks` to match the spec, because I'm nervous about using just "locks" given the existence of all these lock specs:
https://wicg.github.io/keyboard-lock/
https://w3c.github.io/pointerlock/
https://w3c.github.io/screen-wake-lock/

ddbeck (approved):

PR #965, opened 2024-04-25
Add ::scroll-marker as a draft feature

foolip:


foolip:
The purpose of this isn't really to track `::scroll-marker`, but to find a way to pick a tentative identifier for a feature early in its lifecycle, without publishing it in web-features. That's motivated by @jstenback working on use counters for web-features in Chromium. If we want Chromium engineers to add use counters when preparing a feature for shipping (before they send an Intent to Ship), then they need a way to pick an identifier in a very low-overhead way, and leave the rest of the work for web-features maintainers.

ddbeck:
My main worry is that we'll end up with many draft features that never get any follow-up. Suppose `::scroll-marker` gets abandoned? Or `::scoll-marker` becomes `::scroll-widget`? Who makes sure we clean this up later?

It's not quite as light-weight, but what if we had a little editorial section in the YAML file, to attach some accountability to this? Something like…

```
draft: true
editorial:
  editors:
    - foolip
    - jstenback
  draft-on: 2024-04-26
```

My use-case here would be to have a script that shows me all the drafts older than year and who to @-mention about it.

foolip:
I've added a `draft_date` and filed https://github.com/web-platform-dx/web-features/issues/1009 with what I think we should do.

ddbeck (approved):

Issue #955, opened 2024-04-24
Make it obvious when something is in beta


foolip:
In https://github.com/web-platform-dx/web-features/pull/952 I was surprised that Chrome 125 wasn't listed, but it's because that's in beta.

@ddbeck can we put something in dist files that provides a clue that something is in beta, so that we can be confident we got the list of BCD keys correct before it hits stable?

captainbrosset:
I was surprised that see the `support` object being empty too. Would it make sense to still populate that object even if baseline is false?

foolip:
I think it'd be good to populate it with all of the core browser set, and setting the value to false where not supported. Is that what you mean?

captainbrosset:
Yeah, I think that would be ideal. This way, consumers have a consistent list of browsers in each feature's `support` object, whether a feature is supported in 0 or N browsers, and whatever its baseline status (false, low, high):

For example:

* audio-session:
* 
```json
  "audio-session": {
    "compat_features": [ ... ],
    "description": "...",
    "description_html": "...",
    "name": "Audio session",
    "spec": "https://w3c.github.io/audio-session/",
    "status": {
      "baseline": false,
      "support": {
        "chrome": false,
        "chrome_android": false,
        "edge": false,
        "firefox": false,
        "firefox_android": false,
        "safari": "16.4",
        "safari_ios": "16.4"
      }
    }
  }```

* anchor-positioning:

```json
  "anchor-positioning": {
    "caniuse": "css-anchor-positioning",
    "description": "...",
    "description_html": "...",
    "name": "Anchor positioning",
    "spec": "https://drafts.csswg.org/css-anchor-position-1/#anchoring",
    "status": {
      "baseline": false,
      "support": {
        "chrome": "125",
        "chrome_android": "125",
        "edge": "125",
        "firefox": false,
        "firefox_android": false,
        "safari": false,
        "safari_ios": false
      }
    }
  }
```

For comparison, anchor-positioning is currently:

```json
  "anchor-positioning": {
    "caniuse": "css-anchor-positioning",
    "description": "...",
    "description_html": "...",
    "name": "Anchor positioning",
    "spec": "https://drafts.csswg.org/css-anchor-position-1/#anchoring"
  }
```

ddbeck:
OK, I think what needs to happen here is:

- [ ] Change computeBaseline such that the returned `support` `Map` shows all the browsers (even pre-releases) but `toJSON()`'s `support` shows only shipped browsers as a version number
[...]

PR #968, opened 2024-04-25
Add highlight.dist.yml

foolip:


feature-group-definitions/highlight.dist.yml
```diff
@@ -0,0 +1,43 @@
+# Generated from: highlight.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Custom highlights
+description: Custom highlights style arbitrary text ranges, without adding extra elements to the DOM.
+spec: https://drafts.csswg.org/css-highlight-api-1/
+status:
+  baseline: false
+  support:
```
captainbrosset, line 9: Firefox seems to have support for the API in at least Nightly. I see BCD says "preview". I'm not sure what that means though. Should our support mention Firefox?

foolip: This is the same problem as in https://github.com/web-platform-dx/web-features/issues/955, that the definition of Baseline only considers stable releases, but we should make it visible to reviewers that support might be upcoming.

ddbeck (approved):

PR #969, opened 2024-04-25
Make CSS property value vs CSS declaration consistent

foolip:
Both forms were previously used.


captainbrosset (approved):
I like consistency

foolip:
@ddbeck I know you also like consistency, but do you like the term "CSS declaration"?

https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/CSS_Declaration

ddbeck (approved):

ddbeck:
> @ddbeck I know you also like consistency, but do you like the term "CSS declaration"?

I certainly do https://github.com/web-platform-dx/web-features/pull/974

PR #975, opened 2024-04-26
Add a group for CSS layout

foolip:
Only include features that are unambiguously layout features and not
borderline. `anchor-positioning` and `aspect-ratio` are such borderline
cases, which do affect size and/or position, but aren't full layout
models like Grid is.


ddbeck (approved):
Approved, pending dist refreshes.

PR #976, opened 2024-04-26
Add status for MathML, bypassing BCD entirely

foolip:
The tags are removed from BCD here:
https://github.com/mdn/browser-compat-data/pull/22950


ddbeck:
I'm confused here. We're dropping _both_ the compat keys and the tags in BCD? So this effectively only for mapping to caniuse?

ddbeck:
OK, that works for me. I like having written it down, so later on when somebody wants that feature to have keys, we can look back up on this conversation and be able to explain why we did what we did and what that person must do to make it happen. I just hope it's not me. 😄

ddbeck (approved):

foolip:
I also don't want to be on the hook for sorting out MathML :)

I'll file an issue though, maybe a kind soul will feel inspired to go digging.

PR #977, opened 2024-04-26
Add a status for unprefixed line-clamp (not supported at all yet)

foolip:


ddbeck (approved):
Honestly, I do not love the change in https://github.com/mdn/browser-compat-data/pull/22951. It seems to me that spec'd prefixes ought to be their own keys in BCD (i.e., we ought to have _both_ `css.properties.-webkit-line-clamp` and `css.properties.line-clamp`), since in a lot of ways it's more like a shorthand than a prefix.

That said, that change is now in BCD, so let's go.

foolip:
@ddbeck I think that having both `line-clamp` and `-webkit-line-clamp` in BCD might be OK, but it does create some additional questions. I'll file a BCD issue for discussion.

PR #984, opened 2024-04-29
Update name and a.k.a of picture-in-picture

foolip:
Since there's a document picture-in-picture feature and this API is
video-specific, it makes sense to call this video picture-in-picture.

Also add the common PiP shorthand to the description.


foolip:
This doesn't conflict with https://github.com/web-platform-dx/web-features/pull/983 but it will break CI if it lands after that PR, and vice versa. Reviewers, please don't merge, and I'll take care of updating the other PR when one lands.

feature-group-definitions/picture-in-picture.yml
```diff
@@ -1,5 +1,5 @@
-name: Picture-in-picture
-description: The picture-in-picture API allow websites to create a floating, always-on-top video window.
+name: Video picture-in-picture
```
ddbeck, line 1: Suggesting this instead, though I don't think we've done anything like this except for initial support:

```suggestion
name: Picture-in-picture (video)
```

My thinking here is that:

- People already call this "picture-in-picture" (and it directly analogizes to non-web TV technologies) but I couldn't find any good examples of this being called "video picture-in-picture". We probably shouldn't invent a backformation ourselves.
- I suspect "document picture in picture" will be a marked form for some time. Variants like "[Document in Picture-in-Picture](https://developer.chrome.com/docs/web-platform/document-picture-in-picture)" imply that document is stuffing something extra or unusual into the PiP window.

feature-group-definitions/picture-in-picture.yml
```diff
@@ -1,5 +1,5 @@
...
+description: The picture-in-picture API allow websites to create a floating, always-on-top video window. Also known as PiP.
```
ddbeck, line 2:
```suggestion
description: The picture-in-picture API allow websites to create a floating, always-on-top video window. Also known as PiP or pop-out video.
```

Ran across this term while doing some searches for my previous comment. [Vivaldi user docs](https://help.vivaldi.com/desktop/media/pop-out-video/).

feature-group-definitions/picture-in-picture.yml
foolip: The parenthetical works I think, that's enough to make it clear if the two PiP features are seen by name only.

ddbeck (approved):

PR #987, opened 2024-04-29
Add `transferable-arraybuffer` feature and `transferable-objects` group

ddbeck:
I've added in the group in anticipation of transferable streams, etc.


feature-group-definitions/transferable-arraybuffer.dist.yml
```diff
@@ -0,0 +1,22 @@
+# Generated from: transferable-arraybuffer.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Transferable ArrayBuffer
+description: The `transfer()` and `transferToFixedLength()` methods of `ArrayBuffer` move a buffer from one context to another (for example, to a worker).
+group: transferable-objects
```
foolip, line 6: Optional suggestion, ignore if you already considered it:

```suggestion
group: transferable
```

FWIW, these are the `[Transferable]` things defined with Web IDL:

- `OffscreenCanvas`
- `ImageBitmap`
- `MessagePort`
- `MediaSourceHandle`
- `ReadableStream`
- `WritableStream`
- `TransformStream`
- `AudioData`
- `VideoFrame`
- `WebTransportSendStream`
- `WebTransportReceiveStream`

My hunch is that not all of these things will be in features by themselves, and the features they are part have some other primary location in a feature group hierarchy, like `MediaSourceHandle` being a part of Media Source Extensions or some media group.

But we're not publishing groups yet, and we can change things even after we publish groups. Let's find out what works and not.

foolip (approved):

PR #991, opened 2024-04-29
Add `fetch-metadata-request-headers`

ddbeck:
The name is a bit of a mouthful, but it seems to be used fairly widely. We could go with the shorter `fetch-metadata` ID and name, though, if preferred.


feature-group-definitions/fetch-metadata-request-headers.yml
```diff
@@ -0,0 +1,8 @@
+name: Fetch metadata request headers
+description: The headers `Sec-Fetch-Dest`, `Sec-Fetch-Mode, `Sec-Fetch-Site`, and `Sec-Fetch-User` provide extra information about the way a request was made, to help servers reject certain kinds of malicious requests.
+spec: https://w3c.github.io/webappsec-fetch-metadata/
```
foolip, line 3: Let's make the feature identifier `fetch-metadata` to match https://www.w3.org/TR/fetch-metadata/. (Even when TR isn't published, it's always been the repo name without `webappsec-` in cases I remember.)

feature-group-definitions/fetch-metadata-request-headers.yml
```diff
@@ -0,0 +1,8 @@
+name: Fetch metadata request headers
+description: The headers `Sec-Fetch-Dest`, `Sec-Fetch-Mode, `Sec-Fetch-Site`, and `Sec-Fetch-User` provide extra information about the way a request was made, to help servers reject certain kinds of malicious requests.
```
foolip, line 2:
```suggestion
description: The `Sec-Fetch-Dest`, `Sec-Fetch-Mode, `Sec-Fetch-Site`, and `Sec-Fetch-User` HTTP headers provide extra information about the way a request was made, to help servers reject certain kinds of malicious requests.
```

foolip (approved):
This is good to go with a rename and description tweak.

PR #993, opened 2024-04-29
Add :modal

foolip:
This is a Baseline 2022 feature.


feature-group-definitions/modal.yml
```diff
@@ -0,0 +1,6 @@
+name: ":modal"
+description: "The `:modal` pseudo-class TODO."
```
ddbeck, line 2:
```suggestion
description: "The `:modal` pseudo-class selects elements that put other elements into a non-interactive state, such as a fullscreen element or a `<dialog>` element used with `showModal()`."
```

ddbeck (approved):

PR #995, opened 2024-04-29
Add `dirname`

ddbeck:


feature-group-definitions/dirname.yml
```diff
@@ -0,0 +1,8 @@
+name: dirname
```
foolip, line 1: Before I read the description I though that `dirname` must have something to do with directory names, since there are many APIs like https://nodejs.org/docs/latest/api/path.html#pathdirnamepath and https://docs.python.org/3/library/os.path.html#os.path.dirname.

But I _don't_ think we should try to disambiguate this, let's see what kind of feedback we get first.

I've done a quick search of https://github.com/w3c/webref and can't find any other web platform features that contain the string "dirname", which is good.

foolip (approved):

PR #996, opened 2024-04-29
Add `URL.canParse()` feature

ddbeck:


foolip (approved):
If this had been shipped together with `URL.parse()` we definitely would have considered the pair a single feature. Maybe we will in the future. And maybe both will blend into `URL` with time.

Right now, this looks like a distinct feature, however.

PR #998, opened 2024-04-29
Add relative-color.dist.yml

foolip:
Matches https://caniuse.com/css-relative-colors.


ddbeck (approved):
I continue to (not) love how messy this is.

PR #999, opened 2024-04-29
Add `Document.parseHTMLUnsafe()` and friends

foolip:


features/parse-html-unsafe.yml
```diff
@@ -0,0 +1,7 @@
+name: Unsafe HTML parsing methods
```
foolip, line 1: @josepharhar can you help explain what this feature is and how web developers should understand it? I think it will seem a bit strange to say that the web has a new "feature" that allows unsafe (compared to what?) parsing of HTML. Does it only make sense when compared to something else? Should web developers start using this, or stay away forever because it's unsafe?

josepharhar: I just talked to mason about this, so the long term plan is that there will be a parseHTML (and setHTML) method without the "unsafe" which uses the not-yet-finished sanitizer API to make it "safe". The reason that we shipped this "unsafe" version now is because we needed a way to parse declarative shadowdom from script.

https://github.com/otherdaniel/purification/blob/main/explainer.md#proposed-api

@otherdaniel @mfreed7

foolip: I see. Is it web developers who need some way to parse declarative shadowdom from script, or is it really for a polyfill or something? If web developers might use it, then is "unsafe HTML parsing" the right label to put on it?

ddbeck: Stepping in here with some suggestions based on this conversation (definitely not adding any real knowledge of the feature).

One way to name this might be to talk more specifically than "safety" (safe for whom? for what purpose?):

```suggestion
name: Unsanitized HTML parsing methods
```

This would imply a future feature called "Sanitized HTML parsing methods" or similar.

An alternative route would be to ignore the safe/unsafe distinction for now, with the understanding that this feature would become "HTML parsing methods (initial support)" when something like "Sanitized HTML parsing methods" appear later.

 ```suggestion
name: HTML parsing methods
```

File: features/parse-html-unsafe.yml [hunk 2]
```diff
@@ -0,0 +1,7 @@
...
+description: The `Document.parseHTMLUnsafe()` static method and `setHTMLUnsafe()` method of `Element` and `ShadowRoot` TODO.
```
ddbeck, line 2: Just leaving an idea here that if santization or safety is part of the framing for this, then we should just go ahead and give readers a handle on that directly. The last sentence here, for instance:

```suggestion
description: The `Document.parseHTMLUnsafe()` static method parses HTML into a DOM tree, while the `setHTMLUnsafe()` method of `Element` and `ShadowRoot` parses and inserts HTML into an existing tree. No sanitization applies to these methods, so never call them with user-provided HTML strings.
```

features/parse-html-unsafe.yml
mfreed7: > Is it web developers who need some way to parse declarative shadowdom from script, or is it really for a polyfill or something? If web developers might use it, then is "unsafe HTML parsing" the right label to put on it?

This is definitely a developer-facing feature. It can be used for parsing any HTML content, in particular when it contains declarative shadow dom content.

> This would imply a future feature called "Sanitized HTML parsing methods" or similar.

I really like the "Unsanitized" description of this feature, which is exactly what makes it "unsafe". But that's a more specific description of why it's unsafe.

foolip: Let's go with "Unsanitized HTML parsing methods". I am tempted to drop "methods" but that makes it unclear if it's just talking about the main HTML parser. So let's keep that qualifier.

Issue #979, opened 2024-04-27
Authorize `compute-baseline` to grant Baseline statuses to single keys in `@mdn/browser-compat-data`


ddbeck:
Also known as "blessing" single-compat keys' status from compute-baseline.

## Background

Through a combination of governance documentation, Baseline definition, and convention, explicit review and publication is required for a status to be assigned to a web platform feature. This issue proposes that we authorize—under a specific combination of code and data—certain BCD keys to receive statuses without explicit review.

## Motivation

Typically, we expect consumers of Baseline data to show a status for a feature as a whole, taken directly from the `web-features` package. For example, if you go to [the Popover API overview page on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API), then you ought to see a status relating to whole popover feature.

However, this doesn't make sense in every context. If you go to the MDN reference page for a specific popover property or method, then MDN ought to be able to show a Baseline status that reflects that _specific_ property or method's status.

## Proposal

Permit official Baseline statuses to be generated from standalone BCD keys, using an approved method in the `compute-baseline` package.

Some additional limitations:

- Only keys that are part of a feature, group, or snapshot of `web-features` are eligible for a status. Arbitrary, uncategorized keys will not be eligible for computation (e.g., since they might be still under editorial review). This ensures that the underlying data will receive some scrutiny before setting a status.
- Statuses should only be generated from the current or previous releases of `compute-baseline`, `web-features`, and `@mdn/browser-compata-data`. Consumers using this approach should take care to avoid stale statuses.

## Tasks

If adopted, we'd need to do the following:

- [ ] Make required changes to `compute-baseline`: create an authorized single-key method that returns a plain JS object like the data found in the `status` block of `web-features` data
- [ ] Document `web-features` and `compute-baseline` as the only official sources of Baseline status data

Optionally we might:

- [ ] Provide pre-computed JSON for non-JS consumers who need this data
- [ ] Consider synchronizing `compute-baseline`'s version number with `web-features` (from 1.0)

foolip:
> This issue proposes that we authorize—under a specific combination of code and data—certain BCD keys to receive statuses without explicit review.

Would we not consider it reviewed via the feature? With dist files in the style of https://github.com/web-platform-dx/web-features/issues/866#issuecomment-2078907927, the status of each individual BCD key would be checked in and reviewed.

foolip:
[...]

PR #1002, opened 2024-04-29
Add scroll-driven-animations.dist.yml

foolip:


feature-group-definitions/scroll-driven-animations.dist.yml
```diff
@@ -0,0 +1,40 @@
...
+  - css.properties.animation-timeline.scroll
+  - css.properties.animation-timeline.view
+  # https://crbug.com/337580010 tracks this bug, but setting the timeline via
+  # the animation shorthand is not necessary to use the feature.
+  # - css.properties.animation.animation-timeline_included
+  - css.properties.scroll-timeline
+  - css.properties.scroll-timeline-axis
+  - css.properties.scroll-timeline-name
+  # Later addition (Chrome 116)
+  # - css.properties.timeline-scope
```
ddbeck, line 36: I would like to be convinced on this. What's the case for pulling this one out, when they landed just four weeks apart?

foolip: https://chromestatus.com/feature/6752840701706240 says 115 and I expect caniuse would say 115 too if it was added. One bit of an API coming a release later is pretty common I think. Unless it's a terrible accident and the feature is unusable without that bit, I think https://github.com/web-platform-dx/web-features/issues/979 is the way to handle such cases.

ddbeck (approved):

feature-group-definitions/scroll-driven-animations.dist.yml
captainbrosset: Maybe @bramus can share some insights about whether the `timeline-scope` property is a vital part of scroll-driven-animations, or whether the feature is mostly usable without it.

bramus: `timeline-scope` is there for when want to animate an element based on a scroller **that is not a parent of the animated element**.

In most cases though, the animation subject _is_ a child of the scroller – I’ve only used `timeline-scope` a few times in my demos.

PR #1003, opened 2024-04-29
Add `form-associated-custom-elements` (`ElementInternals`)

ddbeck:


feature-group-definitions/form-associated-custom-elements.yml
```diff
@@ -0,0 +1,19 @@
+name: Form-associated custom elements
+description: The `attachInternals()` method of `HTMLElement` allows you to create custom elements that act like built-in form elements.
```
ddbeck, line 2: The description here is a little funny ("allows you", meh) but this is the closest thing to an entry point that gets directly documented. (Truly, it's something closer to "putting an `formAssociated` static property on your custom element allows you to…").

foolip: I presume that the "Custom state pseudo-class" bit in https://html.spec.whatwg.org/multipage/custom-elements.html#elementinternals would also have to go through `attachInternals()`? That wouldn't make this description wrong, but could it be written in a way that would be similar for other uses of `ElementInternals` without claiming `attachInternals()` as the sole entry point?

ddbeck: Hmm, yes, that's a good point. I was trying to make this item searchable by including the `internals` bit, but since this isn't unique, perhaps we can relocate the non-unique bit to the end. How about this?

```suggestion
description: Custom elements may act like built-in form elements, via the the `attachInternals()` method of `HTMLElement` and the `ElementInternals` API.
```

foolip: Yeah, I like that! It could be expanded by naming one of the bits of `ElementInternals` API that's useful for this, but not necessary. I'll approve and let you decide.

foolip (approved):

PR #1006, opened 2024-04-29
Add text-box-trim.dist.yml

foolip:


ddbeck (commented):
Maybe this is also a candidate to be a draft feature? I don't love that there's no implementation at all.

foolip:
This now depends on https://github.com/web-platform-dx/web-features/pull/965.

ddbeck (approved):

PR #1007, opened 2024-04-30
Split text fragments into a group with two features

foolip:
This is very unfortunate, but without partial support there's no
reasonable status to set for the previous feature.

As a consolation, we now get two use counters instead of one.

BCD update: https://github.com/mdn/browser-compat-data/pull/22966/


groups/text-fragments.yml
```diff
@@ -0,0 +1,11 @@
+# This group was originally a feature, which has 3 parts:
+#  1. Scrolling to a text fragment such as #:~:text=snippet.
+#  2. Feature detection for this, document.fragmentDirective. This cannot be
+#     relied upon due to https://webkit.org/b/273466. See also
+#     https://github.com/WICG/scroll-to-text-fragment/issues/257.
```
ddbeck, line 5: Is there a reason we don't have a feature for this? Something like `text-fragment-feature-detection`?

foolip: That could make sense, but I'm not sure what the outcome of https://github.com/WICG/scroll-to-text-fragment/issues/257 will be. I think that if `document.fragmentDirective` remains in the spec and is eventually implemented, it would make the most sense to retroactively consider Safari's initial implementation partial.

ddbeck: OK, that's cool. Thank you!

ddbeck (approved):

PR #1012, opened 2024-04-30
Add ESM worker features

ddbeck:


feature-group-definitions/js-modules-for-workers.dist.yml
```diff
@@ -0,0 +1,20 @@
...
+  support:
+    chrome: "80"
+    edge: "80"
+    firefox: "114"
+    firefox_android: "114"
+    safari: "15"
+    safari_ios: "15"
+compat_features:
+  - api.Worker.Worker.ecmascript_modules
+  - api.SharedWorker.SharedWorker.ecmascript_modules
```
ddbeck, line 20: It wasn't 100% clear to me whether to include `SharedWorker` in this. Without it, this feature is Baseline low (Chrome for Android does not support shared workers.)

feature-group-definitions/js-modules-for-service-workers.yml
```diff
@@ -0,0 +1,7 @@
+name: Service worker JavaScript modules
```
ddbeck, line 1: I'm not sure about these names and IDs. It's unclear to me whether the emphasis is on modules (i.e., JS modules for…) or worker types (i.e., Workers accept JS modules).

feature-group-definitions/js-modules-for-service-workers.yml
```diff
@@ -0,0 +1,7 @@
...
+description: 'The `navigator.serviceWorker.register()` method accepts `{ type: "module" }` to load scripts that use `import` and `export`. Also known as ECMAScript modules or ESM in service workers.'
+spec:
+  - https://html.spec.whatwg.org/multipage/workers.html#dom-worker-dev
+  - https://html.spec.whatwg.org/multipage/workers.html#shared-workers-and-the-sharedworker-interface:dom-sharedworker-2
```
foolip, line 5:
```suggestion
spec: https://w3c.github.io/ServiceWorker/#dom-registrationoptions-type
```

feature-group-definitions/js-modules-for-service-workers.yml
```diff
@@ -0,0 +1,7 @@
...
+compat_features:
+  - api.ServiceWorker.ecmascript_modules
```
foolip, line 7: Yep, that's where it is. I think the more correct place would be under `api.ServiceWorkerContainer.register`, but arguably that's not where people are going to find it. So let's leave this alone until it changes in BCD, which I'm not going to push for.

feature-group-definitions/js-modules-for-service-workers.yml
foolip: My guess is the devs are asking is "can I use ESM (everywhere)?" more than "I'm writing service worker code, can I use ESM?"

I think we should think about ESM as a single feature that's split into multiple pieces only because of how it's rolled out, and strive to merge it eventually or make a composite feature. For the name I suggest.

```suggestion
name: JavaScript modules in service workers
```

Identifier `js-modules-service-workers` so that we can change our mind about "in" easily.

feature-group-definitions/js-modules-for-workers.dist.yml
foolip: Since it does affect the status, I say let's split it. We'll have these 3 "subfeatures" of JS modules:

- JavaScript modules in dedicated workers
- JavaScript modules in service workers
- JavaScript modules in shared workers

foolip (approved):

PR #1015, opened 2024-04-30
Manually set status for Web Bluetooth

foolip:


ddbeck (approved):
I know it doesn't add anything, but if you wanted to have a dist here, then we could get a little closer to making those mandatory too. But feel free to merge with or without.

foolip:
@ddbeck that'll be my next burndown list, and I left it out here knowing that it wouldn't have any computed status or compat keys.

PR #1016, opened 2024-04-30
Add web-midi.dist.yml

foolip:


feature-group-definitions/web-midi.yml
```diff
@@ -2,3 +2,53 @@ name: Web MIDI
...
+  - api.MIDIPort.id
+  - api.MIDIPort.manufacturer
+  - api.MIDIPort.name
+  - api.MIDIPort.open
+  - api.MIDIPort.state
+  - api.MIDIPort.statechange_event
+  - api.MIDIPort.type
+  - api.MIDIPort.version
+  - api.Navigator.requestMIDIAccess
+  # Later additions:
```
captainbrosset, line 51: I suppose we had to bring all of the BCD keys in this file and comment out a few because we don't have #841 yet. Is that it?

foolip: Kind of. If we tagged all of Web MIDI in BCD with "web-features:web-midi" then we would need a mechanism like https://github.com/web-platform-dx/web-features/issues/841.

But after all of the recent work my idea of the ideal design has shifted somewhat, and https://github.com/web-platform-dx/web-features/issues/979 is a path to giving individual BCD keys a Baseline status separate from the overall feature's status. https://github.com/web-platform-dx/web-features/issues/866#issuecomment-2078907927 is a mockup of what I'd like to see when reviewing a feature.

So I think that the way to resolve these "later additions" comments is likely to tag everything in BCD, review the groups of statuses that result, and then designate one of those groups (usually the earliest) as the status for the overall feature.

ddbeck (approved):

PR #1021, opened 2024-05-01
Add array*.dist.yml

foolip:


feature-group-definitions/array-fill.dist.yml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.fill
+  - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%25typedarray%25.prototype.fill
+snapshot: ecmascript-2015
+group:
+  - arrays
+  - typed-arrays
+status:
+  baseline: high
+  baseline_low_date: 2016-03-21
```
foolip, line 15: This changed due to the Firefox and Safari versions changing, because `fill()` didn't ship on array and typed arrays at the same time. @Elchi3 how do you think we should handle this?

foolip:
Closing this, it'll be easier to review if we first add dist, and then update the source in separate PRs.

PR #1024, opened 2024-05-02
Rename array and promise features to singular

foolip:
This is to match the name naming of Map and Set.


ddbeck (approved):
Submitted https://github.com/web-platform-dx/web-features/pull/1026 to go with this.

PR #1033, opened 2024-05-03
Add `origin-private-file-system` feature

ddbeck:


features/origin-private-file-system.yml
```diff
@@ -0,0 +1,6 @@
+name: Origin private file system
+description: The `navigator.storage.getDirectory()` method returns a `FileSystemDirectoryHandle` that is restricted to a specific origin and invisible to the user's actual file system for faster file-based applications, such as SQLite databases.
+spec: https://fs.spec.whatwg.org/#origin-private-file-system
+compat_features:
+  - api.StorageManager.getDirectory
+  - api.FileSystemFileHandle.createSyncAccessHandle
```
foolip, line 6: Is there a principle for which parts of the API surface are counted as this API? Everything that shipped in the same versions of browsers as the entry point would be a good start, but is it just this one method?

ddbeck: Mostly, I just don't know this API well enough to be a good judge. I took a conservative approach here and put only those things that looked (to my inexperienced eye) _exclusive_ to OPFS. A less conservative approach would probably overlap significantly with [`file-and-directory-entries`](https://github.com/web-platform-dx/web-features/blob/main/features/file-and-directory-entries.yml).

I'm happy to do something reasonable here, I just didn't know where to start drawing the lines (or if we even ought to draw such lines).

captainbrosset: `getDirectory` is definitely the right entry-point to the OPFS feature. But I think we should also take a look at the `FileSystemHandle` interface and list what is supported where, and what, from it, OPFS really needs. Some of it is likely only needed for "real" file system access, and doesn't need to be listed here.

I think @tomayac would be a good person to help us list the exhaustive pieces of OPFS that developers really need to use the feature.

tomayac:
For the pure OPFS, you need:

- [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle)
  - [FileSystemFileHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle)
  - [FileSystemDirectoryHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle)
- [FileSystemSyncAccessHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemSyncAccessHandle)
- [FileSystemWritableFileStream](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream)
- [StorageManager.getDirectory()](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/getDirectory)

These are all specified in the [File System Living Standard](https://fs.spec.whatwg.org/).

If you want to look at the user-visible file system as well, which shares the interfaces of [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle), [FileSystemFileHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle), and [FileSystemDirectoryHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle), you want to look additionally at: 

- [Window.showOpenFilePicker()](https://developer.mozilla.org/en-US/docs/Web/API/Window/showOpenFilePicker)
- [Window.showSaveFilePicker()](https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker)
- [Window.showDirectoryPicker()](https://developer.mozilla.org/en-US/docs/Web/API/Window/showDirectoryPicker)
- [DataTransferItem.getAsFileSystemHandle()](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle)

These are specified in the [File System Access Draft Community Group Report](https://wicg.github.io/file-system-access/).

foolip:
@ddbeck if you want to leave a big comment and a TODO for this, that would be fine I think. I'm not sure if the `file-and-directory-entries` feature will survive a closer look, if you're also not sure maybe turn it into a draft?

tomayac:
Correct. File and Directory Entries has nothing to do with the OPFS. It's mostly its spiritual predecessor.

ddbeck:
OK, I've prepared two big changes to this PR:

- I've made `file-and-directory-entries` a draft. That feature got less scrutiny and I'm less confident about its suitability as a feature—it might be one of these odd "infrastructure" items that doesn't do anything in particular on its own. I'd like to revisit it later.
- I've added most of the interfaces that @tomayac mentioned. Thank you! The one thing I left out was `FileSystemWritableFileStream`. This seems to be a later addition that Safari doesn't support yet and doesn't appear to be strictly required to use OPFS. I've left TODO comments until we can make concrete progress on https://github.com/web-platform-dx/web-features/issues/866.

foolip (approved):

PR #1039, opened 2024-05-06
Add a `print` feature

ddbeck:


features/print.yml
```diff
@@ -0,0 +1,8 @@
+name: Printing
+description: The `window.print()` method, along with the `beforeprint` and `afterprint` events, allow you to change the page for printing, open the browser's print dialog, and restore the page after printing.
```
foolip, line 2: Should we say that `@media print` can also be used to adapt the page for printing?

features/print.yml
```diff
@@ -0,0 +1,8 @@
+name: Printing
```
foolip, line 1: With a name this broad, how do we explain that CSS page breaking features aren't a part of it?

I think a group for printing might make more sense, and that we have one feature for `print()`, and another for the events. I assume the events also fire when the user initiates printing, so they're not tied to the API.

ddbeck: Since we're following caniuse here, we could do this (and rename the feature to something like `print-events`):

```suggestion
name: Printing events
```

I think this is defensible since the spec specifies that `beforeprint` and `afterprint` must fire for before and after `window.print()` as well as when the user requests printing.

features/print.yml
ddbeck: Honestly, I think we should _recommend_ `@media print`. 😄 We could be a little opinionated here and say something like:

```suggestion
description: An alternative to `@media print` queries, the `window.print()` method, along with the `beforeprint` and `afterprint` events, allow you to change the page for printing, open the browser's print dialog, and restore the page after printing.
```

Or stronger still:

```suggestion
description: The `window.print()` method, along with the `beforeprint` and `afterprint` events, allow you to change the page for printing, open the browser's print dialog, and restore the page after printing. Consider using `@media print` queries instead.
```

features/print.yml
foolip: `print` + `print-events` + `media-print` features sounds good to me. The last one not in this PR however, we'll need to think more about media query features in general.

features/print.yml
foolip: I think the first is strong enough. I'm sure there are cases where CSS alone just can't do the job, or where it's clearly the wrong tradeoff.

features/print.yml
```diff
@@ -1,8 +1,6 @@
-name: Printing
-description: An alternative to `@media print` queries, the `window.print()` method, along with the `beforeprint` and `afterprint` events, allow you to change the page for printing, open the browser's print dialog, and restore the page after printing.
+name: window.print()
+description: The `window.print()` method opens the browser's print dialog, to give the user the choice to make a physical copy of the page.
```
foolip, line 2: I understand you wanted to not be circular, but print dialogs are often used to save PFDs too, so "make a physical copy" seems overly specific. Minimal fix is to shorten a lot:

```suggestion
description: The `window.print()` method opens the browser's print dialog.
```

foolip (approved):
A thought on the description, but feel free to merge with or without changes.

PR #1045, opened 2024-05-06
Mark scroll snap as Baseline high

foolip:
This was a mistake in https://github.com/web-platform-dx/web-features/pull/1004.


ddbeck (approved):
This suggests a slightly different way of doing such overrides. I filed https://github.com/web-platform-dx/web-features/issues/1046 for it.

Issue #967, opened 2024-04-25
compute-baseline: incomplete import statements?


tidoust:
The NPM `compute-baseline` package does not work on my machine. Node.js complains about import statements with `ERR_MODULE_NOT_FOUND` errors. As far as I can tell, that's because the import statements in the package do not specify the filename extensions whereas [that's mandatory](https://nodejs.org/docs/latest/api/esm.html#mandatory-file-extensions). Also, there's an attempt to import a directory, which is not supported either.

ddbeck:
This is a consequence of using `tsx` to run everything here (and elsewhere I've used this code)—I never see what it looks like as a plain Node.js import. Oops!

I think the steps to fix this are:

- [ ] Add `.js` extensions to all the imports
- [ ] Change `moduleResolution` to `"NodeNext"` (maybe, need to test this)

foolip:
@tidoust can you check if v0.8.1 works for you?

tidoust:
v0.8.1 is `web-features`, not `compute-baseline`, right?

Testing compute-baseline by creating a dependency on a local folder with the latest version of the code in the web-features repository, I still have an import problem: 

> TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "[full path redacted]/@mdn/browser-compat-data/data.json" needs an import attribute of type "json"

That's triggered by `import bcd from "@mdn/browser-compat-data"` statements such as:
https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/browser.ts#L1C1-L5C35

These import statements need something like `with { type: 'json' }` (well, or `assert { type: 'json'}` depending on the runtime version).

ddbeck:
https://github.com/web-platform-dx/web-features/pull/1051

foolip:
> v0.8.1 is `web-features`, not `compute-baseline`, right?

Right... I got myself confused because I saw the hopeful fix for this in the list of changes for web-features.

tidoust:
Closing as I believe this issue was addressed a long time ago through #1027 and #1051. I confirm package now works with Node.js without problem.

PR #1051, opened 2024-05-06
`compute-baseline`: Use JSON import attribute when importing BCD

ddbeck:
I think this will fix the issue reported by @tidoust in https://github.com/web-platform-dx/web-features/issues/967#issuecomment-2096339885.

This change will require one of Node.js ~~v16.14.0+~~ (see https://github.com/web-platform-dx/web-features/pull/1051#issuecomment-2100565931), v18.20.0+, or v20.10.0+.

If you want to test this immediately, do the following:

1. Check out this PR's branch.
2. Run `npm pack --workspace compute-baseline`. A `.tgz` file for the package is built.
3. In your project, install the `.tgz` file. For example, run `npm install ../path/to/compute-baseline-0.1.0.tgz`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3415,7 +3415,7 @@
       }
     },
     "packages/compute-baseline": {
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@js-temporal/polyfill": "^0.4.4",
--- a/packages/compute-baseline/package.json
+++ b/packages/compute-baseline/package.json
@@ -1,6 +1,6 @@
 {
   "name": "compute-baseline",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "A library for computing web-features statuses from browser compatibility data",
   "exports": {
     ".": "./dist/baseline/index.js",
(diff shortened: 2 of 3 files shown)


tidoust:
That seems a good start but there are other import statements that need the same treatment at the top of other files: [`browser.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/browser.ts), [`feature.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/feature.ts), [`supportStatements.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/supportStatements.ts), [`release.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/release.ts), [`typeUtils.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/typeUtils.ts), [`walk.ts`](https://github.com/web-platform-dx/web-features/blob/main/packages/compute-baseline/src/browser-compat-data/walk.ts).

Also, I don't think that will work with v16.x (the syntax then used `assert` instead of `with`) but that's probably fine?

foolip (approved):

PR #1054, opened 2024-05-07
Map Test262 using a new field

foolip:
Identifiers must be from https://github.com/tc39/test262/blob/main/features.txt.

### DIFF
--- a/features/array-at.yml
+++ b/features/array-at.yml
@@ -5,3 +5,6 @@ snapshot: ecmascript-2022
 group:
   - arrays
   - typed-arrays
+test262:
+  - Array.prototype.at
+  - TypedArray.prototype.at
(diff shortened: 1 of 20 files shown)


foolip:
These are the features (files) I couldn't map:

array-copywithin.yml
array-fill.yml
array-find.yml
array-from.yml
array-fromasync.yml
array-isarray.yml
array-iteration-methods.yml
array-iterators.yml
array-of.yml
array-splice.yml
array.yml
stable-array-sort.yml
typed-array-iteration-methods.yml
typed-array-iterators.yml

I look at `copyWithin()` and found that indeed the tests don't name a feature:
https://github.com/tc39/test262/blob/main/test/built-ins/Array/prototype/copyWithin/name.js
https://github.com/tc39/test262/blob/main/test/built-ins/TypedArray/prototype/copyWithin/name.js

I think the possible fixes are:

- Add to https://github.com/tc39/test262/blob/main/features.txt and map within Test262
- Map using the esid in th tests
- Map using directory paths

Parking this here since any of these would be time consuming, and I can't do it now.

captainbrosset:
See https://github.com/tc39/test262/issues/4567

captainbrosset:
And https://github.com/tc39/test262/pull/4599

Issue #1038, opened 2024-05-03
Features that only make sense in one form factor should still be considered Baseline


atopal:
[Screen orientation lock](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock) is currently only available in Chrome on Android, and is therefore marked as not available in Chrome in the MDN banner, but even if it was supported in all mobile browsers, it would still not be considered Baseline [according to our definition](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md). 

I suggest that we extend the definition to consider the form factor, and in cases where the feature is expected to be used almost exclusively used in one form factor, we consider the feature Baseline. That is in cases where developers don't have an expectation for a feature to work in another form factor, we don't state that the feature is not ready yet, as that would go against one of the goals of Baseline: _Baseline status aims to identify features of developers’ contemporary, workaday web_.

foolip:
@ddbeck mentioned that he thinks this could be handled with editorial override. Trying to imagine what that would look like, there, I don't think there's actually some edit to [screen-orientation-lock.yml](https://github.com/web-platform-dx/web-features/edit/main/features/screen-orientation-lock.yml) that would have the intended effect. I've sent https://github.com/web-platform-dx/web-features/pull/1058 to show the information that needs to be conveyed, where N/A would mean "not supported and that's OK".

But do we want to do this per-browser, or divide the browsers into desktop and mobile and have something like `relevant_on_desktop: false`?

foolip:
@LeoMcA until we can figure this out, would you mind suppressing the baseline banner for screen orientation lock?

ddbeck:
When I mentioned that I thought this could be handled with an editorial override, I did mention that it would be blocked on https://github.com/web-platform-dx/web-features/issues/915. If we're going to do an override along the lines of "this doesn't make sense on desktop" then I think we also need a way to say something directly to explain to developers why they'd see both a Baseline status and partial implementation or unimplemented iconography together.

captainbrosset:
Chiming in quickly on this. I do like the N/A approach demonstrated in https://github.com/web-platform-dx/web-features/pull/1058/files, but this should probably come with a note too.

I'm not as keen on separating desktop vs. mobile. I don't think this has the same chances of enduring the passage of time.

Something like:

```yaml
name: Screen orientation lock
status:
  baseline: false
  support:
    chrome:
      value: "N/A"
      note: The feature doesn't make sense on desktop operating systems
```

[...]

PR #1058, opened 2024-05-07
Mark screen orientation lock as a mobile-only feature

foolip:


zifiri99 (approved):

zifiri99 (approved):

jgraham:
We discussed this a bit in the meeting. The feeling was that we should prefer having a mechanism to say that a feature applies to certain form factors and then a mapping from that to individual browsers that can be used in baseline computation. Since #1038 includes a list of many known mobile-only features the conclusion was that we should close this PR for now.

PR #1060, opened 2024-05-08
Publish groups and snapshots

foolip:


captainbrosset:
Discussed in today's WebDX call ([notes](https://docs.google.com/document/d/1ree75ImLZjf60lTZ3BhCaLHygxgywr7SBXp-q0xPs8A/edit#heading=h.4tsa4a527j7v)).

Philip: the bigger topic here is coming up with guidelines for groups. While groups are easy for like 50% of features, it gets harder for other features. Is grid animation part of animations, layout? Which group should structuredCloned be in?
Giving everything a group might get messy. Tags might help.

Patrick: MDN is very interested in groups, for information architecture purposes. We should revive this discussion with MDN and OWD.

foolip:
The key question blocking this for me is whether we should change the default export of the package, or just add named exports for groups and snapshots. In either case we can make the following work:

```js
import { features, groups } from 'web-features';
```

But what should the default export be?

```js
import thing from 'web-features';
```

What is `thing` here? Is it an object with `features`, `groups`, and `snapshots` as keys, or is it still the same thing as currently, the features object?

Both are easy to do, but I'm not sure what's typical for JS modules.

ddbeck:
My preference (though not strongly held) would be to not have a default export at all. If we change what we're exporting, it'll break in a more transparent way for consumers. I'd much rather get a `SyntaxError: The requested module 'web-features' does not provide an export named 'default'` instead of `undefined` from a feature lookup, if I'm upgrading from this.

This also makes it less risky for us to embellish the package in the future. For instance, if we add some `utils` functions, then tree shaking ought to place a lesser burden on consumers that don't need that code. I don't think the same would be true for a default export.

foolip:
@ddbeck I've removed the default export so we only have named exports now.

The remaining thing I need feedback on is the schema. It's still split into defs.schema.json and features.schema.json, but features.schema.json is no longer a great name. I can't quite tell why we have two files. What would need to be done to have a single self-contained schema file that we could then rename?

foolip:
I tinkered and it seems like defs.schema.json already is standalone. I've pushed a commit to remove features.schema.json.

schemas/defs.schema.json
```diff
@@ -143,6 +172,74 @@
...
+    },
+    "WebFeaturesData": {
+      "additionalProperties": false,
+      "properties": {
+        "features": {
+          "additionalProperties": {
+            "$ref": "#/definitions/FeatureData"
+          },
+          "description": "Feature identifiers and data",
+          "type": "object"
```
ddbeck, line 220: The thing you'll lose here is the `patternProperties` constraint from `features.schema.json`:

```
  "patternProperties": {
    "^[a-z0-9-]*$": {
      "$ref": "defs#/definitions/FeatureData"
    }
  }
```

That said, if I rename a file to use underscores, `ajv` seems to not catch the break with the schema. So maybe this doesn't matter, or never has? 🤷

See https://github.com/web-platform-dx/web-features/pull/44#issuecomment-1438664927 for background

foolip: I tried to find a way to express this with TypeScript but failed. It may be possible with https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html but it wouldn't be pretty, and I can't imagine it would get turned into a pretty RegEx in JSON schema.

Maybe https://github.com/web-platform-dx/web-features/pull/1317 instead?

scripts/build.ts
```diff
@@ -19,7 +19,7 @@ function buildPackage() {
   const packageDir = new URL("./packages/web-features/", rootDir);
   const filesToCopy = ["LICENSE.txt", "types.ts"];
 
-  const json = stringify(features);
+  const json = stringify(data);
```
ddbeck, line 22: Ugh, sorry, I only just now noticed this. If we're going to the trouble of having named exports, we ought to generate three JSON files, then change `packages/web-features/package.json` to have something like:

```
  "exports": {
    ".": "./index.js",
    "./features.json": "./features.json",
    "./groups.json": "./groups.json",
    "./snapshots.json": "./snapshots.json",
  },
```

(and soon `"./browsers.json": "./browsers.json"`)

foolip: I can do that, but what's the effect for consumers of the package? Does it change how you can import things?

Also, do you want a single JSON file or multiple ones for GitHub releases?

ddbeck: I think just one JSON is fine. I'm expecting we'll eventually have three release artifacts: npm package, JSON, and full-sized expanded JSON.

foolip: Great, one JSON file is what I'd like too :)

ddbeck: @foolip and I chatted about this a bit. This could be a YAGNI thing, so let's not for now.

ddbeck (approved):

captainbrosset:
Groups and snapshots are published! Yay!

captainbrosset:
You can see a very simple visualization of groups and features that have parent groups at https://web-platform-dx.github.io/web-features-explorer/groups/.

foolip:
That was quick, thank you @captainbrosset! This view immediately makes some odd things clear, like the group for Grid while Flexbox doesn't have one. We'll need guidelines for groups as well :)

PR #1061, opened 2024-05-08
Add feature for `AbortSignal.any()` static method

ddbeck:
This is a Baseline 2024 feature.


features/abortsignal-any.yml
```diff
@@ -0,0 +1,5 @@
+name: AbortSignal.any()
+description: The `AbortSignal.any()` static method combines an iterable of abort signals into a single signal, with abort reason set to the first signal to abort.
```
foolip, line 2: A signal isn't a reason, trying a tweak:

```suggestion
description: The `AbortSignal.any()` static method combines an iterable of abort signals into a single signal, with abort reason taken from the first signal to abort.
```

Any other rewording you prefer would also be OK.

foolip (approved):

PR #1064, opened 2024-05-09
Add 2D canvas convenience method features

ddbeck:
These are all a little small, but notable enough (e.g., see this [web.dev write-up](https://developer.chrome.com/blog/canvas2d)) and are all Baseline 2023 features:

- `createConicGradient()`
- `roundRect()`
- `reset()`


features/canvas-reset.yml
```diff
@@ -0,0 +1,6 @@
+name: reset()
```
foolip, line 1: I find 10 `reset()` methods in webref, so this will need to be disambiguated. "Canvas `reset()`" would be most similar to "Canvas text baselines". I guess let's also do the same for all of these features?

foolip (approved):
In addition to my name comment, would you mind landing https://github.com/web-platform-dx/web-features/pull/1081 and adding the group here too?

PR #1069, opened 2024-05-12
Add lazy-loading images and iframes feature

ddbeck:
This is a Baseline 2023 feature (barely) and matches caniuse.


foolip (approved):
Looks good but I'd prefer `loading-lazy` as the identifier. That follows the pattern of https://github.com/web-platform-dx/web-features/blob/main/features/blocking-render.yml and is closer to the caniuse identifier.

PR #1076, opened 2024-05-14
Add preloading for responsive images feature

ddbeck:


foolip (approved):
The identifier is a bit long but I don't have a suggestion to shorten it :)

PR #1082, opened 2024-05-14
Increment patch version to v0.8.2

foolip:


foolip:
I'd like to make a new release now that we have some new Baseline 2023/24 features.

ddbeck (approved):

foolip:
I'll make a release now.

PR #1083, opened 2024-05-15
Add some resource timing features

ddbeck:
This started out as a look at resource size additions to `PerformanceResourceTiming`, a Baseline 2023 feature.

But I figured there'd be a suite of resource timing features and I was right. I've tried to match caniuse on an initial support feature. It's a little bit messy because both sides are a little bit wrong about everything.


ddbeck:
https://github.com/mdn/browser-compat-data/pull/23093 attempts to fix some of Firefox's data issues, to line up with caniuse.

Meanwhile, caniuse says Chrome 25 supports `PerformanceResourceTiming` while BCD says Chrome 29, which is probably right (though it appears that even [the implementers didn't know when exactly it shipped](https://groups.google.com/a/chromium.org/g/blink-dev/c/OffuiKUc1k4/m/eDB-__0XCAAJ)).

features/resource-timing.yml
```diff
@@ -0,0 +1,27 @@
+name: Resource timing (initial support)
+description: The `serverTiming` property of the `PerformanceResourceTiming` API contains server timing information about network requests.
```
foolip, line 2: There's no `serverTiming` in the BCD keys below, but lots of other things on `PerformanceResourceTiming`. Is this description still correct?

features/resource-size.yml
```diff
@@ -0,0 +1,11 @@
+name: Resource size
+description: The `decodedSize`, `encodedSize`, and `transferSize` properties of the `PerformanceResourceTiming` API reports the size, in octets, of resources loaded.
```
foolip, line 2:
```suggestion
description: The `decodedSize`, `encodedSize`, and `transferSize` properties of the `PerformanceResourceTiming` API reports the size, in bytes, of resources loaded.
```

We can also just not say what the unit is:

```suggestion
description: The `decodedSize`, `encodedSize`, and `transferSize` properties of the `PerformanceResourceTiming` API reports the size of resources loaded.
```

features/resource-timing.yml
ddbeck: Oops, I copied from https://github.com/web-platform-dx/web-features/blob/main/features/server-timing.dist.yml to match its style, then wrote the description into the dist, and clobbered it. How about…

```suggestion
description: `PerformanceResourceTiming` entries report when network events happen while loading a resource, such as when connections start and end. You can use this information to measure loading times.
```

features/resource-size.yml
ddbeck: I cribbed the octets thing from MDN—I thought it was weirdly specific and thus significant. But if it's not, let's go with the unitless version.

features/resource-timing.yml
foolip: I've never used this feature, but that looks reasonably accurate!

foolip (approved):
Good to go with a dist update.

PR #1084, opened 2024-05-15
Remove some overrides and TODOs that have been fixed through BCD

foolip:
This does not change any browser versions.


ddbeck (approved):
I love fewer overrides and TODOs. 🎉

PR #1086, opened 2024-05-15
Add <datalist>

foolip:


foolip:
I'm adding this because it's 2% of the answers in https://2023.stateofhtml.com/en-US/usage/#html_interoperability_features. It looks like developers aren't having a good time with `<datalist>`, and it would be good to dig in more.

features/datalist.yml
```diff
@@ -0,0 +1,24 @@
+name: <datalist>
+description: The `<datalist>` element TODO
```
ddbeck, line 2: This isn't great, but maybe a gesture in the right direction?

```suggestion
description: The `<datalist>` element with `<options>` child elements defines a set of recommended values for an `<input>` element, connected by a `list` attribute set to the ID of the `<datalist>`.
```

features/datalist.yml
```diff
@@ -0,0 +1,24 @@
+name: <datalist>
+description: The `<datalist>` element with `<options>` child elements defines a set of recommended values for an `<input>` element, connected by a `list` attribute set to the ID of the `<datalist>`.
```
foolip, line 2: @ddbeck how about this simplification and elaboration:

```suggestion
description: The `<datalist>` element defines a set of recommended values for an `<input>` element. Browsers may show a dropdown menu of all options, or matching options as the user types.
```

Issue #1088, opened 2024-05-15
Split custom elements


foolip:
The root cause of https://github.com/GoogleChrome/webstatus.dev/issues/285 is that we have a single feature that corresponds to https://caniuse.com/custom-elementsv1, but we don't have a concept of partial support, so instead Safari is treated as not supported custom elements at all. (Firefox not being supported is due `:host-context` I think, which we should also fix.)

I think that most likely we should have a custom elements group with at least two features, called autonomous custom elements and customized built-in elements. Possibly also form-associated custom elements.

PR #1089, opened 2024-05-15
Split custom elements into two features

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/1088.


features/autonomous-custom-elements.yml
```diff
@@ -0,0 +1,11 @@
+name: Autonomous custom elements
+description: Autonomous custom elements are HTML elements with behavior that you define.
```
foolip, line 2: This description should say something about how to create these elements, tag names with a dash in them.

features/customized-built-in-elements.dist.yml
```diff
@@ -0,0 +1,20 @@
+# Generated from: customized-built-in-elements.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+name: Customized built-in elements
+description: Customized built-in elements are HTML elements with behavior that you define.
```
foolip, line 5: This description should mention the `is` attribute.

features/customized-built-in-elements.yml
```diff
@@ -0,0 +1,9 @@
+name: Customized built-in elements
+description: Customized built-in elements are HTML elements with behavior that you define.
```
foolip, line 2: This description should mention the `is` attribute.

features/autonomous-custom-elements.yml
ddbeck: How about….

```suggestion
description: Autonomous custom elements are HTML elements with a hyphenated tag name (like `<example-element>`) that have a behavior you define.
```

features/customized-built-in-elements.yml
ddbeck: How about …

```suggestion
description: Customized built-in elements are HTML elements that extend built-in elements using the `is` attribute, to add new behaviors that you define.
```

ddbeck (commented):
What do you think of a `custom-elements` group to put these together?

features/autonomous-custom-elements.yml
foolip: Look good to me! An example tag name that's shorter and also realistic might be nice, but I struggle to find something that's obviously an example and not defining the scope of the feature. `<login-button>` and `<video-player>` aren't as good as `<example-element>`, so let's go with that.

features/customized-built-in-elements.yml
foolip: Thank you!

foolip:
> What do you think of a `custom-elements` group to put these together?

I did consider that, but I hesitated because maybe we'll want a composite feature later to match the caniuse feature. But let me add the group and deal with that problem if it appears later. Some of the CSS pseudo-classes that aren't part of either feature now should probably go in the same group.

ddbeck (approved):
Thanks, Philip!

(For what it's worth, I also tried to find a more concrete example tag name but also found that it narrowed the apparent utility of the feature.)

foolip:
The WPT side of this is done in https://github.com/web-platform-tests/wpt/pull/46309 and was quite hairy.

PR #1090, opened 2024-05-15
Add Permissions API feature

ddbeck:
This is a Baseline 2022 feature. It matches caniuse.

I created this in the process of investigating `api.PermissionStatus.change_event`, which would be Baseline 2023 by itself. Ultimately, I found that it wasn't a feature in its own right (I could find little evidence that developers recognized this a distinct thing), but the Permissions API was.


features/permissions.yml
```diff
@@ -0,0 +1,12 @@
+name: Permissions
+description: The `navigator.permissions` API checks whether a permission, such as access to geolocation data, has been granted.
+spec: https://w3c.github.io/permissions/
+caniuse: permissions-api
+compat_features:
+  - api.Navigator.permissions
+  - api.PermissionStatus
+  # - api.PermissionStatus.change_event # TODO: later addition
+  # - api.PermissionStatus.name  # TODO: later addition
+  - api.PermissionStatus.state
```
foolip, line 10: https://caniuse.com/permissions-api says Chrome 43, but we say Chrome 44 because of this, which was called `status` in Chrome 43. I think it's reasonable to bump the version to Chrome 44 because of this, because AFAICT you can't query a permission without accessing this property.

Can you add a note explaining the difference between caniuse and BCD?

foolip (approved):

features/permissions.yml
ddbeck: Done with c3a7916.

PR #1096, opened 2024-05-16
Add content-visibility.dist.yml

foolip:
This also adds support in Firefox 124, hidden by the status override.


ddbeck (approved):
I'm pretty sure this one will fail when it lands due to me taking too long to review this—it's actually Firefox 125. See https://github.com/mdn/browser-compat-data/pull/23116 and https://github.com/mdn/browser-compat-data/blob/v5.5.31/css/properties/content-visibility.json.

I'll approve this, but would you rebase, refresh, then merge it?

foolip:
@ddbeck sorted!

PR #1110, opened 2024-05-16
Put more features into the fonts group

foolip:


ddbeck (approved):
This is good, but can you refresh the dists and merge?

(As is, it'll fail the build on merge because `font-face` has a dist now.)

PR #1114, opened 2024-05-17
Generate status for CPU compute pressure

foolip:
Support on Chrome desktop only is correct for now. Edge 125 has not been
released yet, and the feature doesn't work on Android.


foolip:
@captainbrosset this doesn't fix the Edge data, but it does mean that the next release of BCD will.

captainbrosset:
Edge 125 is marked as "current" in BCD now and has been for a few days. So, yeah, I guess we just need to wait for the next BCD release. I thought BCD was released twice a week though. Last time was last week. Do you know if that's expected?

captainbrosset (approved):

Elchi3 (approved):

foolip:
@captainbrosset this is fixed in https://github.com/web-platform-dx/web-features/pull/1148.

PR #1118, opened 2024-05-20
Add CSS will-change

foolip:
This is a Baseline 2020 feature.


features/will-change.yml
```diff
@@ -0,0 +1,11 @@
+name: will-change
+description: "The `will-change` CSS property gives hints to the browser about expected changes to an element's scroll position, contents, or style. When used correctly, these hints can improve performance."
```
ddbeck, line 2: The second sentence here worries me a little. Feels a bit magical. What do you think of something like this:

```suggestion
description: "The `will-change` CSS property gives hints to the browser about expected changes to an element's scroll position, contents, or style. When used sparingly, these hints allow browsers to optimize for upcoming style changes."
```

foolip: I understand why you suggest "sparingly" instead of "correctly" but it doesn't seem quite right still. We recommend that it's used sparingly because it's not easy to get right and using it in the wrong way might waste resources instead of improve performance. But "sparingly" is more guidance on how to use it than a description of how the feature works.

Maybe we just drop the prelude and say "These hints allow browsers to optimize for upcoming style changes"?

ddbeck: Yeah, that works for me. "Use this well" is out of scope for "clearly identify the thing we're talking about."

```suggestion
description: "The `will-change` CSS property gives hints to the browser about expected changes to an element's scroll position, contents, or style. These hints allow browsers to optimize for upcoming style changes."
```

ddbeck (approved):

PR #1121, opened 2024-05-20
Add WebVTT cue alignment feature

foolip:


features/webvtt-cue-alignment.yml
```diff
@@ -0,0 +1,8 @@
+name: WebVTT cue alignment
+description: The WebVTT cue alignment settings control which part of the cue is aligned with the given line and position. Cue alignment is set using `line` and `position` settings in WebVTT files or the `lineAlign` and `positionAlign` properties of `VTTCue` using JavaScript.
```
ddbeck, line 2: Optional suggestion here: What do you think of something like this? I'm trying to avoid (my initial) confusion that this affects overall placement of the cue/region with respect to the video viewport.

```suggestion
description: WebVTT cue alignment settings control how text is inset within a cue box, relative to the writing direction of the cue. Cue alignment is set using `line` and `position` settings in WebVTT files or the `lineAlign` and `positionAlign` properties of `VTTCue` using JavaScript.
```

ddbeck (approved):

features/webvtt-cue-alignment.yml
foolip: Maybe I have misunderstood how this all is supposed to work.

My understanding was that `line` and `position` (the percentages before the comma in the file format) define a point in the video viewport, and by default it's the top of the cue box and a horizontal coordinate that depends on `align` that is lined up with that point.

And then `lineAlign` and `positionAlign` (they keywords after the comma in the file format) change which point of the cue box is lined up with that point in the video viewport.

ddbeck: I think it's genuinely possible I've misunderstood it, but I can't be sure. But I think it's not worth it to spend a time understanding this to write a slightly better description. But if it's _wrong_ then I suspect that it won't be long before someone who cares about and understands WebVTT will run across this and demand a correction.

Which is to say, let's merge it as is.

ddbeck (approved):

PR #1125, opened 2024-05-22
Add hidden="until-found"

foolip:


features/hidden-until-found.yml
```diff
@@ -0,0 +1,7 @@
+name: hidden="until-found"
+description: 'The `hidden="until-found"` attribute TODO.'
```
ddbeck, line 2: I think this is how it's supposed to work?

```suggestion
description: 'The `hidden="until-found"` attribute hides an element until it is found using the browser's find-in-page search or it is directly navigated to by following a URL fragment.
```

ddbeck (approved):

features/hidden-until-found.yml
foolip: Applied, I think this is accurate!

PR #1127, opened 2024-05-22
Add :dir()

foolip:

The versions match https://caniuse.com/css-dir-pseudo.

The identifier is "dir-pseudo" instead of just "dir" because there's
also an HTML dir attribute that we may want to represent later.


features/dir-pseudo.yml
```diff
@@ -0,0 +1,8 @@
+name: :dir()
+description: "The `:dir()` CSS functional pseudo-class TODO."
```
ddbeck, line 2:
```suggestion
description: "The `:dir()` CSS functional pseudo-class matches elements by text direction, either right to left (`rtl`) or left to right (`ltr`)."
```

ddbeck (approved):

PR #1131, opened 2024-05-22
Add display: contents

foolip:


features/display-contents.dist.yml
```diff
@@ -0,0 +1,21 @@
...
+name: "display: contents"
+description: "The `display: contents` CSS declaration TODO."
+spec: https://drafts.csswg.org/css-display/#valdef-display-contents
+caniuse: css-display-contents
+status:
+  baseline: high
+  baseline_low_date: 2020-01-15
+  baseline_high_date: 2022-07-15
+  support:
+    chrome: "65"
```
foolip, line 13: https://caniuse.com/css-display-contents treats all browsers (except Safari iOS?) as having partial support. Maybe we can't even add this feature without notes.

captainbrosset: I think we absolutely need a more about accessibility issues of the feature.

PR #1144, opened 2024-05-27
Increment patch version to v0.8.4

foolip:


foolip:
This is to get https://github.com/web-platform-dx/web-features/pull/1137 released ASAP, since it's effectively a feature removal and I need to use the identifier in WPT.

ddbeck (approved):

Issue #963, opened 2024-04-25
Remove or hide svg2-script-html-equivalence


foolip:
This feature is unusual, in that it's a collection of things that work in HTML, now supported in SVG, but only in Firefox. Do consumers have any interest in a feature like this, or should we remove it?

History of the feature so far:
https://github.com/web-platform-dx/web-features/pull/464
https://github.com/web-platform-dx/web-features/pull/677

If the feature is useful now, should it be archived or hidden at some point in the future?

PR #1145, opened 2024-05-28
Drop SVG equivalence feature

ddbeck:
To quote myself in
https://github.com/web-platform-dx/web-features/pull/677#issuecomment-2002046126:

> [This] is _not_ a feature. It's _why_ async and defer should work in SVGs, but I have a hard time seeing it as a thing one could _use_. As if you'd want the option to insert a script tag into either HTML or SVG dynamically or something? It's too weird.

Fixes https://github.com/web-platform-dx/web-features/issues/963


ddbeck:
I've opened a corresponding PR on BCD https://github.com/mdn/browser-compat-data/pull/23188

foolip (approved):
Yay!

Issue #1018, opened 2024-04-30
Allow ranged dates and versions for non-threshold support information


ddbeck:
This issue is blocked on adopting an earliest-possible-date indicator, as in #623.

@foolip, @rumyra, and I had a chat earlier that veered into this topic. I am trying to capture that conversation, but I invite them to correct me on any misstatements here.

## Proposal

Allow an "at least this early" range indicator (e.g., `≤`) on `status` block version numbers and dates, if those dates and versions are Baseline `high`.

For dates, (i.e., `baseline_low_date` and `baseline_high_date`) a range indicator would be taken to mean "this date but potentially earlier." That is, new or wider availability may have been more historical, but not newer.

For version numbers, a range indicator would be taken to mean "supported this version and later." That is, support in preceding versions would be _unknown_, not unsupported.

## Example

Ranges for version numbers would allow web-features to communicate the distinction between caniuse's red, green, and gray boxes. Take this, for example:

<img width="312" alt="Screenshot 2024-04-30 at 18 37 30" src="https://github.com/web-platform-dx/web-features/assets/64103/bdd7eae0-76e5-4a13-b072-412862d2a770">

The corresponding data would look like this:

```yaml
support:
  chrome: "45"
  edge: "≤15"
  safari: "7.1"
  firefox: "25"
```

Presently, our schema would only be able to show *unsupported* for Edge versions less than 15.

## Motivation

This has a few key benefits:

- Shortens the time and effort to set statuses for older features, even when we're certain they're Baseline high
- Provides a transparent way to handle version ranges from BCD (right now we don't show when we source version ranges from BCD)
- Makes it easy to find and prioritize features for later refinement

Though it comes at increased integration burden to consumers (though it's likely most could strip out the indicator).

@foolip and I are spending quite a lot of time picking apart features and cleaning up data in BCD, so that each version numbers reported by web-features is accurate. So far, we've been pretty meticulous, so that a consumer could take version numbers for a feature's `support` block and report that the feature was supported from those versions continuously to the present _and_ that preceding versions were unsupporting. That's ideal, especially for recent features where it's important for developers to have accuracy through the whole life of the feature.

For older features, however, there's less value in knowing precisely which version marked the beginning of support for a feature while the cost of finding that precise version increases substantially. Whether a feature originally shipped in one of the core browsers in 2017 or 2016 is probably not a make-or-break event for using that feature, as long as we're reasonably certain it's no _later_ than a given date or version number.

[...]

PR #1153, opened 2024-05-29
Add draft features for HTML elements

foolip:


foolip:
This would really benefit from https://github.com/web-platform-dx/web-features/issues/1018, so that we don't need to carefully review exactly when each of these were added.

features/area.yml
```diff
@@ -0,0 +1,54 @@
+draft: true
+name: <area>
```
foolip, line 2: This features should be Image maps with https://html.spec.whatwg.org/multipage/image-maps.html as the spec link.

features/draft/html-elements/video.yml
```diff
@@ -0,0 +1,31 @@
+draft_date: 2024-06-05
```
ddbeck, line 1: Hmm, why doesn't this one have a dist?

foolip:
@ddbeck this is ready for review now. I could keep tinkering with it indefinitely, but it's good enough now.

ddbeck (approved):

PR #1155, opened 2024-05-29
Add XHR

foolip:


foolip:
This depends on https://github.com/web-platform-dx/web-features/issues/1018 to not have to research the correct versions, or trim the feature down to get the earliest possible version for each browser.

foolip:
I don't think we need https://github.com/web-platform-dx/web-features/issues/1018, the dist grouping sufficed for this.

features/xhr.yml
```diff
@@ -0,0 +1,56 @@
+name: XMLHttpRequest
+description: The `XMLHttpRequest` API TODO. Also known as XHR.
```
ddbeck, line 2: I've restrained myself from including something very opinionated here, like "Use `fetch()` instead."

```suggestion
description: The `XMLHttpRequest` API makes HTTP requests. It's the predecessor to `fetch()`. Also known as XHR.
```

ddbeck (approved):

PR #1161, opened 2024-05-30
Add a type to resolve a TypeScript red squiggly

foolip:


ddbeck (approved):

scripts/dist.ts
```diff
@@ -246,7 +246,7 @@ function warnOnNeedlessOverrides(id, overridden, generated) {
 
 function main() {
   // Map from .yml to .dist.yml to filter out duplicates.
-  const sourceToDist = new Map(
+  const sourceToDist: Map<string, string> = new Map(
```
ddbeck, line 249: If you prefer a little more brevity, this also works:

```suggestion
  const sourceToDist = new Map<string, string>(
```

PR #1162, opened 2024-05-30
Reduce dist files to only the additional generated data

foolip:
The duplication of data across source and dist means that any changes to
the source require regenerating dist. This frequently slows down work as
this needs to be done in a checkout. Most commonly, an updated
description from a code review suggestion results in failing checks
because the dist file is out of sync.

The new approach is that a dist file will only contain `compat_features`
when they are sourced from BCD, and only contain `status` when it is
computed as opposed to manually maintained.

The built packages/web-features/index.json is unchanged.


foolip:
> The built packages/web-features/index.json is unchanged.

I've confirmed that this is true, but there is something weird happening with 3 dist files I don't understand:

- anchor-positioning.dist.yml
- compute-pressure.dist.yml
- customized-built-in-elements.dist.yml

The changes in these files look right, and without them packages/web-features/index.json _would_ change, but I don't understand why the changes happen.

foolip:
> A small sacrifice is that Baseline high dates are not checked in, and instead only computed in the build step.

This is what I thought would happen, but for whatever reason it's not like that. I haven't checked why, it's fine if we still have the Baseline high dates.

ddbeck:
@foolip I pushed a couple of changes to show what it would look like to rename the files so that the dists sort after the YAML files. If/when you rebase, this is the procedure I used to rename the dists:

```fish
# untrack old dist files without deleting them
$ git rm --cached features/*.dist.yml

# rename dist files (`brew install rename` if you need this or do something clever with your shell of choice)
$ rename --subst .dist.yml .yml.dist features/*.dist.yml

# track dist files
$ git add features/*.yml.dist
```

foolip:
@ddbeck I have rebased now!

ddbeck (approved):
I'm happy with this, provided you fix the prettier error. Feel free to merge immediately (and we can clean up the mess in other PRs). Or we could schedule a time to merge this (on Friday, maybe?) and I can do my best to reduce the number of open PRs between now and then.

foolip:
This is a very conflict-y PR so I'll get this merged ASAP and just rebase everything else as needed.

foolip:
> The built packages/web-features/index.json is unchanged.

I reconfirmed this is the case after landing by comparing the built index.json from commit f1ebf697e6d025027e6e45826d201ac74b71f700 and c25032062c3211d7ffa6dcbc86bfb631eb648ba3.

Issue #1164, opened 2024-06-03
Baseline "limited availability" for ESM?


lionel-rowe:
Originally reported [in mdn/browser-compat-data](https://github.com/mdn/browser-compat-data/issues/23231).

> Articles for [Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) both claim "limited availability", but the article for [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) has "widely available".
>
> I guess it's something to do with support for typed imports using `with` and the legacy `assert`, but I don't think those are what people would understand by "support for modules" or "support for `import` statements".

I had a look in [js-modules.yml](https://github.com/web-platform-dx/web-features/blob/d3d6f97b7d32b289cbfb87134251e12fb963a54e/features/js-modules.yml#L12-L20) in the current repo, but all the features cited there are widely supported — even `nomodule` looks to be supported in Safari 11 (Sep 2017), [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#browser_compatibility).

foolip:
I think the problem is that `javascript.statements.import` is part of both CSS modules and JS modules. When yari finds a web feature by compat key, it probably finds CSS modules first because it comes first alphabetically. And CSS modules is only supported in Chromium browsers.

This fix should be to remove javascript.statements.import from CSS modules in web-features.

captainbrosset:
> When yari finds a web feature by compat key, it probably finds CSS modules first because it comes first alphabetically.

I didn't know how yari mapped MDN pages to web-features, but this way sounds prone to bugs.

> This fix should be to remove javascript.statements.import from CSS modules in web-features.

Ensuring that we use BCD keys just once in the repo feels like an arbitrary limitation that will be a problem for us in the future.
I'm not apposed to doing this as a quick fix if this unblocks MDN. But would this work long term:

MDN optionally associates articles with web-features (probably by adding a `webFeaturesId: js-modules` entry in the front matter). For articles that have this association, that's how the right web-features dist file is pulled. For others, they continue doing what they do now.
As a bonus: we get MDN links for free, for some of our features.

PR #1174, opened 2024-06-03
Ensure that BCD keys are only used once

foolip:
This prevents a problem like in https://github.com/web-platform-dx/web-features/issues/1164
from happening before we have a way to handle it.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -110,6 +110,9 @@ function convertMarkdown(markdown: string) {
     return { text, html };
 }
 
+// Map from BCD keys/paths to web-features identifiers.
+const bcdToFeatureId: Map<string, string> = new Map();
+
 const features: { [key: string]: FeatureData } = {};
 for (const [key, data] of yamlEntries('features')) {
     // Draft features reserve an identifier but aren't complete yet. Skip them.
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -155,6 +158,19 @@ for (const [key, data] of yamlEntries('features')) {
         }
     }
 
+    // Check that no BCD key is used twice until the meaning is made clear in
+    // https://github.com/web-platform-dx/web-features/issues/1173.
+    if (data.compat_features) {
+        for (const bcdKey of data.compat_features) {
+            const otherKey = bcdToFeatureId.get(bcdKey);
+            if (otherKey) {
+                throw new Error(`BCD key ${bcdKey} is used in both ${otherKey} and ${key}, which creates ambiguity for some consumers. Please see https://github.com/web-platform-dx/web-features/issues/1173 and help us find a good solution to allow this.`);
```
ddbeck, line 167: I really like this message. 👍

ddbeck (approved):

index.ts
foolip: I'm glad you like it :) It's a bit weird to throw and error and still say that we'd like to allow it...

PR #1176, opened 2024-06-03
New ClipboardItem.supports feature

captainbrosset:
Starting with Edge/Chrome 121, the `ClipboardItem` API has a new capability to detect if a given mime-type is supported by the browser when writing to the clipboard. This allows web devs to simplify their code by avoiding attempting to write first and catching the error.


foolip:
@captainbrosset can you add this to BCD? They BCD path would be `api.Clipboard.read.formats_unsanitized_parameter`. Search for "options_latencyHint_parameter" for an entry you can copy-paste as a start.

We could still land this without waiting for the BCD change to be merged or released.

features/clipboard-supports.yml
```diff
@@ -0,0 +1,6 @@
+name: ClipboardItem.supports()
+description: The `ClipboardItem.supports()` static method checks if the browser supports copying a mime-type to the system clipboard.
```
foolip, line 2: The spec doesn't say anything about what this static method does, only how to implement it. Is it really scoped to reading from the clipboard, or can it also be used for check for write support of a type?

Regarding "mime-type" I see that there's the "web "-prefixed custom formats which aren't mime types, so how about wording similar to this?

```suggestion
description: The `ClipboardItem.supports()` static method checks if the browser supports copying data types such as "text/html" or "image/png" to the system clipboard.
```

Or is there a better example? I guess the point of this method is really [optional formats](https://w3c.github.io/clipboard-apis/#optional-data-types) but which of them is supported by some browser but not all?

ddbeck: I like the use of examples here and "media type" is probably the right term, though "data type" is OK if we want to avoid confusing with audio-visual formats. But MIME is old terminology (e.g., [IANA says](https://www.iana.org/assignments/media-types/media-types.xhtml) "formerly known as MIME types") and we should probably avoid it, or restrict it to some kind of "also known as" text.

captainbrosset: Reading the recent [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/supports_static) about this (which @tayloregivens wrote, btw), I think I need to chang the description.

This is about _writing_ to the clipboard (so "copying" is correct, but changing to "writing" would, I think, make more sense). 
Also, mandatory data types are always supported, those are: text/plain, text/html, image/png.
Optional data types are optionally supported by the UA. These include custom formats that start with the "web " prefix.

ddbeck (approved):
This sounds good to me.

The dist situation has changed since the PR started, so to fix it and make this branch mergeable:

- Delete the `.dist.yml` files modified on this branch (`async-clipboard.dist.yml` and `clipboard-supports.dist.yml`).
- Merge `main` into this branch
- Regenerate the dist files for the clipboard features

PR #1178, opened 2024-06-03
Reading unsanitized data from the async clipboard API

captainbrosset:
Since Edge/Chrome 122, devs can optionally pass formats when using the `Clipboard.read()` async API to get the non browser-sanitized version of the data. This is useful for apps that deal with HTML for example.


captainbrosset:
This one has no corresponding keys in BCD, so it's all manual.

foolip:
@captainbrosset can you add this to BCD? They BCD path would be `api.Clipboard.read.formats_unsanitized_parameter`. Search for "options_latencyHint_parameter" for an entry you can copy-paste as a start.

We could still land this without waiting for the BCD change to be merged or released.

features/clipboard-unsanitized-formats.yml
```diff
@@ -0,0 +1,10 @@
+name: Paste unsanitized data
+description: The optional `formats` parameter of the `Clipboard.read()` method pastes unsanitized data from the system clipboard.
```
foolip, line 2: `Clipboard.read()` isn't a static method, but we could do this in this case:

```suggestion
description: The optional `formats` parameter of the `navigator.clipboard.read()` method reads unsanitized data from the system clipboard.
```

(Also "reads" instead of "pastes".)

@ddbeck what style do you think we should use for a parameter of a method of an interface? It's not the parameter that does the reading, it's still the method, but how to phrase that?

foolip: Hmm, wait, am I wrong about paste vs. read? From the user point of view you copy and paste. As the user of the clipboard API you read from or write to the clipboard. Write to copy. Read to paste. Right?

I have no view on what terminology is easier to understand.

captainbrosset: Correct: paste == read.

captainbrosset:
> @captainbrosset can you add this to BCD? They BCD path would be `api.Clipboard.read.formats_unsanitized_parameter`. Search for "options_latencyHint_parameter" for an entry you can copy-paste as a start.

Adding it in https://github.com/mdn/browser-compat-data/pull/23236

Let's wait for this to ship and then I'll add the compat key here.

foolip:
> Let's wait for this to ship and then I'll add the compat key here.

That's OK if you like, but if you want to move faster I'm OK with a TODO here as well.

captainbrosset:
I believe the new BCD key that this PR depends on is in BCD 5.5.32, which is about to be added to the repo.

features/clipboard-unsanitized-formats.yml
```diff
@@ -0,0 +1,6 @@
+name: Paste unsanitized data
```
ddbeck, line 1: I worry a little about creating confusion with _paste_ (which I think of as whatever UI action that leads up to the `paste` event being fired). What do you think of one of these alternatives? (I don't have a strong feeling.)

```suggestion
name: Unsanitized clipboard formats
```

```suggestion
name: Unsanitized clipboard data
```

```suggestion
name: Reading unsanitized clipboard data
```

```suggestion
name: Unsanitized clipboard reads
```

ddbeck (approved):
Excellent. Thank you Patrick!

PR #1179, opened 2024-06-03
New EyeDropper feature

captainbrosset:


features/eyedropper.yml
```diff
@@ -0,0 +1,5 @@
+name: EyeDropper
+description: The EyeDropper API creates an eyedropper tool that allows users to select a color from their screen.
+spec: https://wicg.github.io/eyedropper-api/#eyedropper-interface
+compat_features:
+  - api.EyeDropper
```
foolip, line 5: Can you list all 4 BCD keys that are listed by `npm run traverse | grep EyeDropper` in BCD? (Something like this is usually the starting point for me.)

features/eyedropper.yml
```diff
@@ -0,0 +1,5 @@
+name: EyeDropper
+description: The EyeDropper API creates an eyedropper tool that allows users to select a color from their screen.
+spec: https://wicg.github.io/eyedropper-api/#eyedropper-interface
```
foolip, line 3:
```suggestion
spec: https://wicg.github.io/eyedropper-api/
```

features/eyedropper.yml
```diff
@@ -0,0 +1,5 @@
+name: EyeDropper
+description: The EyeDropper API creates an eyedropper tool that allows users to select a color from their screen.
```
foolip, line 2:
```suggestion
description: The `EyeDropper` API opens an eyedropper tool that allows users to select a color from their screen.
```

foolip (commented):
I'm not sure if I should suggest `eye-dropped` instead of `eyedropper`. The spec strangely says `EyeDropper` instead of `Eyedropper`. But https://helpx.adobe.com/photoshop/using/tool-techniques/eyedropper-tool.html suggests that web developers might be more familiar with "Eyedropper" than "Eye dropper".

features/eyedropper.yml
captainbrosset: Is `api.EyeDropper.secure_context_required` really needed? I thought this kind of usage requirement was handled at a different level in BCD, not as a compat key.

foolip: Oh, that changed the versions from 95 to 96. That doesn't seem right to me, the feature not working over HTTP isn't essential in any way. So let's exclude the key, after all.

ddbeck (approved):
Thank you @captainbrosset! The test failure here is due to https://github.com/web-platform-dx/web-features/pull/1182, so this can now merge failing tests notwithstanding. Thanks again!

PR #1180, opened 2024-06-03
Move draft features into a special directory

foolip:
This is to prepare for a large number of generated draft features,
which we'll put in features/draft/generated/.


index.ts
```diff
@@ -112,11 +112,6 @@ function convertMarkdown(markdown: string) {
 
 const features: { [key: string]: FeatureData } = {};
 for (const [key, data] of yamlEntries('features')) {
-    // Draft features reserve an identifier but aren't complete yet. Skip them.
-    if (data.draft) {
-        continue;
-    }
-
```
ddbeck, line 119: I think the crawler is recursive by default, so we still need a `draft` key or exclude the draft folder from the search.

index.ts
```diff
@@ -20,6 +20,10 @@ const nameMaxLength = 80;
 // The longest description allowed, to avoid them growing into documentation.
 const descriptionMaxLength = 300;
 
+// Internal symbol to mark draft entries, so that using the draft field outside
+// of a draft directory doesn't work.
+const draft = Symbol('draft');
```
ddbeck, line 25: Oh neat, one of those JS features I never think to use.

ddbeck (approved):
Thank you!

PR #1181, opened 2024-06-03
New CSS Painting API feature

captainbrosset:


features/paint.yml
```diff
@@ -0,0 +1,9 @@
+name: paint()
+description: The `paint()` CSS function paints an element's background or border using a custom paint worklet.
```
ddbeck, line 2: I didn't really know what "paint" meant here, so after reading the docs…

```suggestion
description: The `paint()` CSS function creates a custom image, drawn using a paint worklet, as in an element's `background-image` or `border-image` property.
```

I'm not sure that's a good description, but I think it adds a little more info. What do you think?

features/paint.yml
```diff
@@ -0,0 +1,9 @@
...
+spec: https://drafts.css-houdini.org/css-paint-api-1/
+caniuse: css-paint-api
+compat_features:
+  - css.types.image.paint
+  - api.CSS.paintWorklet_static
+  - api.PaintWorkletGlobalScope
+group: css
```
ddbeck, line 9: a real nit here (and I should write a script to format these automatically) but, if you're editing this file anyway, then moving this after `spec` will make this more consistent with other entries.

features/paint.yml
captainbrosset: Thanks, feels better. I'm a bit confused by the "as in" part at the end. I'll think of it and push a new commit.

foolip (approved):

foolip:
Thank you @captainbrosset!

PR #1182, opened 2024-06-03
Refresh `selectlist` dist

ddbeck:
I broke the build when I merged https://github.com/web-platform-dx/web-features/pull/1111. This unbreaks it.


foolip:
I snuck this into #1110.

PR #1191, opened 2024-06-04
Link caniuse feature for grid animation

foolip:


foolip:
Tests are failing because this is hidden on caniuse.com. Not sure why, but OK.

PR #1195, opened 2024-06-05
Add a script to generate a draft feature per spec

foolip:
One of the features, the Vibration API, is added to show what this will look like.


foolip:
https://github.com/web-platform-dx/web-features/pull/1215 shows what the rest of the drafts would look like.

foolip:
Scrolling through https://github.com/web-platform-dx/web-features/pull/1215 I think a clear improvement would be to list which existing features include some keys defined in the spec, to make it very easy to figure out where some keys should be added to an existing feature. Over time most of these files will just contain bits and bobs that should be part of other existing features.

foolip:
There are two things I want to do here:

- Group into directories by standards org / working group
- Generate a comment about which features already include keys from the feature

foolip:
I just discussed this with @tropicadri who is interested in the draft features generation. I'd now like to land this PR as-is and instead treat the following as follow-up work in separate PRs:

- Generate a comment about which features already include keys from the feature. This makes it much easier to spot when most keys for a spec are already associated with a feature.
- A GitHub Actions workflow to update spec drafts weekly. It would send a PR that a human would review and merge.

Regarding grouping into directories by standards org / working group, I took a look at that and think it will be too messy to do  in a general fashion. Working groups don't have nice names that can be used as directory names, and many specs belong to multiple working groups, and some no working group or defunct working groups. All specs in a single directory is OK I think.

foolip:
@ddbeck can you review this?

scripts/update-drafts.ts
```diff
@@ -0,0 +1,103 @@
...
+  const specToCompatFeatures = new Map<object, Set<string>>();
+  for (const feature of compat.walk()) {
+    // Skip any BCD keys already used in web-features.
+    if (usedFeatures.has(feature.id)) {
+      continue;
+    }
+
+    // Skip deprecated and non-standard features.
+    const status = feature.data.__compat.status;
+    if (status?.deprecated || !status?.standard_track) {
```
ddbeck, line 60: I _think_ `standard_track: false` implies that `spec_url` is `undefined`, so that part probably doesn't do anything but also can't hurt.

ddbeck (approved):
Looks good!

autonome:
>  I'd now like to land this PR as-is and instead treat the following as follow-up work in separate PRs:
> 
>     Generate a comment about which features already include keys from the feature. This makes it much easier to spot when most keys for a spec are already associated with a feature.
>     A GitHub Actions workflow to update spec drafts weekly. It would send a PR that a human would review and merge.

@foolip are there issues filed for these? The generated comment bit would certainly speed the workflow up.

foolip:
@autonome, yes, the generated comments were added by @tropicadri in https://github.com/web-platform-dx/web-features/pull/1303.

The GitHub Actions workflow hasn't been done yet. @tropicadri are you still interested in working on that?

PR #1199, opened 2024-06-06
Adding a WebNN feature

captainbrosset:
This one is experimentally available in Edge Canary/Beta, but has no docs, caniuse, bcd entries, or stable support anywhere.
Given all the noise about it recently, it makes sense to track it.


captainbrosset:
I guess I was wrong, there are quite a few BCD entries already, which already have the web-features:webnn tag. So that's a start.

features/webnn.yml
```diff
@@ -0,0 +1,3 @@
+name: WebNN
+description: The WebNN API constructs and executes computational graphs of neural networks by making use of the various machine learning capabilities and hardware accelerators available on the device. Also known as Web Neural Network API.
+spec: https://www.w3.org/TR/webnn/
```
foolip, line 3: Can you add this to the allowed spec lists? Background in https://github.com/mdn/browser-compat-data/pull/22569#issuecomment-1992632118

In https://github.com/w3c/browser-specs/issues/1268 I made an attempt at automating this, but it's not tractable yet.

foolip (approved):

foolip:
Thanks @captainbrosset!

Issue #1173, opened 2024-06-03
Allow overlapping BCD keys without creating ambiguity


foolip:
https://github.com/web-platform-dx/web-features/issues/1164 was caused by `javascript.statements.import` being part of both CSS modules and JS modules. In this situation, looking up a feature by BCD key finds two features, so consumers need something more to determine the status of `javascript.statements.import`.

The `<source>` element is a good illustration of the problem. The `<audio>` and `<video>` elements aren't yet in web-features, but if we add them, it would make sense for the `<source>` element  (and `HTMLMediaElement` interface) to be part of _both_ features, which shipped at the same time. But the `<source>` element is also a part of the `<picture>` element which shipped much later. Rather than picking a feature which gets to claim each BCD key, it would be good if overlap was handled gracefully.

This also has implications for https://github.com/web-platform-dx/web-features/issues/979.

foolip:
It looks like `javascript.statements.import` is the only key that did belong to two features.

foolip:
Rather than having rules that consumers need to handle for this rare situation, it would be good if we can bake the behavior we want into `compute-baseline` in https://github.com/web-platform-dx/web-features/issues/979. Something like:

- Our recommended way of getting the Baseline status of a BCD key is together with a web-features identifier, like `computeBaselineForPartOfFeature('js-modules', 'javascript.statements.import')`. This would clamp the dates so that the part of the feature is never claimed to be older than the overall feature.
- Theres's also a `computeBaselineForCompatKey('javascript.statements.import')` which would find the features that includes the compat key, and clamp to the oldest of the found featuers.

This would end up having the right effect on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import at least.

captainbrosset:
@ddbeck now that #1204 is done, does that bring us closer to being able to use the same BCD keys in multiple features?

ddbeck:
@captainbrosset I'm not sure. This limitation came at MDN's request. I meet with the MDN folks periodically, so I'm going to talk to them next week and see what blockers, if any, remain from their perspective. I don't think any other consumers care.

ddbeck:
I talked to MDN last week. Now that we generate per-key statuses and we've discussed an approach for [pages that have multiple stories](https://github.com/mdn/yari/issues/11546), they don't have any objections _except_ that they'd like to wait until January to deal with the possibility of multiple keys (or the discouraged data).

ddbeck:
[...]

PR #1204, opened 2024-06-06
Provide an approved method to get a status for a single BCD key

ddbeck:
This adds an approved method, called `getStatus()`, to turn a `web-features` ID and a BCD key into a status for that specific BCD key.

Right now, this method is a formalism: it doesn't actually check that the BCD key is part of the named feature (or that the named feature even exists). But this illustrates the overall principle and will let us actually clamp down in follow up PRs, if this is the right approach.

Towards https://github.com/web-platform-dx/web-features/issues/979.


ddbeck:
Note that this probably needs owner's group consensus before merging.

ddbeck:
See also: https://github.com/web-platform-dx/web-features/issues/1173#issuecomment-2145033700

packages/compute-baseline/README.md
```diff
@@ -40,6 +38,33 @@ Run:
 
 ## Usage
 
+### Get a Baseline status for a portion of a feature
+
```
captainbrosset, line 42:
```suggestion

To get a Baseline status for a specific browser compatibility data entry within a `web-features` feature, call `getStatus` with the web feature's ID and the BCD feature key as parameters, as shown below:

```

packages/compute-baseline/README.md
```diff
@@ -40,6 +38,33 @@ Run:
...
+    edge: '14',
+    firefox: '39',
+    firefox_android: '39',
+    safari: '10.1',
+    safari_ios: '10.3'
+  }
+}
+```
+
 ### Check support for a group of compat keys
```
captainbrosset, line 69:
```suggestion

**Note**: the code shown below returns support data that has not received an editorial review. Do not use for presenting a Baseline status. See [Limitations](#limitations).

```

captainbrosset: I think it would be safe to add an intro like this here, as well as below in "Check support for a single support key" and in "Bring your own compatibility data".

packages/compute-baseline/README.md
```diff
@@ -92,17 +117,12 @@ If you don't want to import `@mdn/browser-compat-data` as your data source, you
...
 
 const compat = new Compat(data);
 
-computeBaseline(
-  {
-    compatKeys: ["css.properties.border-color"],
-  },
-  compat,
-);
+getStatus("fetch", "api.Response.json", compat);
```
captainbrosset, line 125: Can you help me understand when someone would need to use their own compat data? Are there other data sources out there that match the BCD schema and which `getStatus` would be able to work with?

If this is a valid use case, then I would suggest not using the `getStatus` method for it, since `getStatus` is for blessed/editorial reviewed Baseline statuses only. But if you use your own data, then we can't guaranty anything about it.

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -44,6 +44,33 @@ interface SupportStatus {
...
+ * feature overall. Then you'd call `getStatus('example-feature',
+ * 'api.ExampleManager.doExample')`.
+ */
+export function getStatus(
+  featureId: string,
+  compatKey: string,
+  compat: Compat = defaultCompat,
+) {
+  // TODO: actually check that featureId is a valid feature
+  // TODO: actually check that compatKey is tagged as featureId in BCD _or_ listed in web-features
```
captainbrosset, line 65: Are you planning on implementing those TODOs in this PR?

captainbrosset (commented):
I'm generally in favor of this approach. Just left a few comments.

packages/compute-baseline/README.md
```diff
@@ -40,6 +38,33 @@ Run:
 
 ## Usage
 
+### Get a Baseline status for a portion of a feature
+
+```javascript
+import { getStatus } from "compute-baseline";
+
+getStatus("fetch", "api.Response.json");
```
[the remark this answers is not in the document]
ddbeck: Done with 032d809.

packages/compute-baseline/README.md
ddbeck: Yes, I was trying to accommodate (for example) MDN upgrading BCD independently of web-features. Really, I'd like to be able to say something like: "You can use any version of BCD >= web-features's version of BCD" but we don't currently provide that information to consumers of web-features.

Another route here is to leave this usage undocumented.

packages/compute-baseline/README.md
ddbeck: Good idea. Done.

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -44,6 +44,33 @@ interface SupportStatus {
...
+  return JSON.parse(
+    computeBaseline(
+      { compatKeys: [compatKey], checkAncestors: true },
```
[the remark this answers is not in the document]
ddbeck: This is already tested with https://github.com/web-platform-dx/web-features/blob/cc5cccb1b17c02dd0884325d184019356b9a609e/packages/compute-baseline/src/baseline/index.test.ts#L78-L104. Do you want another test to confirm that `getStatus` does the same?

packages/compute-baseline/src/baseline/index.ts
ddbeck: I am not. This requires some non-trivial changes to the way the repo and dependencies are set up, which would overwhelm the policy and documentation changes going on here.

ddbeck:
Hi folks, I tagged the current owners for review. I brought this PR up on the last WebDX call (and the idea in general in previous web-features/Baseline calls), since it's a little bit of a departure from how we've done editorial review so far. I suspect this is getting close to landing, so if any of the owners want to personally review it before we move forward, I wanted to make sure there was an explicit opportunity for that. Thank you!

packages/compute-baseline/src/baseline/index.ts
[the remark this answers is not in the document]
foolip: Ah, existing tests are OK.

foolip (approved):

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -44,6 +44,33 @@ interface SupportStatus {
...
+ * For example, suppose you want to show a Baseline status for a specific method
+ * in a feature, which might've been supported earlier or later than the broader
+ * feature overall. Then you'd call `getStatus('example-feature',
+ * 'api.ExampleManager.doExample')`.
+ */
+export function getStatus(
+  featureId: string,
+  compatKey: string,
+  compat: Compat = defaultCompat,
+) {
```
foolip, line 63: Can you add the return type here? It looks like it's not the same as `computeBaseline`, and I wonder if that's intentional.

captainbrosset (approved):

packages/compute-baseline/src/baseline/index.ts
ddbeck: Done with 63e00adaab8af7e5b4f6cdef5847207e60c8e02b. (Sorting out the peer dependency between `compute-baseline` and `web-features` will be fussy. Saving it for a PR _just_ for that.)

atopal (approved):
LGTM!

ddbeck:
Thanks to everyone who had a look at this (or listened to me ramble on about it in the WebDX call last week)!

PR #1208, opened 2024-06-07
Allow use of ranges (≤) before the Baseline low date

foolip:
The `use_ranges_before_baseline_low_date` field controls when generating
the status, but it could also be done in overrides.

Closes https://github.com/web-platform-dx/web-features/issues/1018.

### DIFF
--- a/features/search-input-type.yml
+++ b/features/search-input-type.yml
@@ -5,3 +5,4 @@ group: forms
 caniuse: input-search
 # TODO: investigate minor differences in the early support of the feature
 # between https://caniuse.com/input-search and our generated status.
+use_ranges_before_baseline_low_date: true
--- a/features/search-input-type.yml.dist
+++ b/features/search-input-type.yml.dist
@@ -6,12 +6,12 @@ status:
   baseline_low_date: 2015-07-29
   baseline_high_date: 2018-01-29
   support:
-    chrome: "5"
+    chrome: ≤5
     chrome_android: "18"
     edge: "12"
-    firefox: "4"
+    firefox: ≤4
     firefox_android: "4"
-    safari: "5"
-    safari_ios: "4.2"
+    safari: ≤5
+    safari_ios: ≤4.2
 compat_features:
   - html.elements.input.type_search
--- a/index.ts
+++ b/index.ts
@@ -20,6 +20,9 @@ const nameMaxLength = 80;
 // The longest description allowed, to avoid them growing into documentation.
 const descriptionMaxLength = 300;
 
+// The cutoff point for when ranges (≤) are allowed.
+const allowRangesBeforeDate = Temporal.PlainDate.from('2016-01-01');
+
 // Internal symbol to mark draft entries, so that using the draft field outside
 // of a draft directory doesn't work.
 const draft = Symbol('draft');
@@ -27,9 +30,10 @@ const draft = Symbol('draft');
 // Some FeatureData keys aren't (and may never) be ready for publishing.
 // They're not part of the public schema (yet).
 const omittables = [
+    "group",
     "snapshot",
-    "group"
-]
+    "use_ranges_before_baseline_low_date",
+];
 
 function scrub(data: any) {
     for (const key of omittables) {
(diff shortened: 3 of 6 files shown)


foolip:
I have tried sprinkling this onto the HTML element draft features and really like what it's doing. For example:

```diff
--- a/features/draft/html-elements/h1.yml.dist
+++ b/features/draft/html-elements/h1.yml.dist
@@ -11,5 +11,5 @@ status:
     edge: "12"
     firefox: "1"
     firefox_android: "4"
-    safari: "4"
-    safari_ios: "3.2"
+    safari: ≤4
+    safari_ios: ≤3.2
```

Not having to spend any time updating BCD to show that `<h1>` was actually supported in Safari 1 will be great.

index.ts
```diff
@@ -168,6 +172,16 @@ for (const [key, data] of yamlEntries('features')) {
         }
     }
 
+    // Ensure that version ranges are only used for prehistoric features.
+    if (data.status?.support) {
+        const versions: string[] = Object.values(data.status.support);
+        if (versions.some((v) => v.startsWith('≤'))) {
+            if (data.status.baseline_low_date !== earliestBaselineLowDate) {
```
foolip, line 179: Should I change this to allow for ranges when the Baseline low date is before 2016? I'm on the fence.

index.ts
```diff
@@ -20,16 +20,20 @@ const nameMaxLength = 80;
 // The longest description allowed, to avoid them growing into documentation.
 const descriptionMaxLength = 300;
 
+// The earliest possible Baseline low date, the release date of Edge 12.
+const earliestBaselineLowDate = '2015-07-29';
```
ddbeck, line 24: I don't know if this is better exactly, but it feels more "correct":

```suggestion
// needs import { browser } from "compute-baseline/browser-compat-data";
const earliestBaselineLowDate = browser("edge").version("12").date.toString();
```

index.ts
ddbeck: I like this being a more specific thing to start, then expanded to later uses later (as needed).

index.ts
```diff
@@ -168,6 +172,16 @@ for (const [key, data] of yamlEntries('features')) {
         }
     }
 
+    // Ensure that version ranges are only used for prehistoric features.
+    if (data.status?.support) {
+        const versions: string[] = Object.values(data.status.support);
+        if (versions.some((v) => v.startsWith('≤'))) {
```
ddbeck, line 178: I don't love adding uncertainty to potentially already-certain ranges in BCD, but I expect a future resolution to https://github.com/web-platform-dx/web-features/issues/1211 to fix this at a later date, so it's fine.

ddbeck (approved):

features/search-input-type.yml.dist
```diff
@@ -6,12 +6,12 @@ status:
...
+    chrome: ≤5
     chrome_android: "18"
     edge: "12"
-    firefox: "4"
+    firefox: ≤4
     firefox_android: "4"
-    safari: "5"
-    safari_ios: "4.2"
+    safari: ≤5
+    safari_ios: ≤4.2
```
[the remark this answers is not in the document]
foolip: Yes, which is different from "it may have come earlier _or later_ than we report" which would be a natural outcome of "we're not sure we trust this data and haven't reviewed it carefully." But, when throwing together a bunch of things, the most recent version is the one we use, so there's good reason to expect that the error will usually be mostly in the direction of claiming features are newer than they are.

features/search-input-type.yml
```diff
@@ -5,3 +5,4 @@ group: forms
 caniuse: input-search
 # TODO: investigate minor differences in the early support of the feature
 # between https://caniuse.com/input-search and our generated status.
+use_ranges_before_baseline_low_date: true
```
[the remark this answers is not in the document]
foolip: I think that would provide a useful clue to someone reading the source and wondering if they can use the data through a published package. However, it wouldn't make it any easier for us to filter out such keys, because in order to catch typos we have to enumerate all allowed keys.

index.ts
foolip: I get the discomfort. Should we instead have a single possible combination of ranged browser versions as described in https://github.com/web-platform-dx/web-features/issues/1018#issuecomment-2087553410?

index.ts
ddbeck: OK, if this will eventually become a date, let's leave it as a date.

index.ts
ddbeck: Let's leave this for now too. I think bubbling up BCD's uncertainty will mess with a bunch of values anyway (especially dates), so let's sort that out first.

foolip: Alright, let me make some cosmetic changes around the Edge 12 date and land this.

tidoust (approved):

ddbeck:
@foolip I think this is ready to go, pending a main merge or rebase.

foolip:
This got in the way of grouping comments, see https://github.com/web-platform-dx/web-features/pull/1231/files#r1656826439. I'm considering if we should revert this as it's less important than https://github.com/web-platform-dx/web-features/pull/1231.

PR #1205, opened 2024-06-07
Remove orphaned dist file

foolip:

PR #1209, opened 2024-06-07
Catch orphaned dist files during `--check`

ddbeck:
This prevents `.dist` files from hanging around even after the authored YAML has been deleted.

Inspired by https://github.com/web-platform-dx/web-features/pull/1205

PR #1210, opened 2024-06-07
Make the dist script crawl all features by default

foolip:
The "**" path pattern in package.json doesn't work on at least macOS,
and contributed to an orphaned dist file:
https://github.com/web-platform-dx/web-features/pull/1205.

To avoid such issues, let the command expand directory names to any dist
files within, and make the dist script crawl all features by default.

This also allows simplifiying instructions in a README.


foolip:
This leaves in the check from https://github.com/web-platform-dx/web-features/pull/1209 since that also provides a clear error message.

ddbeck (approved):

PR #1213, opened 2024-06-07
Add dist refresh workflow

ddbeck:
This PR adds a GitHub Actions workflow that lets reviewers request a dist file refresh by adding a `refresh dist` label to the pull request. It only refreshes dists where the dist or authored YAML has changed on the PR, relative to the target branch.

A known limitation: this will only work for owners/peers on the repo, who can set labels on a PR.

### DIFF
--- a/.github/workflows/refresh_dist.yml
+++ b/.github/workflows/refresh_dist.yml
@@ -30,7 +30,7 @@ jobs:
         env:
           CHANGED_FEATURES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          npx tsx scripts/dist.ts ${CHANGED_FEATURES}
+          npm run dist -- ${CHANGED_FEATURES}
 
       - uses: stefanzweifel/git-auto-commit-action@v5
         if: steps.changed-files.outputs.any_changed == 'true'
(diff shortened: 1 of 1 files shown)


captainbrosset (approved):
Seems useful. How about creating a pre-commit hook that checks for the need to re-generate dist files? And even runs the script for you?

.github/workflows/refresh_dist.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - run: npm install
+        if: steps.changed-files.outputs.any_changed == 'true'
+
+      - name: Refresh dist files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        env:
+          CHANGED_FEATURES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          npx tsx scripts/dist.ts ${CHANGED_FEATURES}
```
foolip, line 33:
```suggestion
          npm run dist ${CHANGED_FEATURES}
```

foolip (approved):

ddbeck:
> Seems useful. How about creating a pre-commit hook that checks for the need to re-generate dist files? And even runs the script for you?

@captainbrosset Yes, this is a good idea too, though I don't know know much about pre-commit hooks (I rarely use or make them myself). Though even if we had a hook, we'd still need this for changes applied during GitHub PR reviews or via the web editor.

PR #1214, opened 2024-06-07
Mark dist files as read-only in VS Code

ddbeck:
For contributors using VS Code, this configuration change marks `.dist` files as read-only, so it's harder to accidentally edit a dist file. Less likely now with the compact dists, but a convenience nonetheless.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
   "files.readonlyInclude": {
-    "features/*.yml.dist": true
+    "features/**/*.yml.dist": true
   }
 }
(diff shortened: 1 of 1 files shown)


foolip (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "files.readonlyInclude": {
+    "features/*.yml.dist": true
```
foolip, line 3: Hmm, does this include dist in subdirectories?

ddbeck: Good catch. Fixing this now.

PR #1216, opened 2024-06-07
Add support to "pin" a feature's support to its entrypoint(s)

foolip:


foolip:
After https://github.com/web-platform-dx/web-features/pull/1204 we will need something like this, and I've prepared this PR to show how I expect we'd use it. For it to work, compute-baseline would need to understand it, I think.

@ddbeck is this how you imagined this would work? What changes are needed in compute-baseline to make it work?

features/trusted-types.yml
```diff
@@ -3,17 +3,16 @@ description: Trusted types allow you to lock down insecure parts of the DOM API
 spec: https://w3c.github.io/trusted-types/dist/spec/
 caniuse: trusted-types
 usage_stats: https://chromestatus.com/metrics/feature/timeline/popularity/2722
-# toJSON() is excluded because they are not needed to use trusted types in the
-# idiomatic way. toJSON() shipped in Chrome 90, and the rest in Chrome 83.
+entrypoint: api.trustedTypes
```
ddbeck, line 6: @foolip we talked about this on our call today. Instead of doing this, what do you think of something like:

```yaml
status:
  compute_from:
    - api.trustedTypes
```

I'd expect an error to throw when `Set(data.status.compute_from).isSubsetOf(Set(data.compat_features)) === false`

scripts/dist.ts
```diff
@@ -78,6 +78,8 @@ function toDist(sourcePath: string): YAML.Document {
   );
   const { name: id } = path.parse(sourcePath);
 
+  const entrypoint: yaml.get('entrypoint');
```
ddbeck, line 81: For my earlier suggestion, instead we'd have to check whether `compute_from` is set. If it is, we'd treat it as `overridden.compatFeatures` instead of `compat_features` itself.

foolip:
Closing in favor of https://github.com/web-platform-dx/web-features/pull/1227.

PR #1220, opened 2024-06-10
Update Safari version for typed array methods

foolip:
Safari 9.1 vs 10 was carefully researched here:
https://github.com/mdn/browser-compat-data/pull/23319

Where BCD is still wrong, we have to keep the data manually maintained
until the next BCD release. After that it can be generated as .yml.dist.

This changes the Baseline low date by 1 year.


Elchi3 (approved):

features/array-from.yml.dist
```diff
@@ -0,0 +1,18 @@
...
+
+status:
+  baseline: high
+  baseline_low_date: 2016-09-20
+  baseline_high_date: 2019-03-20
+  support:
+    chrome: "45"
+    chrome_android: "45"
+    edge: "12"
+    firefox: "38"
```
foolip, line 12: Note that this changed from 32 to 38, but I didn't check why because it's long ago and doesn't change the Baseline date.

PR #1225, opened 2024-06-10
Refresh all dists on `package.json` file changes

ddbeck:
This is for PRs like https://github.com/web-platform-dx/web-features/pull/1223 where no individual feature file has changed, but there are still dist updates to be made.

(A follow-on would be a workflow that automatically labels dependabot PRs with refresh dist.)

### DIFF
--- a/.github/workflows/refresh_dist.yml
+++ b/.github/workflows/refresh_dist.yml
@@ -12,28 +12,40 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - id: changed-files
-        uses: tj-actions/changed-files@v44
+      - id: changed-package-json
+        uses: tj-actions/changed-features@v44
+        with:
+          files: |
+            **/package.json
+            **/package-lock.json
+
+      - id: changed-features
+        uses: tj-actions/changed-features@v44
         with:
           files: |
             features/**/*.yml
             features/**/*.yml.dist
 
       - uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh_dist.yml
```diff
@@ -12,28 +12,40 @@ jobs:
...
         env:
-          CHANGED_FEATURES: ${{ steps.changed-files.outputs.all_changed_files }}
+          CHANGED_FEATURES: ${{ steps.changed-features.outputs.all_changed_files }}
         run: |
           npm run dist -- ${CHANGED_FEATURES}
 
+      - name: Refresh all dist files (for package*.json changes)
+        if: steps.changed-package-json.outputs.any_changed == 'true'
+        run: |
+          npm run dist -- features/**/*.yml.dist
```
foolip, line 45: Is there any reason we can't just update everything and not care about what files were changed?

```suggestion
          npm run dist
```

ddbeck: In the "changed-features" case, we don't want to change dists for files not already in-scope for the PR. _Ordinarily_ it shouldn't make a difference, but I was thinking of some of the bad merges that have taken place, where refreshing the dist would add surprising files to the diff of unrelated PRs.

In the "changed-package-json" case, I forgot we changed the dist script to do this by default. I'll fix this.

foolip (approved):

Issue #1160, opened 2024-05-30
Subfeatures for later additions


foolip:
Most features are launched with an initial API shape and then evolve in many steps over time. Fetch is an example of this, where `fetch()` became Baseline in 2017, and the `Request` `destination` property followed in 2018. So far we've just excluded such later additions from our source YAML files, and left a lot of TODOs.

So far we've had two ideas for how to handle this:

- Just add new features with their own identifier, name, description, etc.
- Allow features to be a more living concept, and make it clear what the status of every part is via https://github.com/web-platform-dx/web-features/issues/979, giving the BCD keys individual Baseline statuses.

Because most features evolve in many steps, the overhead of handling minor additions is a major consideration. This is the main drawback of the first option above, that we'd need to name and describe lots of such features.

Brainstorming with @atopal @nairnandu @petele @rachelandrew we have a third option, which is a kind of "light" features. By example:

```yml
subfeatures:
  picture-in-picture-window:
    name: enterpictureinpicture event pictureInPictureWindow property
    compat_features:
      - api.PictureInPictureEvent
      - api.PictureInPictureEvent.PictureInPictureEvent
      - api.PictureInPictureEvent.pictureInPictureWindow
  picture-in-picture-pseudo:
    name: ":picture-in-picture pseudo-class"
    compat_features:
      - css.selectors.picture-in-picture
```

This requires an identifier and a name per subfeature, but nothing else.

@ddbeck WDYT?

captainbrosset:
> This requires an identifier and a name per subfeature, but nothing else.

I think there will be cases where we'll also need to describe those sub-features. Whether we do this by adding a description field, or just by making the name longer, I sense that it will be similar work than option 1 above.

I'm not saying I'm against this solution though. It does seem like in many cases, it will be less work.

captainbrosset:
This came up in our bi-weekly Baseline call today: [notes](https://docs.google.com/document/d/1ree75ImLZjf60lTZ3BhCaLHygxgywr7SBXp-q0xPs8A/edit#heading=h.6gou55b9mud1).

ddbeck:
There are some things that I like and dislike about this idea, but I want to tease apart some goals of this first. What I'm hearing here (and from the discussion on the Baseline call earlier today) is threefold:

[...]

Issue #1159, opened 2024-05-30
Document dist files and how to update them


foolip:
A new contributor just might ask: What are these even? Why is the CI failing?

foolip:
https://github.com/web-platform-dx/web-features/pull/1162 is the most important step I think we could take to make it easier to explain and require fewer dist file updates.

ddbeck:
Brainstorming some topics needed here:

- Concept: What's a dist file? Why do we have them?
- Task: how to create a dist file for the first time (in the course of creating a new feature)
- Task: how to update a dist file
- Reference: what fields are contained in the dist file and what's _not_ in the dist file

foolip:
Where should we put the docs? I'd like to document `compute_from` at least.

ddbeck:
I think either starting a new section in [`CONTRIBUTING.md`](https://github.com/web-platform-dx/web-features/blob/main/docs/CONTRIBUTING.md) or linking to a new file from there would be the way to go. I'd suggest this outline to start if you want to write any part of this or even just seed a markdown file with headings:

- Feature authoring
  - Before you begin
    - Features are authored in YAML [concept: what's a feature file?]
    - Feature files are used to generate dist files [concept: what's a dist file?]
  - Create a new feature
    - How to find and edit a draft [task]
    - Creating a new feature from scratch [task]
  - Update an existing feature [task]
    - Edit authored YAML file [⬅ example `compute_from` usage goes here]
    - Regenerate dist file [task]
  - Feature YAML fields
    - `compute_from` [⬅ more formal `compute_from` description goes here, e.g., type, whether it's published, nitpicky style stuff, etc.]

captainbrosset:
We have this now: https://github.com/web-platform-dx/web-features/blob/main/docs/CONTRIBUTING.md#feature-dist-files

When this was filed originally, the dist file copied everything. Now we don't have this problem anymore, and we have a comment discouraging people from touching dist files.
We could change the comment to also mention the docs.

PR #1227, opened 2024-06-10
Introduce compute_from to pin a feature's status to BCD key(s)

foolip:


foolip:
There are things I'd like to improve here still, but this shows the promise of it.

features/map.yml
```diff
@@ -18,33 +18,5 @@ spec:
...
-  - javascript.builtins.Map.has
-  - javascript.builtins.Map.key_equality_for_zeros
-  - javascript.builtins.Map.keys
-  - javascript.builtins.Map.set
-  - javascript.builtins.Map.size
-  - javascript.builtins.Map.values
-  - javascript.builtins.Map.@@iterator
-  - javascript.builtins.Map.@@species
+  compute_from:
+    - javascript.builtins.Map.Map
```
captainbrosset, line 22: I probably don't have the full context here, but wouldn't something like this be a bit more intuitive:

```yaml
status:
  compat_features:
    - javascript.builtins.Map.*
```

Pros: no need for a new field in the YAML (`compute_from`), and the `*` character is well known to developers.

But, again, maybe this is not reflecting fully the idea of the feature here.

ddbeck:
For @captainbrosset and anyone else following along:

The idea here is to let us pick a subset of a feature's compat keys to compute the Baseline status from. Applications for this would be, for example, to:

- Align a feature to caniuse's reported availability
- Capture a feature's interoperable date despite minor additions to that feature
- Ignore specific faulty/weird/complex/missing keys in BCD while still maintaining a connection to the underlying data

In general, this would let us associate more keys with features without minting¹ new features or moving a feature's Baseline low date upon later additions. Accepting this PR would depend on accepting https://github.com/web-platform-dx/web-features/pull/1204, too, so that consumers could show the most-correct status for a given slice of a feature.

Discussion relating to this idea comes mostly from https://github.com/web-platform-dx/web-features/issues/1160 and (to a lesser extent) https://github.com/web-platform-dx/web-features/issues/866 and (regrettably) a few conversations between Philip and myself.

(@foolip please correct me if I've mischaracterized anything here)

features/map.yml
ddbeck: I tried to add more context with https://github.com/web-platform-dx/web-features/pull/1227#issuecomment-2160997849.

To the specific wildcard suggestion, I think it's contrary to the idea of pinning to specific keys. I also worry that it would be confusing, since it would be a wildcard against the keys in `compat_features` (or tagged in BCD), not all keys matching the pattern.

scripts/dist.ts
```diff
@@ -88,22 +88,34 @@ function toDist(sourcePath: string): YAML.Document {
     status: yaml.toJS().status,
   };
 
+  const allCompatFeatures: string[] = Array.isArray(overridden.compatFeatures)
```
foolip, line 91: I'd like to rebase this on top of https://github.com/web-platform-dx/web-features/pull/1233 because I think it'll look nicer after that change.

foolip:
@ddbeck I'd like to document how to use `compat_features` but I'm not sure where. See https://github.com/web-platform-dx/web-features/issues/1159.

features/trusted-types.yml
```diff
@@ -3,33 +3,6 @@ description: Trusted types allow you to lock down insecure parts of the DOM API
...
-  - api.TrustedTypePolicyFactory.getPropertyType
-  - api.TrustedTypePolicyFactory.isHTML
-  - api.TrustedTypePolicyFactory.isScript
-  - api.TrustedTypePolicyFactory.isScriptURL
-  - api.trustedTypes
-  - http.headers.Content-Security-Policy.require-trusted-types-for
-  - http.headers.Content-Security-Policy.trusted-types
+status:
+  compute_from:
+    - api.trustedTypes
```
ddbeck, line 8: Hmm. On the one hand, I'd say that I'd prefer to have the maximal set of keys in `compute_from`, particularly when we're sidestepping a single key (as suggested by the comment)—we'll see the status change if something alters the story in the broad feature.

On the other hand, maybe https://github.com/web-platform-dx/web-features/pull/1231 will provide us with sufficient visibility into support changes? I'm not sure.

features/webtransport.yml
```diff
@@ -3,27 +3,6 @@ description: The `WebTransport` API transmits data between a client and a server
...
-
-# As of 2023-08-10, the spec calls these "feature at risk"
-# - api.WebTransport.WebTransport.options_congestionControl_parameter
-# - api.WebTransport.congestionControl
-
-# TODO: figure out what the story is with WebTransportReceiveStream
-# This feature has a different compatibility story than closely-related
-# interfaces (i.e., WebTransportBidirectionalStream)
-# - api.WebTransportReceiveStream
-# - api.WebTransportReceiveStream.getStats
```
ddbeck, line 29: Can we preserve these comments or migrate them to an issue? It looks like these keys didn't make into the tag in BCD, so this will just evaporate without being dealt with.

ddbeck (approved):
Happy with this, but I have one straggling thought and a request to preserve comments. Thank you!

features/trusted-types.yml
foolip: I think a minimal set will serve us better than a maximal set and that https://github.com/web-platform-dx/web-features/pull/1231 will indeed add visibility into whether the choice makes sense.

The minimal set is quite easy to determine without checking implementation status, it's just the entrypoint(s) to the feature most of the time.

features/webtransport.yml
foolip: I've restored these comments.

features/trusted-types.yml
foolip: In this case there are more keys that could be listed, like "api.TrustedTypePolicyFactory.createPolicy", but it's hard to say where to draw the line for a minimal set.

ddbeck: Yeah, now that I've slept on this I think keying on entrypoints is the right way to go. It suggests a pretty good test for inclusion even, something close to: "omitting such a key would make the feature unusuable."

features/webtransport.yml.dist
```diff
@@ -1,15 +1,23 @@
...
 compat_features:
-  # ⬇️ Same status as overall feature ⬇️
   # baseline: false
   # support:
   #   chrome: "97"
   #   chrome_android: "97"
   #   edge: "97"
   - api.WebTransport.createUnidirectionalStream
 
+  # ⬇️ Same status as overall feature ⬇️
```
foolip, line 20: I see the sort order here is actually wrong, the following block has more supporting browsers than the previous. I'll send a PR to fix.

foolip: That was https://github.com/web-platform-dx/web-features/pull/1295.

PR #1231, opened 2024-06-11
Group BCD keys in dist by status block

foolip:


foolip:
This depends on https://github.com/web-platform-dx/web-features/pull/1204.

features/trusted-types.yml
```diff
@@ -31,5 +31,5 @@ compat_features:
   - api.TrustedTypePolicyFactory.isScript
   - api.TrustedTypePolicyFactory.isScriptURL
   - api.trustedTypes
-  - http.headers.Content-Security-Policy.require-trusted-types-for
-  - http.headers.Content-Security-Policy.trusted-types
+  # - http.headers.Content-Security-Policy.require-trusted-types-for
```
foolip, line 34: I had to comment these out because `http.headers.Content-Security-Policy` has non-real values.

features/abortable-fetch.yml.dist
```diff
@@ -13,3 +13,18 @@ status:
...
+  #   chrome: "66"
+  #   chrome_android: "66"
+  #   edge: "16"
+  #   firefox: "57"
+  #   firefox_android: "57"
+  #   safari: "11.1"
+  #   safari_ios: "11.3"
+  - api.fetch.init_signal_parameter
+  # Same status as overall feature
+  - api.Request.signal
```
ddbeck, line 30: White space to separate "generations" of keys would help readability, I think

```suggestion
  #   safari_ios: "11.3"
  - api.fetch.init_signal_parameter

  # Same status as overall feature
  - api.Request.signal
```

features/abortsignal-any.yml.dist
```diff
@@ -12,3 +12,6 @@ status:
     firefox_android: "124"
     safari: "17.4"
     safari_ios: "17.4"
+compat_features:
+  # Same status as overall feature
```
ddbeck, line 16: A nice-to-have thing would be to omit this when there's only one group of keys (e.g., every single single-entry feature, but also features where all of the keys landed at the same time).

ddbeck (commented):
This is really cool!

features/abortable-fetch.yml.dist
foolip: I agree, but wasn't sure how to achieve it. Do you know?

scripts/dist.ts
```diff
@@ -123,25 +123,65 @@ function toDist(sourcePath: string): YAML.Document {
...
       }
-      insertStatus(dist, JSON.parse(status.toJSON()));
+      const tmp = JSON.parse(status.toJSON());
+      statusYaml = YAML.stringify(tmp);
+      insertStatus(dist, tmp);
     }
   }
 
+  // Sort first by Baseline low date, TODO.
+  const compatFeatures = Array.isArray(overridden.compatFeatures) ? overridden.compatFeatures : taggedCompatFeatures;
```
foolip, line 133: I'd like to rebase this on top of https://github.com/web-platform-dx/web-features/pull/1233 because I think it'll look nicer after that change.

features/abortable-fetch.yml.dist
foolip: I found a way!

features/trusted-types.yml
foolip: Fixed by https://github.com/mdn/browser-compat-data/pull/23352 it seems.

features/search-input-type.yml.dist
```diff
@@ -14,4 +14,15 @@ status:
     safari: ≤5
     safari_ios: ≤4.2
 compat_features:
+  # baseline: high
```
foolip, line 17: This notably does not say "Same status as overall feature" because the `use_ranges_before_baseline_low_date` logic only runs for the overall feature. @ddbeck I see a few options:

- Add a `useRangesBeforeBaselineLowDate` parameter to `computeBaseline`. This would fix the problem locally, but users of `computeBaseline` don't know what to pass in. You shouldn't really be able to pick if you want the ranges or not as a consumer, that's something we do as providers of the data.
- Revert https://github.com/web-platform-dx/web-features/pull/1208 to not block progress on this more important PR and worry about it later.

WDYT?

foolip: I've sent https://github.com/web-platform-dx/web-features/pull/1286.

ddbeck: I agree on the revert and I just approved it. I need to think on the broader problem. I think doing https://github.com/web-platform-dx/web-features/issues/1211—which would just make ranges a dumb fact of the status calculation—is the right way to go, but that's more of a guess right now.

foolip:
The main open question is the sort order of blocks.

We should sort first by Baseline date, oldest first. But among blocks with the same Baseline date, what's the secondary sort key?

One approach is to use all of the (between 0 and 7) releases dates to sort. The most recent date is the Baseline low date, and if that's the same we move on to compare the 2nd most recent date, and so on. The issue is that two browsers can have the same release date. For example, Chrome 126 and Firefox 127 were both released on 2024-06-11, and this seems to be common now. Using dates only won't produce unique sort keys in every case where browser versions differ.

Any way of breaking that tie should be fine, but I don't have a concrete way in mind right now...

ddbeck:
Huh, this is complicated. I think I'd like to see the overall sort look like this:

1. Baseline low date blocks, oldest first.
2. After all Baseline blocks, sorted by number of engines.

Break ties within 1 and 2 by:

a. Number of compat keys, greater first.
b. Next last-implementer release date, older first.
c. Lexical ordering of the first compat key in each block.

Taken together, this ought to the most-widely available things first, with bigger slices before smaller ones. Then it's back to date ordered, with something _very_ arbitrary as a fallback.

foolip:
@ddbeck I've implemented a sort order now, it's the same as you suggested except the tie breaking. I wanted to avoid this in particular:

> Next last-implementer release date, older first.

This information isn't visible in the dist files, and I think it might be hard to understand the order if it uses inferred information. Adding dates in comments would address this, if you think this is important.

Instead I just sorted by the first non-equal version when the number of supporting browsers is the same. This is done even if it's not the same browsers in two groups.

The resulting order feels _alright_ but could probably be improved with experience reviewing with groups.

features/aspect-ratio.yml.dist
```diff
@@ -14,6 +14,31 @@ status:
...
+  #   chrome: "79"
+  #   chrome_android: "79"
+  #   edge: "79"
+  #   firefox: "71"
+  #   firefox_android: "79"
+  #   safari: "15"
+  #   safari_ios: "15"
+  - html.elements.img.aspect_ratio_computed_from_attributes
+
+  # Same status as overall feature
```
ddbeck, line 43: OK, I'm now thinking that this comment ought to show the full status as well—it's some distance from the status it's referencing, which makes it hard to review (also because there's no diff to the status itself, I have to click to expand it).

What do you think of something like this, so it stands out, too?

```suggestion
  # ⬇️ Same status as overall feature ⬇️
  # baseline: high
  # …
```

features/border-image.yml.dist
```diff
@@ -14,14 +14,87 @@ status:
...
+  # baseline_low_date: 2015-07-29
+  # baseline_high_date: 2018-01-29
+  # support:
+  #   chrome: "15"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "15"
+  #   firefox_android: "15"
+  #   safari: "6"
+  #   safari_ios: "6"
```
ddbeck, line 40: Ah this is cool! This is very likely a bug in the data, right? It seems deeply unlikely that Firefox shipped `border-image-width` before `border-image-source`.

features/broadcast-channel.yml.dist
```diff
@@ -13,10 +13,22 @@ status:
...
+  # baseline_low_date: 2022-03-14
+  # support:
+  #   chrome: "54"
+  #   chrome_android: "54"
+  #   edge: "79"
+  #   firefox: "38"
+  #   firefox_android: "38"
+  #   safari: "15.4"
+  #   safari_ios: "15.4"
   - api.BroadcastChannel
```
ddbeck, line 26: Neat! We can use `compute_from: api.BroadcastChannel` to sync up with caniuse.

features/conic-gradients.yml.dist
```diff
@@ -13,3 +13,24 @@ status:
...
+
+  # baseline: false
+  # support:
+  #   chrome: "72"
+  #   chrome_android: "72"
+  #   firefox: "83"
+  #   firefox_android: "83"
+  #   safari: "12.1"
+  #   safari_ios: "12.2"
+  - css.types.image.gradient.conic-gradient.doubleposition
```
ddbeck, line 36: Something weird has happened here. How did we get to `baseline: high` without Edge?

features/container-queries.yml.dist
```diff
@@ -12,3 +12,51 @@ status:
...
+  # baseline_low_date: 2022-09-12
+  # support:
+  #   chrome: "105"
+  #   chrome_android: "105"
+  #   edge: "105"
+  #   firefox: "101"
+  #   firefox_android: "101"
+  #   safari: "16"
+  #   safari_ios: "16"
+  - css.properties.contain.inline-size
```
ddbeck, line 50: These keys have no business being there and these comments make it obvious. Nice! Sent https://github.com/web-platform-dx/web-features/pull/1290

ddbeck (commented):
@foolip I'm really appreciating the new sort. I reviewed the dist files from A to C. I think this is really close, but I noticed a couple of oddities.

features/conic-gradients.yml.dist
ddbeck: Filed https://github.com/mdn/browser-compat-data/pull/23603

features/aspect-ratio.yml.dist
foolip: I've had the same thought, so I'll make it so.

foolip: Done.

features/border-image.yml.dist
foolip: Yep. This grouping will throw up a lot of issues like this. If we try to fix them I'm sure we'll move slower instead of faster. Do we need some fast way to mark issues like this as future TODOs?

features/conic-gradients.yml.dist
foolip: I've sent https://github.com/web-platform-dx/web-features/pull/1294, we shouldn't use both `checkAncestors: true` and `checkAncestors: false` in the dist script.

PR #1235, opened 2024-06-12
Remove chromestatus.com usage stats

foolip:
Chrome usage stats for web-features is now being maintained in the
Chromium repository:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom;drc=41e539e573351d8e57160f2de3201c28e4a3cc19

All of the 87 features touched here have an enum entry in the above
webdx_feature.mojom file, confirmed by transforming the feature
identifier to kCamelCase and looking for an exact match in the list.

### DIFF
--- a/features/accent-color.yml
+++ b/features/accent-color.yml
@@ -1,4 +1,3 @@
 name: accent-color
 description: The `accent-color` CSS property sets a color for checkboxes, radio buttons, and other form controls.
 spec: https://drafts.csswg.org/css-ui-4/#widget-accent
-usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/695
--- a/features/animation-composition.yml
+++ b/features/animation-composition.yml
@@ -1,4 +1,3 @@
 name: animation-composition
 description: The `animation-composition` CSS property chooses how to combine animations that affect the same property.
 spec: https://drafts.csswg.org/css-animations-2/#animation-composition
-usage_stats: https://chromestatus.com/metrics/css/timeline/popularity/743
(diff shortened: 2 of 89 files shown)


foolip:
This is draft because it depends on https://github.com/web-platform-dx/web-features/pull/1234.

foolip:
@jstenback FYI, removing this from web-features as discussed. It realized I could transform the identifiers quickly in VS Code, so I just did the check again the enum directly, and found one outlier, the draft speech recognition feature.

foolip:
> So should we keep the `usage_stats` field at all?

I wanted to play it safe and remove it later if it's been unused for some time, because removing it and adding back would be churn. But I can remove it if you think it'll turn into unused cruft, which honestly is likely.

jstenback:
I went ahead and created a CL to add the speech recognition one to Chromium but realized in that process that it's already there. So AFAICT this is all good to go.

ddbeck (approved):
> > So should we keep the `usage_stats` field at all?
> 
> I wanted to play it safe and remove it later if it's been unused for some time, because removing it and adding back would be churn. But I can remove it if you think it'll turn into unused cruft, which honestly is likely.

@foolip I'm happy either way. If you choose to merge this without removing the field, please file an issue so we get to it later.

foolip:
> > > So should we keep the `usage_stats` field at all?
> > 
> > 
> > I wanted to play it safe and remove it later if it's been unused for some time, because removing it and adding back would be churn. But I can remove it if you think it'll turn into unused cruft, which honestly is likely.
> 
> @foolip I'm happy either way. If you choose to merge this without removing the field, please file an issue so we get to it later.

I'll just remove it entirely now.

PR #1244, opened 2024-06-14
New overlay feature

captainbrosset:
The `overlay` CSS property is special in that it can't be used directly, but only as one of the transitioned property, and only if the transition behavior is set to `allow-discrete`. It's useful when animating top-layer things (popover, dialogs) to prevent the browser from too quickly removing them from the top layer, before the animation has ended.

Let me know if the description makes sense.


features/overlay.yml
```diff
@@ -0,0 +1,5 @@
+name: overlay
+description: The `overlay` CSS property, used as an `allow-discrete` CSS transition, prevents a top layer element, such as a popover or a `<dialog>`, from being removed from the top layer before it has finished animating. You can't set the value of the `overlay` property, only use it as transition property.
```
ddbeck, line 2: I think this description is really good! I have the teeniest, tiniest style nit.

```suggestion
description: The `overlay` CSS property, used as an `allow-discrete` CSS transition, prevents a top layer element, such as a popover or a `<dialog>`, from being removed from the top layer before it has finished animating. You can't set the value of the `overlay` property; only use it as transition property.
```

ddbeck (approved):

PR #1245, opened 2024-06-14
New document.caretPositionFromPoint feature

captainbrosset:


features/document-caretpositionfrompoint.yml
```diff
@@ -0,0 +1,5 @@
+name: document.caretPositionFromPoint()
+description: The `document.caretPositionFromPoint()` method returns a DOM node and an offset within that node for a given point in the viewport.
```
ddbeck, line 2: Perhaps a little additional context here would be helpful to folks (like me) who are unfamiliar with this?

```suggestion
description: The `document.caretPositionFromPoint()` method finds an insertion point, represented by a DOM node and an offset within that node, for given coordinates in the viewport.
```

ddbeck (approved):

PR #1246, opened 2024-06-14
New JSON module script feature

captainbrosset:


features/json-modules.yml
```diff
@@ -0,0 +1,8 @@
+name: JSON module scripts
+description: 'JSON module scripts allow other modules to load JSON data by using `import ... with {type: "json"}`.'
```
ddbeck, line 2: I'm concerned about the name here (and potentially for CSS modules)—for example, there are no results on Stack Overflow with this name as the search string. I _think_ this is better known by the syntax ("import attributes"), but I suppose there's a possibility that we're talking about two distinct but related things. Take my following suggestions with a grain of salt.

```suggestion
name: JSON import attributes
description: 'Module `import ... with { type: "json" }` statements load JSON data. Also known as JSON module scripts and formerly known as import assertions.'
```

captainbrosset: You're right that this feature is all about the import attribute, with type `json`. "JSON module scripts" comes from the spec, rather than what people have been calling it. The only problem is that I don't think that devs have been calling it anything specific so far, other than "importing JSON into modules". This feature has evolved over time, and there are other ways to load JSON. So naming this is hard.

I like your proposed description, because it places the focus on the module (the receiver of the data), rather than on the JSON (the data itself). In the end, this is a feature for modules to use. So in that sense, your proposed name makes sense too.

How about `JSON-type import attribute`? Or is that too weird?

captainbrosset: And then, `CSS-type import attribute` for the CSS module scripts one too.

ddbeck: Sorry it took me so long to come back to this. I think "JSON-type import attribute" is OK with me. Personally, I'd probably rank these (tentatively) like so:

1. JSON import attributes
2. JSON-type import attributes
3. JSON(-type) import attribute (see https://github.com/web-platform-dx/web-features/pull/1026#issuecomment-2090254267 about pluralization)

Though not strongly. One thing I'm not completely sure about is whether attributes other than `type` are expected. I'm assuming we're not likely to see something like `{ type: "json", foo: "json" }`, so leaving out "type" is a nice simplification—but if that's a bad guess (or we want to be more risk-averse), then let's leave in the word "type." I'll defer to your judgment here. (Either way, with my suggested description we'll have "type" be searchable.)

Whatever you do, I think the same applies to CSS.

Now, I just wanted to point out that this is a funny case. You can stop reading here because this doesn't matter. But there's a task that web developers want to do: importing JSON as modules. This has caused two features to come into existence (sort of): import assertions and import attributes. It happens that in this case one feature is the clear successor to the other. But it's easy to imagine an alternate timeline where both came into existence in parallel and we would've ended up with a group for attributes and assertions: "importing JSON."

ddbeck (approved):

PR #1247, opened 2024-06-14
New object-view-box feature

captainbrosset:


features/object-view-box.yml
```diff
@@ -0,0 +1,5 @@
+name: object-view-box
+description: The `object-view-box` CSS property defines the subset within an image that should be drawn in the content box of the image.
```
ddbeck, line 2: I found "subset … drawn in the context box" to be pretty abstract. What do you think of something like this, that gets at the result more than the method?

```suggestion
description: The `object-view-box` CSS property crops and zooms to an inset area of an image.
```

features/object-view-box.yml
```diff
@@ -0,0 +1,5 @@
...
+spec: https://drafts.csswg.org/css-images-5/#the-object-view-box
+compat_features:
+  - css.properties.object-view-box
```
ddbeck, line 5: Should the descendant key `css.properties.object-view-box.none` be a part of this? If not, why not?

captainbrosset: `none` is the initial value of the property and does nothing. I would expect any browser that supports `object-view-box` to also support the `none` value, from the get go.

I think it would be more interesting to add the `inset` and `rect` value types once they get added to BCD (I'll take a look and open a PR in BCD for it).

ddbeck: OK, that's fine. I suspect we may end up tagging `none` into this feature as well (and the other values, as they get added) as we merge the tools to consume BCD more comprehensively, but this is fine for now. Thank you!

ddbeck (approved):

PR #1252, opened 2024-06-14
Refresh dist with forks

ddbeck:
This PR is meant to improve the behavior of the dist refresh workflow, when the contributor's PR is from a fork.

This is needed because we can't (safely) trust the branch PR to modify labels on the web-features repo or assume that the PR is coming from a branch on this repository.

It's a bit convoluted, but here's how it works:

1. The first workflow starts when you label a PR with _refresh dist_. This workflow's permissions vary, depending on the author, originating repo, and the "Allow edits and access to secrets by maintainers" checkbox. In some cases, it won't have permissions to the `web-platform-dx/web-features` repo, so we can't do anything that assumes such permission.
1. It checks out contributor's repo and branch (whether that's `web-platform-dx/web-features` or a fork).
1. It refreshes the affected dist files, commits them, and pushes the change to the contributor's repo and branch.
1. It records which PR the refresh dist workflow ran against, as an artifact.
1. The completion of the previous workflow triggers (via `workflow_run`) a workflow to modify labels. This workflow has repository permissions (like `pull_request_target` would) and runs (safely) in the context of the base branch.
1. The workflow reads the artifact (the PR URL) from the workflow that triggered it, then removes the label from that PR.

Prompted by https://github.com/web-platform-dx/web-features/pull/1242.

### DIFF
--- a/.github/workflows/refresh_dist.yml
+++ b/.github/workflows/refresh_dist.yml
@@ -11,6 +11,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
 
       - id: changed-package-json
         uses: tj-actions/changed-files@v44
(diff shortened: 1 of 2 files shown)


.github/workflows/refresh_dist.yml
```diff
@@ -49,13 +52,15 @@ jobs:
...
 
-      - name: Remove label
-        uses: actions/github-script@v6
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow
+      - name: Save PR URL
+        env:
+          PR_URL: https://github.com/${{ github.repository }}/pull/${{ github.event.number }}
+        run: |
+          mkdir -p ./pr
+          echo $PR_URL > ./pr/pr_url
```
foolip, line 61: Could we use https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable here instead of an artifact?

ddbeck: Apparently not. This ultimately borrowed heavily from the [GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow). It's really strange that this is the official method.

That said, I did spend a bit of time trying to figure out if there was a less indirect way to get the PR number from the triggering workflow. [This comment](https://github.com/orgs/community/discussions/25220?sort=new#discussioncomment-8697399) was as close as I could find (from a long discussion about this curious omission in the API), but I wasn't able to get it working myself. Maybe you'd like to take a crack at it?

foolip: Ah, the data needs to cross workflows. Never mind then.

foolip (approved):

PR #1263, opened 2024-06-19
Create a `mouse-events` feature

ddbeck:
This is an old feature, but created to capture `api.MouseEvent.movement{X,Y}` that were under the `pointer-lock` feature. The properties work independently of pointer lock, so it makes sense for them to live somewhere else, with a better support history than pointer lock's.

~~This will need follow up after https://github.com/web-platform-dx/web-features/pull/1227 (or something like it) merges. See the diff for details.~~

This was prompted by https://github.com/mdn/content/issues/34263

cc: @bsmth

### DIFF
--- a/features/mouse-events.yml
+++ b/features/mouse-events.yml
@@ -1,5 +1,5 @@
 name: Mouse events
-description: Mouse events, such as `click`, `mousedown`, or `mousemove`, fire when users interact with a pointing device such a mouse or trackpad.
+description: Mouse events, such as `click`, `mousedown`, or `mousemove`, fire when users interact with an input or pointing device such as a mouse, trackpad, or touchscreen.
 spec: https://w3c.github.io/uievents/#events-mouse-types
 # TODO: With the exception of mouseenter and mouseleave, it's likely that the
 # mouse events were supported from the initial release in each browser. It'd be


features/mouse-events.yml
```diff
@@ -0,0 +1,39 @@
...
+  - api.MouseEvent.clientX
+  - api.MouseEvent.clientY
+  - api.MouseEvent.ctrlKey
+  - api.MouseEvent.metaKey
+  - api.MouseEvent.pageX
+  - api.MouseEvent.pageY
+  - api.MouseEvent.relatedTarget
+  - api.MouseEvent.screenX
+  - api.MouseEvent.screenY
+  - api.MouseEvent.shiftKey
```
bsmth, line 25: Most of these look good to me based on the spec and looks like everything predating pointer events -> https://www.w3.org/TR/2019/WD-pointerevents3-20191212/#intro

features/mouse-events.yml
```diff
@@ -0,0 +1,52 @@
+name: Mouse events
+description: Mouse events, such as `click`, `mousedown`, or `mousemove`, fire when users interact with a pointing device such a mouse or trackpad.
```
foolip, line 2: For site compat reasons the events also fire when interacting with touch screens, although in that case touch events are also fired. And pointer events. I'm not sure how much to say, WDYT?

features/mouse-events.yml.dist
```diff
@@ -0,0 +1,191 @@
+# Generated from: mouse-events.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: high
+  baseline_low_date: 2015-07-29
+  baseline_high_date: 2018-01-29
+  support:
+    chrome: "2"
```
foolip, line 9: Without doing any research I'm confident this should be the earliest release of all browsers. Mouse events are ancient, and there's no chance that they were so broken in any browser released since 2003 (Safari 1) that we should consider them not supported.

At the same time I don't want to spend any time researching the small differences here. How about just `compute_from: api.MouseEvent` and a comment/TODO?

foolip (approved):

PR #1289, opened 2024-07-01
Increment patch version to v0.9.0

foolip:


packages/web-features/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "web-features",
   "description": "Curated list of Web platform features",
-  "version": "0.8.6",
+  "version": "0.8.7",
```
ddbeck, line 4: I think the removal of `usage_stats` means this ought to be a minor version, `"0.9.0"`.

foolip: You're right, I'll update.

ddbeck (approved):

captainbrosset (approved):

PR #1290, opened 2024-07-01
`container-queries`: remove similarly-named but irrelevant compat keys

ddbeck:
Discovered thanks to https://github.com/web-platform-dx/web-features/pull/1231/ (https://github.com/web-platform-dx/web-features/pull/1231#discussion_r1660925264).


foolip:
Including these keys wasn't accidental, it's because container queries depends on size containment. However, size containment is its own feature, so it makes sense to remove these.

foolip (approved):

PR #1292, opened 2024-07-01
`compute-baseline`: Add a couple of conveniences for spec and status data

ddbeck:
It's pretty clunky to do the `feature.data.__compat` + cleanup. I'd rather not think about the BCD schema, if I can help it. 😅


foolip (approved):

Elchi3 (approved):
Yes! I was about to propose this as well. :)

PR #1294, opened 2024-07-01
Enable checkAncestors when computing Baseline status

foolip:


features/conic-gradients.yml.dist
```diff
@@ -2,13 +2,10 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 status:
-  baseline: high
-  baseline_low_date: 2020-11-17
-  baseline_high_date: 2023-05-17
+  baseline: false
```
foolip, line 5: This will be fixed by https://github.com/mdn/browser-compat-data/pull/23603.

features/clipboard-unsanitized-formats.yml.dist
```diff
@@ -4,6 +4,4 @@
 status:
   baseline: false
   support:
-    chrome: "122"
```
foolip, line 7: The cause of this is here:

https://github.com/mdn/browser-compat-data/blob/ebb58c142ff3f1dcade8857de96d16fb8e2f80f1/api/Clipboard.json#L46-L71

The notes don't actually say what isn't supported...

features/js-modules-shared-workers.yml.dist
```diff
@@ -8,7 +8,7 @@ status:
     edge: "80"
     firefox: "114"
     firefox_android: "114"
-    safari: "15"
-    safari_ios: "15"
+    safari: "16"
```
foolip, line 11: The 16 comes from https://github.com/mdn/browser-compat-data/blob/ebb58c142ff3f1dcade8857de96d16fb8e2f80f1/api/SharedWorker.json#L35.

It seems like this shouldn't pass BCD's lints, there claim that ES modules is supported in Safari 15 is at odds with the data for `SharedWorker`.

features/clipboard-unsanitized-formats.yml.dist
ddbeck: Hmm, yeah, that's iffy data. Two thoughts here:

- The partial seems to be the requirement to have the (non-standard?) `clipboard-read` permission. I can't find that permission specified anywhere, which makes it hard to understand.

- Those media types should be subfeatures of `read` instead of notes. Probably something like `text/plain` and `image/png` ought to be in `compute_from` list?

features/conic-gradients.yml.dist
ddbeck: Should we site on this, until it's closer to the release? Or should we just avoid doing a release until this is fixed upstream?

features/js-modules-shared-workers.yml.dist
ddbeck: I filed https://github.com/mdn/browser-compat-data/issues/23624

features/conic-gradients.yml.dist
foolip: Let's wait for the next BCD release before merging this PR.

features/js-modules-shared-workers.yml.dist
foolip: I see. I've tried to figure out if TypeScript can constrain strings to a specific format but don't know the answer. How about we just make `yamlEntries` throw an error if an identifier isn't in the form we like?

ddbeck (approved):
With https://github.com/web-platform-dx/web-features/pull/1332 merged, I think this is ready for a rebase and merge.

ddbeck:
OK, how about this: we turn off the ancestor check, but add an issue (here? on BCD?) saying we ought to do something about the keys we find via `find-troublesome-ancestors.ts`?

Seems weird that we should know about the problem but not do anything about it.

foolip:
Let's do this in https://github.com/web-platform-dx/web-features/pull/1485 instead.

PR #1299, opened 2024-07-02
Update and publish referrer policy feature

foolip:
Trim the BCD keys to just HTTP header and HTML attributes. Pin the
status to the same but ignoring the reflectd API since that's not key to
using this.

Note that support in <script> came a bit later but the Baseline year is
2020 regardless, so include it in compute_from.

Fetch and SVG integration is excluded, as it's not part of the feature
description.

All the entries for specific values like "no-referrer-when-downgrade"
are also removed since the data for that looks messy and would need
detailed review to have confidence in it.


features/referrer-policy.yml
```diff
@@ -0,0 +1,27 @@
...
+  - api.HTMLImageElement.referrerPolicy
+  - api.HTMLLinkElement.referrerPolicy
+  - api.HTMLScriptElement.referrerPolicy
+  - html.elements.a.referrerpolicy
+  - html.elements.area.referrerpolicy
+  - html.elements.iframe.referrerpolicy
+  - html.elements.img.referrerpolicy
+  - html.elements.link.referrerpolicy
+  - html.elements.script.referrerpolicy
+  - http.headers.Referrer-Policy
```
ddbeck, line 27: Without `http.headers.Referrer-Policy.default_strict-origin-when-cross-origin` we end up pretty far out of line with caniuse.

```suggestion
  compute_from:
    - html.elements.a.referrerpolicy
    - html.elements.area.referrerpolicy
    - html.elements.iframe.referrerpolicy
    - html.elements.img.referrerpolicy
    - html.elements.link.referrerpolicy
    - html.elements.script.referrerpolicy
    - http.headers.Referrer-Policy
    - http.headers.Referrer-Policy.default_strict-origin-when-cross-origin
compat_features:
  - api.HTMLAnchorElement.referrerPolicy
  - api.HTMLAreaElement.referrerPolicy
  - api.HTMLIFrameElement.referrerPolicy
  - api.HTMLImageElement.referrerPolicy
  - api.HTMLLinkElement.referrerPolicy
  - api.HTMLScriptElement.referrerPolicy
  - html.elements.a.referrerpolicy
  - html.elements.area.referrerpolicy
  - html.elements.iframe.referrerpolicy
  - html.elements.img.referrerpolicy
  - html.elements.link.referrerpolicy
  - html.elements.script.referrerpolicy
  - http.headers.Referrer-Policy
  - http.headers.Referrer-Policy.default_strict-origin-when-cross-origin
```

ddbeck:
> I noticed that, but thought it's strange to make the default part of the feature. After all, the default is what you get when you don't use the feature (header or attributes) at all. On the other hand, as a matter of spec and implementation, the default is part of the feature.
> 
> I'll add some comment or TODO about this.

FWIW, I agree that it's a bit weird. I think the TODO might be to ask caniuse to make the change. Aligning with caniuse is pretty important, I think, since that's often how developers know what a feature _is_, in terms of compat.

features/referrer-policy.yml.dist
```diff
@@ -0,0 +1,114 @@
+# Generated from: referrer-policy.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: high
+  baseline_low_date: 2021-09-20
+  baseline_high_date: 2024-03-20
+  support:
+    chrome: "85"
```
foolip, line 9: This aligns with https://caniuse.com/referrer-policy.

PR #1300, opened 2024-07-02
Add Web Audio API

foolip:


features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+    chrome: "35"
+    chrome_android: "35"
+    edge: "12"
+    firefox: "25"
+    firefox_android: "25"
+    safari: "14.1"
+    safari_ios: "14.5"
+compat_features:
+  # baseline: high
+  # baseline_low_date: 2015-07-29
```
foolip, line 18: @ddbeck claiming that these bits of the API have been Baseline since 2015 while the whole feature only since 2021 doesn't seem right. You can't use any of this without a prefix unless the unprefixed `AudioContext` constructor is supported.

Should `getStatus` perhaps clamp support to the support of the overall feature? The downside is that if we do that for a feature where the Baseline status (false/low/high) isn't the same for the feature and its earliest parts, that would end up being confusing on MDN, as compat tables could claim support but the Baseline badge says not Baseline, as well as getting the low/high connection to dates wrong, although that's not as noticeable.

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+  - api.BaseAudioContext.audioWorklet
+
+  # baseline: false
+  # support:
+  #   chrome: "41"
+  #   chrome_android: "41"
+  #   edge: "14"
+  #   safari: "14.1"
+  #   safari_ios: "14.5"
+  - api.OfflineAudioContext.suspend
```
foolip, line 671: This is because of https://bugzilla.mozilla.org/show_bug.cgi?id=1265406. The reason that `resume()` isn't listed here is that it's on `BaseAudioContext` in Gecko source, but should be on `AudioContext` and `OfflineAudioContext` separately per spec. I'm not sure if calling `resume()` can ever do anything.

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+
+  # baseline: false
+  # support:
+  #   chrome: "52"
+  #   chrome_android: "52"
+  #   edge: "79"
+  #   safari: "14.1"
+  #   safari_ios: "14.5"
+  - api.AudioListener.forwardX
```
foolip, line 680: This matches https://developer.mozilla.org/en-US/docs/Web/API/AudioListener#browser_compatibility and the notes make me trust the data.

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+  - api.AudioListener.upZ
+
+  # baseline: false
+  # support:
+  #   chrome: "57"
+  #   chrome_android: "57"
+  #   edge: "79"
+  #   safari: "14.1"
+  #   safari_ios: "14.5"
+  - api.AudioParam.cancelAndHoldAtTime
```
foolip, line 697: This and the next is correct, see https://github.com/mdn/browser-compat-data/pull/23628.

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+  #   firefox_android: "79"
+  - api.AudioContext.outputLatency
+
+  # baseline: false
+  # support:
+  #   chrome: "14"
+  #   edge: "12"
+  #   safari: "6"
+  #   safari_ios: "6"
+  - api.AudioParam.exponentialRampToValueAtTime
```
foolip, line 723: BCD has notes linking to bugs for these.

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+  - api.AudioParam.automationRate
+
+  # baseline: false
+  # support:
+  #   chrome: "102"
+  #   chrome_android: "102"
+  #   edge: "102"
+  #   firefox: "70"
+  #   firefox_android: "79"
+  - api.AudioContext.outputLatency
```
foolip, line 715: This is a FIXME in WebKit:
https://github.com/WebKit/WebKit/blob/312f06335038c0c43733c76223f9a53806a002c1/Source/WebCore/Modules/webaudio/AudioContext.idl#L38-L39

features/web-audio.yml.dist
```diff
@@ -0,0 +1,732 @@
...
+  #   safari: "6"
+  #   safari_ios: "6"
+  - api.AudioParam.exponentialRampToValueAtTime
+  - api.AudioParam.linearRampToValueAtTime
+
+  # baseline: false
+  # support:
+  #   firefox: "68"
+  #   firefox_android: "68"
+  - api.AudioContext.createMediaStreamTrackSource
```
foolip, line 730: This seems correct. I've sent https://github.com/mdn/browser-compat-data/pull/23629 to make it even more easily verifiable next time around.

features/web-audio.yml.dist
foolip: Turns out that `resume()` always rejects:

https://github.com/mozilla/gecko-dev/blob/bb682ab37e12bf256f4180693568bf1325aab2be/dom/media/webaudio/AudioContext.cpp#L1062-L1066

foolip: I've sent https://github.com/mdn/browser-compat-data/pull/23641, which should end up moving `resume()` into this block.

features/web-audio.yml
```diff
@@ -0,0 +1,216 @@
+name: Web Audio
+description: The Web Audio API TODO.
+spec: https://webaudio.github.io/web-audio-api/
+caniuse: audio-api
+status:
+  compute_from: api.AudioContext.AudioContext
```
captainbrosset, line 6: Oh I see, you're using the `compute_from` keyword to get the browser compat data for that key only, while still listing all (most) other compat keys below.

Sorry, I've not been following the `compute_from` work very closely. Can you remind me how we should think about this? To what end do we list `compat_features` if they don't play a role in the final status?

features/web-audio.yml
```diff
@@ -0,0 +1,216 @@
...
+  - api.AudioWorkletGlobalScope.registerProcessor
+  - api.AudioWorkletGlobalScope.sampleRate
+  - api.AudioWorkletNode
+  - api.AudioWorkletNode.AudioWorkletNode
+  - api.AudioWorkletNode.parameters
+  - api.AudioWorkletNode.port
+  - api.AudioWorkletNode.processorerror_event
+  - api.AudioWorkletProcessor
+  - api.AudioWorkletProcessor.AudioWorkletProcessor
+  - api.AudioWorkletProcessor.port
```
captainbrosset, line 101: I would tend to get rid of this and perhaps add it as a separate feature. To me, being able to run webaudio graphs in a worklet is an nice-to-have optimization technique that doesn't prevent using the API at all, and should be seen as a later addition.

features/web-audio.yml
```diff
@@ -0,0 +1,216 @@
...
+  - api.BaseAudioContext.createPeriodicWave
+  - api.BaseAudioContext.createStereoPanner
+  - api.BaseAudioContext.createWaveShaper
+  - api.BaseAudioContext.currentTime
+  - api.BaseAudioContext.decodeAudioData
+  - api.BaseAudioContext.destination
+  - api.BaseAudioContext.listener
+  - api.BaseAudioContext.sampleRate
+  - api.BaseAudioContext.state
+  - api.BaseAudioContext.statechange_event
```
captainbrosset, line 127: Same for the BaseAudioContext. My understanding is that it was introduced for the benefit of the OfflineAudioContext interface, which is a later addition too, and doesn't prevent using the API.

features/web-audio.yml
```diff
@@ -0,0 +1,216 @@
...
+  - api.MediaStreamTrackAudioSourceNode
+  - api.MediaStreamTrackAudioSourceNode.MediaStreamTrackAudioSourceNode
+  - api.OfflineAudioCompletionEvent
+  - api.OfflineAudioCompletionEvent.OfflineAudioCompletionEvent
+  - api.OfflineAudioCompletionEvent.renderedBuffer
+  - api.OfflineAudioContext
+  - api.OfflineAudioContext.OfflineAudioContext
+  - api.OfflineAudioContext.complete_event
+  - api.OfflineAudioContext.length
+  - api.OfflineAudioContext.startRendering
```
captainbrosset, line 182: See my comment about the BaseContext.

features/web-audio.yml
foolip: The purpose of listing many BCD keys is to provide a (reviewed) Baseline status for individual BCD keys so that that status can be shown on the corresponding MDN page.

The way to think about features now is as more "living" and evolving and not set in stone once created. Small additions to a feature can be added without it affecting the feature's overall status.

This will require more judgment calls about what to include and exclude, as the resulting status is no longer a guardrail.

For this feature, I do wonder if we should split out any of the newer parts, in particular the worklet caught my eye. The nice thing is that we don't need to make final decisions, we can break some things out into separate feedback in response to feedback.

features/web-audio.yml.dist
ddbeck: > Should `getStatus` perhaps clamp support to the support of the overall feature? The downside is that if we do that for a feature where the Baseline status (false/low/high) isn't the same for the feature and its earliest parts, that would end up being confusing on MDN, as compat tables could claim support but the Baseline badge says not Baseline, as well as getting the low/high connection to dates wrong, although that's not as noticeable.

So I'm guessing there's four scenarios, for any given key:

1. It's newer than the overall feature (by baseline low date). You _do_ want to show the newer status.
2. It's the same as the overall feature (by baseline low date, even if the specific versions differ). No conflict.
3. It's older than the overall feature. It works in isolation, so you want to show the older status.
4. It's older than the overall feature. It depends on another part of the feature, so you want to show the newer status.

It's scenario 4 that's the danger here, right? If that's much more common than scenario 3, we should just clamp the status. If scenario 3 is much more common, then that's probably not a good idea.

Brainstorming here, there are a few ways to deal with that situation. Most are on the consumer side:

- If it's in scenario 3 or 4, change the appearance on MDN to say, "This is part of another feature, with a different status." (I'd like to see the "this is part of…" thing in the widget, but that's an even more general matter.)
- If it's in scenario 3 or 4, make the page author decide which status to show. (I prefer this! These are the people who ought to know best! But I don't think the MDN folks are hyped about this idea, though I could be wrong.)

The only part I think we can do on the authoring side would be to do the work of declaring some keys as being tightly coupled, to opt-in to clamping just those keys. That'd still leave apparent conflicts on MDN, but they'd be good scenario 4 conflicts, where Baseline tells a _more_ correct story than the compat table.

So you might have something like this:

```yaml
coupled_compat_features:
  - [api.AnalyserNode, api.AudioContext]
```

That would say, "for these keys specifically, clamp to the worst availability of them, otherwise, proceed normally."

If this is rare, I'd hate to have to proactively research every feature for this. But it'd be an escape hatch for "broken" features, as they get reported.

(I suppose one other option would exist, which is maybe more correct but harder: do this annotation in BCD.)

features/web-audio.yml
captainbrosset: Makes sense, thank you.
I would definitely break out the Worklet stuff now. Looking at the Chrome and Firefox implementations, as well as the spec changes over time, worklets didn't exist at first, around 2013, and only really became a thing in 2018.

I know this is ancient history, that both AudioContext and AudioWorklet are baseline hight, and so none of this might matter anyway, but to be more truthful to what happened, I think it would make sense to break a feature called AudioWorklet out of this one.

features/web-audio.yml
foolip: I've split out worklets.

features/web-audio.yml
foolip: Most of what's on `BaseAudioContext` is part of the origin API accessed through the `AudioContext` constructor. `AudioContext` was made to inherit from the new `BaseAudioContext` and a lot of things were moved, but it was a backwards compatible change. This note in BCD is about this:
https://github.com/mdn/browser-compat-data/blob/8e86a205cfde203fc91f5548c9ffe142ede2270d/api/BaseAudioContext.json#L16

It also looks like the status of the `BaseAudioContext` members here ends up being clamped to `BaseAudioContext`, despite that being worked around in BCD with partial implementation. I'll file an issue.

captainbrosset: Makes sense. Let's keep this in.

foolip: I've filed https://github.com/web-platform-dx/web-features/issues/1314 now.

features/web-audio.yml
```diff
@@ -0,0 +1,216 @@
+name: Web Audio
+description: The Web Audio API TODO.
```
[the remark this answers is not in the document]
foolip: Thank you @captainbrosset!

features/web-audio.yml
```diff
@@ -0,0 +1,190 @@
+name: Web Audio
```
foolip, line 1: Should the name be "Web Audio" with Audio capitalized? "Web audio" looks wrong to me because it's not about audio generically, but this specific API. "Web Audio API" is the spec name.

features/web-audio.yml
```diff
@@ -0,0 +1,190 @@
...
+description: The Web Audio API creates graphs of audio nodes used to handle sound inputs, synthesize sounds, apply effects, create visualizations, and output to audio devices.
```
foolip, line 2: Should we name `AudioContext` in the description? Usually we do that early in the description, but saying "Also known as Web Audio API" at the end in this case would be embarrassing, since that's how the feature is usually referred to.

features/web-audio.yml
captainbrosset: Web Audio feels much better. That's how everybody talks about it.

features/web-audio.yml
foolip: OK, let's leave it out for now.

features/web-audio.yml.dist
foolip: @ddbeck those four scenarios look correct, although I would really consider any differences in status, not just differences in Baseline low date. I'd like a way to override data that don't make sense together, even if it's only at the level of browser versions.

And yes, scenario 4 is the issue here. It should be common for prefixed features, especially APIs with multiple interfaces but where the prefix is in a single place, the entrypoint.

For scenario 3, I'm struggling to think of when we'd decide to author a feature like this. I can imagine it in theory, that some marginally useful part of a feature shipped first, followed by the real thing that was motivating it to begin with. So perhaps for `Set` this would happen if we decide that `size` is an essential part of the feature, leaving the initial parts in Firefox under scenario 3.

I wonder if a variation of `use_ranges_before_baseline_low_date` could be revived for this, perhaps `use_ranges_before_initial_support` that would use ranges for everything before the browser version in the overall feature.

An important part of this is to cut down on the amount of research we need to do, to move faster.

ddbeck: > I wonder if a variation of `use_ranges_before_baseline_low_date` could be revived for this, perhaps `use_ranges_before_initial_support` that would use ranges for everything before the browser version in the overall feature.

I like this idea. We should try it, once we're surfacing ranges.

Heck, we could try clamping _anything_ pre-`compute_from` as `≤` with respect to the Baseline low date and versions, by default. Then, for those cases where things might be useable anyway (`Set` without `size`), we opt into unclamping?

Filing this as https://github.com/web-platform-dx/web-features/issues/1335.

foolip:
I think this is OK to land as-is.

features/web-audio.yml
ddbeck: Sorry, I'm late to this with a style nits (avoiding "used to").

```suggestion
description: The Web Audio API creates graphs of audio nodes that handle sound inputs, synthesize sounds, apply effects, create visualizations, and output to audio devices.
```

ddbeck (approved):
I'm happy with this, though please fix the style nit I noted in a comment. Thank you!

PR #1302, opened 2024-07-02
Moving the <time> feature from drafts

tropicadri:
I reviewed the <time> element file. The baseline date seems recent, but looking through Chrome status, it was added in [Chrome 62](https://chromestatus.com/feature/5633937149788160)


foolip:
@tropicadri thanks for moving this along! Can you rebase or merge to resolve the conflict in time.yml.dist? Whatever the conflict is doesn't really matter, just make sure features/draft/html-elements/time.yml.dist doesn't exist, and run `npm run dist features/time.yml` so that the proper dist files does.

Here's what it looks like for me locally:

```yaml
status:
  baseline: high
  baseline_low_date: 2018-10-02
  baseline_high_date: 2021-04-02
  support:
    chrome: "62"
    chrome_android: "62"
    edge: "18"
    firefox: "22"
    firefox_android: "22"
    safari: "10"
    safari_ios: "10"
compat_features:
  # baseline: high
  # baseline_low_date: 2017-10-24
  # baseline_high_date: 2020-04-24
  # support:
  #   chrome: "62"
  #   chrome_android: "62"
  #   edge: "14"
  #   firefox: "22"
  #   firefox_android: "22"
  #   safari: "10"
  #   safari_ios: "10"
  - api.HTMLTimeElement
  - api.HTMLTimeElement.dateTime

  # baseline: high
  # baseline_low_date: 2018-10-02
  # baseline_high_date: 2021-04-02
  # support:
  #   chrome: "62"
  #   chrome_android: "62"
  #   edge: "18"
  #   firefox: "22"
  #   firefox_android: "22"
  #   safari: "7"
  #   safari_ios: "4"
  - html.elements.time
  - html.elements.time.datetime
```

It's surprising that the status of the API and element isn't the same, this is probably wrong somehow.

The Baseline low date 2018-10-02 is the release date of Edge 18 (found by searching in BCD) so the most important question is whether Edge 14 or 18 is correct for Edge.

I doesn't look like the Safari data will change the Baseline low date, but figuring out which is the correct data would still be good.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/23638 to align the Edge data.

https://bugs.webkit.org/show_bug.cgi?id=148840 also confirms that HTMLTimeElement came later than `<time>` in WebKit, so we can leave that data as-is I think. It's not going to define the Baseline low date anyway.

foolip:
In order to not have to wait for the next BCD release, adding this to the source YAML would work:

```yaml
# TODO: remove after https://github.com/mdn/browser-compat-data/pull/23638
status:
  compute_from: api.HTMLTimeElement
```

tropicadri:
I added the comment and regenerated the dist file.

features/time.yml
```diff
@@ -0,0 +1,12 @@
+# TODO: remove after https://github.com/mdn/browser-compat-data/pull/23638
+status:
```
foolip, line 2: No order is enforced, but I'll move this to where it is by convention in most files, closer to the end.

foolip (approved):
Looks good, thanks @tropicadri!

@ddbeck can you review the description and merge?

features/time.yml
```diff
@@ -0,0 +1,12 @@
+name: <time>
+description: The `<time>` HTML element represents a specific period in time. It may include the `datetime` attribute to translate dates into machine-readable format.
```
ddbeck, line 2: I like this! What do you think of adding some examples for the "period in time", like this?

```suggestion
description: The `<time>` HTML element represents a time, such as a calendar date, clock time, or duration. It may include the `datetime` attribute to translate dates into machine-readable format.
```

PR #1303, opened 2024-07-03
Add comments for spec draft features that are already part of web features

tropicadri:
update-drafts script now adds a comment with the features that are already part of existing web features at the end of the yml file.

webtt1.yml is an example of the current comment.

@foolip is something like this what you had in mind?


scripts/update-drafts.ts
```diff
@@ -35,9 +35,14 @@ async function main() {
   const compat = new Compat();
 
   // Build a set of used BCD keys.
```
foolip, line 37:
```suggestion
  // Build a map of used BCD keys to feature.
```

scripts/update-drafts.ts
```diff
@@ -1,8 +1,8 @@
 import { Compat } from "compute-baseline/browser-compat-data";
 import fs from "node:fs/promises";
 import { fileURLToPath } from "node:url";
+import {Document} from "yaml";
```
foolip, line 4: There are some whitespace differences to surrounding code, but that's fine. I'll add this file to the Prettier config after this which will reformat stuff for consistency.

scripts/update-drafts.ts
```diff
@@ -35,9 +35,14 @@ async function main() {
   const compat = new Compat();
 
   // Build a set of used BCD keys.
-  const usedFeatures = new Set<string>(
-    Object.values(features).flatMap((data) => data.compat_features)
-  );
+  const webFeatures = new Map<string,string>();
```
foolip, line 38: I'd like some other name than `webFeatures` since that sounds like the project as a whole and I'd guess they keys are web-features identifiers. This is a reversed map of web-features. How about `compatFeatureToName`, also making it more obvious that it's the feature name, not ID, being used here.

scripts/update-drafts.ts
```diff
@@ -35,9 +35,14 @@ async function main() {
...
+  Object.values(features).map((data) => {
+    if(data.compat_features){
+    for(const compatFeature of data.compat_features){
+      webFeatures.set(compatFeature, data.name);
```
foolip, line 42: Anticipating that we might solve https://github.com/web-platform-dx/web-features/issues/1173, can you throw an error here if there is overlap?

```js
if (compatFeatureToName.has(compatFeature)) {
  throw new Error(`${compatFeature} seen in multiple features`);
}
```

or similar.

scripts/update-drafts.ts
```diff
@@ -50,17 +55,20 @@ async function main() {
   // Iterate BCD and group compat features by spec.
   const specToCompatFeatures = new Map<object, Set<string>>();
   for (const feature of compat.walk()) {
-    // Skip any BCD keys already used in web-features.
-    if (usedFeatures.has(feature.id)) {
+
+    // Skip deprecated and non-standard features.
+    const status = feature.data.__compat.status;
```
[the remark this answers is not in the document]
tropicadri: I did get this code from some other PR because the check before was skipping all of the features, so would it be ok to delete all this check knowing that it will be added afterwards?

tropicadri:
Other than the potential conflict, I updated the PR with the review comments.

foolip (approved):
I tweaked the formatting a bit and will land this now. It works as I had hoped, thank you @tropicadri!

scripts/update-drafts.ts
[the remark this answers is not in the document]
foolip: I wasn't sure what would happen without trying it myself, so I tried simplifying this again like in https://github.com/web-platform-dx/web-features/pull/1292 and confirmed it produces the same resulting draft features written to disk. I'll commit that change.

PR #1304, opened 2024-07-03
Adding a few new Sensor features

captainbrosset:
Took a stab at covering Sensor features. Let me know if this organization makes sense.


features/accelerometer.yml
```diff
@@ -0,0 +1,7 @@
+name: Accelerometer
+description: The `Accelerometer` and `LinearAccelerationSensor` APIs read the acceleration applied to a device in three dimensions, either including the effect of gravity or not, respectively.
+spec: https://w3c.github.io/accelerometer/#accelerometer-interface
+group: sensors
+compat_features:
+  - api.Accelerometer
```
foolip, line 6: Here and elsewhere, can you include the interface members too? My starting point for a list would be `npm run traverse | grep Accelerometer` and similar.

groups/sensors.yml
```diff
@@ -0,0 +1 @@
+name: Sensors
```
foolip, line 1: Can you add a comment to say if the intended scope of the group is only the APIs built on https://w3c.github.io/sensors/, or also older features that are also exposing sensor data, like device motion events?

Luis-2305 (approved):

groups/sensors.yml
captainbrosset: No strong opinion on this. I would tend to make this group pretty generic and accommodate all types of features that expose sensor data. We could then perhaps create subgroups for new generic sensor API-based things, and older parts.

features/orientation.yml
```diff
@@ -0,0 +1,12 @@
+name: Orientation Sensor
```
foolip, line 1:
```suggestion
name: Orientation sensor
```

groups/sensors.yml
foolip: OK, let's put `device-orientation-events` into this group too then.

features/accelerometer.yml
foolip: The constructor is missing. I'll add it.

foolip:
Closing and reopening, see https://github.com/web-platform-dx/web-features/issues/1320.

foolip (approved):

PR #1305, opened 2024-07-03
New Web Audio API initial feature

captainbrosset:
To create this feature, I started by adding all of the Web Audio API-related interfaces that are listed on MDN as compat_features in the yaml file. Then, I removed the ones that seemed like later additions, or like additional nice-to-have features, one by one, until I somewhat matched [Can I Use](https://caniuse.com/audio-api).

I can remove more compat_features if you think that's better. In fact, just having api.AudioContext would be enough.

Here is the list of compat_features I removed, and why:

* api.AudioSinkInfo because it's experimental.
* api.AudioProcessingEvent because it's deprecated.
* All the api.AudioWorklet* keys because handling audio in a separate worklet thread is a nice to have later addition.
* api.AudioScheduledSourceNode and api.ConstantSourceNode because you can still use the API without them, they're convenient in certain scenarios.
* api.BaseAudioContext because it's only useful for OfflineAudioContext, which is a later addition (I also removed the api.OfflineAudio* entries for that reason).
* api.IIRFilterNode because it's a later addition to do tone control and graphic equalizers.
* api.StereoPannerNode, also not vital to start using the API.
* api.MediaStream* because this is for WebRTC integration.

Note that Can I Use actually goes further back and shows Chrome 14-33 as supported, but with the webkit prefix. Same for Safari 6-14. I don't think covering these ranges is useful, and starting at Chrome 34 and Safari 14.1 is enough (especially given that the feature is baseline high either way).

One discrepancy is that Can I Use shows Chrome 34 as being fully supported (after the 14-33 webkit prefix range), but our data starts at 35. Not sure where this 1 release delta comes from.


captainbrosset:
Oh my  ... #1300 already covers this ... should have checked the list of PRs more carefully before.

foolip:
@captainbrosset review of https://github.com/web-platform-dx/web-features/pull/1300 would be much appreciated, I took a slightly different (and less documented) approach that gave different results. There's also a question in there about the early parts of the feature, if you want to take a look.

PR #1312, opened 2024-07-03
`compute-baseline`: add method to expand a support statement into per-release support information

ddbeck:
This is another incremental step toward handling version ranges more completely, by taking a support statement and mapping it to _every_ release that a browser has.

More specifically, this PR:

- Adds `toReleaseSupportMap()` method to `SupportStatement`s. It's very much like `supportedBy()`, except it has something to say about every version, not just the supporting versions. I've made some of `supportedBy()` reusable. Eventually, but not today, we'll rip out `supportedBy()` entirely.

- Adds an `inRange(start, end?)` method to `Releases`s. This is a not-infrequent thing to check and remembering what `>= 0` means is hard.

- Remove some never-used code relating to ranges.

- Handle bugged `version_removed` values in BCD. Basically, some values should never happen according to the schema, but are happening anyway. This won't be completely fixed until it gets fixed upstream.

The next step after this will be to reconcile multiple support statements at the feature level (i.e., filtering out "unsupported" entries when there's a competing "supported" entry) and do it across browsers (e.g., to create something like `Map<Browser, Map<Release, Supported | Unsupported | UnknownSupport>>`).

### DIFF
--- a/packages/compute-baseline/src/browser-compat-data/release.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/release.test.ts
@@ -108,10 +108,19 @@ describe("Release", function () {
 
     it("handles closed ranges", function () {
       const cr = browser("chrome");
+
+      // Start of range is inclusive
       assert.equal(
         cr.version("1").inRange(cr.version("1"), cr.version("125")),
         true,
       );
+
+      // End of range is exclusive
+      assert.equal(
+        cr.version("20").inRange(cr.version("1"), cr.version("20")),
+        false,
+      );
+
       assert.equal(
         cr.version("1").inRange(cr.version("10"), cr.version("15")),
         false,
--- a/packages/compute-baseline/src/browser-compat-data/release.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/release.test.ts
@@ -87,4 +87,30 @@ describe("Release", function () {
       assert.equal(safariPreview.isPrerelease(), true);
     });
   });
+
+  describe("inRange()", function () {
+    it("handles closed ranges", function () {
+      const cr = browser("chrome");
+      assert.equal(
+        cr.version("1").inRange(cr.version("1"), cr.version("125")),
+        true,
+      );
+      assert.equal(
+        cr.version("1").inRange(cr.version("10"), cr.version("15")),
+        false,
+      );
+      assert.equal(
+        cr.version("100").inRange(cr.version("10"), cr.version("15")),
+        false,
+      );
+    });
+
+    it("handles open ranges", function () {
+      const cr = browser("chrome");
+      assert.equal(cr.version("1").inRange(cr.version("1")), true);
+      assert.equal(cr.version("1").inRange(cr.version("10")), false);
+      assert.equal(cr.version("100").inRange(cr.version("10")), true);
+      assert.equal(cr.version("preview").inRange(cr.version("10")), true);
+    });
+  });
 });
--- a/packages/compute-baseline/src/browser-compat-data/release.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/release.test.ts
@@ -40,6 +40,16 @@ describe("Release", function () {
   });
 
   describe("compare()", function () {
+    it("throws when comparing between two browsers", function () {
+      const cr = browser("chrome");
+      const ed = browser("edge");
+
+      assert.throws(
+        () => cr.version("100").inRange(ed.version("79"), cr.version("125")),
+        Error,
+      );
+    });
+
     it("returns 0 for equivalent releases", function () {
       const chrome100 = browser("chrome").version("100");
       assert.equal(chrome100.compare(chrome100), 0);
@@ -89,6 +99,13 @@ describe("Release", function () {
   });
 
   describe("inRange()", function () {
+    it("throws when comparing between two browsers", function () {
+      const cr = browser("chrome");
+      const fx = browser("firefox");
+
+      assert.throws(() => cr.version("50").inRange(fx.version("50")), Error);
+    });
+
     it("handles closed ranges", function () {
       const cr = browser("chrome");
       assert.equal(
--- a/packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
@@ -96,9 +96,17 @@ describe("statements", function () {
         assert.equal(s.version_removed, "2");
       });
 
-      it("returns false for undefined", function () {
-        const s = new SupportStatement({ version_added: "1" });
+      it("returns false", function () {
+        const s = new SupportStatement({
+          version_added: "1",
+          version_removed: false,
+        });
         assert.equal(s.version_removed, false);
+      });
+
+      it("returns undefined", function () {
+        const s = new SupportStatement({ version_added: "1" });
+        assert.equal(s.version_removed, undefined);
       });
     });
 
--- a/packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
@@ -109,105 +109,6 @@ describe("statements", function () {
         assert.equal(s.version_removed, undefined);
       });
     });
-
-    describe("toReleaseSupportMap()", function () {
-      it("expands false values to unsupported", function () {
-        const cr = browser("chrome");
-        const statement = new SupportStatement({ version_added: false }, cr);
-        const supportMap = statement.toReleaseSupportMap();
-
-        assert.equal(supportMap.size, cr.releases.length);
-        for (const { supported } of supportMap.values()) {
-          assert.equal(supported, false);
-        }
-      });
-
-      it("expands null values to unknown support", function () {
-        const cr = browser("chrome");
-        const statement = new SupportStatement({ version_added: null }, cr);
-        const supportMap = statement.toReleaseSupportMap();
-
-        assert.equal(supportMap.size, cr.releases.length);
-        for (const { supported } of supportMap.values()) {
-          assert.equal(supported, null);
-        }
-      });
-
-      it("expands true values to supported in current and future releases", function () {
-        const cr = browser("chrome");
-        const statement = new SupportStatement({ version_added: true }, cr);
-        const supportMap = statement.toReleaseSupportMap();
-
-        assert.equal(supportMap.size, cr.releases.length);
-        assert.equal(supportMap.get(cr.current())?.supported, true);
-        assert.equal(
-          supportMap.get(cr.releases.at(-1) as any)?.supported,
-          true,
-        );
-      });
-
-      it("expands open-ended statements to (unsupported, …, supported, …)", function () {
-        const cr = browser("chrome");
-        const statement = new SupportStatement({ version_added: "100" }, cr);
-        const supportMap = statement.toReleaseSupportMap();
-        const entries = [...supportMap.entries()];
-
-        assert.equal(supportMap.size, entries.length);
-        assert.equal(supportMap.get(cr.version("1"))?.supported, false);
-        assert.equal(supportMap.get(cr.version("99"))?.supported, false);
-        assert.equal(supportMap.get(cr.version("100"))?.supported, true);
-        assert.equal(supportMap.get(cr.version("101"))?.supported, true);
-        assert.equal(supportMap.get(cr.current())?.supported, true);
-        assert.equal(
-          supportMap.get(cr.releases.at(-1) as any)?.supported,
-          true,
-        );
-      });
-
-      it("expands ranged open-ended statements to (unknown, …, supported, …)", function () {
-        const cr = browser("chrome");
-        const statement = new SupportStatement({ version_added: "≤100" }, cr);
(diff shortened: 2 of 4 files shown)


packages/compute-baseline/src/browser-compat-data/release.test.ts
```diff
@@ -87,4 +87,30 @@ describe("Release", function () {
       assert.equal(safariPreview.isPrerelease(), true);
     });
   });
+
+  describe("inRange()", function () {
```
foolip, line 91: Since the API allows for comparing versions from different browsers, can you add a test showing what it does? Throws an error?

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+    return value;
+  }
+
+  /**
+   * Expand this support statement into a `Map` from `Release` objects to objects
+   * describing whether the release is supporting, unsupporting, or unknown.
+   */
+  toReleaseSupportMap(): ReleaseSupportMap {
+    if (this.browser === undefined) {
+      throw Error("This support statement's browser is unknown.");
```
foolip, line 124:
```suggestion
      throw new Error("This support statement's browser is unknown.");
```

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -140,17 +205,14 @@ export class RealSupportStatement extends SupportStatement {
...
+  get version_removed(): string | false | undefined {
+    return super.version_removed as string | false | undefined;
   }
 
-  // TODO: `supportedBy()` ought to be (partially) implemented on non-real value
-  // support statements. For example, `"version_added": true` should allow for
-  // returning `[this.browser.current()]` at least.
   supportedBy(): { release: Release; qualifications?: Qualifications }[] {
     if (this.browser === undefined) {
       throw Error("This support statement's browser is unknown.");
```
foolip, line 218:
```suggestion
      throw new Error("This support statement's browser is unknown.");
```

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+  /**
+   * Expand this support statement into a `Map` from `Release` objects to objects
+   * describing whether the release is supporting, unsupporting, or unknown.
+   */
+  toReleaseSupportMap(): ReleaseSupportMap {
+    if (this.browser === undefined) {
+      throw Error("This support statement's browser is unknown.");
+    }
+
+    if (this.version_added === false || this.version_removed === true) {
```
foolip, line 127: If `version_removed` is true, support is unknown for everything except the `version_added` version.

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+      );
+    }
+
+    const result = new Map();
+
+    let start: Release;
+    let startRanged = false;
+    if (this.version_added === true) {
+      startRanged = true;
+      start = this.browser.current();
```
foolip, line 145: This is correct, but if this code is actually used, our dist files will keep changing. Should we instead refuse to work with true values and wait for them to be fixed in BCD?

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+    }
+
+    return value;
+  }
+
+  /**
+   * Expand this support statement into a `Map` from `Release` objects to objects
+   * describing whether the release is supporting, unsupporting, or unknown.
+   */
+  toReleaseSupportMap(): ReleaseSupportMap {
```
foolip, line 122: It looks like this expands a single support statement (not an array) into a map. That means the resulting map will always be simple, typically with an initial run of false values followed by a run or true values. Sometimes unknown support before a point, but never flip-flopping. Is a map really necessary to represent this, or should we just parse `version_added` and `version_removed` into a form that makes it easy to get support for a specific release?

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+    } else if (this.version_added.startsWith("≤")) {
+      startRanged = true;
+      start = this.browser.version(this.version_added.slice(1));
+    } else {
+      start = this.browser.version(this.version_added);
+    }
+
+    let end: Release | undefined;
+    if (this.version_removed) {
+      end = this.browser.version(this.version_removed);
```
foolip, line 155: What about ranges here?

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -107,9 +90,91 @@ export class SupportStatement {
...
+      } else if (r.inRange(start, end)) {
+        // Supported without qualification
+        result.set(r, { supported: true });
+      } else if (startRanged && !r.inRange(start)) {
+        // Support unknown (before a ≤ version)
+        result.set(r, { supported: null });
+      } else {
+        // Unsupported (outside a hard range)
+        result.set(r, { supported: false });
+      }
```
foolip, line 174: How about putting this into a method to get support for a specific release? That's cheaper when the caller doesn't want to know about all releases.

Luis-2305 (approved):

packages/compute-baseline/src/browser-compat-data/release.test.ts
ddbeck: Done. There's now a test showing that this throws when you compare across browsers.

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -140,17 +138,113 @@ export class RealSupportStatement extends SupportStatement {
...
-    return super.version_removed as string | false;
+  get version_removed(): string | false | undefined {
+    return super.version_removed as string | false | undefined;
+  }
+
+  /**
+   * Find out whether this support statement says a given browser release is
+   * supported (with or without qualifications), unsupported, or unknown.
+   */
+  supportedIn(release: Release): Supported | Unsupported | UnknownSupport {
```
ddbeck, line 153: @foolip I've refactored this as a method that just checks whether one release is supported. If one still wants to know the situation for every release, then that's a one-liner away:

`new Map(browser.releases.map(r => [r, statement.supportedIn(r)]))`

Also, doing this for any arbitrary support statement, while possible, was getting very tricky and hard to follow. Since _we_ don't need this I decided it wasn't worth the effort, so it's only available on `RealSupportStatement`s.

(I've kept some test cases and work from the all-values version, but it's not urgent, so I'll re-introduce that when there's an actual need or a spare couple of hours for it.)

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
ddbeck: This is a problem in a bunch of places. I'll open a follow-up PR to fix this across the board.

packages/compute-baseline/src/browser-compat-data/release.test.ts
```diff
@@ -87,4 +97,37 @@ describe("Release", function () {
...
+    it("throws when comparing between two browsers", function () {
+      const cr = browser("chrome");
+      const fx = browser("firefox");
+
+      assert.throws(() => cr.version("50").inRange(fx.version("50")), Error);
+    });
+
+    it("handles closed ranges", function () {
```
foolip, line 109: Can you add a test for when the version equals the end version, to show the range is exclusive of the end point?

packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
```diff
@@ -186,5 +195,147 @@ describe("statements", function () {
...
+        assert.equal(unranged.supportedIn(cr.version("99")).supported, false);
+        assert.equal(unranged.supportedIn(cr.version("100")).supported, true);
+        assert.equal(unranged.supportedIn(cr.version("101")).supported, true);
+        assert.equal(unranged.supportedIn(cr.version("124")).supported, true);
+        assert.equal(unranged.supportedIn(cr.version("125")).supported, false);
+
+        assert.equal(ranged.supportedIn(cr.version("99")).supported, null);
+        assert.equal(ranged.supportedIn(cr.version("100")).supported, true);
+        assert.equal(ranged.supportedIn(cr.version("101")).supported, true);
+        assert.equal(ranged.supportedIn(cr.version("124")).supported, true);
```
foolip, line 260: Why not the 125 assert here, so it's the same asserts for unranged and ranged? (You might remove 101 if you want to reduce the number of rows.)

packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
```diff
@@ -186,5 +195,147 @@ describe("statements", function () {
...
+
+      it("returns unknown support when release is after version_added and before ranged version_removed", function () {
+        const cr = browser("chrome");
+        const rangedEnd = new RealSupportStatement(
+          { version_added: "100", version_removed: "≤125" },
+          cr,
+        );
+
+        assert.equal(rangedEnd.supportedIn(cr.version("100")).supported, true);
+        assert.equal(rangedEnd.supportedIn(cr.version("124")).supported, null);
```
foolip, line 289: Can you also test 101 here to make it clear that there's just a single release we're sure about?

packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
```diff
@@ -186,5 +195,147 @@ describe("statements", function () {
...
+        const cr = browser("chrome");
+        const statement = new RealSupportStatement(
+          { version_added: false },
+          cr,
+        );
+
+        const allReleases = cr.releases.map((r) => statement.supportedIn(r));
+        for (const { supported } of allReleases) {
+          assert.equal(supported, false);
+        }
```
foolip, line 303: This could be simplified:

```suggestion
        for (const release of cr.releases) {
          assert.equal(statement.supportedIn(release).supported, false);
        }
```

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -198,3 +278,30 @@ export class RealSupportStatement extends SupportStatement {
...
+  if (statement.flags.length) {
+    qualifications.flags = statement.flags;
+  }
+  if (statement.partial_implementation) {
+    qualifications.partial_implementation = statement.partial_implementation;
+  }
+  return qualifications;
+}
+
+function isRangedVersion(s: any): s is string {
```
foolip, line 301: I've only seen this syntax once before, but does `s is string` really do the right thing if this method returns false? Won't TypeScript infer that the argument wasn't a string even though it may have been?

ddbeck: Yes, this really does the right thing. Type predicates are (by themselves) one-way: if the function returns `true`, then the type is satisfied. The inverse (`false` → s is not string) is not inferred. [Demo](https://www.typescriptlang.org/play/?#code/DYUwLgBGIM5gXBArgOwNYoPYHcUQLwQBEmmAFgDQRwBOAligOYCeRA3AFCiQqIpIBbAEYgaBCAEYATAGYALJ27UwNRLQaNxRdUzqj2EDhwBmqAMZg6mPHRgBlFRuYAKGIlQYcKAJRqIt5XomCABvDggIiBpwJBo8MGYABxBMY2oCfEJtRyYiCAAyfOoAOgYzYCQAE1hnbKCWIm9OAF8jOjTnBOTUqFhITKydRkbQ8Mj2iE6+jMIfUciFiAB6JYhRGkxVCCwITAA3UWAAQ0SIETBsEBA8IYgjlErtwREaMYjW1o4J51sHepdoHBvN55uMOoD+pltiCwotIis1jQNlsdvtDiczuBLtdAho7g8nsJRG8IB81sAYCBQRFvhCZoEYSSFgidutNpizEckJTenB-DBlHRgMAIAByDxYXCikkfDhAA).

AFAICT, even in general TypeScript doesn't actually track the list of types something _isn't_, just the set of remaining possible types.

packages/compute-baseline/src/browser-compat-data/release.test.ts
ddbeck: Done with 7c2bbadf46f97d8ab98e20bc2b23202461cde0cf.

packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
ddbeck: Done with 4e0110b52d7ae60a9ced2e0c19f54ea58b8fa0cb.

packages/compute-baseline/src/browser-compat-data/supportStatements.test.ts
ddbeck: Done with 4e0110b52d7ae60a9ced2e0c19f54ea58b8fa0cb.

foolip (approved):

PR #1313, opened 2024-07-04
Add later features for Web Audio

foolip:


features/offline-audio-context.yml
```diff
@@ -0,0 +1,15 @@
+name: OfflineAudioContext
+description: The `OfflineAudioContext` API TODO.
```
ddbeck, line 2:
```suggestion
description: The `OfflineAudioContext` API represents an audio processing graph that writes to an `AudioBuffer` instead of an output device.
```

ddbeck (commented):
Suggesting some descriptions, as someone who has never used this API.

features/audio-worklet.yml
```diff
@@ -0,0 +1,19 @@
+name: AudioWorklet
+description: The `AudioWorklet` API TODO.
```
ddbeck, line 2:
```suggestion
description: The `AudioWorklet` API runs module code in a separate thread, specifically for non-blocking, low latency audio processing.
```

ddbeck:
@foolip We're still missing descriptions and my suggestions were marked as resolved. Is there a missing commit here, maybe?

foolip:
@ddbeck weird, I don't know what happened. I'll fix it up.

ddbeck (approved):

Issue #1249, opened 2024-06-14
Async Clipboard API is "all red"


captainbrosset:
Reported by @tomayac here: https://toot.cafe/@tomayac/112614061442371457

See the status: https://web-platform-dx.github.io/web-features-explorer/features/async-clipboard/

Compare to Can I Use: https://caniuse.com/async-clipboard

Here are the BCD keys used to compute our status:

- api.Clipboard
- api.Clipboard.read
- api.Clipboard.readText
- api.Clipboard.write
- api.Clipboard.writeText
- api.ClipboardEvent
- api.ClipboardEvent.ClipboardEvent
- api.ClipboardEvent.clipboardData
- api.ClipboardItem
- api.ClipboardItem.ClipboardItem
- api.ClipboardItem.getType
- api.ClipboardItem.presentationStyle
- api.ClipboardItem.types
- api.Navigator.clipboard
- api.Permissions.permission_clipboard-read

ddbeck:
The overall non-support is due to BCD reporting that:

- Firefox and Safari don't support the `clipboard-read` permission (`api.Permissions.permission_clipboard-read`).
- Chrome does not support `ClipboardItem`'s `presentationStyle` property and only partially supports `read` and `readText` methods (`api.Clipboard.read*`).

It's unclear to me whether BCD, caniuse, or both have errors. BCD and caniuse differ significantly. Major differences are:

- `readText`: BCD says this is partially supported in Chrome, caniuse says Firefox doesn't support it.
- `writeText`: BCD says this is fully supported in Safari for iOS, but caniuse says it's unsupported.

Resolving this would be a three step process:

- [ ] Test and (if applicable) correct the data in BCD.

  The following issues are open on BCD for clipboard data:

  - https://github.com/mdn/browser-compat-data/issues/22096
  - https://github.com/mdn/browser-compat-data/issues/18140
  - https://github.com/mdn/browser-compat-data/issues/7991
  
  I'd expect that resolving them would go a long way toward resolving this issue in general.

- [ ] If this feature is still truly partially supported in any of the core browsers, then this is a web-features bug and it's blocked by https://github.com/web-platform-dx/web-features/issues/915. When that's done, mark this as partially supported and explain why. If there are no partials, you can skip this step.

- [ ] Check if there are any remaining discrepancies with caniuse. If so, file an issue (or PR) against to caniuse to resolve those differences.

foolip:
[...]

PR #1315, opened 2024-07-04
Pin status of async clipboard to navigator.clipboard

foolip:


foolip:
This leaves `navigator.clipboard.read()` and `readText()` as not Baseline, and that needs fixing in BCD.

ddbeck (approved):

ddbeck:
Should this actually fix https://github.com/web-platform-dx/web-features/issues/1249? There's still some question about the data quality, I think.

foolip:
No, it doesn't fully fix https://github.com/web-platform-dx/web-features/issues/1249, I'll unlink it.

foolip:
Holding off merging. https://caniuse.com/async-clipboard treats Firefox and Safari iOS as having partial support, and the Chrome versions aren't the same.

foolip:
https://github.com/web-platform-dx/web-features/pull/1346 results in some changes here...

foolip:
Let's go with https://github.com/web-platform-dx/web-features/pull/1349 instead.

PR #1319, opened 2024-07-05
New URLPattern feature

captainbrosset:


features/urlpattern.yml.dist
```diff
@@ -0,0 +1,41 @@
...
+  - api.URLPattern.search
+  - api.URLPattern.test
+  - api.URLPattern.username
+
+  # baseline: false
+  # support:
+  #   chrome: "107"
+  #   chrome_android: "107"
+  #   edge: "107"
+  - api.URLPattern.URLPattern.ignoreCase_option
```
captainbrosset, line 34: Later addition (2022): https://chromestatus.com/feature/5206436850696192

features/urlpattern.yml.dist
```diff
@@ -0,0 +1,41 @@
...
+
+  # baseline: false
+  # support:
+  #   chrome: "122"
+  #   chrome_android: "122"
+  #   edge: "122"
+  - api.URLPattern.hasRegExpGroups
```
captainbrosset, line 41: Later addition (2023): https://chromestatus.com/feature/4761338127843328

features/urlpattern.yml
```diff
@@ -0,0 +1,19 @@
+name: URLPattern
+description: The `URLPattern` API creates patterns that can be matched against URLs or URL components.
+spec: https://urlpattern.spec.whatwg.org/
+status:
+  compute_from: api.URLPattern
```
foolip, line 5: For `compute_from` I would recommend the entrypoint to the feature, the thing you have to use in order to use the feature at all. In this case it's the `URLPattern` constructor:

```suggestion
  compute_from: api.URLPattern.URLPattern
```

I see that's also missing from `compat_features`. My approach is to use `npm run traverse | grep -i urlpattern` or similar, and start from that list.

captainbrosset: Yeah, I've used this here, but didn't keep the constructor. For some reason, I was under the impression that we had an old guideline about not including constructors. But, thinking about it again, I think that was only about events.

foolip: Ah, indeed, because so many event constructors were added long after the interfaces themselves.

Now that features are "living" however, we could include event constructors as well.

foolip (approved):

PR #1325, opened 2024-07-05
Use BCD tags for anchor positioning

foolip:


captainbrosset (approved):
Seems adequate.

foolip:
> Seems adequate.

Adequate is what I strive for :D

PR #1326, opened 2024-07-05
New Contact Picker feature

captainbrosset:
The reality is that the API landed without a way to get addresses, in Chrome for Android 80. Then addresses were added, and landed in Chrome for Android 84 (https://chromestatus.com/feature/6316401599971328).

I decided to list all BCD keys in just one feature, and use `compute_from` to stick the status to version 80.

We could, instead:

* Either move the address stuff into a later addition feature.
  I don't think this is necessary seeing how the two things landed at around the same general timeframe, and this timeframe is long past.

* Or remove `compute_from` and make the overall feature start with 84.
  This would make it disagree with what's written at the bottom of https://developer.mozilla.org/en-US/docs/Web/API/Contact_Picker_API


features/contact-picker.yml
```diff
@@ -0,0 +1,22 @@
+name: Contact Picker
```
foolip, line 1:
```suggestion
name: Contact picker
```

foolip (approved):
I agree with your decision to pin this to Chrome Android 80.

foolip:
CI is failing due to https://github.com/web-platform-dx/web-features/issues/1331. I'll fix it up here and suggest how we should fix it more generally in that issue.

PR #1329, opened 2024-07-05
Add a comment for empty dist files

foolip:
This is to have something to search for, and another step towards always
having a dist file.


features/avif.yml.dist
```diff
@@ -2,3 +2,5 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 {}
+
+# Nothing is generated, nothing to see here.
```
ddbeck, line 6: What do you think of linking out to some documentation (you can just stub a heading for now, if you like) to handle the whole "why is it empty?" question?

foolip: A link to https://github.com/web-platform-dx/web-features/blob/5f17741fd7c73119a8e25270ca2070d04d73cc0c/docs/CONTRIBUTING.md#feature-dist-files after it's landed, perhaps?

foolip: Although, maybe it's better to link to the docs at the top of all dist files?

captainbrosset: I agree with this. Linking to the docs on line 3 of all dist files (right after asking people not to edit the file by hand) would give them a friendly way to learn more, especially for folks who prefer to jump in the code first, and read docs later.

foolip: Cool, I've sent https://github.com/web-platform-dx/web-features/pull/1345 as a draft, to land after https://github.com/web-platform-dx/web-features/pull/1330.

If the docs mention `{}` then I don't think this PR is needed at all.

PR #1330, opened 2024-07-05
New contributor docs

### LINKED ISSUE #1057: Contributing guidelines
Someone asked me about guidelines for adding features, and I noticed that https://github.com/web-platform-dx/web-features/blob/main/docs/CONTRIBUTING.md doesn't say anything about this.

We should at least point to https://github.com/web-platform-dx/web-features/tree/main/docs#feature-guidelines and say how to pick `compat_features` and generate a .dist.yml file.

cc @ddbeck

captainbrosset:
This is an attempt to fix #1057.
Draft for now, I've not yet written everything I wanted to write. Arguably the most important section is still just a TODO.

Pinging @ddbeck and @foolip though, since they might have early feedback about the TODO at the bottom of the file.

To more reasily review, use this rendered markdown: https://github.com/web-platform-dx/web-features/blob/docs/docs/CONTRIBUTING.md


docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,147 @@
...
+
+For example, the `fetch` API feature is described in the [`features/fetch.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/fetch.yml) file.
+
+The name of the file also constitutes the unique ID of the feature. In the example above, the unique ID of the feature is `fetch`.
+
+Feature IDs must be unique as they are used in other projects to reference the feature. For example, the [browser-compat-data](https://github.com/mdn/browser-compat-data/) project references feature IDs in their compatibility data.
+
+### Feature dist files
+
+The YAML files that are used to author features are also used to generate _dist_ files. Dist files are also YAML files, but they end with the `.yml.dist` extension.
```
foolip, line 44: Say somewhere that dist is short for distribution, and it's part of the data that we publish.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,147 @@
...
+
+For example, the `fetch` feature file is `features/fetch.yml`, and its dist file is `features/fetch.yml.dist`.
+
+You **never edit a dist file** directly. Dist files are generated from the authored YAML files, and are used at build time to generate the final data bundle that's used by other projects.
```
foolip, line 48: Mention `npm run dist`?

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,147 @@
-# Contributing
+# Contributing to the web-features project
 
-Thanks for your interest in contributing to this project!
-Before you contribute, consider the following:
+Thanks for your interest in contributing to this project! Before you contribute, consider the following:
```
ddbeck, line 3: For what it's worth, I like (but will not insist on) semantic line breaks. https://sembr.org/ 😄

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,147 @@
...
+In other cases, you might want to help but don't have a specific feature in mind. That's okay too. You can find inspiration in the following places:
+
+* [The list of issues with the **feature definition** label](https://github.com/web-platform-dx/web-features/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+definition%22).
+* [The list of draft features](https://github.com/web-platform-dx/web-features/tree/main/features/draft).
+
+A third case is if you've found a feature that's incorrectly defined. For example if its description is misleading, or you believe the browser support data is incorrect.
+
+In any case, before starting to work on a feature, make sure it isn't already being worked on by checking the list of [pull requests with the **feature definition** label](https://github.com/web-platform-dx/web-features/pulls?q=is%3Aopen+is%3Apr+label%3A%22feature+definition%22).
+
+## Contributing to the repository
```
ddbeck, line 65: Let's narrow the scope here:

```suggestion
## Contributing features to the repository
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,147 @@
...
+
+### Feature file name and format
+
+Features in this repository are authored as **YAML files** that are stored in the [`features`](https://github.com/web-platform-dx/web-features/tree/main/features) directory. Each file corresponds to a single feature, and contains metadata about the feature, such as its name, description, specification, or support data.
+
+For example, the `fetch` API feature is described in the [`features/fetch.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/fetch.yml) file.
+
+The name of the file also constitutes the unique ID of the feature. In the example above, the unique ID of the feature is `fetch`.
+
+Feature IDs must be unique as they are used in other projects to reference the feature. For example, the [browser-compat-data](https://github.com/mdn/browser-compat-data/) project references feature IDs in their compatibility data.
```
ddbeck, line 40: This would be a good place for a table or definition list with all of our fields in the YAML files. A TODO comment is fine though, if you want to focus on the narrative bits for now.

ddbeck (commented):
I think this is a really strong start—I've made a note about the one omission that hasn't already been mentioned. I'll have some nits to pick when it gets to that stage, but so far so good.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+BCD keys describe all constituents of a feature, whether they are vital to the feature, or later additions to it. For example, developers have been able to use the Web Audio API for many years even if the AudioWorklet or OfflineAudioContext APIs were added later, and only recently became baseline.
+
+It's important to ensure that the BCD keys that your feature is associated with matches how most developers use the feature.
+
+To fix data discrepancies in your dist file, open the dist file and, under `compat_features`, review each individual section.
+
+Each section corresponds to a group of BCD keys that have the same baseline and support status.
+
+_TODO: continue documenting this section (consider splitting BCD keys into different features (later addition), or using just silencing some BCD keys by using `compat_from`)._
```
captainbrosset, line 257: This is the last part that I haven't yet documented. And it also happens to be the most tricky. I could use help figuring out how far to go. I feel like a lot of the times, this will be resolved by discussions in PRs, so I'm not sure how far I should go in trying to document each and every case.

captainbrosset:
For info, the content is easier to review by looking at the rendered markdown here: https://github.com/web-platform-dx/web-features/blob/docs/docs/CONTRIBUTING.md

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+To generate your feature's dist file, once your feature is associated with one or more BCD keys:
+
+1. Run the command `npm run dist`.
+
+1. Check that a new file, named after your feature ID, and with the `.yml.dist` file extension exists, and open the file.
+
+1. Check that the data in the dist file is correct. In particular, verify that the data is consistent with what developers would find when searching for the same feature on caniuse.com or MDN.
+
+   * Check the `baseline` status. If the feature isn't yet supported in all browsers of the core browsers set, it should be `false`. If the feature is supported on all browsers, it should either be `low` or `high`, depending on how long has passed.
+   * Check the `baseline_low_date` and `baseline_high_date` fields.
```
ddbeck, line 242: What are we checking for? I'd probably say something like…

```suggestion
   * Check the `baseline_low_date`. Does it seem like the feature should be older? Newer? Or just right?
```

docs/CONTRIBUTING.md
ddbeck: Yeah, I feel like this might be simplified a bit for now, until we have a better handle on common cases. I might simplify this whole section as having three key parts: too old, too new, and too wrong.

1. If the feature's overall status is too old (e.g., it's a new feature being reported as long-established), then you might have missing compat keys. Check that you've covering the complete feature by looking for missing interfaces, CSS property values, and so on.
1. If the feature's overall status is too young (e.g., a long-established feature is reported as newly available), then you might have later additions holding the feature back unfairly. Consider finding a small number of entry points into the feature that are representative of the whole (e.g., the `api.fetch` key can stand in for support for Fetch in general), then using those entry points in the `compute_from` field.
1. If the feature's overall status _and_ individual keys are yielding incorrect results, the underlying data may have caveats or errors. Check `mdn/browser-compat-data`. Important caveats include features behind prefixes or flags, or partial implementations. Errors include missing support (e.g., a feature shown as supported in a desktop browser but not mobile) or wrong version numbers.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+1. Open the file in your favorite text editor, and add the following minimum content:
+
+   ```yaml
+   name: "My feature"
+   description: "A short description of the feature."
+   spec: https://urlofthespec.com
+   ```
+
+   For guidance, see the [Names](./guidelines.md#names) and [Descriptions](./guidelines.md#descriptions) sections of the feature guidelines.
```
tidoust, line 132: Another place where a highllight could be used.
It would also be cool if we had guidance on `spec`: what to link to, when to have mulitple URLs.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+1. Optionally add a `group` field to the feature.
+
+   The `group` field is used to categorize features into groups. For example, the Async Clipboard API feature is in the `Clipboard` group. Groups are maintained in the [`groups`](https://github.com/web-platform-dx/web-features/tree/main/groups) directory, and each group is a YAML file.
```
tidoust, line 136:
```suggestion
   The `group` field is used to categorize features into groups. For example, the Async Clipboard API feature is in the `Clipboard` group. Groups are maintained in the [`groups`](https://github.com/web-platform-dx/web-features/tree/main/groups) directory, and each group is a YAML file. The name of the file defines the unique ID of a group.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+   * If one of the existing groups fits your feature, add the `group` field to your feature, and set it to the name of the group.
```
tidoust, line 138:
```suggestion
   * If one of the existing groups fits your feature, add the `group` field to your feature, and set it to the ID of the group.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+   * If none of the existing groups fit your feature, but you believe a new group should be greated, then create a new group file in the `groups` directory, and add the feature to that group.
```
tidoust, line 139:
```suggestion
   * If none of the existing groups fit your feature, but you believe a new group should be greated, then create a new group file in the `groups` directory, then add the `group` field to your feature set to the ID of the new group.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+   ```yaml
+   name: "My feature"
+   description: "A short description of the feature."
+   spec: https://urlofthespec.com
+   ```
+
+   For guidance, see the [Names](./guidelines.md#names) and [Descriptions](./guidelines.md#descriptions) sections of the feature guidelines.
+
+1. Optionally add a `group` field to the feature.
```
tidoust, line 134: I don't have text to suggest immediately, but it would be great to have some sort of purpose here. That is, without knowing what groups are for (other than "to categorize features"), it's hard to understand whether they're worth spending time on.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+1. Review the existing draft features by looking at the YAML files in the [`features/draft`](https://github.com/web-platform-dx/web-features/tree/main/features/draft) directory and sub-directories.
+
+1. Find a draft that you want to work on.
+
+1. Move the file to the `features` directory.
+
+1. Remove the `draft_date` field from the file.
+
+1. Review the feature ID, name, description, and spec fields.
```
tidoust, line 157: Consider splitting `description` here. Most draft features have `TODO` as a description. That requires more than a review!

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+After you've prepared your new feature file with its unique ID, `name`, `description`, and `spec` fields, you must provide the necessary information for this feature's status to be calculated.
+
+A feature's status consists of:
+
+* Whether the feature is considered baseline or not, and which level of baseline it has achieved:
+
+  * Baseline _low_ means that the feature is now available on all browsers listed in the core browser set.
+  * Baseline _high_ means that the feature has been available on all browsers listed in the core browser set for long enough that it's considered _widely_ available.
+
+  For more information about baseline, see [Baseline](./baseline.md).
```
tidoust, line 172: I suppose you don't want to duplicate the baseline high definition here as it could then become out of sync. It may be worth emphasizing that this is where the notion of "long enough" is captured:

```suggestion
  For more information about baseline and the actual definition of _long enough_, see [Baseline](./baseline.md).
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+BCD contains browser support data for individual constituent of web features, such as CSS properties, JavaScript statements, HTML elements and attributes, DOM interfaces, methods, and events.
+For example, BCD contains the list of browsers, and their versions, that support the [`grid-template-rows` CSS property](https://github.com/mdn/browser-compat-data/blob/main/css/properties/grid-template-rows.json), which is a constituent of the [grid](https://github.com/web-platform-dx/web-features/blob/main/features/grid.yml) feature.
+
+We refer to each entry that contains browser support data in BCD as a _BCD key_. In the `grid-template-rows` example, the BCD key is `css.properties.grid-template-rows`.
+
+Features in the web-features project are associated to one or more BCD keys. For example, the grid feature is associated to multiple BCD keys, which together, describe the overall grid feature: `css.properties.display.grid`, `css.properties.display.inline-grid`, `css.properties.grid`, `css.properties.grid-area`, and more.
```
tidoust, line 189: Nit: the documentation uses both "associated to" and "associated with". I think the latter is more correct. It would be good to be consistent throughout.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
 
-* Go to [feature guidelines](./README.md#feature-guidelines).
-* Go to [open issues](https://github.com/web-platform-dx/web-features/issues) to propose new feature groups or report a bug.
-* Go to [open PRs](https://github.com/web-platform-dx/web-features/pulls) to review incoming feature groups and updates.
+In any case, **thank you** for wanting to help. This document will guide you through the process of contributing to the web-features repository.
 
-Also consider joining [the WebDX Community Group](https://www.w3.org/community/webdx/).
+Also consider joining the [WebDX Community Group](https://www.w3.org/community/webdx/).
+
+## What makes a feature
```
tidoust, line 25: It could be worth completing the section to say that, at least for the time being, we're more directly focusing on existing features: those that have some level of support somewhere, and that are typically already somewhat documented on MDN (and/or Can I Use).

What triggered this comment is that the part on setting the status below suggest to look at compatibility tables at the end of MDN pages related to the new feature at hand.

tidoust (commented):
Excellent start! I added a few comments. One general comment is that features have two heads:

1. There's one head that is more about "naming": choosing an ID, a name, setting the description, the URLs to the spec(s). Although technical knowledge is no doubt needed to understand how to scope a feature, that seems to be doable by a large number of people.
2. There's one head that is more about the support data. That part requires more intimate knowledge about BCD, browser versions and history. That's more for "experts".

I'm wondering whether we might want to separate the two in the submission process: Want to propose a feature? First step, give us the "naming" part and don't worry too much about the support data for now if you're not familiar enough. Second step, which could be done by others, focus on the support data.

Now, I do realize that the two heads are intertwined: the way you name and describe a feature depends on the exact set of compat keys that get associated with it, so not sure there's much that can be done in practice.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+As such, a feature has no specific size. Some features might cover single CSS properties while other features cover entire JavaScript APIs with multiple interfaces, methods, and properties. As long as a web developer would use this to achieve a specific goal, it's a feature. 
+
+For example, the `fetch` API is a feature, the `:has()` CSS pseudo-class function too, and so is the Web Audio API.
+
+### Feature file name and format
+
+Features in this repository are authored as **YAML files** that are stored in the [`features`](https://github.com/web-platform-dx/web-features/tree/main/features) directory. Each file corresponds to a single feature, and contains metadata about the feature, such as its name, description, specification, or support data.
+
+For example, the `fetch` API feature is described in the [`features/fetch.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/fetch.yml) file.
```
foolip, line 37:
```suggestion
For example, the `fetch()` API feature is described in the [`features/fetch.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/fetch.yml) file.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+Feature IDs must be unique as they are used in other projects to reference the feature. For example, the [browser-compat-data](https://github.com/mdn/browser-compat-data/) project references feature IDs in their compatibility data.
+
+### Fields in a feature file
+
+This table lists the fields that can be found in a feature file, and provides a brief description of each field:
+
+| Field | Description | Type | Mandatory |
+|---|---|---|---|
+| `name` | The name of the feature. | String | Yes |
+| `alias` | An alternative name for the feature. | String, or array of strings | No |
```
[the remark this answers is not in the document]
foolip: We use it when changing feature IDs, with the idea that users of web-features could detect renames. However, AFAIK nobody has used this, and I think we should consider just removing it before 1.0 and coming up with a more explicit scheme that doesn't require building a reverse map from old name to new name.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+| `description` | A short description of the feature. | String | Yes |
+| `spec` | One or more specification URLs for this feature | String, or array of strings | Yes |
+| `group` | An optional group, or list of groups that this feature belongs to | String, or array of strings | No |
+| `snapshot` | An optional snapshot, or list of snapshots that this feature belongs to | String, or array of strings | No |
+| `caniuse` | The feature's ID on the [Can I Use](https://caniuse.com/) website. | String | No |
+| `compat_features` | The list of browser-compat-data entries that make up this feature. In many cases, the corresponding browser-compat-data entries already map to the feature ID, and this is therefore not needed. | Array of strings | No |
```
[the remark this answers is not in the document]
foolip: Very good questions, @tidoust!

`compat_features` is the easiest way to author a new feature, but then we want tags in BCD for the long term, so that web-features doesn't have to react to renames and the like.

I have a script in https://github.com/web-platform-dx/web-features/pull/473 to migrate tags, but I've only run it 2 or 3 times, there's no schedule.

A situation which is likely to get contributors confused is when BCD has tags but you need to add or remove something. Then you need to list all of the keys that should be included in `compat_features`, completely ignoring BCD. But how do you get back in sync with BCD? The current answer is by hand, with some pain.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+| `status` | An optional object which describes whether the feature is considered baseline, when it achieved this baseline status, and the version number of supported browsers. In the majority of cases, this is calculated from the `compat_features` list, and therefore not needed. | Object | No |
```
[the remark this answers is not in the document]
foolip: This is also where we need to describe `status.compute_from`.

docs/CONTRIBUTING.md
foolip, line 136:
```suggestion
   The `group` field is used to categorize features into groups. For example, the Async Clipboard API feature is in the `clipboard` group. Groups are maintained in the [`groups`](https://github.com/web-platform-dx/web-features/tree/main/groups) directory, and each group is a YAML file.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+To start from an existing draft feature:
+
+1. Go through the steps in [Fork the repository and set up your local environment](#fork-the-repository-and-set-up-your-local-environment).
+
+1. Review the existing draft features by looking at the YAML files in the [`features/draft`](https://github.com/web-platform-dx/web-features/tree/main/features/draft) directory and sub-directories.
+
+1. Find a draft that you want to work on.
+
+1. Move the file to the `features` directory.
```
foolip, line 153:
```suggestion
1. Move the `.yml` and `.yml.dist` files to the `features` directory.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+1. Remove the `draft_date` field from the file.
```
foolip, line 155:
```suggestion
1. Remove the `draft_date` field from the `.yml` file.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+  ```yaml
+  name: Grid
+  description: CSS Grid is a two-dimensional layout system, which lays content out in rows and columns.
+  spec: https://drafts.csswg.org/css-grid-3/
+  group: grid
+  caniuse: css-grid
+  ```
+
+  You should always check first if BCD doesn't already map to your feature ID by searching for `web-features:<your-feature-id>` in the BCD source code.
```
foolip, line 209: Instead of searching BCD, one can just run `npm run dist` and see if that results in some computed status.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+  compat_features:
+    - css.at-rules.counter-style
+    - css.at-rules.counter-style.additive-symbols
+    - css.at-rules.counter-style.fallback
+    - css.at-rules.counter-style.negative
+   ```
+
+  To identify the BCD keys that you need, check the browser compatibility tables displayed at the end of the MDN pages that document your new feature, and then search the BCD source code for the names that appear in the table.
+
+  You can also use the `npm run traverse | grep <keyword>` command in BCD to list all BCD keys that match a certain keyword.
```
foolip, line 229:
```suggestion
  You can also use the `npm run traverse | grep -i <keyword>` command in BCD to list all BCD keys that match a certain keyword.
```

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+  spec: https://drafts.csswg.org/css-grid-3/
+  group: grid
+  caniuse: css-grid
+  ```
+
+  You should always check first if BCD doesn't already map to your feature ID by searching for `web-features:<your-feature-id>` in the BCD source code.
+
+* Or BCD doesn't already reference your new feature. In this case, you need to list the BCD keys that make up your feature under the `compat_features` field of your new feature file.
+
+  For example, here are the first few lines of the @counter-style feature, which defines a list of BCD keys:
```
foolip, line 213:
```suggestion
  For example, here are the first few lines of the `@counter-style` feature, which defines a list of BCD keys:
```

docs/CONTRIBUTING.md
foolip: My approach is to look for the feature's entrypoint in the dist file. If that entrypoint doesn't have the "⬇️ Same status as overall feature ⬇️" comment I use `compute_from` to make it so. Whether the feature was too old or too new is something you find out this way, at least I don't do a different thing depending on whether I think it's too old or too new initially.

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
captainbrosset: Sounds like I should remove it from the docs for now then.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,271 @@
...
+
+The YAML files that are used to author features are also used to generate _dist_ files. The term _dist_ is short for distribution.
+
+Dist files are also YAML files, but they end with the `.yml.dist` extension. Dist files are part of the data that we publish.
+
+For example, the `fetch` feature file is `features/fetch.yml`, and its dist file is `features/fetch.yml.dist`.
+
+You **never edit a dist file** directly. Dist files are generated from the authored YAML files, and are used at build time to generate the final data bundle that's used by other projects. Dist files are generated by using the `npm run dist` command, which you use when creating or updating a feature. To learn more, see [Set the status of a feature](#set-the-status-of-a-feature), below.
+
+It's important to note that, even if dist files are generated artifacts, they are still checked-in to the repository. This is normal and expected. Because dist files contain the compatibility data and status information for a feature, we review them as part of the pull request process.
```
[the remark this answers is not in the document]
captainbrosset: Funny, I almost have the opposite reaction to these kinds of banners. I tend to think that users skip over them, associating them with ads or other useless content and continue reading the text which looks similar to the one before the banner. I think I tend to do this myself. I quickly scan a document, found out where the content is, and scan through it quickly, skipping over anything that doesn't look like normal content.

I'll try the Important highlight and we can see what it looks like for real.

docs/CONTRIBUTING.md
captainbrosset: I don't think we have a purpose for groups yet. We need to work with consumers to learn what they could do with groups. If they're not going to use groups, then my personal opinion would be to turn them into a learning resource: something you can use to logically learn what's available in the web platform. But that's just one way to group features.

Perhaps we can remove this section for now, and only add it once we know what we want to do with groups?

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
captainbrosset: This goes back to your general comment @tidoust, stating that it might make sense to almost have 2 contribution guides (or 2 sections). One where you focus on creating the feature at all, with a good ID, name, description. And one, more involved, where you spend time identifying the right BCD keys and which ones the status should be computed from.
I'm not against that. I like beginner-friendly guides that help newcomers get from nothing to a PR.
Since this is a larger piece of work, how about we do this in a separate PR?

foolip:
@captainbrosset this is looking really good to me! Is there something left you want to do before marking this ready for review (not draft)?

captainbrosset:
> @captainbrosset this is looking really good to me! Is there something left you want to do before marking this ready for review (not draft)?

Yes, I want to resolve the TODO in ** Fix data discrepancies in your feature's dist file**. 
There's more to do, but I think it's probably enough of a start, and we can continue to improve it over time.

captainbrosset:
@ddbeck @tidoust @foolip thank you for the reviews so far. I'm now done with the content I wanted to write. Again, there's more to do here, such as:

- Guidance about specs.
- Guidance about groups.
- Splitting the docs into 2 parts: one, simpler, part to just get a feature started. And another, more involved, part to resolve data discrepancies.
- Better documentation for the compute_from field.
- Adding missing docs for completely overriding the status.

I'd prefer to track this in a new issue so we can add the missing docs over time in subsequent PRs.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,278 @@
...
+In any case, **thank you** for wanting to help. This document will guide you through the process of contributing to the web-features repository.
 
-Also consider joining [the WebDX Community Group](https://www.w3.org/community/webdx/).
+Also consider joining the [WebDX Community Group](https://www.w3.org/community/webdx/).
+
+## What makes a feature
+
+Think of features as individual components of the web platform that web developers can use to achieve specific tasks.
+
+As such, a feature has no specific size. Some features might cover single CSS properties while other features cover entire JavaScript APIs with multiple interfaces, methods, and properties. As long as a web developer would use this to achieve a specific goal, it's a feature.
```
autonome, line 29: This may be a good point to note that web features can also span any of HTML, JS, CSS.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,278 @@
...
+
+For example, the `fetch()` API is a feature, the `:has()` CSS pseudo-class function too, and so is the Web Audio API.
+
+For the time being, this repository focuses only on features that have at least some level of browser support and that are already documented on MDN.
```
autonome, line 33: Re "some level" - the W3C process doc kinda/sorta says at least two implementers, and that often plays out in practice, so might be a good place to start.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,278 @@
...
+### Fields in a feature file
+
+This table lists the fields that can be found in a feature file, and provides a brief description of each field:
+
+| Field | Description | Type | Mandatory |
+|---|---|---|---|
+| `name` | The name of the feature. | String | Yes |
+| `description` | A short description of the feature. | String | Yes |
+| `spec` | One or more specification URLs for this feature | String, or array of strings | Yes |
+| `group` | An optional group, or list of groups that this feature belongs to | String, or array of strings | No |
```
autonome, line 54: Should explain or link to explanation of `groups` in this context.

docs/CONTRIBUTING.md
```diff
@@ -1,21 +1,278 @@
...
+| `snapshot` | An optional snapshot, or list of snapshots that this feature belongs to | String, or array of strings | No |
```
autonome, line 55: ditto as for `group`

docs/CONTRIBUTING.md
foolip: We've been adding features with only one implementation, from all three engines. We could make this edit to make that explicit:

```suggestion
For the time being, this repository focuses only on features that are implemented in some browser and that are already documented on MDN.
```

autonome: gotcha, in that case +1 to your edit with a s/focuses only on/only includes/.

docs/CONTRIBUTING.md
captainbrosset: Removed this row for now, as we've not really used snapshots so far.

captainbrosset:
I think I've addressed most comments on this first version, and I'd like to merge this sooner rather than later to avoid blocking other docs changes/merging conflicts for a long period of time.
I've seen enough positive reactions here to make me confident that this is enough of a useful start to merge, and then continue iterating over subsequent PRs.

@ddbeck do you agree?

Issue #1331, opened 2024-07-08
Features entering Baseline high will cause CI to fail


foolip:
`npm run test:dist` can start failing at any time due to the `isFuture()` call here:

https://github.com/web-platform-dx/web-features/blob/a2a874aba8096b9dcb0a0f88a883c5960df90407/packages/compute-baseline/src/baseline/index.ts#L231

This just happened in https://github.com/web-platform-dx/web-features/pull/1326 with `npm run dist` producing this diff:

```diff
diff --git a/features/font-synthesis.yml.dist b/features/font-synthesis.yml.dist
index 1ea2d98a..88064a0a 100644
--- a/features/font-synthesis.yml.dist
+++ b/features/font-synthesis.yml.dist
@@ -2,8 +2,9 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 status:
-  baseline: low
+  baseline: high
   baseline_low_date: 2022-01-06
+  baseline_high_date: 2024-07-06
   support:
     chrome: "97"
     chrome_android: "97"
```

foolip:
To avoid this we need to not use `Temporal.Now.plainDateISO()`. We need some other definition of "now" that depends on the state of the tree, and I would suggest the most recent date for any current browser release.

PR #1333, opened 2024-07-08
Use a cutoff date from BCD for Baseline high

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/1331.

### DIFF
--- a/packages/compute-baseline/src/baseline/date-utils.ts
+++ b/packages/compute-baseline/src/baseline/date-utils.ts
@@ -1,12 +1,6 @@
 import { Temporal } from "@js-temporal/polyfill";
 import { BASELINE_LOW_TO_HIGH_DURATION } from "./index.js";
 
-type LowDate = Temporal.PlainDate | string;
-
-export function isFuture(date: Temporal.PlainDate): boolean {
-  return Temporal.PlainDate.compare(Temporal.Now.plainDateISO(), date) < 0;
-}
-
 export function toHighDate(
   lowDate: Parameters<typeof Temporal.PlainDate.from>[0],
 ): Temporal.PlainDate {
--- a/packages/compute-baseline/src/baseline/index.ts
+++ b/packages/compute-baseline/src/baseline/index.ts
@@ -7,7 +7,7 @@ import { Compat, defaultCompat } from "../browser-compat-data/compat.js";
 import { feature } from "../browser-compat-data/feature.js";
 import { Release } from "../browser-compat-data/release.js";
 import { browsers } from "./core-browser-set.js";
-import { isFuture, toDateString, toHighDate } from "./date-utils.js";
+import { toDateString, toHighDate } from "./date-utils.js";
 import { support } from "./support.js";
 
 interface Logger {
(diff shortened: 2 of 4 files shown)


features/font-synthesis.yml.dist
```diff
@@ -2,9 +2,8 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 status:
-  baseline: high
+  baseline: low
   baseline_low_date: 2022-01-06
-  baseline_high_date: 2024-07-06
```
foolip, line 7: Even though this date is in the past, this change is rolled back, since the latest release in BCD is Safari 17.6, released 2024-06-17.

packages/compute-baseline/src/baseline/date-utils.ts
```diff
@@ -1,12 +1,6 @@
 import { Temporal } from "@js-temporal/polyfill";
 import { BASELINE_LOW_TO_HIGH_DURATION } from "./index.js";
 
-type LowDate = Temporal.PlainDate | string;
-
-export function isFuture(date: Temporal.PlainDate): boolean {
-  return Temporal.PlainDate.compare(Temporal.Now.plainDateISO(), date) < 0;
```
foolip, line 7: Removing this line is key. Everything else was the follow-on effects.

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -101,10 +106,12 @@ export function computeBaseline(
   const keystoneDate = findKeystoneDate(
     statuses.flatMap((s) => [...s.support.values()]),
   );
-  const { baseline, baseline_low_date, baseline_high_date, discouraged } =
+  const discouraged = statuses.some((s) => s.discouraged);
```
foolip, line 109: I pulled out `discouraged`, previously it was passed into `keystoneDateToStatus` only to be returned unchanged.

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -123,24 +130,12 @@ export function computeBaseline(
...
  * for a single compat key.
  */
-function calculate(compatKey: string, compat: Compat): SupportDetails {
+function calculate(compatKey: string, compat: Compat) {
   const f = feature(compatKey);
-  const s = support(f, browsers(compat));
-  const keystoneDate = findKeystoneDate([...s.values()]);
-
-  const { baseline, baseline_low_date, baseline_high_date, discouraged } =
-    keystoneDateToStatus(keystoneDate, f.deprecated ?? false);
```
foolip, line 132: Eliminating this call to `keystoneDateToStatus` made this change a bit simpler, and the information wasn't used anyway.

foolip (commented):
Self-review to make this easier to follow.

packages/compute-baseline/src/baseline/index.ts
```diff
@@ -90,6 +90,11 @@ export function computeBaseline(
   },
   compat: Compat = defaultCompat,
 ): SupportDetails {
+  // findKeystoneDate is used here because it returns the latest date from an
+  // array of releases. The result isn't a keystone date of any feature.
+  const cutoffDate = findKeystoneDate(browsers(compat).map((b) => b.current()));
+  assert(cutoffDate !== null);
+
```
ddbeck, line 97: This adds a new condition to the Baseline high date: nothing can reach Baseline high until 30 months after Baseline low _and_ at least one browser has released on or since that date.

If the aim here is to pin ourselves to BCD's state then I'd like to suggest some alternatives:

- BCD's `__meta.timestamp` value (e.g., we can never get ahead of BCD's release schedule)
- The last recorded release date in BCD, even if it's a future release date (something like `browsers(compat).map((b) => b.releases.at(-1))`).

foolip: I've gone with `__meta.timestamp`. The latest release in the releases array can be a planned release months into the future, so it's not a good approximation of now. Currently, Edge and Firefox have planned releases in September in BCD.

ddbeck (approved):

PR #1337, opened 2024-07-09
Add <audio> and <video> features

foolip:
The overall status matches caniuse with these exceptions:

- Chrome 4 is the earliest possible version in caniuse, so it cannot
  match BCD's claim of Chrome 3.
- Partial support due to parts of the API being unavailable or not
  working isn't reproduced. This is clear from the statuses of the
  individual BCD keys instead.

HTMLMediaElement is made part of <video> only for now.

Everything is Baseline high except for volume and playbackRate.


features/audio.yml
```diff
@@ -1,9 +1,12 @@
-draft_date: 2024-06-05
 name: <audio>
 description: The `<audio>` element TODO.
```
ddbeck, line 2:
```suggestion
description: The `<audio>` element embeds a sound media player in a document.
```

features/video.yml
```diff
@@ -0,0 +1,77 @@
+name: <video>
+description: The `<video>` element TODO.
```
ddbeck, line 2:
```suggestion
description: The `<video>` element embeds video media player in a document.
```

features/video.yml.dist
```diff
@@ -0,0 +1,339 @@
...
+  #   firefox: "3.5"
+  #   firefox_android: "4"
+  #   safari: "3.1"
+  - api.HTMLMediaElement.volume
+
+  # baseline: false
+  # support:
+  #   safari: "3.1"
+  #   safari_ios: "3"
+  - api.HTMLMediaElement.playbackRate
```
ddbeck, line 339: FWIW, both of these have dodgy-looking data. Someday someone might come along and fix it 😄

foolip: `playbackRate` is about only Safari supporting a negative playback rate. The spec does allow for throwing an exception "if the given value is not supported by the user agent" but doesn't require any specific playback rates to be supported.

`volume` is about it not doing anything on iOS.

features/audio.yml
```diff
@@ -1,9 +1,12 @@
-draft_date: 2024-06-05
 name: <audio>
-description: The `<audio>` element TODO.
+description: The `<audio>` element embeds a sound media player in a document.
```
foolip, line 2: @ddbeck the "embeds in a document" part of this strikes me as odd, as all elements add something to a document. It gives it a plugin feel, as if it's not quite integrated into the platform.

How about...

```suggestion
description: The `<audio>` element plays sound such as audio effects or music, optionally with controls provided by the browser.
```

features/video.yml
```diff
@@ -0,0 +1,77 @@
+name: <video>
+description: The `<video>` element embeds a video media player in a document.
```
foolip, line 2: Maybe...

```suggestion
description: The `<video>` element plays videos or movies, optionally with controls provided by the browser.
```

foolip (commented):
@ddbeck tweaking the descriptions a bit...

features/audio.yml
ddbeck: Yeah, I had a time with this myself. In the end, I aligned to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio, which wasn't _great_.

Setting aside embeds, I think the big alternative here is _represents_. As in:

```suggestion
description: The `<audio>` element represents a sound file or stream, such as audio effects or music. Optionally, the browser provides playback controls.
```

We haven't done "`<element>` represents…" with other elements so far, but we _could_. It's a bit abstract—representation doesn't _do_ anything, without further explanation—but a lot of elements _are_ abstract in this way (e.g.,`<em>` or `<div>`).

features/video.yml
ddbeck: Continuing on my commentary above:

```suggestion
description: The `<video>` element represents a video file or stream, a moving picture. Optionally, the browser provides playback controls. 
```

features/audio.yml
foolip: I'm fine with "represents" if we can point to the thing it represents. Because of MSE (`MediaSource`) and how `getUserMedia()` plugs into `<video>` with `srcObject` or a special URL, media elements don't necessarily represent files. You could say they represent media resources, but that's quite abstract.

ddbeck (approved):
Thank you!

foolip:
A summary of a side-channel chat between me and @ddbeck. The reason to avoid "plays" in the description here would be the precedent that we need to pick a verb for every element, leading to things like "`<div>` shows content". However, we'll just not do that, "play" works for media elements and for other elements we might have to use a different structure.

PR #1339, opened 2024-07-09
`compute-baseline`: Add per-release support checks to `Feature`s

ddbeck:
The main thing this does is gives us a way to quickly and unambiguously report `true`, `null,` or `false` for _unqualified_ support relationship between a release and a feature. There's some untidiness to clean up along the way. I'll self-review with commentary.

### DIFF
--- a/packages/compute-baseline/src/baseline/support.ts
+++ b/packages/compute-baseline/src/baseline/support.ts
@@ -19,7 +19,7 @@ export function support(feature: Feature, browsers: Browser[]): Support {
 
     // const reverseChronological = b.releases.slice().reverse();
     // let previousRelease: string | undefined;
-    // for (let index = 0; index < reverseChronological.length; index++) {
+    // for (let index = b.releases.length - 1; index >= 0; index--) {
     //   const release = reverseChronological[index] as Release;
     //   const current = feature.flatSupportedIn(release);
 
--- a/packages/compute-baseline/src/baseline/support.ts
+++ b/packages/compute-baseline/src/baseline/support.ts
@@ -13,6 +13,18 @@ export function support(feature: Feature, browsers: Browser[]): Support {
   const support: Support = new Map();
   for (const b of browsers) {
     const releases = feature.supportedBy({ only: [b] });
+    // TODO:
+    // let lastInitial: Release | undefined;
+    // let lastInitialBoundary: "" | "≤" | undefined;
+
+    // const reverseChronological = b.releases.slice().reverse();
+    // let previousRelease: string | undefined;
+    // for (let index = 0; index < reverseChronological.length; index++) {
+    //   const release = reverseChronological[index] as Release;
+    //   const current = feature.flatSupportedIn(release);
+
+    //   // Check if current has changed, etc.
+    // }
 
     const unqualifiedReleases = [];
     const qualifiedReleases = [];
--- a/packages/compute-baseline/src/browser-compat-data/feature.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/feature.test.ts
@@ -128,7 +128,7 @@ describe("features", function () {
         assert.equal(lineClamp[0]?.qualifications?.prefix, "-webkit-");
       });
 
-      it("returns mixed results for (un)prefixedfeatures", function () {
+      it("returns mixed results for (un)prefixed features", function () {
         const fx = browser("firefox");
         const actual = feature(
           "css.types.image.gradient.repeating-linear-gradient",
--- a/packages/compute-baseline/src/browser-compat-data/feature.test.ts
+++ b/packages/compute-baseline/src/browser-compat-data/feature.test.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 
 import { feature } from "./feature.js";
-import { browser } from "./index.js";
+import { browser, Compat } from "./index.js";
 
 describe("features", function () {
   describe("feature()", function () {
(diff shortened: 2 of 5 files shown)


packages/compute-baseline/src/baseline/support.ts
```diff
@@ -13,6 +13,18 @@ export function support(feature: Feature, browsers: Browser[]): Support {
...
+    // let lastInitialBoundary: "" | "≤" | undefined;
+
+    // const reverseChronological = b.releases.slice().reverse();
+    // let previousRelease: string | undefined;
+    // for (let index = 0; index < reverseChronological.length; index++) {
+    //   const release = reverseChronological[index] as Release;
+    //   const current = feature.flatSupportedIn(release);
+
+    //   // Check if current has changed, etc.
+    // }
```
ddbeck, line 27: This is just to show where I'm going with this: walk backward through releases, then find the support-to-non-support boundary. If we stepped into a `null`, then that's `≤X`.

packages/compute-baseline/src/browser-compat-data/feature.test.ts
```diff
@@ -69,5 +69,86 @@ describe("features", function () {
...
+
+      it("returns null for unknown support", function () {
+        const edge = browser("edge");
+        const f = feature("svg.elements.animate"); // { version_added: "≤79" }
+
+        assert.equal(f.flatSupportedIn(edge.version("12")), null);
+        assert.equal(f.flatSupportedIn(edge.version("79")), true);
+        assert.equal(f.flatSupportedIn(edge.version("80")), true);
+      });
+    });
```
ddbeck, line 108: This is the flat version, which ignores prefixes and partials.

packages/compute-baseline/src/browser-compat-data/feature.test.ts
```diff
@@ -69,5 +69,86 @@ describe("features", function () {
...
+        const f = feature("svg.elements.animate");
+        const unknown = f.supportedIn(edge.version("12"));
+        assert.equal(unknown.length, 1);
+        assert.equal(unknown[0]?.supported, null);
+
+        const known = f.supportedIn(edge.version("79"));
+        assert.equal(known.length, 1);
+        assert.equal(known[0]?.supported, true);
+      });
+    });
```
ddbeck, line 152: To find good test cases for the flat version, I ended up needing an un-flat version. We can dump this if we like, if we think we won't use it. OTOH, I did use it to write test cases. A conundrum.

packages/compute-baseline/src/browser-compat-data/feature.ts
```diff
@@ -72,34 +75,65 @@ export class Feature {
...
     if (statementOrStatements === undefined) {
-      throw Error(`${this} contains no support data for ${browser.name}`);
+      throw new Error(`${this} contains no support data for ${browser.name}`);
     }
 
-    const rawStatements = Array.isArray(statementOrStatements)
+    return Array.isArray(statementOrStatements)
       ? statementOrStatements
       : [statementOrStatements];
+  }
```
ddbeck, line 92: This happens in a few places, so I consolidated it.

packages/compute-baseline/src/browser-compat-data/feature.ts
```diff
@@ -72,34 +75,65 @@ export class Feature {
...
+
+  supportStatements(browser: Browser): SupportStatement[] {
+    return this.rawSupportStatements(browser).map((raw) =>
+      statement(raw, browser, this),
+    );
+  }
```
ddbeck, line 98: This too happens in a few places, now consolidated.

packages/compute-baseline/src/browser-compat-data/feature.ts
```diff
@@ -72,34 +75,65 @@ export class Feature {
...
     const result = [];
-    for (const raw of rawStatements) {
-      const s = statement(raw, browser, this);
+    for (const s of this.supportStatements(release.browser)) {
+      this.assertRealSupportStatement(s, release.browser);
 
-      if (!(s instanceof RealSupportStatement)) {
-        throw Error(
-          `${this.id} contains non-real values for ${browser.name}. Cannot expand support.`,
-        );
```
ddbeck, line 100: This also happens in a couple of places, so I consolidated that too—it's at the end. I have no idea how to order things. 😬

packages/compute-baseline/src/baseline/support.ts
```diff
@@ -13,6 +13,18 @@ export function support(feature: Feature, browsers: Browser[]): Support {
   const support: Support = new Map();
   for (const b of browsers) {
     const releases = feature.supportedBy({ only: [b] });
+    // TODO:
+    // let lastInitial: Release | undefined;
+    // let lastInitialBoundary: "" | "≤" | undefined;
+
+    // const reverseChronological = b.releases.slice().reverse();
+    // let previousRelease: string | undefined;
+    // for (let index = 0; index < reverseChronological.length; index++) {
```
foolip, line 22: Since we're iterating an index anyway, I think we can just iterate `b.releases` backwards like this:

```suggestion
    // for (let index = b.releases.length - 1; index >= 0; index--) {
```

packages/compute-baseline/src/browser-compat-data/feature.ts
foolip: Order is hard. End is fine.

packages/compute-baseline/src/browser-compat-data/feature.test.ts
```diff
@@ -69,5 +69,86 @@ describe("features", function () {
...
+        // { version_added: "…", prefix: "-webkit-" }
+        const lineClamp = feature("css.properties.line-clamp").supportedIn(
+          cr.version("100"),
+        );
+        assert.equal(lineClamp.length, 1);
+        assert.equal(lineClamp[0]?.supported, true);
+        assert.equal(lineClamp[0]?.qualifications?.prefix, "-webkit-");
+      });
+
+      it("returns mixed results for (un)prefixedfeatures", function () {
```
foolip, line 131:
```suggestion
      it("returns mixed results for (un)prefixed features", function () {
```

packages/compute-baseline/src/browser-compat-data/feature.test.ts
foolip: Unless it gets exported to compute-baseline users, let's have it. Although if the flat version is the one we'll use most, maybe give that the nice name and rename the other one to a longer name?

foolip (approved):

packages/compute-baseline/src/browser-compat-data/feature.test.ts
ddbeck: Done with c07b62db0f812f87bb3e51d8f7091692239fb12d.

PR #1341, opened 2024-07-09
Move `<base>` out of drafts

ddbeck:


features/base.yml
```diff
@@ -1,12 +1,15 @@
...
 group: html
+status:
+  compute_from: html.elements.base
 compat_features:
   - api.HTMLBaseElement
   - api.HTMLBaseElement.href
   - api.HTMLBaseElement.target
   - html.elements.base
   - html.elements.base.href
+  - html.elements.base.href.forbid_data_javascript_urls
```
ddbeck, line 13: Seeing this in a burndown caused me to include it here.

This is one of many, many behavioral features in BCD that's a bit backwards: the old behavior ought to be a deprecated "feature" that gets removed in a couple of years after everything has settled. This is never something you'd "use" about the feature—rather, it's something that defines what the feature _isn't_.

For completeness, I've included it, but I'm sorely tempted to create an issue in BCD to start tracking all of these backwards features, so we can eventually flip them around and purge them.

That said, this specific one is new, so I could be persuaded to go flip this one now.

foolip: Should we just remove behavioral features from all of our burndown tracking? They never have their own MDN pages, so what is anyone ever going to do with the statuses?

foolip: Meh, ignore my suggestion. Sometimes they'll be useful in `compute_from` I'm sure, and providing a status for them is harmless.

ddbeck: They do get their own pages on caniuse though: https://caniuse.com/mdn-html_elements_base_href_forbid_data_javascript_urls

foolip (approved):
There's a tiny mismatch in early Safari data about `<base>` vs `HTMLBaseElement` that I'm pretty sure is wrong in BCD, but it's inconsequential.

PR #1342, opened 2024-07-09
Add `indexeddb` draft

ddbeck:


ddbeck:
Actually, I shouldn't do this sort of thing late in the evening. Pausing this for a bit—I _know_ there's a bunch of oddities with this, but was on autopilot..

features/indexeddb.yml.dist
```diff
@@ -0,0 +1,378 @@
...
+  - api.IDBDatabase.deleteObjectStore
+  - api.IDBDatabase.name
+  - api.IDBDatabase.objectStoreNames
+  - api.IDBDatabase.transaction
+  - api.IDBDatabase.version
+  - api.IDBDatabase.versionchange_event
+  - api.IDBFactory
+  - api.IDBFactory.cmp
+  - api.IDBFactory.deleteDatabase
+  - api.IDBFactory.open
```
ddbeck, line 51: This is the weird one. I'm guessing that BCD should be saying that this feature is partially supported on iOS from 14.5 to 15 (given the set of releases BCD tracks). But then we'd break alignment with caniuse.

https://bugs.webkit.org/show_bug.cgi?id=226547

foolip (approved):

PR #1343, opened 2024-07-09
Require a dist file for every feature

foolip:
This will mean that running just `npm run dist` will be enough even if
the dist file doesn't exist yet.


captainbrosset (approved):
I've not looked at the implementation of the dist script before, so can't really judge the technical soundness of this, but if this makes it possible to run `npm run dist` without having to pass the feature file name with it, then great!!

PR #1345, opened 2024-07-10
Link to docs from dist files

foolip:


foolip:
This depends on https://github.com/web-platform-dx/web-features/pull/1330 to land first.

captainbrosset:
Discussed on the WebDX CG call just now. We decided to close this. It's not that hard to ask people to run `npm run dist`.
Also we don't necessarily want to update all dist files right now.

Issue #715, opened 2024-03-20
Release the schema files in the the GitHub Release


jcscottiii:
Building on https://github.com/web-platform-dx/web-features/issues/714

It would be nice to include the schema files in the GitHub Release so that I know the format of the data.json I am about to parse.

I wouldn't consider this a high priority unless there are breaking changes to the schema.

foolip:
@jcscottiii do you use the two JSON schema files directly, or do you need a single JSON schema file that you can validate the whole `data.json` with?

foolip:
@jcscottiii can you share the quicktype invocation that would get us a single JSON schema file? If that just works, we can publish that and save you some trouble.

ddbeck:
We now publish the JSON Schema files alongside the data, both in the npm package and GitHub releases as artifacts, starting with [v1.0.0](https://github.com/web-platform-dx/web-features/releases/tag/v1.0.0). Let me know if you need anything else here. Cheers!

PR #1348, opened 2024-07-10
Rename schema file to data.schema.json and publish it

foolip:
This will include the schema in the NPM package. The idea is to also 
publish it with GitHub releases.


foolip:
The idea is that this would help with https://github.com/web-platform-dx/web-features/issues/715. We can publish web-features.schema.json in both the NPM package and the GitHub release. The name of the data files are different (index.json vs. data.json) so matching either of those names would still leave a mismatch in the other place.

ddbeck (approved):
Ah, so the end-goal here is to export this on the package? `import { schema } from "web-features"`? In any case, this is fine.

foolip:
I've sent https://github.com/web-platform-dx/web-features/pull/1470, and if we do that I think we should call this `data.schema.json` instead.

package.json
```diff
@@ -20,8 +20,8 @@
     "dist": "tsx scripts/dist.ts",
     "feature-init": "tsx scripts/feature-init.ts",
     "format": "npx prettier --write .",
-    "schema-defs:write": "npm run schema-defs -- --out ./schemas/defs.schema.json",
-    "schema-defs": "ts-json-schema-generator --tsconfig ./tsconfig.json --type WebFeaturesData --path ./types.ts --id defs",
+    "schema:write": "npm run schema -- --out ./schemas/data.schema.json",
+    "schema": "ts-json-schema-generator --tsconfig ./tsconfig.json --type WebFeaturesData --path ./types.ts",
```
ddbeck, line 24: This is no longer needed either (it has its own name one level down in definitions).

```suggestion
    "schema": "ts-json-schema-generator --tsconfig ./tsconfig.json --path ./types.ts",
```

ddbeck (approved):
It'd be nice to update `packages/web-features/README.md` with an example `import` statement for this, but I won't block on it. One minor suggestion in a line comment.

foolip:
I've documented the import as `import schema from "web-features/data.schema.json" with { type: "json" }` and will merge this now.

PR #1349, opened 2024-07-10
Override async clipboard status awaiting fix in BCD

foolip:
The manually added status is what will result from treating all 4
methods on navigator.clipboard as required parts of the feature.

Comment out read() and readText() to not bless the incorrect status
computed from BCD.

Fixes https://github.com/web-platform-dx/web-features/issues/1249.


ddbeck (approved):
This is fine, though we're now reporting Firefox as supporting while caniuse doesn't (I think because of the no-longer `clipboard-read` permission). I saw on https://github.com/mdn/browser-compat-data/pull/23593 that you've got some permissions follow ups anyway. Do you need to file an issue on caniuse for this too?

foolip:
I've sent https://github.com/Fyrd/caniuse/pull/7106 to align Chromium versions, and will file an issue about Firefox. I'll link to everything in a new comment before landing.

PR #1353, opened 2024-07-11
Add features for `@import` and `@supports`

ddbeck:
I picked these up because `api.CSSImportRule.supportsText` is newly available.


features/feature-queries.yml.dist
```diff
@@ -0,0 +1,77 @@
...
+  - api.CSSImportRule.supportsText
+
+  # baseline: false
+  # support:
+  #   chrome: "122"
+  #   chrome_android: "122"
+  #   edge: "122"
+  #   firefox: "115"
+  #   firefox_android: "115"
+  - css.at-rules.import.supports
```
foolip, line 77: This and api.CSSImportRule.supportsText are the same feature, and any differences are either errors in BCD, or mishaps in shipping. The chromestatus entry is https://chromestatus.com/feature/5899007704694784.

@danielsakhapov, does it make sense that the API part shipped first in Chrome 121?

ddbeck: Looking at https://github.com/mdn/browser-compat-data/pull/21603 and https://github.com/mdn/browser-compat-data/pull/22164 and https://issues.chromium.org/issues/40244647#comment15 it's also possible that it all shipped in 122?

(It's unclear to me whether the BCD collector correctly found `supportsText` in Chrome 121 beta, or if @Elchi3 added it because Chrome Status incorrectly said it was shipping.)

foolip: Hmm, 121 is the correct version according to https://chromiumdash.appspot.com/commit/f9bbbf13060df98c24385bc83da7693d33f3650d

foolip: @danielsakhapov helped me figure out what happened here. https://chromiumdash.appspot.com/commit/6ca8ac7e347b73c62609643164a444719f9d26f7 is the commit that actually shipped the feature, in Chrome 122.

The CSSOM bits were accidentally exposed in Chrome 121 and don't do anything. That's actually my fault, as the reviewer of that CL.

foolip: For Safari, we'd need to check if the feature is enabled, or if they made the same mistake of exposing the API first.

foolip: Hmm, there's is a mismatch in Firefox too...

@ddbeck I'd suggest just putting this subfeature to the side for now.

ddbeck: OK, I've commented it out and left a note back to this discussion. We can get it to when we're in the long tail, I guess?

features/feature-queries.yml
```diff
@@ -0,0 +1,16 @@
+name: Feature queries
```
foolip, line 1: Let's make the identifier `support`. I'm not sure about the name, whatever seems most common for web developers is OK.

foolip (approved):
Identifier should change, but OK to go after that.

ddbeck:
OK, I changed the `@supports` identifier and change the name and description. I used an "Also known as feature queries" to capture the alternate name.

PR #1354, opened 2024-07-11
Add `Intl.Segmenter` feature (and `Intl` group)

ddbeck:
I picked this up because `javascript.builtins.Intl.Segmenter` is newly available.


features/intl-segmenter.yml
```diff
@@ -0,0 +1,13 @@
+name: Intl.Segmenter
+description: "`Intl.Segmenter()` creates a locale-aware text splitter that can separate a string into meaningful graphemes, words, or sentences."
```
foolip, line 2: I tried `Intl.Segmenter()` and it's a constructor, so `new` is required. But putting that in here seems silly. This would align with what we did for `URLPattern`, which is also a constructor:

```suggestion
description: "The `Intl.Segmenter` API creates a locale-aware text splitter that can separate a string into meaningful graphemes, words, or sentences."
```

foolip (approved):

features/intl-segmenter.yml
ddbeck: Yes, I like this and being consistent. 👍

PR #1360, opened 2024-07-11
Add gravity sensor feature

foolip:
Chrome 91 is confirmed by https://chromestatus.com/feature/5384099747332096


features/gravity-sensor.yml
```diff
@@ -0,0 +1,6 @@
+name: Gravity sensor
+description: The `GravitySensor` API read the gravity applied to a device in three dimensions.
+spec: https://w3c.github.io/accelerometer/#gravity-sensor-model
+compat_features:
```
captainbrosset, line 4:
```suggestion
group: sensors
compat_features:
```

features/gravity-sensor.yml
```diff
@@ -0,0 +1,7 @@
+name: Gravity sensor
+description: The `GravitySensor` API reads the gravity applied to a device in three dimensions.
+spec: https://w3c.github.io/accelerometer/#gravity-sensor-model
+group: sensors
+compat_features:
+  - api.GravitySensor
+  - api.GravitySensor.GravitySensor
```
captainbrosset, line 7: BCD doesn't have keys for the `x`, `y`, `z` instance properties. Should we add them there, and reference them here?

captainbrosset (approved):

features/gravity-sensor.yml
foolip: This is because `GravitySensor` inherits from `Accelerometer` and that's where the properties are. It feels like something missing in this case because you absolutely need `x`, `y` and `z`, but it's structurally the same as why we don't list the `nodeValue` property (from `Node`) on every element interface (like `HTMLTitleElement`).

PR #1361, opened 2024-07-11
Heading elements

petele:
Adds `<h1>` through `<h6>` elements


petele:
@ddbeck & @foolip  - this is my first attempt at headings, hopefully it's not too far off!

features/headings.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
-name: <h1>
-description: The `<h1>` element TODO.
+name: Section headings elements
+description: The `<h1>` through `<h6>` elements represent six levels of section headings.
 spec:
```
foolip, line 3: There's just a single section in the spec here, with many IDs. Let's use just https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements, which is what the ToC links to.

features/headings.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
-name: <h1>
-description: The `<h1>` element TODO.
+name: Section headings elements
```
foolip, line 1: I'm unsure about naming for collections of elements. For something like `<picture>` and `<table>` I think we should name the feature after those elements, even though other elements are involved. Here, just`<h1>` doesn't represent all elements. But how about "&lt;h1>-&lt;h6>"? @ddbeck WDYT?

foolip (commented):
The status here is wrong, I'm sure the elements were supported in Safari 1. But the data uses ranges in BCD, so this should get turned into ranges here too after some work that @ddbeck is doing. So let's leave it alone.

petele:
Addressed feedback.

features/headings.yml
petele: done

features/headings.yml
petele: done

features/headings.yml
```diff
@@ -0,0 +1,12 @@
+name: <h1> through <h6>
+description: The `<h1>` through `<h6>` HTML elements represent six levels of section headings, in order of decreasing importance.
+spec:
+  - https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements
+compat_features:
```
captainbrosset, line 5: Let's add group:html like we've done for the other html element features.

captainbrosset (changes requested):

features/headings.yml
petele: done

features/headings.yml
```diff
@@ -0,0 +1,12 @@
+name: <h1> through <h6>
+description: The `<h1>` through `<h6>` HTML elements represent six levels of section headings, in order of decreasing importance.
+spec: https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements
+group: html
```
captainbrosset, line 4:
```suggestion
group: html-elements
```

A new group just landed. So let's use it here.

captainbrosset (approved):

captainbrosset (approved):

ddbeck:
Sorry about that. The way to fix this is to merge `main` into your branch and run `npm run dist` and commit the result. The conflicts to `.dist` files will get wiped out by `npm run dist`.

petele:
> Sorry about that. The way to fix this is to merge `main` into your branch and run `npm run dist` and commit the result. The conflicts to `.dist` files will get wiped out by `npm run dist`.

I think I did that, but not sure I did the right thing, it looks pretty messy now, and it still has the conflict. Here's what I ran to get here.

```
git checkout main
git fetch upstream
git merge upstream/main
git checkout elem-headings
git rebase main
git ci
npm run dist
git rebase
git push
```

petele:
Replaced by #1463 after messing this one up.

PR #1363, opened 2024-07-11
Moving background-sync from draft to features

tropicadri:
Edge support version is 79, even though Chrome support is 49. I believe the discrepancy comes from Edge moving to Chromium but since webkit hasn't implemented it, so we don't have a Baseline support the discrepancy between Edge and Chrome is not meaningful.


tropicadri:
@ddbeck tagging you as a reviewer until we get github write access I can't add you via the interface.

features/background-sync.yml
```diff
@@ -0,0 +1,13 @@
+name: Background sync
+description: Background synch allows web applications to defer server synchronization work to their service worker to handle at a later time, if the device is offline.
```
captainbrosset, line 2: A couple of thoughts:

* We usually try to name the entry point of the feature in the description. In this case it would be ServiceWorkerRegistration.sync. Although it's a bit long, and devs don't actually write ServiceWorkerRegistration in their code, but instead get a registration object by doing `await navigator.serviceWorker.ready`.  So I'm not too sure what to do here.
* The Background Sync API should be used for small data exchanges with the server, because the SW needs to be alive for the entire duration. It's advised to use Background Fetch for longer duration downloads. I feel like this should be mentioned in there somewhere.
* Also, if I'm not mistaken, you don't _have_ to use this API only for server-side requests. You can use it to run arbitrary JavaScript service worker code.

```suggestion
description: The `ServiceWorkerRegistration.sync` API registers a service worker callback to run only when the device is no longer offline. You can use this to let users continue using your app while offline, and synchronize with a server later when, when a connection is established.
```

captainbrosset (changes requested):

features/background-sync.yml
tropicadri: I merged your suggestion with the entry point, I'd go with the same. As far as the recommendations, when to use it (vs background fetch) and for server-side requests. I feel that veers into documentation for the feature and how to use it, which if I understand correctly is this would not be the right source. 
I would not add it because then devs might expect to have more information in this repo than its intent.

foolip: I've filed https://github.com/web-platform-dx/web-features/issues/1388 about how to name the entrypoint in this case. My take for this specific case is that the `sync` property entrypoint isn't necessary to spell out in this case, that specific piece of code will appear in different places depending on coding style, and is too short and generic to refer to the feature. (In contrast to something like `requestVideoFrameCallback()`.) The name of the `ServiceWorkerRegistration` interface also never shows up in code, so naming it isn't useful IMHO.

In other words, I think "The background sync API ..." is the best tradeoff here.

features/background-sync.yml
```diff
@@ -0,0 +1,13 @@
+name: Background sync
+description: The `ServiceWorkerRegistration.sync` API registers a service worker callback to run only when the device is no longer offline. You can use this to let users continue using your app while offline, and synchronize with a server later when, when a connection is established.
```
foolip, line 2:
```suggestion
description: The background synchronization API registers a service worker callback to run only when the device is no longer offline. You can use this to let users continue using your app while offline, and synchronize with a server later when, when a connection is established.
```

See https://github.com/web-platform-dx/web-features/pull/1363#discussion_r1675652337. @captainbrosset can you review? I don't want to have a suggestion battle :D

captainbrosset: I agree with your comment. The word SericeWorkerRegistration won't appear in code, and sync is too short. Let's stick with background synchronization API. More recognizable.

captainbrosset (approved):

features/background-sync.yml
```diff
@@ -0,0 +1,13 @@
+name: Background sync
+description: The background synchronization API registers a service worker callback to run only when the device is no longer offline. You can use this to let users continue using your app while offline, and synchronize with a server later when, when a connection is established.
```
ddbeck, line 2: Fixing the double "when" nudged me toward this suggestion for brevity. In any case, if the double "when" is fixed, then this is ready to merge.

```suggestion
description: The background synchronization API registers a service worker callback to run only when the device is no longer offline. You can use this to let users continue using your app while offline and synchronize with a server after reconnecting.
```

ddbeck:
@tropicadri do you plan to come back to this, or would you like someone to take it over?

PR #1364, opened 2024-07-11
Add <br> element

kosamari:
This PR adds `<br>` element as a feature


captainbrosset (changes requested):

captainbrosset (approved):

foolip (approved):
I think there's a good chance that we will later want to bundle this up with some other things into a bigger feature like "Basic HTML markup" but that's no reason to not add it as a feature. (We already have other features that should be combined with the passage of time.)

PR #1365, opened 2024-07-11
Adds `<canvas>` HTML element

petele:
Adds HTML `<canvas>` element


petele:
@ddbeck & @foolip - this one prob needs a little extra scrutiny for how I handled the `transferToOffscreen` method

foolip:
@petele I think `transferControlToOffscreen` ought to be in the offscreen canvas feature. It shipped with `OffscreenCanvas` and is one way of getting a `OffscreenCanvas` instance, in addition to the `OffscreenCanvas` constructor.

It's quite common for an API to have small bits on interfaces that mostly belong to another feature in this way. For example, https://w3c.github.io/selection-api/ has `window.getSelection()` and `document.getSelection()`, and those make sense together, not in whatever feature we end up putting `Window` and `Document` in.

features/canvas.yml
```diff
@@ -1,15 +1,16 @@
-draft_date: 2024-06-05
-name: <canvas>
-description: The `<canvas>` element TODO.
+name: Canvas API
+description: The canvas API provides a means for drawing graphics via JavaScript and the HTML `<canvas>` element.
 spec: https://html.spec.whatwg.org/multipage/canvas.html#canvas
 group: html
```
foolip, line 4:
```suggestion
group:
  - canvas
  - html
```

features/canvas.yml
```diff
@@ -1,15 +1,16 @@
-draft_date: 2024-06-05
-name: <canvas>
-description: The `<canvas>` element TODO.
+name: Canvas API
+description: The canvas API provides a means for drawing graphics via JavaScript and the HTML `<canvas>` element.
```
foolip, line 2: Should this feature represent just the HTML `<canvas>` element, or also 2D canvas?

I think we should add features for 2D canvas and WebGPU and put them on equal footing, even though 2D canvas is much older and shipped together with `<canvas>`. (Developers shouldn't need to know this in 2024.)

That would leave `<canvas>` as a feature that does nothing by itself, just a way to access other features. I think that's OK, but the name and description would be a bit different. @ddbeck thoughts?

features/canvas.yml
```diff
@@ -1,15 +1,16 @@
...
 spec: https://html.spec.whatwg.org/multipage/canvas.html#canvas
 group: html
 compat_features:
```
foolip, line 5: @petele can you add `compute_from: html.elements.canvas` in a `status` block here to get the computed status for the feature to be correct?

features/canvas.yml
ddbeck: Yes, I agree with this. The name here would return to `<canvas>` and the description might look something like this, where we gesture in the direction of other features you would use with this one:

```suggestion
name: Canvas API
description: The `<canvas>` HTML element represents a space to draw graphics in, using the 2D canvas API or 3D WebGPU API.
```

features/canvas.yml
petele: done

petele:
integrated feedback

features/offscreen-canvas.yml
```diff
@@ -4,7 +4,6 @@ spec: https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-int
 group: canvas
 caniuse: offscreencanvas
 compat_features:
-  - api.HTMLCanvasElement.transferControlToOffscreen
```
foolip, line 7: Please leave this feature with offscreen canvas.

features/canvas.yml
```diff
@@ -1,15 +1,21 @@
...
-name: <canvas>
-description: The `<canvas>` element TODO.
+name: Canvas API
+description: The `<canvas>` HTML element represents a space to draw graphics in, using the 2D canvas API or 3D WebGPU API.
 spec: https://html.spec.whatwg.org/multipage/canvas.html#canvas
-group: html
+group:
+  - canvas
+  - html
+caniuse: canvas
```
foolip, line 7:
```suggestion
# Note: https://caniuse.com/canvas covers both the <canvas> element and the 2D
# rendering context. These are separated in web-features, so no caniuse link here.
```

features/canvas.yml
```diff
@@ -1,15 +1,21 @@
...
+status:
+  compute_from: html.elements.canvas
 compat_features:
   - api.HTMLCanvasElement
+  - api.HTMLCanvasElement.captureStream
```
foolip, line 12: Let's leave this out and add a feature corresponding to https://caniuse.com/mediacapture-fromelement for it.

foolip: I'm adding that feature in https://github.com/web-platform-dx/web-features/pull/1456.

petele: Removed

features/canvas.yml
```diff
@@ -1,8 +1,13 @@
-draft_date: 2024-06-05
 name: <canvas>
-description: The `<canvas>` element TODO.
+description: The `<canvas>` HTML element represents a space to draw graphics in, using the 2D canvas API or 3D WebGPU API.
```
foolip, line 2:
```suggestion
description: The `<canvas>` HTML element is a space to draw graphics in, using the 2D canvas API, WebGL, or WebGPU.
```

PR #1366, opened 2024-07-11
Add list elements including `<ol>`, `<ul>`, and `<li>`

petele:


features/list-elements.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
-name: <li>
-description: The `<li>` element TODO.
+name: List elements.
```
foolip, line 1: The name of this is challenging in a similar way to `<h1>` through `<h6>`. I think that "&lt;ol> and &lt;ul>" similar to https://github.com/web-platform-dx/web-features/pull/1374 would be OK here, since `<li>` only makes sense in the context of another element. The description should mention all elements and the word "list" so that it can be found in search. The name of the file would be `ol-ul.yml`.

petele:
Updated based on feedback

features/list-elements.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
-name: <li>
-description: The `<li>` element TODO.
+name: <ol>, <ul>, and <li>.
```
captainbrosset, line 1: (this comment is here only because github doesn't let me comment on lines you didn't change)

How about using:

```
compute_from:
  - html.elements.ol
  - html.elements.ul
  - html.elements.li
```

This way we'd ignore the `reversed` attribute which makes the feature start at Chrome 18, Firefox 18, and Safari 6, but which really isn't needed in 99.9% of the cases.

captainbrosset (changes requested):

petele:
@ddbeck - would you please share the command line steps I need complete to merge in main, so I don't mess it up royally this time? 😄

features/list-elements.yml
```diff
@@ -1,11 +1,15 @@
...
+description: The elements to represent an ordered or unordered list.
 spec:
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element
 group: html
```
captainbrosset, line 7: Can you change this to belong to html-elements instead? This new group contains all HTML elements (same for your other PRs, if you can).

captainbrosset (approved):

features/list-elements.yml
ddbeck: @captainbrosset I opened https://github.com/web-platform-dx/web-features/pull/1467 to clean up the ones like this that have already merged.

petele: done

Issue #971, opened 2024-04-26
Composite features


foolip:
With https://github.com/web-platform-dx/web-features/pull/970 we'll have a 5 features relating to font synthesis control. That makes sense right now because the individual bits are all relatively and with different Baseline low dates.

When all of these features are Baseline high, it would make sense to create a composite font synthesis feature, and for consumers to show that instead of the individual features by default. One could also think of it as hidden subfeatures that could be revealed.

As we continue to expand web-features I'm sure we'll find cases like this that are ~3 years in the past where this would make sense, so we don't have to wait for font synthesis to be widely available before tackling this problem.

Related discussions:
https://github.com/web-platform-dx/web-features/issues/217
https://github.com/web-platform-dx/web-features/issues/495
https://github.com/web-platform-dx/web-features/issues/648
https://github.com/web-platform-dx/web-features/issues/866

foolip:
@ddbeck asked in https://github.com/web-platform-dx/web-features/pull/957#discussion_r1580805589:

> There's an open question of how consumers would see a composite feature. Would the composite feature contain a list of compat keys, the union of the subordinate features? Or would the composite feature contain references to the subordinate features only.

We should probably enumerate the features that make up a composite feature with a field like `composed_from`. But I'm not sure if the computed status of a composite feature should be based on the union of all compat keys, or if it starts from the status of the features being composed without ever touching BCD.

If we want composed features that have the "initial support" and "later additions" aspect being discussed in https://github.com/web-platform-dx/web-features/issues/866, then I suspect sticking to web-features identifiers might be better. It'd be just another level of the same kind of summarization. Or perhaps we can have syntax that means "expand the BCD keys of this feature here" and that could allow us to efficiently point to the initial/later parts of a feature, despite the computation happening in terms of BCD keys.

ddbeck:
Adding some terms to make this easier to find: feature combination, feature composition.

PR #1368, opened 2024-07-11
Add `<b>`, `<i>`, `<u>`, `<em>`, and `<strong>` elements

petele:


features/b.yml
```diff
@@ -0,0 +1,7 @@
+name: Bold element
+description: The `<b>` element is used to draw the reader's attention to the element's contents.
```
foolip, line 2: I don't think we should be shy about describing the feature in the way it's easiest to understand: it makes text bold. The definitions of some elements were contorted in the HTML5 area to not have anything purely presentational, but those semantics are quite abstract. I'd go with:

```suggestion
description: The `<b>` element makes text bold in order to draw the reader's attention to it.
```

And for the name, just "&lt;b>".

foolip (commented):
I wonder if we should throw these text formatting elements into a single feature or not? And what about `<s>`? I can't make up my mind. @ddbeck?

features/b.yml
captainbrosset: Haha, I haven't used or read about these elements in a very long time. Reading the MDN docs for `b`, `i`, and `u` shows how vague they now are: 

* "Bring Attention"
* "Idiomatic Text"
* "Unarticulated Annotation"

`s` is much clearer: "Strikethrough".

I think we should group them under just one feature only, and have a note in the `description` field about how they once were for text styling, but that CSS ways of achieving the same styles should now be used, and that semantic elements such as `strong` or `em` should be considered too.

I'm OK with being clear about the fact that, e.g., `b` makes the text bold, because that's what it's always done, and I don't see a user agent ever deviating from this. But I do think it's important for us to also say something about there being other, more current ways, of doing the same.

captainbrosset (changes requested):

features/b.yml
ddbeck: @foolip and I talked about the broader question earlier today: when should individual elements get their own feature? Are we really going to have individual features for these long-established things?

I think, in the near term, I favor individual features as a matter convenience until we recognize groups. I suspect we'll actually find our way to composite features (https://github.com/web-platform-dx/web-features/issues/971). So these elements may eventually find their way into something like "text annotation elements."

features/b.yml
```diff
@@ -0,0 +1,7 @@
...
+spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-b-element
+group: html
+compat_features:
+  - html.elements.b
+  - html.elements.strong
```
ddbeck, line 7: I'm _very_ uncomfortable with these being lumped together. `<b>` and `<strong>` have different semantics, despite the same default styling.

features/i.yml
```diff
@@ -1,7 +1,7 @@
-draft_date: 2024-06-05
-name: <i>
+name: Italic element
 description: The `<i>` element TODO.
 spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-i-element
 group: html
 compat_features:
   - html.elements.i
+  - html.elements.em
```
ddbeck, line 7: Same here. I think we'll get an earful if folks find out we're combining `<i>` and `<em>`. 😅

petele:
I split these each into their individual elements, and updated the descriptions based on the feedback provided.

foolip:
@petele there are conflicts due to https://github.com/web-platform-dx/web-features/pull/1482. Can you rebase and update dist?

features/b.yml.dist
```diff
@@ -0,0 +1,17 @@
+# Generated from: b.yml
```
foolip, line 1: @petele can you remove the orphaned features/draft/html-elements/b.yml.dist file?

features/em.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <em>
-description: The `<em>` element TODO.
+description: The `<em>` element element makes text italic in order to identify idomatic content.
```
foolip, line 2: It would be good to use "emphasis" or "emphasize" in the description for searchability. I'll also suggest descriptions that put the styling effect first for `<b>`, `<i>`, and `<u>`, but the semantics first for `<em>` and `<strong>`.

So in this case:

```suggestion
description: The `<em>` HTML element emphasizes or stresses the content, rendering text as italic by default.
```

features/u.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <u>
-description: The `<u>` element TODO.
+description: The `<u>` HTML element makes text appear with an underline.
```
foolip, line 2: Perhaps even simpler, but also accounting for things other than text:

```suggestion
description: The `<u>` HTML element underlines text and other content.
```

features/strong.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <strong>
-description: The `<strong>` element TODO.
+description: The `<strong>` element brings attention to the element's contents, usually rendered as bold text.
```
foolip, line 2:
```suggestion
description: The `<strong>` HTML element brings attention to the content, rendering text as bold by default.
```

features/i.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <i>
-description: The `<i>` element TODO.
+description: The `<i>` element element makes text italic in order to identify idomatic content.
```
foolip, line 2:
```suggestion
description: The `<i>` HTML element makes text italic in order to identify idiomatic content.
```

features/b.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <b>
-description: The `<b>` element TODO.
+description: The `<b>` element makes text bold in order to draw the reader's attention to it.
```
foolip, line 2:
```suggestion
description: The `<b>` HTML element makes text bold in order to draw the reader's attention to it.
```

foolip (approved):
@ddbeck do you want to review the descriptions too? I think we'll be setting some precedents here.

features/b.yml.dist
petele: done!

features/em.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <em>
-description: The `<em>` element TODO.
+description: The `<em>` HTML element emphasizes or stresses the content, rendering text as italic by default.
```
ddbeck, line 2:
```suggestion
description: The `<em>` HTML element emphasizes or stresses the content, styling text as italic by default.
```

@foolip asked me to chime in here, on the convention we're going to establish via https://github.com/web-platform-dx/web-features/pull/1403. I'm putting forward "styling" instead of "rendering" because I'd like to emphasize that this is a style that can be overridden by the page author and not solely the domain of the browser's rendering engine.

features/strong.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <strong>
-description: The `<strong>` element TODO.
+description: The `<strong>` HTML element brings attention to the content, rendering text as bold by default.
```
ddbeck, line 2:
```suggestion
description: The `<strong>` HTML element brings attention to the content, styling text as bold by default.
```

features/b.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <b>
-description: The `<b>` element TODO.
+description: The `<b>` HTML element makes text bold in order to draw the reader's attention to it.
```
ddbeck, line 2: Optionally, we could align all of these to use the same pattern, like this:

```suggestion
description: The `<b>` HTML element draws reader's attention to text, styling text as bold by default.
```

features/i.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <i>
-description: The `<i>` element TODO.
+description: The `<i>` HTML element makes text italic in order to identify idiomatic content.
```
ddbeck, line 2: Optionally, we could align all of these to use the same pattern, like this:

```suggestion
description: The `<i>` HTML element identifies idiomatic content, styling text as italic by default.
```

features/u.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <u>
-description: The `<u>` element TODO.
+description: The `<u>` HTML element underlines text and other content.
```
ddbeck, line 2: Optionally, we could align all of these to use the same pattern, like this:

```suggestion
description: The `<u>` HTML element makes non-textual annotations, styling text as underlined by default.
```

features/em.yml
petele: Done for all.

foolip (approved):
Let's do it! Styling text by default!

foolip:
@captainbrosset do you want to review this again? You have previously requested changes, and I'd need to dismiss that review in order to merge.

ddbeck (approved):

PR #1369, opened 2024-07-11
Adds table and related elements

petele:


foolip:
@petele can you use `compute_from: html.elements.table` to anchor the feature to the earliest support of `<table>`? Right now it's dragged forward by a single method.

features/table.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: Tables
-description: The `<table>` element TODO.
+description: The `<table>` HTML element represents tabular data—that is, information presented in a two-dimensional table comprised of rows and columns of cells containing data.
```
ddbeck, line 2: My suggestions here are toward brevity and acknowledging that this is more than just one element. What do you think of this instead?

```suggestion
description: The `<table>` HTML element, with several related elements, represents tabular data in rows and columns of cells.
```

petele:
Added `compute_from`

foolip (approved):
I've checked that `createTBody` was indeed shipped in Chrome 20 using http://mdn-bcd-collector.gooborg.com/tests/api/HTMLTableElement/createTBody, so that outlier isn't an error in BCD.

PR #1370, opened 2024-07-11
Add <wbr> element

petele:


foolip:
`<wbr>` shows up in https://github.com/mdn/browser-compat-data/issues/23796 so I'm not sure about the support in Edge. @petele what I would do here is try `<wbr>` in Edge 18 in BrowserStack to see if BCD is wrong. Do you have access to BrowserStack?

petele:
I don't think I have BrowserStack access :(

foolip:
OK, I copied the MDN example into https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=12893 and tried that in Edge 15. It works. I doesn't work in IE11 however.

I think the best guess we can make is that support was added in Edge 12 since that's when most of the changes between IE and Edge happened.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/23804 to update BCD. That will change the Baseline low date, so we need to wait for that to reach us via a BCD release.

features/wbr.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <wbr>
-description: The `<wbr>` element TODO.
+description: The `<wbr>` HTML element represents a word break opportunity.
```
ddbeck, line 2: Do you think it'd be excessive to elaborate on this?

```suggestion
description: The `<wbr>` HTML element represents a word break opportunity, to explicitly mark a place within a word where a line might be broken.
```

foolip:
Still waiting for https://github.com/mdn/browser-compat-data/pull/23804...

petele:
@foolip - can you add blocked to this one please?

foolip:
The Edge version fix came in with https://github.com/web-platform-dx/web-features/pull/1482. @petele can you rebase and update dist?

petele:
done

foolip (approved):
There's a minor discrepancy in the Safari version with caniuse, but too small to investigate I think.

PR #1371, opened 2024-07-11
Add <figure> and <figcaption>

petele:


features/figure.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <figure>
-description: The `<figure>` element TODO.
+description: The `<figure>` HTML element represents self-contained content, potentially with an optional `<figcaption>`.
 spec:
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-figcaption-element
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-figure-element
```
foolip, line 5:
```suggestion
  - https://html.spec.whatwg.org/multipage/grouping-content.html#the-figure-element
  - https://html.spec.whatwg.org/multipage/grouping-content.html#the-figcaption-element
```

features/figure.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <figure>
-description: The `<figure>` element TODO.
+description: The `<figure>` HTML element represents self-contained content, potentially with an optional `<figcaption>`.
```
foolip, line 2: I think this would benefit from being more concrete, like "picture with text under it" concrete, even `<figure>` can be used for other things too. I'll leave it to @ddbeck to suggest better wording though :)

ddbeck: How about this?

```suggestion
description: The `<figure>` and `<figcaption>` HTML elements represent an illustration, diagram, text, or other self-contained reference content, with an optional caption.
```

features/figure.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <figure>
```
ddbeck, line 1:
```suggestion
name: <figure> and <figcaption>
```

petele:
addressed feedback.

ddbeck (approved):

PR #1373, opened 2024-07-11
Add <div> element

petele:


foolip (commented):
The one we've all been waiting for! @petele can you throw in `compute_from: html.elements.div` so that the status is the earliest version of every browser?

(It's possible that the `HTMLDivElement` interface was originally not there, and we don't need to research that because it doesn't matter.)

petele:
added, ready for review

captainbrosset (approved):

features/div.yml
```diff
@@ -1,8 +1,9 @@
-draft_date: 2024-06-05
 name: <div>
-description: The `<div>` element TODO.
+description: The `<div>` HTML element is the generic container for flow content.
```
ddbeck, line 2: I'm guessing few developers could give a clear definition of "flow content." Feels a bit like spec speak. What do you think of one of these?

Just being more casual:

```suggestion
description: The `<div>` HTML element is the generic container for content.
```

Or defining the term a bit:

```suggestion
description: The `<div>` HTML element is the generic container for flow content (text and most elements typically found in `<body>`).
```

petele: I went with the more generic since people use div's for more than just text.

foolip (approved):

foolip:
And it's done! Now people can search for `<div>` and be impressed that we've covered it :D

PR #1374, opened 2024-07-11
Add <sub> and <sup>

petele:


features/sub-sub.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <sub> and <sup>
-description: The `<sub>` and `<sup>` elements TODO.
+description: The `<sub>` and `<sup>` elements specifies inline text which should be displayed as subscript or superscript for solely typographical reasons.
```
foolip, line 2: @ddbeck this runs afoul of https://github.com/web-platform-dx/web-features/tree/main/docs#specifies but "sets" wouldn't work here. Do you think this description is OK and do you want to change the guideline?

foolip (approved):

features/sub-sub.yml
petele: changed to sets

petele:
i think this is good to merge! 😄

foolip:
Oops, I see there's a typo in the filename. Will fix.

PR #1378, opened 2024-07-11
Add <title> element

petele:


features/title.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <title>
-description: The `<title>` element TODO.
+description: The `<title>` element defines the document's title that is shown in a browser's title bar or a page's tab.
```
captainbrosset, line 2: Might be worth noting that the title is also a very important accessibility aspect of an HTML document, which users of assistive tech use to quickly know what the page is about before reading more of the content (which is more time consuming).
The title is also used in the titlebar of an installed desktop PWA (not sure about mobile). So perhaps we can generalize a bit here.
Finally, let's try to avoid `define`, see https://github.com/web-platform-dx/web-features/tree/main/docs#defines.

```suggestion
description: The `<title>` element sets the title of the document and exposes it to users via the UI of the browser or app they're using and through assistive technology.
```

I'm not sold on my suggestion, but having trouble finding something clearer for now. I'm sure @ddbeck can help find something too.

captainbrosset (changes requested):

petele:
Updated based on feedback.

features/title.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <title>
-description: The `<title>` element TODO.
+description: The `<title>` element sets the title of the document and exposes it to users via the UI of the browser or app they're using and through assistive technology.
 spec: https://html.spec.whatwg.org/multipage/semantics.html#the-title-element
 group: html
 compat_features:
```
captainbrosset, line 5: Somehow GitHub doesn't show me the content of the dist file here, only that it moved. So I didn't see this the first time I reviewed. But it seems like adding the two `api.*` keys in here causes the feature to appear as supported starting with Safari 3 only.
It's OK since the feature is baseline high either way, but adding `compute_from: html.elements.title` would make the feature start at the earliest version of all browsers, which would make more sense here.

captainbrosset (approved):
Approving with a comment to add compute_from.

features/title.yml
petele: done!

captainbrosset (approved):

PR #1379, opened 2024-07-11
Add <samp>, <small>, and <var>

petele:


features/small.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <small>
-description: The `<small>` element TODO.
+description: The `<small>` element represents side-comments and small print, like copyright and legal text, independent of its styled presentation.
```
foolip, line 2: Can you also say that it makes text smaller?

features/samp.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <samp>
-description: The `<samp>` element TODO.
+description: The `<samp>` element is used to enclose inline text which represents sample (or quoted) output from a computer program.
```
foolip, line 2: Can you say what the default styling is? Monospace.

ddbeck: I think it'd be nice to have a conventional way to describe default styling. Copying myself from another PR, but how about this pattern: _Typically styled as_ [description] _by default_.

(And [avoiding is used to](https://github.com/web-platform-dx/web-features/blob/fbd5bca14baac3c2312d230d58deedd541b96139/docs/README.md#is-used-to), sorry I will be a broken record about this.)

```suggestion
description: The `<samp>` element represents an in-line sample (or quoted) output from a computer program. Typically styled as monospaced type by default.
```

features/small.yml
ddbeck:
```suggestion
description: The `<small>` element represents side-comments and small print, like copyright and legal text, independent of its styled presentation. Typically styled in a reduced font size by default.
```

petele:
updated based on feedback

ddbeck (approved):

PR #1380, opened 2024-07-11
Adding the Range as a feature

tropicadri:


features/range.yml
```diff
@@ -0,0 +1,33 @@
+name: Range
+description: The Range interface represents a fragment of a document that can contain nodes and parts of text nodes.
```
foolip, line 2: `Range` is defined in https://dom.spec.whatwg.org/#interface-range, but on its own it doesn't really do much, it's mostly in support of https://w3c.github.io/selection-api/.

I we may need a DOM feature for most old parts of https://dom.spec.whatwg.org/ that other specs depend on. @ddbeck WDYT?

foolip:
@tropicadri I think we can land the `XMLSerializer` feature before we figure out what to do with `Range`. Can you separate the `Range` bits into a separate PR?

features/range.yml
ddbeck: Until we're backed into a corner, I think we ought to put such interfaces with their most longstanding dependent features (e.g., unless someone has an older use-case, `Range` can go with the selection API—not sure what use `Range` would be alone).

That said, if there's no such feature or it has use alone—supposing I'm wrong about `Range` here—then I'd be OK with a foundational DOM feature. It might be quite boring, but make a lot of sense to stand in contrast with the Shadow DOM feature, with "Also known as light DOM" description text.

features/range.yml
```diff
@@ -0,0 +1,33 @@
+name: Range
```
ddbeck, line 1: Since we haven't got an answer to whether this should standalone yet, but I also don't want to let this sit idle for forever, here's my idea for how to proceed with this:

Let's move this in to the `draft` folder, add the draft date, and add a comment that links to this PR's discussion. That way, when we do get an answer to whether or not we need some "foundational DOM" feature (or features), then we have at least this part of the work documented somewhere.

ddbeck (changes requested):

PR #1381, opened 2024-07-11
Add <cite>, <p>, <pre>, and <q>

petele:


features/p.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <p>
-description: The `<p>` element TODO.
+description: The `<p>` element represents a paragraph.
```
foolip, line 2:
```suggestion
description: The `<p>` element represents a paragraph of text.
```

features/pre.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <pre>
-description: The `<pre>` element TODO.
+description: The `<pre>` element epresents preformatted text which is to be presented exactly as written in the HTML file.
```
foolip, line 2:
```suggestion
description: The `<pre>` element represents preformatted text which is to be presented exactly as written in the HTML file.
```

I wonder if it would be better to focus on how whitespace is preserved rather than saying that it's exactly as written. Unlike `<xmp>`, the normal rules of parsing aren't suspended inside `<pre>`.

features/q.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <q>
-description: The `<q>` element TODO.
+description: The `<q>` element  indicates that the enclosed text is a short inline quotation.
```
foolip, line 2:
```suggestion
description: The `<q>` element represents a short inline quotation.
```

features/cite.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <cite>
-description: The `<cite>` element TODO.
+description: The `<cite>` element is used to mark up the title of a cited creative work.
```
foolip, line 2: I wonder if we should say that it makes the text italics by default?

ddbeck: In general, I don't think we need to be consistent with MDN. As far as I know, there's never been a systematic review of those descriptions since the wiki days, so there's a good chance they're not well-considered to begin with.

I have a few style ideas here:

- [Avoiding "is used to"](https://github.com/web-platform-dx/web-features/blob/fbd5bca14baac3c2312d230d58deedd541b96139/docs/README.md#is-used-to)
- Avoiding circular descriptions (e.g., if you're not sure what "\<cite>" — the name — refers to, then our description ought to help)
- Giving an example
- Represents for consistency with the other items here

```suggestion
description: The `<cite>` element represents the title of a quoted, referenced, or mentioned creative work, such as a film or book.
```

If we want to mention default styling, we could do something like:

```suggestion
description: The `<cite>` element represents the title of a quoted, referenced, or mentioned creative work, such as a film or book. Typically styled in italic type by default.
```

features/pre.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <pre>
-description: The `<pre>` element TODO.
+description: The `<pre>` element represents a block element of preformatted text which is presented exactly as written, including all white space.
```
ddbeck, line 2:
```suggestion
description: The `<pre>` element represents a block element of preformatted text that is presented exactly as written, including all white space.
```

petele:
updated based on feedback

petele:
@foolip & @ddbeck - this is failing with the error: 
`/features/cite: must have required property 'status'`

Do we need to add `status` to all features now? And if so, what is the value we should set there?

foolip:
@petele yes, `status` is required. But you don't need to write one yourself, just generate the dist file with `npm run dist` and commit it. In this PR they've been removed instead of moved, and that's the problem.

petele:
> @petele yes, `status` is required. 

Ah - got it. From the error message, it wasn't clear if the error was in the `.yml` file, or the `.yml.dist` file, and I assumed it should be in the `.yml`.

Fixed.

features/cite.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <cite>
-description: The `<cite>` element TODO.
+description: The `<cite>` element represents the title of a quoted, referenced, or mentioned creative work, such as a film or book.
 spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-cite-element
 group: html
```
ddbeck:
```suggestion
group: html-elements
```

features/p.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <p>
-description: The `<p>` element TODO.
+description: The `<p>` element represents a paragraph of text.
 spec: https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
 group: html
```
ddbeck:
```suggestion
group: html-elements
```

features/pre.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <pre>
-description: The `<pre>` element TODO.
+description: The `<pre>` element represents a block element of preformatted text that is presented exactly as written, including all white space.
 spec: https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element
 group: html
```
ddbeck:
```suggestion
group: html-elements
```

features/q.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <q>
-description: The `<q>` element TODO.
+description: The `<q>` element represents a short inline quotation.
 spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-q-element
 group: html
```
ddbeck:
```suggestion
group: html-elements
```

ddbeck (approved):
OK, I'm suggesting the group changes for consistency with https://github.com/web-platform-dx/web-features/pull/1467. But after those changes are applied, this is ready.

ddbeck:
@petele you read my mind on the group changes! Thank you!

PR #1382, opened 2024-07-11
Add <head> and <meta> tags

petele:


features/head.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <head>
-description: The `<head>` element TODO.
+description: The `<head>` element contains machine-readable information (metadata) about the document.
```
foolip, line 2: For web developers, isn't scripts and stylesheets the most important things in `<head>`? I'll leave this description to @ddbeck's review.

features/meta.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <meta>
-description: The `<meta>` element TODO.
+description: The `<meta>` element represents metadata that cannot be represented by other HTML meta-related elements.
```
foolip, line 2: This description mainly says what `<meta>` isn't, not what it is. What's the most common use of `<meta>` and can we put that in the description?

petele:
Tweaked description based on feedback.

features/meta.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <meta>
-description: The `<meta>` element TODO.
+description: The `<meta>` element represents metadata used by the browser or search engines about the page including description, keywords, character sets, etc.
```
ddbeck, line 2: Ooh, nice update!

Style nits:

- Keep "about the page" with its referent
- Avoid Latin abbreviations

```suggestion
description: The `<meta>` element represents metadata about the page used by the browser or search engines, including description, keywords, and character sets.
```

Happy to see this merge soon!

ddbeck (approved):
Appro

petele:
Updated based on feedback, I think its ready to go.

petele:
Note - I split `theme-color` into its own feature. it landed much later, is only supported on mobile, or in standalone web apps, and it was cause meta to be considered not baseline.

features/meta-theme-color.yml.dist
```diff
@@ -0,0 +1,11 @@
+# Generated from: meta-theme-color.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome_android: "80"
```
foolip, line 7: https://chromestatus.com/feature/5398016231997440 says Chrome 39, and https://caniuse.com/meta-theme-color doesn't have data for Chrome Android. Is 80 the correct version?

features/meta-theme-color.yml
```diff
@@ -0,0 +1,7 @@
+name: <meta> Theme color
+description: The `<meta>` theme color is used to customize the display of the page or the surrounding browser user interface.
```
foolip, line 2: We already have features named `<link rel="modulepreload">` and `<input type="search">` so since this is also about a specific attribute value, we could call it `<meta name="theme-color">`:

```suggestion
name: '<meta name="theme-color">'
description: 'The "theme-color" `<meta>` name allows you to customize the display of the page or the surrounding browser user interface.'
```

It is pushing the limits of how syntax-y names are reasonable, but I don't know what else to suggest.

features/meta.yml
```diff
@@ -19,4 +20,3 @@ compat_features:
   - html.elements.meta.name
   - html.elements.meta.name.color-scheme
```
foolip, line 21: Can you exclude color-scheme? It was introduced around the time of the CSS property by the same name and should probably be bunched in with that feature, or be a new feature.

features/meta-theme-color.yml
petele: Will move theme color to its own PR so to not block this one.

features/meta.yml
petele: done.

features/meta-theme-color.yml.dist
petele: moving to other PR

PR #1383, opened 2024-07-11
Add Description List as a feature 

kosamari:
This PR adds description list as a feature. 

Note: in draft features for HTML, this was labled as `dd` element, so I hanged name and file name as well.


features/description-list.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
-name: <dd>
-description: The `<dd>` element TODO.
+name: Description list
+description: Elements to represent a description list.
```
captainbrosset, line 2:
```suggestion
description: The `<dl>`, `<dt>`, and `<dd>` elements create a list of terms and their associated descriptions. Use a description list to to create glossaries or other similar key-value pair lists.
```

Just a suggestion to make it more straightforward to understand. Not completely set on it though.

One more thing to consider: accessibility doesn't seem to be the best with DL. See https://adrianroselli.com/2022/12/brief-note-on-description-list-support.html and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#accessibility
In particular, it looks like keyboard navigation isn't supported everywhere.

Baseline currently doesn't cover assistive tech (see https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#non-goals) so this is probably fine as is, but it might be good to do a bit of research for a11y best practices with DL and add a short mention of this in the description.

captainbrosset (changes requested):

PR #1385, opened 2024-07-11
Add <hr>

petele:


features/hr.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <hr>
-description: The `<hr>` element TODO.
+description: The `<hr>` element represents a thematic break between paragraph-level elements.
```
ddbeck, line 2: Two ideas here:

- to aim for brevity on the main concept here (paragraphs)
- to acknowledge the historic abbreviation (for search) and the default behavior

```suggestion
description: The `<hr>` element represents a thematic break between paragraphs, often shown as a horizontal rule.
```

features/hr.yml
```diff
@@ -1,6 +1,5 @@
...
 spec: https://html.spec.whatwg.org/multipage/grouping-content.html#the-hr-element
 group: html
```
ddbeck, line 4: Not a blocker, but I'm wondering if there's a second group to be had here, perhaps with `<br>` (https://github.com/web-platform-dx/web-features/pull/1364)? I'm struggling to come up with a name though. "Text breaks"?

ddbeck (approved):
Thanks for jumping in on feature authoring! A couple of ideas to consider in comments here, but this would be OK to merge.

features/hr.yml
petele: I lean towards keeping them separate, as they have different visual and semantic meanings.

petele:
Addressed feedback.

captainbrosset (approved):

PR #1386, opened 2024-07-11
Adding Keyboard interface feature

tropicadri:


foolip:
@tropicadri can you split this into two features, one for `navigator.keyboard.lock()` (https://wicg.github.io/keyboard-lock/) and another for `navigator.keyboard.getLayoutMap()` (https://wicg.github.io/keyboard-map/)? Even though they shipped around the same time in Chrome, they have distinct use cases. Describing them is probably easier as separate features.

tropicadri:
@foolip done the splitting thanks!

features/keyboard-lock.yml
```diff
@@ -0,0 +1,8 @@
+name: Keyboard lock
+description: Keyboard lock allows websites to capture keys that are normally reserved by the underlying host operating system. It is intended to be used by web applications that provide a fullscreen immersive experience.
```
foolip, line 2: https://github.com/web-platform-dx/web-features/tree/main/docs#allows discourages use of "allows" in most cases except "allows you". https://github.com/web-platform-dx/web-features/tree/main/docs#descriptions also recommends "The Thing TypeOfThing" for most descriptions. Applying both here:

```suggestion
description: The `navigator.keyboard.lock()` API allows you to capture keys that are normally reserved by the operating system. It can be used to provide an immersive fullscreen experience such as games.
```

features/keyboard-map.yml
```diff
@@ -0,0 +1,6 @@
+name: Keyboard map
+description: Keyboard map allows websites to convert from a given `code` value to a valid `key` value that can be shown to the user to identify the given key. It is intended to be used by web applications that want to treat the keyboard as a set of buttons and need to describe those buttons to the user.
```
foolip, line 2:
```suggestion
description: The `navigator.keyboard.getLayoutMap()` API returns a map from key codes to human readable key names. It can be used to describe buttons to the user when using the keyboard as a set of buttons, such as in games.
```

foolip (approved):
Some description suggestions, but the feature groups are exactly what I had in mind, and the computed statuses too.

foolip (approved):

PR #1390, opened 2024-07-12
Introduce a scrolling group

captainbrosset:
This is a proposal to create a scrolling group. I was originally going to create it only for the scroll color, gutter, width properties, but then found out that we had many more features that have to do with scrolling.

This group doesn't represent a single spec, but I think it might make sense to web devs.
Of course, this depends on how we think consumers would use groups.


features/loading-lazy.yml
```diff
@@ -2,4 +2,5 @@ name: Lazy-loading images and iframes
 description: |
   The `loading="lazy"` attribute for `<img>` and `<iframe>` elements blocks loading the external resource until the user scrolls to that element's part of the page.
 spec: https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes
+group: scrolling
```
foolip, line 5: There is a connection to scrolling with this feature, but I think it's primarily a loading performance feature. If we had a primary group for that I think it'd be OK to also list it in the scrolling group, but to have it as only scrolling feels odd to me. WDYT?

foolip (approved):

features/loading-lazy.yml
captainbrosset: Agreed, I was on the fence with this one. Ok to remove.

PR #1391, opened 2024-07-12
Update spec drafts

foolip:


foolip:
This was to test if merging a bunch of PRs and then updating drafts would produce a sensible result. And it does. One case that isn't handled yet but hasn't happened here is if all keys in a spec are claimed. Those files should be deleted, but aren't.

ddbeck (approved):
Extremely satisfying! Let me know if you want a re-review if you tend to deal with the not-deleted file case, otherwise please merge.

PR #1392, opened 2024-07-12
New HTML elements and web components groups

captainbrosset:
Looking at groups and ungrouped features at https://web-platform-dx.github.io/web-features-explorer/groups/, I felt compelled to create these 2 new groups.
Let me know if you like this.


groups/web-components.yml
```diff
@@ -0,0 +1 @@
+name: Web Components
```
ddbeck, line 1: Interesting. I kinda think of web components as a composite feature: a web component is a distinct thing you can make or use, made up of smaller things you can use, rather than a group related-but-independent things.

See also: https://github.com/web-platform-dx/web-features/issues/971

groups/custom-elements.yml
```diff
@@ -1,3 +1,3 @@
 # https://html.spec.whatwg.org/multipage/custom-elements.html
 name: Custom elements
-parent: html
+parent: web-components
```
ddbeck, line 3: I think it's weird that custom elements are not part of `elements`.

groups/web-components.yml
captainbrosset: To be honest, I'm having a hard time getting a sense of the differences between a composite feature and a group.

Thinking more, I can see some differences:

- Groups don't get a status, they're mostly a way to navigate features. Composite features will get a status (once #971 is resolved).
- Groups can be very broad, like "html", whereas composite features remain at the individual thing a dev can use level.

If I picture myself navigating the list of features, I think I'd expect to find a group that's about web components somehow. Maybe that group should be the custom-elements group, available under `HTML > Elements > Custom elements`. But that wouldn't give me the list of everything I need to know in order to make a custom element. Shadow DOM isn't part of that group for example, nor is `<template>`.

Maybe I'm thinking about groups too much as a learning tool.

Maybe we try this web components group for now, and then re-evaluate once we have composite features?

groups/custom-elements.yml
foolip: We could allow groups to have multiple parents if that's required to match reality here. However, I would consider custom elements as a sibling group to built-in elements.

I'm not sure about the relationship between custom elements and web components. To start out, can we leave custom elements with HTML as the parent group, and use both custom elements and web components as groups for the features that belong in both?

groups/elements.yml
```diff
@@ -0,0 +1,2 @@
+name: HTML elements
```
foolip, line 1: How should the media elements group relate to this group? `<audio>` and `<video>` are elements, but the other things in the media elements group are `HTMLMediaElement` APIs.

captainbrosset: I would perhaps leave the media-elements group untouched, but make `<audio>` and `<video>` be both in elements and media-elements.
I guess my first intent here is, with the large number of new HTML elements being added, to have a nice group that contains them all.

foolip: That makes sense to me!

foolip:
Sorry I didn't consider this in the first round of review, but I think it would be prudent to name this group `html-elements`. It seems likely that we'll have features for some SVG elements like `<radialGradient>`, and we may want an `svg-elements` group for those. Same for MathML.

captainbrosset:
> Sorry I didn't consider this in the first round of review, but I think it would be prudent to name this group `html-elements`. It seems likely that we'll have features for some SVG elements like `<radialGradient>`, and we may want an `svg-elements` group for those. Same for MathML.

Good point. Done.

PR #1400, opened 2024-07-16
Uncomment keys and compute_from (features A-C)

ddbeck:
I'm running a little experiment here: I've searched for keys we've commented out and started uncommenting them and adding a `compute_from` override. This is what it looks like for the first 8 features (found in my editor with the search for `# ` in `features/{a,b,c}*.yml`).


features/counter-style.yml.dist
```diff
@@ -21,7 +32,9 @@ compat_features:
   - api.CSSCounterStyleRule.pad
   - api.CSSCounterStyleRule.prefix
   - api.CSSCounterStyleRule.range
+  - api.CSSCounterStyleRule.speakAs
   - api.CSSCounterStyleRule.suffix
+  - api.CSSCounterStyleRule.symbols
```
ddbeck, line 37: This is by far the weirdest bit. I don't know if we should be troubled by properties for unsupported descriptors, but it's at least true (it made me stop to check that these are indeed on `CSSCounterStyleRule.prototype`).

features/async-clipboard.yml
```diff
@@ -20,9 +20,8 @@ status:
     safari_ios: "13.4"
 compat_features:
   - api.Clipboard
-  # TODO: include after https://github.com/mdn/browser-compat-data/pull/23593
```
captainbrosset, line 23: I have a feeling we should keep this comments, and others like it. There's some good research in there, which I would hate to lose.

features/constructed-stylesheets.yml
```diff
@@ -3,13 +3,12 @@ description: The `CSSStyleSheet` constructor creates a new stylesheet which can
...
   - https://drafts.csswg.org/cssom-1/#dom-documentorshadowroot-adoptedstylesheets
+status:
+  compute_from: api.CSSStyleSheet.CSSStyleSheet
 compat_features:
   - api.CSSStyleSheet.CSSStyleSheet
   - api.CSSStyleSheet.replace
   - api.CSSStyleSheet.replaceSync
   - api.Document.adoptedStyleSheets
   - api.ShadowRoot.adoptedStyleSheets
-  # Setting the base URL is probably not important for typical uses of
```
captainbrosset, line 12: Same here, and elsewhere. It would be nice if we had a way to keep these editorial notes in there somewhere, maybe in a more structured form.
It would be nice to see them in the dist files in fact.What do you think?

ddbeck: > It would be nice if we had a way to keep these editorial notes in there somewhere, maybe in a more structured form.

I agree that it would be nice, though I think `compute_from` is going to handle the most common case, by marking the boundary after which things are later additions. So I'm doubtful that every comment saying "this is a later addition" will make sense to maintain.


Additionally, a lot of `compat_features` lists will eventually live as tags in BCD. So we'd need a separate way to list compat key commentary.

captainbrosset: Agreed that just having a comment that says "later addition" isn't useful. I'm more concerned about the various links and pieces of info that we spent time digging up when researching a feature.

captainbrosset (approved):
Looks good to me for these features.
I don't think my proposal to have structured comments for BCD keys is practical. But I do think that we should continue to document our findings as comments, and probably put them near `compute_from`, for the more complicated cases.

features/async-clipboard.yml.dist
```diff
@@ -56,6 +56,23 @@ compat_features:
   #   safari_ios: "13.4"
   - api.ClipboardItem.ClipboardItem
 
+  # baseline: false
```
foolip, line 59: This is wrong, because https://github.com/mdn/browser-compat-data/pull/23593 hasn't reached us through a BCD release yet. Let's wait for that and not publish an incorrect status in the interim.

features/counter-style.yml.dist
foolip: It's surprising that this has been exposed in CSSOM in both Chrome and Firefox if they don't actually work. But if that's true, I think they ought to be `partial_implementation` in BCD. I think it'd be best to not include the keys until this is investigated.

foolip (changes requested):

features/counter-style.yml.dist
ddbeck: I looked at this more closely.

For `symbols`, the CSSOM story is accurate: the CSSOM exposure works in Firefox and Chrome, but the descriptor is only partially implemented. Implementers other than Safari don't support `<image>` values, and this seems to be correct.

For `speak-as`, I've sent https://github.com/mdn/browser-compat-data/pull/23857.

We can wait on the latter PR, or I can comment out the key until it lands. Do you have a preference (and maybe we should turn that preference into a policy)?

features/async-clipboard.yml.dist
ddbeck: Fixed with e322f9f4e934960d5f7a9fd7cef2f33f8c44b036 and 9d9776682d58898f8fe81f90fd9821ef9c72ab15.

features/counter-style.yml.dist
foolip: I think waiting for BCD slows us down and we need a policy that allows us to think about these things once. And the risk to minimize is the risk that whatever we assume will happen doesn't happen, and we never notice we got it wrong.

I think these won't be Baseline anyway, so OK to just proceed,

foolip (approved):

PR #1403, opened 2024-07-16
Add description guidelines for default styles

ddbeck:
Inspired by https://github.com/web-platform-dx/web-features/pull/1381/ and https://github.com/web-platform-dx/web-features/pull/1379, this would bless a specific way to talk about default styles, when common and relevant.


docs/README.md
```diff
@@ -192,6 +192,17 @@ Avoid.
 See [sets](#sets).
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635491))
 
+##### typically styled as … by default
+
+Use this phrase (or a closely similarly one like "typically styled in") to note unspecified but conventional browser default styling.
```
foolip, line 197: The default styling is actually specified in detail and quite interoperable. For `<small>`, it's here:

https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3

I think I'd prefer us to say just "by default" for these cases.

foolip (changes requested):

docs/README.md
ddbeck: Oh fascinating! That's worth amending over. Changes forthcoming

docs/README.md
```diff
@@ -192,6 +192,17 @@ Avoid.
 See [sets](#sets).
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635491))
 
+##### styled as … by default
```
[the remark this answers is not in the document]
ddbeck: I've suggested a small change of your approach and updated this PR to reflect it. I will monitor for additional feedback. 😄

foolip (approved):

PR #1404, opened 2024-07-16
Add .nvmrc file to specify node version

petele:
Quickly select the right version of node when using `nvm` for the repo.


ddbeck:
Huh, does nvm support `.tool-versions`? I use asdf for this and have it ignored, but it would be nice to check in something like it.

(I'll see if asdf could use .nvmrc too.)

petele:
Interesting - I don't use asdf, and am not familiar with it. But, it doesn't look like it does. :(

.nvmrc
```diff
@@ -0,0 +1 @@
+22
```
ddbeck, line 1: OK, I couldn't find a way to consolidate on a single file. How about we just set this to be consistent with `.node-version` and be done with it?

```suggestion
18
```

petele: I tried running it with 18.16.1 and get an error that I don't get with 22.4.1. @tropicadri mentioned she was having problems with 18 as well and that 22 worked for her. I'm OK with going with 18, but we'll need to specify a subversion as well. 

Which version are you running?

```
file:///Users/petele/Sites/web-features/packages/compute-baseline/dist/browser-compat-data/compat.js:1
import bcd from "@mdn/browser-compat-data" with { type: "json" };
                                           ^^^^

SyntaxError: Unexpected token 'with'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)

Node.js v18.16.1
```

ddbeck: Ah ha. It needs to be >=18.20, which is when import attributes landed in Node.js.

ddbeck (approved):

PR #1406, opened 2024-07-16
Adds `<a>` element

petele:


features/a.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <a>
-description: The `<a>` element TODO.
+description: The `<a>` element, creates a hyperlink to anything accessible via a URL such as external web pages, files, email addresses, locations in the same page, and more.
 spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element
 group: html
 compat_features:
```
captainbrosset, line 5: We should probably clamp this to html.elements.a to avoid having very late support dates.
Please add `compute_from: html.elements.a` and re-generate the dist file. I suppose this would make it be supported at the first version of every browser.

features/a.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <a>
-description: The `<a>` element TODO.
+description: The `<a>` element, creates a hyperlink to anything accessible via a URL such as external web pages, files, email addresses, locations in the same page, and more.
```
captainbrosset, line 2:
```suggestion
description: The `<a>` element creates a hyperlink to any resource that's accessible via a URL, such as web pages, files, email addresses, locations within the same page.
```

Not super happy with my description, but I think this is a bit of an improvement still.

captainbrosset (changes requested):

features/a.yml
petele: I've added that, my one concern here is that there are some attributes that are only supported in one browser, or got support much later, and that won't be called out. But, saying it's not Baseline isn't the right answer either! (whee)

captainbrosset: I agree with your concern, and I think `<a>` is actually one of the few features for which we might want to create additional later features. The one that comes to mind is `download` because it was added substantially later and has some interesting partial support notes. There might be other.

captainbrosset (approved):

foolip:
As in https://github.com/web-platform-dx/web-features/pull/1427#issuecomment-2232853350, I suggest breaking out `download` and `ping`.

foolip:
https://github.com/web-platform-dx/web-features/pull/1449 are they keys to exclude.

petele:
Removed `download` and `ping`.

features/a.yml
```diff
@@ -1,11 +1,11 @@
-draft_date: 2024-06-05
 name: <a>
-description: The `<a>` element TODO.
+description: The `<a>` element creates a hyperlink to any resource that's accessible via a URL, such as web pages, files, email addresses, locations within the same page.
```
foolip, line 2:
```suggestion
description: The `<a>` element creates a hyperlink to any resource that's accessible via a URL, such as web pages, files, email addresses, or locations within the same page.
```

foolip (approved):

Issue #1316, opened 2024-07-04
Publish result of getStatus() as JSON


foolip:
MDN needs the Baseline statuses of individual BCD keys in order to show Baseline badges. @ddbeck's idea was to publish a "fat" JSON to GitHub releases, and that sounds good to me.

captainbrosset:
This was done. We've gone through the full lifecycle here. We publish the big JSON file now.

Issue #685, opened 2024-03-13
Release web-features as GitHub release with JSON artefact


foolip:
We can probably copy the approach that BCD uses, with an attached `data.json`.
https://github.com/mdn/browser-compat-data/releases/tag/v5.5.15

Requested by @jcscottiii.

foolip:
I have manually made a release in the shape I think we should automate:
https://github.com/web-platform-dx/web-features/releases/tag/web-features%2F0.6.0

ddbeck:
I think we now need two artifacts.

The first is as described, including `data.json` without the npm package wrapper on GitHub releases.

The second would be a more comprehensive file containing the single-key status results of `getStatus()` for each feature. This has come by MDN's request, since they don't want a JS dependency for this data.

For the latter, I think we'll need some extension or variation on the schema for the purposes of this one file. Something like:

```yaml
status:
  baseline: false
  support:
    chrome: "95"
    edge: "95"
  by_compat_features_key:
    api.example.method:
      status:
        baseline: false
        support:
          chrome: "95"
          edge: "95"
    # etc.
```

foolip:
That looks good I think, in particular that BCD keys are grouped under features, so that we have the option of having the status of a specific BCD key differ depending on the feature it's part of in the future. This depends on https://github.com/web-platform-dx/web-features/issues/1173 of course.

ddbeck:
This is now done automatically as of https://github.com/web-platform-dx/web-features/pull/1559. I think this is well and truly done now. Closing as complete.

PR #1407, opened 2024-07-16
Add extended JSON file build output

ddbeck:
This is a step toward https://github.com/web-platform-dx/web-features/issues/685 and https://github.com/web-platform-dx/web-features/issues/1316.

Unlike the regular npm package, this artifact includes computed statuses for each compat key, if a feature has a `status` and `compat_features`. This has also been known casually as the "fat" JSON artifact. This is intended for consumers that cannot or do not wish to use `compute-baseline`.

To make this artifact, use `npm run build:extended`.


scripts/build.ts
```diff
@@ -39,3 +45,25 @@ function buildPackage() {
     encoding: "utf-8",
   });
 }
+
+function buildExtendedJSON() {
+  // TODO: Validate the resulting JSON against a schema.
```
foolip, line 50: Can you do this locally? I suspect that it doesn't match the schema. See below.

scripts/build.ts
```diff
@@ -39,3 +45,25 @@ function buildPackage() {
...
+  for (const [id, featureData] of Object.entries(data.features)) {
+    if (Array.isArray(featureData.compat_features) && featureData.status) {
+      const by_compat_key = {};
+
+      for (const key of featureData.compat_features) {
+        by_compat_key[key] = { status: getStatus(id, key) };
```
foolip, line 56: I think this has to be unwrapped to match the schema:

```suggestion
        by_compat_key[key] = getStatus(id, key);
```

scripts/build.ts
```diff
@@ -39,3 +45,25 @@ function buildPackage() {
...
+      }
+
+      if (Object.keys(by_compat_key).length) {
```
foolip, line 59: Is this check really needed? How about just asserting `featureData.compat_features.length` earlier on?

(I don't love `Object.keys(...).length` so am looking for a way to avoid it.)

ddbeck: Fixed with c95aefd74bf5ea6226215b85bdff985dd6aab09f

package.json
```diff
@@ -16,12 +16,13 @@
...
   },
   "scripts": {
+    "build:extended": "tsx scripts/build.ts extended-json",
     "build": "tsx scripts/build.ts package",
     "dist": "tsx scripts/dist.ts",
     "feature-init": "tsx scripts/feature-init.ts",
     "format": "npx prettier --write .",
     "schema:write": "npm run schema -- --out ./schemas/data.schema.json",
-    "schema": "ts-json-schema-generator --tsconfig ./tsconfig.json --path ./types.ts",
+    "schema": "ts-json-schema-generator --tsconfig ./tsconfig.json --path ./types.ts --type=WebFeaturesData",
```
ddbeck, line 25: This bit is pretty serious:

I figured out what that top-level `$ref` was for. Without it, it was just a list of definitions without asserting anything using those definitions. Without this option, _any_ object will pass validation against our schema. 🙄

scripts/build.ts
```diff
@@ -14,14 +26,23 @@ yargs(process.argv.slice(2))
...
   const packageDir = new URL("./packages/web-features/", rootDir);
   const filesToCopy = ["LICENSE.txt", "types.ts", "schemas/data.schema.json"];
 
+  if (!valid(data)) {
+    logger.error("Data failed schema validation. No package built.");
+    process.exit(1);
+  }
+
   const json = stringify(data);
-  // TODO: Validate the resulting JSON against a schema.
```
ddbeck, line 24: Since I had to figure out how to validate the built JSON against the schema for the extended JSON, I did the same for the packaged one.

scripts/build.ts
```diff
@@ -39,3 +60,42 @@ function buildPackage() {
...
+    stringify(data),
+  );
+}
+
+function valid(data: any): boolean {
+  const valid = validate(data);
+  if (!valid) {
+    // TODO: turn on strictNullChecks, fix all the errors, and replace this with:
+    // const errors = validate.errors;
+    const errors = (valid as any).errors as DefinedError[];
```
ddbeck, line 94: This is a bit ugly because ajv can't produce (useful) types without `strictNullChecks` set to `true` in `tsconfig.json`. Fixable, but it requires touching a dozen files not germane to this PR.

scripts/validate.ts
```diff
@@ -0,0 +1,21 @@
+import Ajv from "ajv";
+import addFormats from "ajv-formats";
+import assert from "node:assert/strict";
+
+import * as schema from "../schemas/data.schema.json" with { type: "json" };
+
+export function validate(data: any) {
+  const ajv = new Ajv({ allErrors: true, allowUnionTypes: true });
```
ddbeck, line 8: I consolidated compiling the schema here, so that we can have one place with settings (and to check that we don't do something foolish like validate nothing at all).

package.json
foolip: Oh... let's put it back!

foolip (approved):

PR #1410, opened 2024-07-16
Add `<textarea>` element

petele:
Couple of notes:

I used `compute_from: html.elements.textarea`, which works for most things, but there are a few newer attributes that aren't baseline, for example: `html.elements.textarea.placeholder.line_breaks` and `html.elements.textarea.wrap.hard`


petele:
@ddbeck - same question about groups

foolip (approved):

PR #1412, opened 2024-07-16
Add `<span>` element

petele:


foolip:
Can you use `compute_from: html.elements.span` here so that the initial browser versions are used?

foolip (approved):

PR #1413, opened 2024-07-16
Add `<select>`, `<optgroup>`, and `<option>` elements

petele:
Explicitly chose to use `compute_from` on `html.elements.select` as `html.elements.select.size` and `html.elements.select.hr_in_select` are not baseline.


petele:

features/select.yml
```diff
@@ -1,11 +1,14 @@
-draft_date: 2024-06-05
 name: <select>
-description: The `<select>` element TODO.
+description: The `<select>` element provides a menu of options for a user to select from and is usually represented by a dropdown list.
```
foolip, line 2:
```suggestion
description: The `<select>` element provides a menu of options for a user to pick from, typically rendered as a dropdown list.
```

https://github.com/web-platform-dx/web-features/blob/main/docs/README.md#provides doesn't apply here IMO, so leaving it.

foolip (approved):

PR #1415, opened 2024-07-16
Add `<script>` and `<noscript>` elements

petele:


foolip:
Since `<noscript>` only makes sense to use together with `<script>` (as a fallback) I think we should merge these into just `<script>`, and _maybe_ mention `<noscript>` briefly in the description.

petele:
@foolip - done. Did not mention noscript in description, but added to the name.

features/script.yml
```diff
@@ -1,10 +1,9 @@
-draft_date: 2024-06-05
-name: <script>
-description: The `<script>` element TODO.
-spec:
-  - https://html.spec.whatwg.org/multipage/scripting.html#script
-  - https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element
-group: html
+name: <script> and <noscript>
+description: The `<script>` element is used to embed executable code or data; this is typically used to embed or refer to JavaScript code.
```
ddbeck, line 2:
```suggestion
description: The `<script>` element contains or loads data or executable code. This is typically used to load JavaScript code. The `<noscript>` element represents alternative content to show when scripting is not allowed.
```

ddbeck (approved):

PR #1417, opened 2024-07-16
Add `<picture>` and `<source>` elements

petele:
I couldn't find an element for `<source>` so I added it here. The description probably needs a little tweak.


features/picture.yml
```diff
@@ -1,9 +1,9 @@
...
 name: <picture>
-description: The `<picture>` element TODO.
+description: The `<picture>` element is used for responsive images.
 spec: https://html.spec.whatwg.org/multipage/embedded-content.html#the-picture-element
 group: html
 caniuse: picture
 compat_features:
   - api.HTMLPictureElement
   - html.elements.picture
+  - html.elements.source
```
foolip, line 9: The `<source>` element was first introduced for `<video>`, so it should either be part of that feature, or possibly its own feature.

features/picture.yml
```diff
@@ -1,9 +1,9 @@
-draft_date: 2024-06-05
 name: <picture>
-description: The `<picture>` element TODO.
+description: The `<picture>` element is used for responsive images.
```
foolip, line 2: I think needs to be elaborated a little bit. https://web.dev/learn/design/picture-element looks like a good intro to the element, maybe a description can be condensed from that.

petele: Updated

features/picture.yml
petele: Split out into a separate element

petele:
Creates a _separate_ entry for `<source>`. 

On `<source>` I used `compute_from` because the `height` and `width` properties only landed in late 2022. I'm open to putting them into their own feature, but not convinced that's required.

features/source.yml
```diff
@@ -0,0 +1,19 @@
+name: <source>
+description: The `<source>` element specifies a media resource for the `<picture>`, `<video>`, or `<audio>` elements.
```
ddbeck, line 2: I like this, but there's a couple of style issues.

First, see [the guidance on _specifies_](https://github.com/web-platform-dx/web-features/blob/main/docs/README.md#specifies.

The current phrasing with "or" also sound ambiguous to me, as if it's not known which of the elements it applies to. One of these alternatives might do.

```suggestion
description: The `<source>` element sets a media resource for an `<picture>`, `<video>`, or `<audio>` element.
```

```suggestion
description: The `<source>` element sets a media resource for the `<picture>`, `<video>`, and `<audio>` elements.
```

petele: done.

ddbeck (approved):

PR #1419, opened 2024-07-16
Add `<meter>` element

petele:


petele:
Should this one get `caniuse: meter` 

It has both: https://caniuse.com/meter and https://caniuse.com/mdn-html_elements_meter

foolip:
> Should this one get `caniuse: meter`
> 
> It has both: https://caniuse.com/meter and https://caniuse.com/mdn-html_elements_meter

Yes, please use `caniuse: meter`. We should never link to the BCD-derived entries starting with "mdn-" as those are just the BCD data in another form.

foolip:
The BCD data here looks fairly messy and implausible. I don't believe that the `<meter>` element and `HTMLMeterElement` interface shipped at different times, at least not far apart.

The Baseline low data ends up being defined by Edge 18, and all of the ≤18 values in https://github.com/mdn/browser-compat-data/blob/main/html/elements/meter.json. That will need fixing in BCD before we can add this feature.

petele:

features/meter.yml
```diff
@@ -1,8 +1,9 @@
-draft_date: 2024-06-05
 name: <meter>
-description: The `<meter>` element TODO.
+description: The `<meter>` element represents either a scalar value within a known range or a fractional value, and is typically rendered as a progress bar.
```
ddbeck, line 2: Copy editing for brevity

```suggestion
description: The `<meter>` element represents a scalar value within a known range or a fractional value, typically shown as a progress bar.
```

ddbeck:
I've merged main into this, lightly revised the description, and sent a PR upstream, https://github.com/mdn/browser-compat-data/pull/25777, (building on Florian's https://github.com/mdn/browser-compat-data/pull/24692) to truly sort out the Baseline low date.

@Elchi3 would you like to have a look at this, since you've seen this from every other angle so far? 😄

Elchi3 (approved):
Yes, let's ship this 👍

PR #1421, opened 2024-07-16
Add the `<menu>` element

petele:


petele:
@ddbeck - this is now failing because it also includes `- api.HTMLLIElement` which is also include in `list-elements.yml`.  Looking at #1173, we still don't have a graceful way to resolve duplicate elements. My inclination is to comment out the `- api.HTMLLIElement` items, and add a comment. Objections?

features/menu.yml
```diff
@@ -0,0 +1,13 @@
+name: <menu>
+description: The `<menu>` element represents an unordered list of items (`<li>`), and is a semantic alternative to the `<ul>` element.
```
ddbeck, line 2:
```suggestion
description: The `<menu>` element represents an unordered list of action items (`<li>`), such as a toolbar. It is a semantic alternative to the `<ul>` element.
```

ddbeck (approved):
> My inclination is to comment out the `- api.HTMLLIElement` items, and add a comment. Objections?

Yes, this is exactly right. Please include the URL for the #1173 in the comment. Look at `audio.yml` or `json-modules.yml`, if you want an example to copy from.

I've also offered up a revised description, if you'd like to apply that alone or with the comment.

Approved, provided both of the changes are dealt with. Thank you!

PR #1422, opened 2024-07-16
Add the `<main>` element

petele:


features/main.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <main>
-description: The `<main>` element TODO.
+description: The `<main>` element represents the dominant content of the `<body>` of a document.
```
foolip, line 2: I'm not sure that mentioning `<body>` makes it clearer. But I wonder if we should also mention how it's typically used together with `<header>` and `<footer>`?

```suggestion
description: The `<main>` element represents the dominant content of a document. It is commonly used together with `<header>` and `<footer>`.
```

petele: WFM

foolip (approved):

PR #1423, opened 2024-07-16
Add `<link>` element

petele:
I'm not sure how I feel about using `compute_from: html.elements.link` on this, but without using it, the link element isn't properly supported until at least chrome 80...


features/link.yml
```diff
@@ -1,8 +1,9 @@
-draft_date: 2024-06-05
 name: <link>
-description: The `<link>` element TODO.
+description: The `<link>` element specifies relationships between the current document and an external resource, for example stylesheets, favicons, and more.
```
ddbeck, line 2:
```suggestion
description: The `<link>` element creates a relationship between the current document and an external resource, such as a stylesheet or favicon.
```

features/link.yml
```diff
@@ -1,8 +1,9 @@
...
 spec: https://html.spec.whatwg.org/multipage/semantics.html#the-link-element
-group: html
+group: html-elements
+status:
+  compute_from: html.elements.link
```
ddbeck, line 6: Yes, this is correct. The foundational use here—stylesheets and favicons—are as old as dirt, so it's good to not make the overall element seem too new.

ddbeck (approved):
Thank you!

PR #1424, opened 2024-07-16
Add `<kbd>` element

petele:


features/kbd.yml
```diff
@@ -0,0 +1,6 @@
+name: <kbd>
+description: The `<kbd>` element represents a span of inline text denoting textual user input and is usually rendered in a monospace font.
```
ddbeck, line 2: Two thoughts here:

- Include the word "keyboard" since MDN refers to it as the "keyboard input element."
- Following https://github.com/web-platform-dx/web-features/pull/1403 though it hasn't merged yet

```suggestion
description: The `<kbd>` element represents textual user input, such as keyboard or voice input, styling text in monospaced type by default.
```

ddbeck (approved):

PR #1426, opened 2024-07-16
Add `<iframe>` element

petele:
Used `compute_from: html.elements.iframe` since a number of the `sandbox` attributes are later in life


foolip:
Caniuse has https://caniuse.com/iframe-srcdoc and https://caniuse.com/iframe-sandbox as separate features. These are quite old by now, but I wonder if we still shouldn't just create separate features for these for now? @ddbeck WDYT?

foolip:
I've sent PRs for separate features for `sandbox` and `srcdoc`: 
https://github.com/web-platform-dx/web-features/pull/1468
https://github.com/web-platform-dx/web-features/pull/1469

@petele can you exclude those keys from this feature?

petele:
Removed `sandbox` and `srcdoc`, which means `compute_from` was no longer necessary, so removed that too.

foolip (approved):

PR #1425, opened 2024-07-16
Add `<img>` element

petele:
Used `compute_from:` otherwise `html.elements.img.decoding` pushed it into later support.

PR #1427, opened 2024-07-16
Add image maps - `<map>` and `<area>`

petele:
Uses `compute_from: html.elements.area` since `ping` pushes it to be not baseline.


foolip:
@petele can you exclude the `download` and `ping`? I think we should create separate features for these, aligning with https://caniuse.com/download and https://caniuse.com/ping.

foolip:
I think these keys from https://github.com/web-platform-dx/web-features/pull/1425 should also be part of this feature:

  - api.HTMLImageElement.isMap
  - api.HTMLImageElement.useMap
  - html.elements.img.ismap
  - html.elements.img.usemap

foolip:
Can you also exclude `srcset` and `sizes` so that we can have a feature matching https://caniuse.com/srcset? I would also remove api.HTMLImageElement.currentSrc since that's related to `srcset`.

foolip (approved):

PR #1431, opened 2024-07-16
Add `<blockquote>`

petele:


features/blockquote.yml
```diff
@@ -1,8 +1,7 @@
-draft_date: 2024-06-05
 name: <blockquote>
-description: The `<blockquote>` element TODO.
+description: The `<blockquote>` element indicates that the enclosed text is an extended quotation, usually rendered as a block element that is indented.
```
ddbeck, line 2:
```suggestion
description: The `<blockquote>` element represents an extended quotation, styling contents as an indented block by default.
```

(Also, I've put in https://github.com/web-platform-dx/web-features/pull/1521 to capture this as a reasonable default when an element doesn't obviously "do" anything.)

ddbeck (approved):

PR #1432, opened 2024-07-16
Add `<article>` element

petele:


features/article.yml
```diff
@@ -0,0 +1,6 @@
+name: <article>
+description: The `<article>` element is a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable.
```
ddbeck, line 2: Making a suggestion here to trade in some concision to make room for illustrating something abstract with concrete examples:

```suggestion
description: The `<article>` element represents self-contained content, which is intended to be independently distributable or reusable, such as a comment, a blog post, or news article.
```

ddbeck (commented):
Looking good so far—just need some work on the description.

ddbeck (approved):

PR #1433, opened 2024-07-16
Add `<button>` element

petele:


petele:
Added `compute_from`

petele:
@ddbeck - should this get 

```
group:
  - forms
  - html-elements
```

or just `forms`?

foolip:
@petele let's put it in both groups, similar to `<video>` which is in both `media-elements` and `html-elements`.

foolip:
@petele there's a conflict due to https://github.com/web-platform-dx/web-features/pull/1482, and the `compute_from` bit will now do the right thing. Can you rebase and update dist?

foolip (approved):

PR #1434, opened 2024-07-16
Add the `<data>` element

petele:
There's a discrepancy between the versions for `api.HTMLDataElement` and `html.elements.data` that probably needs to be fixed in the BCD


foolip:
The Edge version will be ≤18 after https://github.com/web-platform-dx/web-features/pull/1398 and indeed we need to fix this in BCD. Just trusting the API data is probably OK for something this new, and given that the element has an attribute. (For elements without any attributes, not giving them their own interface has very little impact, and has probably happened much more often.)

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/23830. If you'd like to use `compute_from` to get the right versions with a TODO pointing to https://github.com/mdn/browser-compat-data/pull/23830, that's OK.

petele:
> I've sent [mdn/browser-compat-data#23830](https://github.com/mdn/browser-compat-data/pull/23830). If you'd like to use `compute_from` to get the right versions with a TODO pointing to [mdn/browser-compat-data#23830](https://github.com/mdn/browser-compat-data/pull/23830), that's OK.

I'm OK to leave it here and wait for that to land.

petele:
@foolip & @ddbeck - the test is failing on this one because of an invalid date, looking at the `dist` file, I see an odd date...

`baseline_low_date: ≤2018-10-02`

foolip:
@petele ranges in versions and dates are a result of https://github.com/web-platform-dx/web-features/pull/1398. Can you rebase and then run `npm run dist` again to get it all in order?

petele:
@foolip - done.

foolip (approved):

PR #1435, opened 2024-07-16
Add `<ins>` and `<del>` elements

petele:


petele:
Build error: `api.HTMLModElement is used in both del and ins`

Since they're somewhat different rendering and semantic meaning, I split them into two elements... But they both share the API. We could merge them back, but seems like maybe not the best option. Open to suggestions on how to proceed here.

foolip:
@petele I suggest stick the keys in `<del>` and adding a TODO pointing to https://github.com/web-platform-dx/web-features/issues/1173 for `<ins>`.

features/del.yml
```diff
@@ -0,0 +1,11 @@
+name: <del>
+description: The `<del>` element element represents a range of text that has been deleted from a document, usually rendered as strikethrough.
```
foolip, line 2:
```suggestion
description: The `<del>` element element represents a range of text that has been deleted from a document, styling text as strikethrough by default.
```

features/ins.yml
```diff
@@ -0,0 +1,13 @@
+name: <ins>
+description: The `<ins>` element element represents a range of text that has been inserted into a document, usually rendered with an underline.
```
foolip, line 2:
```suggestion
description: The `<ins>` element element represents a range of text that has been inserted into a document, styling text as underlined by default.
```

foolip (approved):
Aligning descriptions with https://github.com/web-platform-dx/web-features/pull/1403.

PR #1436, opened 2024-07-16
Add `<dfn>` element

petele:


features/dfn.yml
```diff
@@ -1,7 +1,6 @@
-draft_date: 2024-06-05
 name: <dfn>
-description: The `<dfn>` element TODO.
+description: The `<dfn>` element indicates a term to be defined, usually rendered in italics.
```
ddbeck, line 2: This is a weird element to use, thus hard to describe. Suggesting to avoid "indicates" and conventional "styling" text.

```suggestion
description: The `<dfn>` element marks a term to be defined, styling text as italic by default.
```

ddbeck (approved):

PR #1437, opened 2024-07-16
Add `<embed>` element

petele:
Need extra review

* `html.elements.embed.type` is available somewhat new
* doesn't address older usage of embed element


features/embed.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <embed>
-description: The `<embed>` element TODO.
+description: The `<embed>` element embeds external content at the specified point in the document.
```
foolip, line 2: What is the most common reason to use `<embed>` in 2024 and can we give that as an example?

In some quick testing I see that `<embed>` can be used to embed HTML and SVG, and MDN uses video embedding as the example. But those use cases are better served `<iframe>`, ``, and `<video>`.

https://stackoverflow.com/questions/291813/recommended-way-to-embed-pdf-in-html mentions `<embed>` and PDF is in a way the last remaining plugin that browsers support. Are there other reasons to use `<embed>` still around?

petele: After asking around, I don't think there are any other _good_ reasons to use it... 

Similar to `<object>`, I lean towards keeping the description generic for now. Maybe in the future it'll have a better use, but its baseline now, and isn't deprecated, yet.

foolip: Let's settle the description in https://github.com/web-platform-dx/web-features/pull/1438/files#r1682583440 and then use something very similar here.

foolip: That's done. Copy-pasting the wording:

```suggestion
description: The `<embed>` element represents an external resource such as a PDF or SVG document. It was historically used for plugins such as Shockwave Flash.
```

foolip (approved):

PR #1438, opened 2024-07-16
Add `<object>` element

petele:
Prob needs a little extra review

* Used `compute_from: html.elements.object` due to newer attributes
* obsolete?


features/object.yml
```diff
@@ -1,10 +1,11 @@
-draft_date: 2024-06-05
 name: <object>
-description: The `<object>` element TODO.
+description: The `<object>` element represents an external resource such as an external plugin or nested browsing context.
 spec:
   - https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element
   - https://html.spec.whatwg.org/multipage/obsolete.html#param
```
foolip, line 5: Since the `param` attribute isn't listed, can you remove this?

foolip (commented):
Like with https://github.com/web-platform-dx/web-features/pull/1437 I'm not sure how `<object>` is used today, if we can give an example.

petele:
There don't seem to be many legit use cases, the only one I've heard so far is embedding an SVG file will execute any embedded scripts, where as using the `` tag will not run the scripts.

Based on that, I think I'd lean towards keeping the generic description, its rarely used, hasn't been deprecated, but is still Baseline.

features/object.yml
```diff
@@ -1,10 +1,9 @@
-draft_date: 2024-06-05
 name: <object>
-description: The `<object>` element TODO.
-spec:
-  - https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element
-  - https://html.spec.whatwg.org/multipage/obsolete.html#param
+description: The `<object>` element represents an external resource such as an external plugin or nested browsing context.
```
foolip, line 2: "browsing context" is spec language that I wouldn't expect most web developers to be able to parse. Avoiding that and elaborating on the real reason this element exists:

```suggestion
description: The `<object>` element embeds an external resource such as a PDF or SVG document. It was historically used for plugins such as Shockwave Flash.
```

@ddbeck can you review so we don't have more rounds of description review after this?

foolip: Also consider that the description should work for `<embed>`. The word "embeds" would make it a bit circular, but I can't come up with anything better. Back to "represents"?

petele: Building on @foolip's description, I'd make one minor tweak to avoid the word _embed_.

> The `<object>` element inserts an external resource such as a PDF or SVG file into the document. It was historically used for plugins such as Shockwave Flash.

ddbeck: I like @foolip's original, though `<embed>` will need to say "represents." It's not perfectly consistent, but produces a good outcome: you search for "embed" and you'll get both features.

petele: > produces a good outcome: you search for "embed" and you'll get both features

I'm unsure why would that be a good outcome? My expectation of searching for "embed" would only return the `<embed>` element.

ddbeck: I think because they're closely related and easily confused (at least they are for me)—if I searched for one I'd appreciate being reminded of the other. But I don't have my heart set on this either; "represents" is fine too.

petele: OK, I've updated the description to keep represents and use the rest of @foolip's description.

foolip (approved):

PR #1439, opened 2024-07-16
Add `<fieldset>` and `<legend>`

petele:


petele:
@ddbeck - same question about groups.

foolip:
@petele can you add `compute_from` html.elements.fieldset and html.elements.legend here? The status is too new as it is.

foolip (approved):

PR #1440, opened 2024-07-16
Add `<form>` element

petele:
Used `compute_from: html.elements.form`


foolip:
This is a bit similar to `<table>` in that it's a collection of elements, but unlike `<table>` some of the elements can be used outside of `<form>`, like `<button>`.

@petele if we called this "Forms" like "Tables", are there any additional bits you think should be included, that don't make sense as standalone features?

Also paging @captainbrosset @ddbeck for thoughts on feature organization.

petele:
I'll advocate for creating a separate feature for each of the [`<form>`  related elements](https://html.spec.whatwg.org/multipage/form-elements.html).

* It would be a large feature with 11 elements, and many APIs included (leaving out `<button>`, including `<form>`)
* Most of the features landed at the very early, but some are significantly later, for example `<datalist>` which is Chrome 69. If there's a new form element in the future, it'll need its own entry.
* We may want to split `<input>` into the different types of input it supports, eg search, date, password, etc
* `<button>` would logically go in here, but can be used outside forms, so it would be in an awkward place
* We are already tagging these as `group: forms`

ddbeck:
What about splitting the difference a little bit? I could imagine some sort foundational _Forms_ feature consisting of a really, really stripped down thing:

- `html.elements.form` and descendants
- `api.HTMLFormElement` and descendants
- `html.elements.input.type_text`
- `html.elements.input.type_submit`

It'd be part of the forms group, but it'd let us be free to handle the various input types as separate features.

ddbeck (approved):
This looks good. Safari showing DOM API support before the element itself is a bit weird. I filed https://github.com/mdn/browser-compat-data/issues/24135 as an experiment. Hoping that Florian et al. have some feedback on whether I/we should be doing that routinely.

ddbeck:
Also, for posterity: I haven't changed my mind exactly on the organization here, but I do feel like we might compose a feature, as suggested in https://github.com/web-platform-dx/web-features/issues/971. I think it'll be easier to sort out when we have all of the forms stuff complete.

PR #1441, opened 2024-07-16
Add the `<input>` element

petele:
Used `compute_from: html.elements.input` since there are a number of attributes that are recent, or *very* recent.


petele:

captainbrosset:
Using compute_from the way you did seems like the right thing to do. A majority of the BCD keys are baseline high anyway, so might as well stick to the earliest of them that makes the feature be supported since the first versions of browsers.
Now, with that said, let's review the keys that are not baseline high:

* api.HTMLInputElement.cancel_event: seems like a detail that's OK to ignore
* html.elements.input.type_color: I could almost see this as a separate feature. As a developer, I would use a text input and a color input as two very different things. 
* api.HTMLInputElement.stepDown and api.HTMLInputElement.stepUp: seem ok to ignore to me.
* html.elements.input.type_week: seems very specific, and I could see this being part of a date/time input feature, instead of here.
* html.elements.input.type_password.insecure_login_handling: seems ok to ignore for now.

captainbrosset:
So, perhaps the right thing to do here is to rename this to something like Text input, and then create other features for Color inputs and Date and time inputs. What do you think?

petele:
> So, perhaps the right thing to do here is to rename this to something like Text input, and then create other features for Color inputs and Date and time inputs. What do you think?

I took it a step further and created features for the special input types where its a completely different input type or something other than text, using `caniuse` as a guideline.

features/input-date-time.yml
```diff
@@ -0,0 +1,16 @@
+name: <input> date, datetime-local, time, and week
```
autonome, line 1: Should find a way to construct the title that is consistent across the single and multiple types. The multiple-types version should also be clear that these are values of the `type` attribute, like the single-type version does.

features/input-date-time.yml
```diff
@@ -0,0 +1,16 @@
...
+description: The `<input>` element allows the user to input date and time information to a form.
```
autonome, line 2: (non-blocking feedback)

Use of `allows` in all these descriptions seems ok per https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#allows.

However, the attribute value seems more like a *constraint* on the input the user is able to enter so it reads a bit ambiguously as opposed to "allows the user to *only* input date and time..." or "restricts input to..." or something like that.

autonome (changes requested):
The feature name part I think should be fixed - if not made consistent, at least to specify that it's the `type` attribute in the multiple-types cases.

ddbeck:
Pete, we landed your work as part of #2044 (after some revisions). Thank you for doing the foundational work here!

PR #1442, opened 2024-07-16
Add `<output>` element

petele:


petele:
@ddbeck - should this get 

```
group:
  - forms
  - html-elements
```

or just `forms`?

petele:
@ddbeck & @foolip - I've merged to main, but I'm still getting a weird value in the `output.yml.dist` file for the baseline dates. I'm not sure what's going on here. Can you take a peek?

foolip:
@petele if you're not getting the same results locally as the CI, it's probably because compute-baseline JS files haven't been updated. Running npm install inside packages/compute-baseline should fix it. @ddbeck can we avoid this someone?

As for the range symbol, the problem is that BCS data about Edge is uncertain and affects the date. We'll need to test this in Edge and update BCD to be sure an out the correct dates.

petele:
> @petele if you're not getting the same results locally as the CI, it's probably because compute-baseline JS files haven't been updated. Running npm install inside packages/compute-baseline should fix it. @ddbeck can we avoid this someone?

I'm getting the same error locally as well, so its not just in the CI

foolip:
Oh... it could be that this is the first feature with a range on the date, and that fails in a weird way. We should either disallow it and fail with a clear message, or just handle it.

ddbeck (commented):
I've opened https://github.com/web-platform-dx/web-features/pull/1496 to deal with the error here. I've also got a suggestion for the description.

PR #1443, opened 2024-07-16
Add `<track>` element

petele:


features/track.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <track>
-description: The `<track>` element TODO.
+description: The `<track>` element is used as a child of the media elements that lets you specify a timed text track to be displayed in parallel with the media element.
 spec: https://html.spec.whatwg.org/multipage/media.html#the-track-element
 group: html
```
foolip, line 4:
```suggestion
group: media-elements
```

features/track.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-06-05
 name: <track>
```
foolip, line 1: I think this feature should be named "Text tracks" to match the existing "Audio and video tracks" feature and include the generic (non-WebVTT) text track API as well. This is a good start, from `npm run traverse | grep TextTrack`:

api.HTMLMediaElement.addTextTrack
api.TextTrack
api.TextTrack.activeCues
api.TextTrack.addCue
api.TextTrack.cuechange_event
api.TextTrack.cues
api.TextTrack.id
api.TextTrack.inBandMetadataTrackDispatchType
api.TextTrack.kind
api.TextTrack.label
api.TextTrack.language
api.TextTrack.mode
api.TextTrack.removeCue
api.TextTrackCue
api.TextTrackCue.endTime
api.TextTrackCue.enter_event
api.TextTrackCue.exit_event
api.TextTrackCue.id
api.TextTrackCue.pauseOnExit
api.TextTrackCue.startTime
api.TextTrackCue.track
api.TextTrackCueList
api.TextTrackCueList.getCueById
api.TextTrackCueList.length
api.TextTrackList
api.TextTrackList.addtrack_event
api.TextTrackList.change_event
api.TextTrackList.getTrackById
api.TextTrackList.length
api.TextTrackList.removetrack_event

This feature would cover both external text tracks from `<track>` and in-band text tracks.

features/track.yml
```diff
@@ -1,6 +1,5 @@
...
-description: The `<track>` element TODO.
+description: The `<track>` element is used as a child of the media elements that lets you specify a timed text track to be displayed in parallel with the media element.
 spec: https://html.spec.whatwg.org/multipage/media.html#the-track-element
```
foolip, line 3:
```suggestion
spec: https://html.spec.whatwg.org/multipage/media.html#timed-text-tracks
```

features/track.yml
petele: done

features/track.yml
```diff
@@ -0,0 +1,53 @@
...
+  - api.HTMLTrackElement.srclang
+  - api.HTMLTrackElement.track
+  - api.TextTrack
+  - api.TextTrack.activeCues
+  - api.TextTrack.addCue
+  - api.TextTrack.cuechange_event
+  - api.TextTrack.cues
+  - api.TextTrack.id
+  # Excluded because of a non-real value
+  # - api.TextTrack.inBandMetadataTrackDispatchType
```
petele, line 23: @foolip - I think this is the right thing here, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextTrack/inBandMetadataTrackDispatchType) this is supported, but it has a limited value. If I don't comment out the line, the whole feature is not baseline because that is apparently not supported in Chrome.

foolip: Does `npm run dist` throw an error if you include it? By non-real value we mean things like `true` in BCD that we can't handle and throw an error. I agree with excluding it, just want to understand.

To not have to fine tune the comment, you can just delete the lines. This whole block will go away if we move the tags to BCD.

foolip:
@petele can you also rename the file to `text-tracks.yml` in line with `audio-video-tracks.yml`?

features/track.yml
petele: No, it does not throw an error.

Looking a bit deeper, this property it's not supported by Chrome, and there's [no plan to support it](https://issues.chromium.org/issues/41338648). 

I uncommented that line, and used `compute_from`, but I don't think that's the right answer either. Open to suggestions here. @ddbeck @captainbrosset

features/text-tracks.yml
```diff
@@ -2,6 +2,8 @@ name: Text tracks
 description: The `<track>` element is used as a child of the media elements that lets you specify a timed text track to be displayed in parallel with the media element.
 spec: https://html.spec.whatwg.org/multipage/media.html#timed-text-tracks
 group: media-elements
```
foolip, line 4:
```suggestion
group:
  - media-elements
  - html-elements
```

foolip (approved):

PR #1448, opened 2024-07-17
Add srcset and sizes draft

foolip:
Publishing this requires fixing Edge ≤18 data in BCD.


features/draft/srcset.yml
```diff
@@ -0,0 +1,15 @@
+draft_date: 2024-07-17
+name: srcset and sizes
+description: The `srcset` and `sizes` attributes TODO.
```
ddbeck, line 3:
```suggestion
description: The `srcset` and `sizes` attributes for `<img>` elements set a list of possible sources for the image and a corresponding list of size conditions for choosing a source, to show a responsive image that fits the size of the display.
```

features/draft/srcset.yml
```diff
@@ -0,0 +1,15 @@
+draft_date: 2024-07-17
```
ddbeck, line 1: A comment here on why we didn't immediately publish would be nice to have

ddbeck (approved):

features/draft/srcset.yml
foolip: I've added a comment up top.

PR #1449, opened 2024-07-17
Add draft download and ping features

foolip:


features/draft/download.yml
```diff
@@ -0,0 +1,14 @@
+draft_date: 2024-07-17
+name: download
+description: TODO
```
ddbeck, line 3:
```suggestion
description: The `download` attribute for `<a>` elements loads the target URL as a file to be saved, instead of navigating to it. An optional attribute value sets a default file name.
```

features/draft/ping.yml
```diff
@@ -0,0 +1,14 @@
+draft_date: 2024-07-17
+name: ping
+description: TODO
```
ddbeck, line 3:
```suggestion
description: The `ping` attribute for `<a>` elements sets a URL to send a POST request to when the `<a>` is clicked. Typically used for outbound link reporting.
```

ddbeck (approved):

features/draft/ping.yml
```diff
@@ -0,0 +1,14 @@
+draft_date: 2024-07-17
+name: ping
+description: The `ping` attribute for `<a>` elements sets a URL to send a POST request to when the `<a>` is clicked. Typically used for outbound link reporting.
```
foolip, line 3: @ddbeck do you have strong feelings about full sentences? I don't think it really makes it clearer, but would you prefer this?

```suggestion
description: The `ping` attribute for `<a>` elements sets a URL to send a POST request to when the `<a>` is clicked. This is typically used for outbound link reporting.
```

I have no preference.

ddbeck: Good catch! I prefer the "This is…" version. We have a few cases where don't include the subject where it might standalone (e.g., "Typically styled as…" or "Also known as…"), but I don't think we should make it a habit for one-offs.

foolip: Alright, adding a subject to the sentence!

Issue #735, opened 2024-03-25
Requirements for 1.0


foolip:
What do we need to settle before releasing 1.0?

foolip:
Specific issues to consider for 1.0:

- https://github.com/web-platform-dx/web-features/issues/84
- https://github.com/web-platform-dx/web-features/issues/91
- https://github.com/web-platform-dx/web-features/issues/217
- https://github.com/web-platform-dx/web-features/issues/557
- https://github.com/web-platform-dx/web-features/issues/623
- https://github.com/web-platform-dx/web-features/issues/648 (and snapshots)
- https://github.com/web-platform-dx/web-features/issues/666 (how much do we burn down?)
- https://github.com/web-platform-dx/web-features/issues/715
- https://github.com/web-platform-dx/web-features/issues/736

foolip:
Why should we eventually release a 1.0? Only so that consumers can better use our version numbers to know when there are breaking changes.

foolip:
In https://github.com/GoogleChrome/webstatus.dev/pull/509 webstatus.dev was affected by our schema change when introducing groups and features. In order to be able to signal such breaking changes going forward, I'd like to release 1.0.0 ASAP and then proceed with semver like any project.

I had been "saving" 1.0 for something special, but I think the practical need to signal breaking changes is here right now. There are major versions after 1 that we can use for special events.

PR #1451, opened 2024-07-17
Release web-features 1.0.0

foolip:
Hoorah.

Closes https://github.com/web-platform-dx/web-features/issues/735.

### DIFF
--- a/packages/web-features/package-lock.json
+++ b/packages/web-features/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "web-features",
-  "version": "0.10.0",
+  "version": "1.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "web-features",
-      "version": "0.10.0",
+      "version": "1.0.0",
       "license": "Apache-2.0",
       "devDependencies": {
         "tsup": "^8.0.1",
--- a/packages/web-features/package.json
+++ b/packages/web-features/package.json
@@ -1,7 +1,7 @@
 {
   "name": "web-features",
   "description": "Curated list of Web platform features",
-  "version": "0.10.0",
+  "version": "1.0.0",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",


foolip:
https://github.com/web-platform-dx/web-features/pull/1398 should also be merged before 1.0.0, of course.

captainbrosset:
Definitely not against going to 1.0.0 if that helps us communicate breaking changes more easily.
Unlike more traditional libraries, though, which you can still keep on using for a while if you don't have time to update your code to adapt to the breaking change, you can't really keep using web-features 1.0.0 if 2.0.0 is out. Otherwise you risk displaying outdated browser version and baseline information.
I'm not saying we shouldn't do it. Let's do all, by all means, but maybe have something in the README that tells consumers how each version comes with updated dates/versions/baseline statuses, and that it's a best practice (or really, the only practice) to keep updating their dependency.

captainbrosset:
What about #1450? It's a schema change. Should we wait for that to land before going to 1.0.0?

foolip:
@captainbrosset, yes, that too! I sent it now because it's a breaking change, sorry I didn't link them together.

ddbeck:
> Let's do all, by all means, but maybe have something in the README that tells consumers how each version comes with updated dates/versions/baseline statuses, and that it's a best practice (or really, the only practice) to keep updating their dependency.

Yes, let's do this. Also, we can mark the previous releases on npm as deprecated. It might even be a good idea to do that on a rolling basis (e.g., if you're using a version of web-features more than a month old, you start getting the deprecation warning from npm).

foolip:
https://github.com/web-platform-dx/web-features/pull/1398 and https://github.com/web-platform-dx/web-features/pull/1450 have been merged now.

One more thing we can do while we're add it is include the schema in the release. I'll do that now.

foolip:
https://github.com/web-platform-dx/web-features/pull/1348 is also done now.

ddbeck (approved):

PR #1453, opened 2024-07-17
Add WebSockets API

petele:


features/websocket.yml
```diff
@@ -1,7 +1,8 @@
-draft_date: 2024-07-11
-name: WebSockets Standard
-description: TODO
+name: WebSockets
+description: WebSockets provides a way to open a two-way communication channel between the user's browser and a server.
 spec: https://websockets.spec.whatwg.org/
+status:
+  compute_from: api.WebSocket
```
foolip, line 5: Let's use the entrypoint, the thing you must use in order to use the feature at all:

```suggestion
  compute_from: api.WebSocket.WebSocket
```

It doesn't make any difference in this case, but for some features it does, most notably when the constructor has been prefixed.

features/websocket.yml
```diff
@@ -1,7 +1,8 @@
-draft_date: 2024-07-11
-name: WebSockets Standard
-description: TODO
+name: WebSockets
+description: WebSockets provides a way to open a two-way communication channel between the user's browser and a server.
```
foolip, line 2: https://github.com/web-platform-dx/web-features/tree/main/docs#provides recommends against "provides" in most cases, and this is a case where I think it's simpler without it:

```suggestion
description: The `WebSocket` API opens a two-way communication channel between the user's browser and a server.
```

@ddbeck do you approve?

features/websocket.yml
```diff
@@ -1,7 +1,8 @@
...
 spec: https://websockets.spec.whatwg.org/
```
foolip, line 3:
```suggestion
spec: https://websockets.spec.whatwg.org/
caniuse: websockets
```

features/websocket.yml
```diff
@@ -1,7 +1,8 @@
-draft_date: 2024-07-11
-name: WebSockets Standard
-description: TODO
+name: WebSockets
```
foolip, line 1: Please keep the filename as `websockets.yml` in line with https://websockets.spec.whatwg.org/ and https://caniuse.com/websockets.

As for the name, the spec is clearly intentional about never saying "web socket" without a space, so I think "WebSockets" as you've use here will work.

petele: Renamed.

foolip (approved):

PR #1457, opened 2024-07-18
Sort compat_features in public API

foolip:


ddbeck (approved):
Great. This means we can reorder items in dist files without it producing noisy, meaningless changes in the consumers' `compat_features` array.

PR #1460, opened 2024-07-18
Update `anonymous-iframe` feature

### LINKED ISSUE #1389: Iframe credentialless
### Specification

https://wicg.github.io/anonymous-iframe/

### Description

Iframe credentialless (formerly known as [anonymous iframe](https://developer.chrome.com/blog/anonymous-iframe-origin-trial)) enables embedding third-party iframes that do not set [Cross-Origin-Embedder-Policy (COEP)](https://web.dev/security-headers#coep), which is a requirement for websites to use privileged features including [SharedArrayBuffer](https://developer.chrome.com/blog/enabling-shared-array-buffer)

### Browser support

Chromium only at this point, see https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless

swissspidy:
Followed [the instructions](https://github.com/web-platform-dx/web-features/blob/6aa25aef9f98535d599fe4e3d57e6f50821100d6/features/draft/spec/README.md) to add a description and move the feature out of the `draft` folder. Manually removed the `draft_date` line because it seemed superfluous now.

Fixes #1389


foolip:
@swissspidy I see the spec calls this feature "Iframe credentialless", and that name makes sense to me since the entry point to this is the `credentialless` attribute on `<iframe>`. Can you rename the file to `iframe-credentialless.yml` to match?

swissspidy:
Done!

features/iframe-credentialless.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-07-11
 name: Iframe credentialless
-description: TODO
+description: When using cross-origin isolation, this allows embedding third-party iframes that do not set COEP.
```
foolip, line 2: Here's a take on the description more in the style of https://github.com/web-platform-dx/web-features/blob/main/docs/README.md#descriptions:

```suggestion
description: The `credentialless` attribute for `<iframe>` loads third-party content in an ephemeral context and does not send any credentials such as cookies. When using cross-origin isolation, this allows you to embed content that do not send COEP headers.
```

I don't know if I accurately described what it does, but I think it makes sense to say what it does in addition to the main use case for it. Also not sure if the wording around COEP is accurate, or if I broke it.

Finally, is this feature sometimes called "anonymous iframes" or is that an old and defunct name? If it is still relevant, consider adding "Also known as anonymous iframes." to the description.

swissspidy: Ah, thanks, I wasn't aware of those guidelines. Appreciate the suggestion 👍 

> Finally, is this feature sometimes called "anonymous iframes" or is that an old and defunct name?

AFAIK it's an old name. It was renamed after some initial feedback from other browser.

features/iframe-credentialless.yml
```diff
@@ -0,0 +1,7 @@
+name: Iframe credentialless
+description: The `credentialless` attribute for `<iframe>` loads third-party content in an ephemeral context and does not send any credentials such as cookies. When using cross-origin isolation, this allows you to embed content that do not send COEP headers.
```
ddbeck, line 2: A couple style nits here:

- "the `<foo>` HTML element" for consistency with other descriptions
- Subject-verb agreement
- (perhaps controversially) expanding "COEP" to what it stands for

```suggestion
description: The `credentialless` attribute for the `<iframe>` HTML element loads third-party content in an ephemeral context and does not send any credentials such as cookies. When using cross-origin isolation, this allows you to embed content that does not send `Cross-Origin-Embedder-Policy` headers.
```

foolip (approved):

PR #1468, opened 2024-07-19
Add draft feature for <iframe sandbox>

foolip:


features/draft/iframe-sandbox.yml
```diff
@@ -0,0 +1,31 @@
+draft_on: 2024-07-19
+# Revisit publishing this when the following are done:
+# https://github.com/mdn/browser-compat-data/pull/23862
+# https://github.com/mdn/browser-compat-data/pull/23858
+# https://github.com/mdn/browser-compat-data/pull/23861
+# https://github.com/mdn/browser-compat-data/pull/23864
+name: Iframe sandbox
```
ddbeck, line 7: I think capital-I Iframe is weird and maybe exceptional. If I'm alone on this, I won't worry about it though.

features/draft/iframe-sandbox.yml
```diff
@@ -0,0 +1,31 @@
...
+description: The `sandbox` attribute for the `<iframe>` HTML element TODO.
```
ddbeck, line 8:
```suggestion
description: The `sandbox` attribute for the `<iframe>` HTML element sets many security restrictions on the iframe, such as preventing form submissions or opening modal dialogs. Optional `allow-` values relax specific restrictions.
```

features/draft/iframe-sandbox.yml
foolip: It's a bit like "Canvas reset()" but I agree it's weirder, because "Iframe" doesn't look as much like a word.

https://github.com/web-platform-dx/web-features/issues/1389 is also in this stye, I'd be happy to change them to another style, like "sandbox for <iframe>".

ddbeck: Having a look around Stackoverflow and the like, I think "Sandboxed iframe" is most common, followed closely by "Sandbox iframe." Credentialless is newer and less definitive (with the oddly ungrammatical "Iframe credentialless" appearing), but "Credentialless iframe" is a plausible option too. "\_ for <iframe>" would be fine too, though a touch inventive on our part.

ddbeck: > So... WDYT about "Sandboxed iframes", "Credentialless iframes", and "srcdoc"?

Yes, let's do it. I think this is all in the spirit of prioritizing names that developers actually use.

foolip:
```suggestion
name: Sandboxed iframes
```

foolip: I've sent https://github.com/web-platform-dx/web-features/pull/1504 too.

features/draft/iframe-sandbox.yml
```diff
@@ -0,0 +1,31 @@
+draft_on: 2024-07-19
```
ddbeck, line 1:
```suggestion
draft_date: 2024-07-19
```

ddbeck (approved):

PR #1469, opened 2024-07-19
Add `<iframe srcdoc>` feature

foolip:
The versions match https://caniuse.com/iframe-srcdoc.


features/iframe-srcdoc.yml
```diff
@@ -0,0 +1,7 @@
+name: srcdoc
+description: The `srcdoc` attribute for the `<iframe>` HTML element sets inline HTML to embed instead of loading the document from the `src` attribute.
```
ddbeck, line 2: I found this a little difficult to follow, both in that it conveys a lot of information at once and "inline" didn't very well clue me into the idea that this is a literal document (like `data:text/html`). I've tried splitting these facts into two sentences. Does this still make sense?

```suggestion
description: The `srcdoc` attribute for the `<iframe>` HTML element sets a string of HTML to embed in the document. The value of `srcdoc` overrides loading a document from the `src` attribute.
```

ddbeck (approved):

PR #1471, opened 2024-07-19
Document Node.js-less web-features import

ddbeck:
Following https://github.com/web-platform-dx/web-features/pull/1470


foolip (approved):
I haven't tested this to confirm it works, but I trust that it does!

PR #1472, opened 2024-07-19
Add `<meta>` theme-color name

petele:


petele:
From @foolip:
> https://chromestatus.com/feature/5398016231997440 says Chrome 39, and https://caniuse.com/meta-theme-color doesn't have data for Chrome Android. Is 80 the correct version?

Initially it was only available on Android in Chrome 39, it landed on Desktop in Chrome 80.

foolip:
@petele it sounds like BCD is wrong here then:

https://github.com/mdn/browser-compat-data/blob/06a0b03c229b66aaedf5f54d6f750d2f5ecb7f92/html/elements/meta.json#L440

Care to send a PR for that?

foolip:
Added the blocked label since BCD needs to be updated.

PR #1473, opened 2024-07-19
Fix import syntax to work with Deno

foolip:
This was tested with the 0.10.0 version of the package from NPM, but with index.json instead of data.json.

The previous syntax didn't work because these aren't named exports.


ddbeck (approved):
Thank you! I was just about to open this myself. 😄

PR #1476, opened 2024-07-19
Mark dist YAMLs as generated

szepeviktor:
> Excluded from stats, hidden in diffs

https://github.com/github-linguist/linguist/blob/master/docs/overrides.md#summary


foolip:
If this suppresses dist files in diffs, it won't work. The statuses in the dist files are important to review, and we often attach review comments to them.

szepeviktor:
I see! Statuses are not coming from non-dist files.

PR #1485, opened 2024-07-22
dist script: use `checkAncestors` consistently

ddbeck:
Rarely, the dist script can produce different results in the comments than in the overall computed status. This is because `computeBaseline()` is called with `checkAncestors: false` while `getStatus()` implies `checkAncestors: true`.


features/js-modules-shared-workers.yml.dist
```diff
@@ -8,8 +8,8 @@ status:
     edge: "80"
     firefox: "114"
     firefox_android: "114"
-    safari: "15"
-    safari_ios: "15"
+    safari: "16"
```
foolip, line 11: The change here is because of https://github.com/mdn/browser-compat-data/issues/12982.

The new data is correct, Safari 16 did ship shared workers:
https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes#New-Features

foolip (approved):

Issue #1320, opened 2024-07-05
Dist refresh workflow does not trigger tests after pushing


foolip:
On https://github.com/web-platform-dx/web-features/pull/1304, after the push and the label was removed, the checks don't start. This is almost certainly because a push by GitHub Actions cannot trigger another GitHub Actions workflow.

The workaround is to close and reopen the PR.

@ddbeck FYI

captainbrosset:
@tidoust and I faced the same problem in https://github.com/web-platform-dx/web-features-explorer/pull/14

ddbeck:
Yeah, I've been banging my head against this problem. Maybe y'all would have a look at https://github.com/web-platform-dx/web-features/issues/1158 and give your thoughts on approaches there?

Issue #1158, opened 2024-05-30
Automatically update PRs with dist changes


foolip:
Running `npm run dist` after changes on a PR is tedious. Let's have a GitHub Actions workflow which updates them and pushes a new commit.

ddbeck:
There's been several PRs toward satisfying this, though it's not working perfectly just yet:

- https://github.com/web-platform-dx/web-features/pull/1213
- https://github.com/web-platform-dx/web-features/pull/1225
- https://github.com/web-platform-dx/web-features/pull/1226
- https://github.com/web-platform-dx/web-features/pull/1252
- https://github.com/web-platform-dx/web-features/pull/1253

I suspect once all of these have merged, we'll have a better sense of whether this is _done_ done.

ddbeck:
OK, I regret spending as much time as I have on this. GitHub Actions: here be dragons. I'll try to summarize the situation here: without a dedicate bot user or using someone's personal access token (thus making it appear as if that person is the author of the commit), this can only be done in a partial, not all that useful way. 💢 

Here's the longer version:

Automatically committing requires elevated permissions, but accepting contributions from forks means the risk of arbitrary code execution from untrusted sources. To deal with this safely, you need to refresh dists in separate steps:

1. In a reduced permissions environment (i.e., in a `pull_request` event workflow), run untrusted code to generate an artifact with the things we care about (e.g., a patch that contains the results of `npm run dist`).
2. In an elevated permissions environment, get the patch artifact, apply it, and commit it to the PR branch. The commit is what requires permissions (even though we're not committing the repository itself—one _could_). _Don't execute the results._ This is safe!
3. In a reduced permissions environment (in a `pull_request` event workflow), run the tests with the new commit applied.

This is a little indirect, but it's manageable. The trouble is that step 2, using the GitHub Actions token, cannot re-trigger the `pull_request` event ([example](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs)). This to prevent endlessly recursive workflows.

The only way to override this specific limitation is to use another user to do the commit.

Other approaches would be:

- Silently refresh dists during tests (e.g., always produce a patch and, if the test workflow says it would pass, apply the patch and assume that's valid). Slightly unsafe.
- Generate a comment that says what to do/what changes would result, but don't automatically commit. In effect, give up.

ddbeck:
[...]

PR #1486, opened 2024-07-22
Remove refresh dist workflow

ddbeck:
This just didn't work out. See https://github.com/web-platform-dx/web-features/issues/1158#issuecomment-2195182111 for why this approach won't work.

Closes https://github.com/web-platform-dx/web-features/issues/1320
Closes https://github.com/web-platform-dx/web-features/issues/1158

(After merging, it'd be nice to delete the refresh dist label too.)


foolip (approved):

foolip:
I've removed the label too!

PR #1492, opened 2024-07-23
Retitle "input event" feature to "input (event)"

ddbeck:
This is a special case. We use only the name of a lone event elsewhere (e.g., "scrollend"), but this event shares a name with the element ("\<input\>"). This moves the disambiguation text to parentheses to distinguish the name of the thing ("input").

Extracted from https://github.com/web-platform-dx/web-features/pull/1026.

### DIFF
--- a/features/input-event.yml
+++ b/features/input-event.yml
@@ -1,4 +1,4 @@
-name: input event
+name: input (event)
 description: The `input` event fires when a form control changes or an element with the `contenteditable` attribute changes.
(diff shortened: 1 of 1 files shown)


foolip:
To avoid setting a precedent too broad, for me a key reason is to avoid conflation with input events, referring to the whole class of events that fire on user input.

I'd like to see more cases before turning the distinction between name and disambiguation, and the use of parenthesis, into a firm guideline.

foolip (approved):

PR #1500, opened 2024-07-24
Add Canvas2D context

petele:
Canvas 2D rendering context!


features/canvas-2d-context-alpha.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `alpha` parameter
```
petele, line 1: The `alpha` parameter is also used in WebGL, WebGL2, and the bitmaprender, but support for it is drastically different across the different contexts (ie confirmed no Safari support here, but there is support in Safari in the WebGL context), and it landed at very different times across the different contexts. I am going to separate them out for each context.

features/canvas-2d-context-alpha.yml.dist
```diff
@@ -0,0 +1,13 @@
+# Generated from: canvas-2d-context-alpha.yml
```
petele, line 1: confirmed no Safari support currently.

features/canvas-2d-context-colorspace.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `colorSpace` parameter
```
petele, line 1: See comment above about the `alpha` parameter.

features/canvas-2d-context-desynchronized.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `desynchronized` parameter
```
petele, line 1: See comment above about the `alpha` parameter.

features/canvas-2d-context-willreadfrequently.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `willReadFrequently` parameter
```
petele, line 1: See comment above about the `alpha` parameter.

features/canvas-2d-font-rendering.yml
```diff
@@ -0,0 +1,19 @@
+name: Canvas 2D advanced font rendering
```
petele, line 1: I'm not in love with the name here, but these are a set of newer font properties that group together nicely.

features/canvas-2d-image-smoothingquality.yml.dist
```diff
@@ -0,0 +1,13 @@
+# Generated from: canvas-2d-image-smoothingquality.yml
```
petele, line 1: Confirmed no support in Firefox.

features/canvas-2d.yml
```diff
@@ -0,0 +1,68 @@
...
+  - api.CanvasRenderingContext2D.createLinearGradient
+  - api.CanvasRenderingContext2D.createPattern
+  - api.CanvasRenderingContext2D.createRadialGradient
+  - api.CanvasRenderingContext2D.drawFocusIfNeeded
+  - api.CanvasRenderingContext2D.drawImage
+  - api.CanvasRenderingContext2D.ellipse
+  - api.CanvasRenderingContext2D.fill
+  - api.CanvasRenderingContext2D.fillRect
+  - api.CanvasRenderingContext2D.fillStyle
+  - api.CanvasRenderingContext2D.fillText
```
petele, line 27: I deleted `canvas-text.yml` and moved the keys from there to here as it's all part of the same feature, and those keys did not change the baseline low/high dates significantly. 

Note that the keys in `canvas-text.yml` are actually generated from the BCD tags. The tool didn't yell at me for having duplicate keys. So hopefully this won't break anything too seriously or cause other headaches.

petele:
One final comment - I did not set a `compute_from` on this, I think there should be one, but I don't feel like there's a good property to set it to. Open to suggestions here.

features/canvas-2d-context-willreadfrequently.yml
ddbeck: By name, I'd expect this be "willReadFrequently" all by itself, though if we have to specify the context:

```suggestion
name: `willReadFrequently` for 2D canvases
```

Again though, if this is a hint, do we need to scope this to a specific context?

features/canvas-2d-context-willreadfrequently.yml
```diff
@@ -0,0 +1,6 @@
...
+description: Switches to software acceleration to improve performance when calling getImageData() frequently.
```
ddbeck, line 2:
```suggestion
description: The optional `willReadFrequently` parameter of a canvas's `getContext()` method permits the browser to optimize for frequent `getImageData()` calls by avoiding hardware acceleration.
```

features/canvas-2d-context.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D getContextAttributes() method
+description: Returns an object with the attributes used by the context.
+spec: https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-canvas-getcontextattributes-2
+group: canvas
+compat_features:
+  - api.CanvasRenderingContext2D.getContextAttributes
```
ddbeck, line 6: I'm very skeptical this is a distinct feature, rather than a later addition to the canvas API. I'd fold this into the 2D canvas and use a `compute_from` such that this is a later addition.

features/canvas-2d-filter.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `filter` property
```
ddbeck, line 1: Again, I'm skeptical that this is a distinct feature, rather than a later addition to canvas generally. I could be convinced otherwise, but I haven't seen much about this apart from browser vendor bug/issues.

features/canvas-2d-font-rendering.yml
ddbeck: Did you look into how developers are referring to this?

FWIW, I think "advanced" font rendering implies a basic font rendering. Is this standing in contrast to something else?

If not and there's no obvious name for this from developers, then I'd suggest something really plain like:

```suggestion
name: Font properties for canvases
```

features/canvas-2d-image-smoothingquality.yml
```diff
@@ -0,0 +1,6 @@
+name: Canvas 2D `imageSmoothingQuality` property
```
ddbeck, line 1: Maybe something like this? Though I'm a little concerned that this is not just a later addition to whatever feature `imageSmoothingEnabled` is a part of. It'd be nice to have some evidence of developers recognizing this as a distinct feature, or should have a reason to believe it so (e.g., blog posts and such promoting this to developers as a standalone thing).

```suggestion
name: imageSmoothingQuality for 2D canvases
```

features/canvas-2d.yml.dist
```diff
@@ -0,0 +1,259 @@
...
+  #   chrome: "31"
+  #   chrome_android: "31"
+  #   edge: "79"
+  #   firefox: "36"
+  #   firefox_android: "36"
+  #   safari: "10.1"
+  #   safari_ios: "10.3"
+  - api.CanvasRenderingContext2D.resetTransform
+
+  # ⬇️ Same status as overall feature ⬇️
```
ddbeck, line 247: I'd say it's very weird to say that 2D canvases have only been supported since 2020, when IE could do a bunch of this stuff. I think we need a `compute_from` such that the Baseline low date is 2018-01-29, even though it'll be a bit arbitrary which key(s) you pick.

ddbeck (commented):
A lot to look over here—sorry for the volume of comments and number of questions. Thank you, @petele!

features/canvas-2d-context-alpha.yml
petele: Yes, I do think Devs will think of this separately, adding transparency to an image is one of the key reasons developers choose image formats (jpg vs png). And because of the limited support here it's not something we want to lump in together with the rest of the features.

features/canvas-2d-context-colorspace.yml
petele: I went back and forth on the naming for these a handful of times. I went with "Canvas 2D \`colorSpace\` parameter" to try to be consistently for all of the special context parameters (colorSpace, alpha, desynchronized, willReadFrequently). I think the naming makes it clear its for the canvas 2d rendering context, and what parameter it is.

features/canvas-2d-font-rendering.yml
petele: I went with "advanced" because it's not all of the font rendering properties, but only the new ones that have landed in the last few years. I didn't want to call it "new" or "new-ish". 

The other alternative I had was to lump kerning, stretch, and spacing into a feature called "font spacing", then  separate features for direction, caps, and textRendering, but that feels like a bunch of one-off features. 

By grouping them into this one group, we've got a handful of font features that all landed closely enough together that they can be considered a feature.

features/canvas-2d.yml.dist
petele: 100% agree, didn't specify one yet until we get this a little closer. As you said, it's somewhat arbitrary.

features/canvas-2d-context-colorspace.yml
```diff
@@ -0,0 +1,6 @@
...
+description: Defines the color space of the rendering context.
```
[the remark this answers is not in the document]
petele: I'm a little hesitant to include the "srgb" or "display-p3", as there might be additional profiles later, and we don't want to have to update in the future. 

How about:

```
The optional `colorSpace` parameter of a canvas's `getContext()` method specifies the color representation.
```

I used specifies here explictly, as "chooses" feels awkward to me.

features/canvas-2d-context-desynchronized.yml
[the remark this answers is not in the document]
petele: The desynchronized feature landed in different contexts at different times, so I think we need to be explicit about what context we're referring to, and about separating it into individual features.

features/canvas-2d-context-desynchronized.yml
```diff
@@ -0,0 +1,6 @@
...
+description: Reduces latency by desynchronizing the canvas paint cycle from the event loop.
```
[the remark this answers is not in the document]
petele: same comment about different contexts above

features/canvas-2d-context-willreadfrequently.yml
petele: Yes, it is only supported in Canvas 2D

features/canvas-2d-context.yml
petele: It landed as such a distinctly different time that it doesn't feel like it is part of the initial feature. Yes, it is odd to have a single key like this, but it would be confusing to say it was supported when Canvas 2D first launched, when it wasn't.

features/canvas-2d-filter.yml
petele: The big reason I put it as it's own is the limited support. It's not in Safari, and I didn't want to create the confusion around something that isn't supported across browsers.

petele:
@ddbeck - updated with discussed changes, ready for review.

features/canvas-2d-context-colorspace.yml
```diff
@@ -0,0 +1,6 @@
+name: Color management for 2D canvas
```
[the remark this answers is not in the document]
petele: To be consistent, are you OK with `canvas-2d-color-management.yml` so it matches the WebGL1/2 naming?

features/canvas-2d-context-desynchronized.yml
```diff
@@ -0,0 +1,6 @@
+name: Desynchronized 2D canvas
```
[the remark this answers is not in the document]
petele: We've gone with `canvas-2d-(something).yml` in most cases, and I think it's more consistent. 

Are you OK with `canvas-2d-desychronized.yml`?

features/canvas-2d-context-willreadfrequently.yml
```diff
@@ -0,0 +1,6 @@
+name: willReadFrequently
```
petele, line 1: @ddbeck - for consistency with naming from above does how about `canvas-2d-willreadfrequently.yml`?

features/canvas-2d-context-alpha.yml
```diff
@@ -0,0 +1,6 @@
+name: Opaque 2D canvas
```
petele, line 1: OK with renaming to `canvas-2d-alpha.yml` for consistency?

petele, line 1: Hmm, I accepted this, but am now having second thoughts, and want to go back to `Alpha` or alternatively I could be OK with `2d canvas opacity`. Every canvas is opaque unless you use this parameter. A quick web search returns better results for either Alpha or opacity.

ddbeck: To give you a sense of my thinking here: the default value is `true` (otherwise the browsers introducing the option would suddenly break a bunch of existing canvases), so the _feature_ is opting into a higher-performance opaque canvas, even though the API is (ostensibly) a neutral control over alpha/opacity.

That said, I am convinced by the search results yielding better for opacity, so "2D canvas opacity" would be OK with me.

petele: done!

ddbeck (approved):

PR #1502, opened 2024-07-24
Adding weekly run to update draft features by spec

tropicadri:


.github/workflows/update_draft_features_weekly.yml
```diff
@@ -0,0 +1,32 @@
+name: Update Draft Features
+
+on:
+  schedule:
+# Runs at midnight on Mondays 05:30 UTC
```
foolip, line 5: Yep, that's as good a time as any! I'll also add `workflow_dispatch` so that we can trigger this manually at any time. Often useful for ensuring everything works after a change to the script or workflow.

foolip (approved):
Thank you @tropicadri! Let's land this and get spec drafts updating.

foolip:
@tropicadri looks like there's a problem. In https://github.com/web-platform-dx/web-features/actions/runs/10117116884/job/27981430193#step:7:92 the error is "GitHub Actions is not permitted to create or approve pull requests".

I haven't seen this before, but I guess we need to give it permissions...

tropicadri:
Ah yes, I enabled write permissions for the test repo that I created, but it seems the organization might also need to be configured to allow write pull requests, but it also needs to be a repo admin, so I couldn't set it up.

Go to https://github.com/organizations/web-platform-dx/settings/actions and check Allow GitHub Actions to create and approve pull requests
Then go to https://github.com/web-platform-dx/web-features/settings/actions and check the box that previously was grayed out

foolip:
@tropicadri do you know if https://github.com/web-platform-dx/web-features/pull/1517 might work, or if one also has to enable it in the GitHub settings? I don't have access to https://github.com/organizations/web-platform-dx/settings/actions, I'd need to ask W3C staff.

PR #1507, opened 2024-07-25
Put some things in the Transferable group

ddbeck:
Originally suggested in https://github.com/web-platform-dx/web-features/pull/987#discussion_r1583180916. I grouped any non-draft that _contained_ a transferable object, to see how things looked.

@foolip originally wrote:

> My hunch is that not all of these things will be in features by themselves, and the features they are part have some other primary location in a feature group hierarchy, like `MediaSourceHandle` being a part of Media Source Extensions or some media group.

My feeling is that canvas is probably rightly grouped here, but the other two are much more uncertain.

(I've been on a tab-closing spree today, so apologies if this seems out of the blue.)

Closes https://github.com/web-platform-dx/web-features/pull/1512

### DIFF
--- a/features/offscreen-canvas.yml
+++ b/features/offscreen-canvas.yml
@@ -1,7 +1,9 @@
 name: Offscreen canvas
 description: The `OffscreenCanvas` API provides a canvas that can be drawn to off screen, with no dependencies on the DOM, which can be used to run heavy rendering operations inside a worker context.
 spec: https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface
-group: canvas
+group:
+  - canvas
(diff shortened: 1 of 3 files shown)


foolip:
This group strikes me as quite odd, a bit like having a group for any APIs that use promises. Are you sure it'll be useful for someone to filter features by "includes something transferable"?

ddbeck:
> Are you sure it'll be useful for someone to filter features by "includes something transferable"?

Nope! I just saw the suggestion on the previous PR and wondered what it would look like. I think it would be totally find to get rid of the transferable group entirely. https://github.com/web-platform-dx/web-features/pull/1512 is what that would look like.

PR #1516, opened 2024-07-26
Add WebGL2 Canvas Context

petele:
I did not use `compute_from`, since it's already `high`.


features/webgl2-context-colorspace.yml
```diff
@@ -0,0 +1,6 @@
+name: WebGL2 Color Space
```
petele, line 1: I'm open to splitting this into two features, one has baseline low, the other is only supported in Chrome/Edge.

It also has the same support versions as the same functionality in WebGL1, so we could alternatively merge it that way.

features/webgl2-context-desynchronized.yml
```diff
@@ -0,0 +1,5 @@
+name: WebGL2 `desynchronized` parameter
```
petele, line 1: Has same support as WebGL1 `desynchronized` so we could merge these two. I feel like WebGL1/WebGL2 are very different and keeping them separate makes sense. But can be convinced otherwise.

features/webgl2-context-drawingbuffer.yml
```diff
@@ -0,0 +1,6 @@
+name: WebGL2 drawing buffer
```
petele, line 1: Has same support as WebGL1 `drawingBufferXXX` so we could merge these two. I feel like WebGL1/WebGL2 are very different and keeping them separate makes sense. But can be convinced otherwise.

features/webgl2-context-powerpreference.yml
```diff
@@ -0,0 +1,5 @@
+name: WebGL2 `powerPreference` parameter
```
petele, line 1: Has similar support as WebGL1 `powerPreference`, with Safari being significantly different (10.1/3 vs 15). We could merge these two, but based on the significant difference, I'd lean towards no.

features/webgl2.yml
```diff
@@ -241,25 +234,6 @@ compat_features:
   - api.WebGL2RenderingContext.viewport
   - api.WebGL2RenderingContext.waitSync
   - api.WebGLQuery
```
petele, line 236: Even though this is titled `WebGLXXX` it is explicitly WebGL2... Same for some of the other top level APIs

petele:
Replaced by #1584

PR #1521, opened 2024-07-29
Write "represents" guideline for HTML elements

ddbeck:
Also, discourage "indicates" to describe things that are not indicators.

This is inspired by several reviews of HTML elements. Sometimes an element does something (the `<br>` element _breaks_ lines), but sometimes they're abstract containers "represents" is a reasonable default.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -162,6 +162,11 @@ Don't use it as a coordinating conjunction; start a new sentence instead.
 For example, this is an example.
 ([#738](https://github.com/web-platform-dx/web-features/pull/738#discussion_r1537762579), [#742](https://github.com/web-platform-dx/web-features/pull/742))
 
+##### indicates
+
+Avoid when you mean sets or represents.
+See [sets](#sets) and [represents](#represents).
+
 ##### is used to
 
 Omit "is used" where there's no loss in meaning.
@@ -180,9 +185,13 @@ Avoid, especially with gerunds.
 For example, prefer the "The feature writes to…" over "The feature provides writing to…."
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635491))
 
+##### represents
+
+If an HTML element or other type doesn't obviously do anything, then it probably _represents_ something.
+
 ##### sets
 
-Prefer this over multisyllabic alternatives, such as "defines", "determines", or "specifies".
+Prefer this over multisyllabic alternatives, such as "defines", "determines", "indicates", or "specifies".
 Use "The property sets…" but never "The property defines…."
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635491))
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -183,6 +183,11 @@ For example, prefer the "The feature writes to…" over "The feature provides wr
 ##### represents
 
 If an HTML element or other type doesn't obviously do anything, then it probably _represents_ something.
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -180,9 +185,13 @@ Avoid, especially with gerunds.
 For example, prefer the "The feature writes to…" over "The feature provides writing to…."
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635491))
 
+##### represents
+
+If an HTML element or other type doesn't obviously do anything, then it probably _represents_ something.
```
petele, line 190: I'd really like to see a Good/Bad example here. I'm still struggling with this one a bunch, and having a couple of examples would help a lot.

docs/README.md
```diff
@@ -162,6 +162,11 @@ Don't use it as a coordinating conjunction; start a new sentence instead.
 For example, this is an example.
 ([#738](https://github.com/web-platform-dx/web-features/pull/738#discussion_r1537762579), [#742](https://github.com/web-platform-dx/web-features/pull/742))
 
+##### indicates
```
foolip, line 165: It looks like this was motivated by https://github.com/web-platform-dx/web-features/pull/1436. If it's only popped up once I don't think we need to document it.

Note that it's used in a way allowed by this guideline here:

https://github.com/web-platform-dx/web-features/blob/ee7fe7940c1e25b649236c12bbb2b3349aa49ea7/features/modulepreload.yml#L2

docs/README.md
foolip: It's hard to think of bad examples, rather I'd treat "represents" as a default for HTML elements which we diverge from when we think some other verb is clearer.

docs/README.md
ddbeck: OK, I've removed "indicates" from this PR.

docs/README.md
ddbeck: Yes, @foolip captured my meaning here, as using it as a default. That said, I was able to come up with some examples, so I added them. This necessitated some edits to actual descriptions, of course.

docs/README.md
```diff
@@ -180,6 +180,15 @@ Avoid, especially with gerunds.
...
 
+##### represents
+
+If an HTML element or other type doesn't obviously do anything, then it probably _represents_ something.
+Use _represents_ as a default for describing the behavior of HTML elements when no more specific verb makes sense.
+
+  - 👍 Recommended: The `<html>` element represents the top level of an HTML document…
+  - 👍 Recommended: The `<br>` element breaks lines in text.
+  - 👎 Not recommended: The `<br>` element represents a line break in text.
+
```
ddbeck, line 191: New elaboration on the "represents" guidance

captainbrosset (approved):

captainbrosset (approved):
I think this is fine and should be merged.

foolip:
@ddbeck can you rebase this to follow the move of guidelines into docs/guidelines.md?

docs/README.md
ddbeck: I dropped the `<html>` example with 7454cb34.

ddbeck:
Actually, now that I'm looking over this again, I'd like to withdraw it.

Following [Florian's comment](https://github.com/web-platform-dx/web-features/pull/1521#discussion_r1745418213), I think in lots of cases we could rewrite _represents_ as _is_, with no loss of clarity. I think we ought to do that, but it's not very pressing.

But more importantly, this work is basically done. We have no more third-level `html.element.*` keys to assign. This guideline didn't merge before it outlived it's usefulness. I'm going to close this instead.

PR #1526, opened 2024-07-30
Add the Presentation API

foolip:


foolip:
Relevant chromestatus entries:
https://chromestatus.com/feature/6676265876586496
https://chromestatus.com/feature/5414209298890752

I didn't split the feature into two, because I suspect that the receiver part isn't really relevant for most devs.

features/presentation-api.yml
```diff
@@ -1,7 +1,8 @@
-draft_date: 2024-07-11
 name: Presentation API
-description: TODO
+description: The Presentation API allows you to show and control content on another device, such as a network-connected TV.
```
ddbeck, line 2: I was not previously aware of this API. "another device" made me think of remote control of another browser (in spec terms, 2-UA). Suggesting a change here to capture the 1-UA sense better:

```suggestion
description: The Presentation API shows or controls content on another display, such as a network-connected TV or projector.
```

ddbeck (approved):
I am happy with this, though I have a small suggestion for the description for you to consider. Thank you!

PR #1535, opened 2024-07-31
Replace "defines" with "sets"

foolip:
https://github.com/web-platform-dx/web-features/blob/main/docs/README.md#defines


features/datalist.yml
```diff
@@ -1,5 +1,5 @@
 name: <datalist>
-description: The `<datalist>` element defines a set of recommended values for an `<input>` element. Browsers may show a dropdown menu of all options, or matching options as the user types.
+description: The `<datalist>` element sets a set of recommended values for an `<input>` element. Browsers may show a dropdown menu of all options, or matching options as the user types.
```
foolip, line 2: "sets a set" is no good. I think defines here is quite OK, and "provides" would also work.

features/color-function.yml
```diff
@@ -1,4 +1,4 @@
 name: color()
-description: "The `color()` function defines a color within a given color space. Wide gamut color spaces like `display-p3` allow showing more vibrant and saturated colors than the standard `srgb` color space."
+description: "The `color()` function sets a color within a given color space. Wide gamut color spaces like `display-p3` allow showing more vibrant and saturated colors than the standard `srgb` color space."
```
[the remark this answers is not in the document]
ddbeck: That said, "defines" isn't very good either: the _color space_ defines the colors, and the `color()` function picks a color from the color space. Though I suppose that itself might provide an answer:

```suggestion
description: "The `color()` function picks a color from a given color space. Wide gamut color spaces like `display-p3` allow showing more vibrant and saturated colors than the standard `srgb` color space."
```

features/datalist.yml
ddbeck: "Defines" here is probably closer to the thing that the element does. That said, "sets" is still possible:

```suggestion
description: The `<datalist>` element sets recommended values for an `<input>` element. Browsers may show a dropdown menu of all options, or matching options as the user types.
```

Though I'm thinking the whole thing though is a bit messy. The second sentence is unclearly connected to the first and does a weird change in verb form. I'd maybe rewrite it like this:

```suggestion
description: The `<datalist>` element sets recommended values for an `<input>` element. Browsers may show a dropdown menu of all values or match values as the user types.
```

ddbeck (approved):

PR #1542, opened 2024-08-01
export all types in web-features package

LeoMcA:
We currently have to do some typescript shenanigans to get access to certain types from `web-features` for use [in our code](https://github.com/mdn/yari/blob/d930863abcb20d52170ab95b670a2676e26f7451/client/src/document/baseline-indicator.tsx#L10-L14):

```ts
// web-features doesn't export these types directly so we need to do a little typescript magic:
import type { features } from "web-features";
type SupportStatus = (typeof features)[keyof typeof features]["status"];
type BrowserIdentifier =
  keyof (typeof features)[keyof typeof features]["status"]["support"];
```

It would be far nicer if we could simply import the types directly:

```ts
import type { SupportStatus, BrowserIdentifier } from "web-features";
```

This PR enables that, by exporting all the types, and renaming `browserIdentifier` to the more type-like `BrowserIdentifier`

### DIFF
--- a/packages/web-features/index.ts
+++ b/packages/web-features/index.ts
@@ -8,4 +8,5 @@ const { features, groups, snapshots } = JSON.parse(
   readFileSync(jsonPath, { encoding: "utf-8" }),
 ) as WebFeaturesData;
 
+export * from "./types";
 export { features, groups, snapshots };
--- a/types.ts
+++ b/types.ts
@@ -28,11 +28,11 @@ export interface FeatureData {
     compat_features?: string[];
 }
 
-type browserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
+export type BrowserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
 
-type BaselineHighLow = "high" | "low";
+export type BaselineHighLow = "high" | "low";
 
-interface Status {
+export interface Status {
     /** Whether the feature is Baseline (low substatus), Baseline (high substatus), or not (false) */
     baseline: BaselineHighLow | false;
     /** Date the feature achieved Baseline low status */
@@ -41,11 +41,11 @@ interface Status {
     baseline_high_date?: string;
     /** Browser versions that most-recently introduced the feature */
     support: {
-        [K in browserIdentifier]?: string;
+        [K in BrowserIdentifier]?: string;
     };
 }
 
-interface SupportStatus extends Status {
+export interface SupportStatus extends Status {
     /** Statuses for each key in the feature's compat_features list, if applicable. Not available to the npm release of web-features. */
     by_compat_key?: Record<string, Status>
 }
@@ -53,7 +53,7 @@ interface SupportStatus extends Status {
 /** Specification URL
  * @format uri
 */
-type specification_url = string;
+export type specification_url = string;
 
 export interface GroupData {
     /** Short name */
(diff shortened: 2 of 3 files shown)


packages/web-features/index.ts
```diff
@@ -8,4 +8,5 @@ const { features, groups, snapshots } = JSON.parse(
   readFileSync(jsonPath, { encoding: "utf-8" }),
 ) as WebFeaturesData;
 
+export * from "./types";
```
ddbeck, line 11: This is surprising to me. Can you explain what this does? When I read the documentation on this, I was under the impression that the `"types"` field in `package.json` was the correct way to provide the types.

types.ts
```diff
@@ -28,11 +28,11 @@ export interface FeatureData {
     compat_features?: string[];
 }
 
-type browserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
+export type BrowserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
```
ddbeck, line 31: If you're going to export this stuff, then it needs a description in the schema. You'll need to do something like this, for each new exported type:

```suggestion
/** Browser identifiers (e.g., "chrome", "firefox", "safari") */
export type BrowserIdentifier = "chrome" | "chrome_android" | "edge" | "firefox" | "firefox_android" | "safari" | "safari_ios";
```

types.ts
```diff
@@ -28,11 +28,11 @@ export interface FeatureData {
...
 
-type BaselineHighLow = "high" | "low";
+export type BaselineHighLow = "high" | "low";
 
-interface Status {
+export interface Status {
```
ddbeck, line 35: I think this is a little weird. `Status` + `SupportStatus` is a little bit of a hack to avoid recursion in the generated JSON Schema (and I don't want to suggest that the `by_compat_key` data is a tree) or writing mostly duplicate types. I'd suggest exporting only `SupportStatus` and see what this looks like after regenerating the schema.

types.ts
```diff
@@ -41,19 +41,19 @@ interface Status {
...
+export interface SupportStatus extends Status {
     /** Statuses for each key in the feature's compat_features list, if applicable. Not available to the npm release of web-features. */
     by_compat_key?: Record<string, Status>
 }
 
 /** Specification URL
  * @format uri
 */
-type specification_url = string;
+export type specification_url = string;
```
ddbeck, line 56: This type only exists so that we get the `uri` format in JSON Schema. In TypeScript, a type alias for `string` seems kinda pointless (especially with an ugly name like this one). Maybe we can leave this out too?

ddbeck (commented):
> Ah, I wasn't aware this would change `data.schema.json`, maybe that's not wanted?

That's supposed to happen. We generate the JSON Schema from TypeScript types, so (almost) any change in `types.ts` will change the schema.

Some specific commentary in line comments.

packages/web-features/index.ts
LeoMcA: Indeed, which sets `index.d.ts` as the type declaration file for the package: however, that's automatically generated by typescript from `index.ts`, so if `index.ts` doesn't export a type, `index.d.ts` also doesn't export a type; so in order to be able to access all the types exported in `types.ts`, we have to export everything from that in `index.ts`.

ddbeck: Ah ha! OK, thank you—that explains it.

ddbeck:
@LeoMcA This is a bit stale (one or two of the changes you proposed have been picked up already, I think). Do you want to come back to it or open a superseding PR? Or is this something someone else should take up?

ddbeck:
This has gone a bit stale (and some of the proposed changes have already been picked up). Feel free to reopen or send a new PR if this is still relevant. Thank you!

PR #1561, opened 2024-08-08
Adds JS Date object

petele:


petele:
Note: I did not remove the Date methods from `features/draft/spec/ecmascript.yml` - can do that if its not automatically updated.

foolip:
> Note: I did not remove the Date methods from `features/draft/spec/ecmascript.yml` - can do that if its not automatically updated.

No need. https://github.com/web-platform-dx/web-features/pull/1568 should get the updater workflow running.

features/date.yml.dist
```diff
@@ -0,0 +1,185 @@
...
+  # baseline: high
+  # baseline_low_date: 2015-07-29
+  # baseline_high_date: 2018-01-29
+  # support:
+  #   chrome: "1"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "1"
+  #   firefox_android: "4"
+  #   safari: "4"
```
foolip, line 84: This surprised me since I thought `Date.now()` was prehistoric, but it seems right:
https://github.com/WebKit/WebKit/commit/c6b29b42c1efbfe6507323ab0d7ba3eed6096d6d

features/date.yml.dist
```diff
@@ -0,0 +1,185 @@
...
+  # baseline: false
+  # support:
+  #   chrome: "1"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "1"
+  #   firefox_android: "4"
+  #   safari: "1"
+  #   safari_ios: "1"
+  - javascript.builtins.Date.getYear
```
foolip, line 183: These are deprecated and `npm run dist` will omit a warning. Can you omit them so that we don't get a "Not Baseline" banner on these MDN pages? It's technically true that it's not Baseline, but it really should be a different category which doesn't exist yet.

features/date.yml
```diff
@@ -0,0 +1,69 @@
+name: Date
+description: The `Date` object represent a single moment in time in a platform-independent format.
```
ddbeck, line 2:
```suggestion
description: The `Date` object represents a single moment in time.
```

petele:
@ddbeck - updated with group, ready to go.

features/date.yml
```diff
@@ -0,0 +1,67 @@
...
+  - javascript.builtins.Date.toLocaleDateString.locales_parameter
+  - javascript.builtins.Date.toLocaleDateString.options_parameter
+  - javascript.builtins.Date.toLocaleString
+  - javascript.builtins.Date.toLocaleString.iana_time_zone_names
+  - javascript.builtins.Date.toLocaleString.locales_parameter
+  - javascript.builtins.Date.toLocaleString.options_parameter
+  - javascript.builtins.Date.toLocaleTimeString
+  - javascript.builtins.Date.toLocaleTimeString.iana_time_zone_names
+  - javascript.builtins.Date.toLocaleTimeString.locales_parameter
+  - javascript.builtins.Date.toLocaleTimeString.options_parameter
```
ddbeck, line 62: Given the recent discussion on `Number` and @Elchi3's work on `Intl`, would you mind dropping these and refreshing the dist?

```suggestion
```

ddbeck (changes requested):

Elchi3 (approved):

ddbeck (approved):

PR #1563, opened 2024-08-08
Add WeakMap and WeakSet

petele:


ddbeck (commented):
I think these features look fine, but I'm less certain about my proposed descriptions. @Elchi3 as resident JS docs expert, would you like to have a look at my suggestions? (If you don't want to, that's fine too—I can apply my suggestions and merge instead.)

Elchi3 (approved):
Thank you! 👍

Elchi3 (commented):
Optional: Do we want to add the to the "collections" group?

PR #1564, opened 2024-08-08
CSS Trig Functions Name Tweak

petele:
Name created confusion between CSS functions and JavaScript Math functions, changed to be explict about CSS functions


features/trig-functions.yml
```diff
@@ -1,3 +1,3 @@
-name: sin(), cos(), tan(), asin(), acos(), atan(), and atan2()
+name: CSS trigonometric functions
```
ddbeck, line 1: We have a number of other CSS math features which don't lend themselves to a naming scheme like this. If we must disambiguate here, then let's do something like this:

```suggestion
name: sin(), cos(), tan(), asin(), acos(), atan(), and atan2() (CSS)
```

Otherwise, we'd need to revisit all of the CSS math features and do some research to make sure we're using names recognized by web developers.

ddbeck (changes requested):

ddbeck (approved):
Thank you, @petele!

PR #1565, opened 2024-08-08
Add RegExp methods

petele:
There are a handful of deprecated methods that I've included, but think we should probably remove? Looking for confirmation please.

* `compile`
* `input`
* `lastMatch`
* `lastParent`
* `leftContext`
* `rightContext`


features/regexp.yml
```diff
@@ -0,0 +1,44 @@
+name: Regular Expressions (RegExp)
+description: The RegExp object is used for matching text with a regular expression.
+spec: https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp-regular-expression-objects
+status:
+  compute_from: javascript.builtins.RegExp
+compat_features:
+  - javascript.builtins.RegExp
+  - javascript.builtins.RegExp.RegExp
+  - javascript.builtins.RegExp.compile
```
petele, line 9: this method is deprecated - should we drop it from this list?

features/regexp.yml
```diff
@@ -0,0 +1,44 @@
...
+  - javascript.builtins.RegExp.dotAll
+  - javascript.builtins.RegExp.exec
+  - javascript.builtins.RegExp.flags
+  - javascript.builtins.RegExp.global
+  - javascript.builtins.RegExp.global.prototype_accessor
+  - javascript.builtins.RegExp.hasIndices
+  - javascript.builtins.RegExp.ignoreCase
+  - javascript.builtins.RegExp.ignoreCase.prototype_accessor
+  - javascript.builtins.RegExp.input
```
petele, line 18: this method is deprecated - should we drop it from this list?

features/regexp.yml
```diff
@@ -0,0 +1,44 @@
+name: Regular Expressions (RegExp)
+description: The RegExp object is used for matching text with a regular expression.
```
ddbeck, line 2: Feature names are sentence case and also-known-as text goes in the description (if needed, though not in this case)

```suggestion
name: Regular expressions
description: The `RegExp` object represents a regular expression, a notation for matching text patterns.
```

features/regexp.yml
```diff
@@ -0,0 +1,44 @@
...
+  - javascript.builtins.RegExp.toString
+  - javascript.builtins.RegExp.toString.escaping
+  - javascript.builtins.RegExp.unicode
+  - javascript.builtins.RegExp.unicodeSets
+  - javascript.builtins.RegExp.@@match
+  - javascript.builtins.RegExp.@@matchAll
+  - javascript.builtins.RegExp.@@replace
+  - javascript.builtins.RegExp.@@search
+  - javascript.builtins.RegExp.@@species
+  - javascript.builtins.RegExp.@@split
```
ddbeck, line 44: We also need `javascript.grammar.regular_expression_literals` and all of the non-deprecated `javascript.regular_expressions.*` keys. It's possible there are some distinct features in there, but it'll be easier to pick them out with the dist file.

ddbeck (changes requested):

petele:
Added additional keys, and removed all deprecated functions.

ddbeck (approved):
Thank you, Pete!

PR #1517, opened 2024-07-26
Give the draft updater workflow pull requests write permission

foolip:

PR #1568, opened 2024-08-09
Revert "Give the draft updater workflow pull requests write permission"

foolip:
Reverts web-platform-dx/web-features#1517

### DIFF
--- a/.github/workflows/update_draft_features_weekly.yml
+++ b/.github/workflows/update_draft_features_weekly.yml
@@ -6,9 +6,6 @@ on:
     - cron: '30 5 * * 1'
   workflow_dispatch:
 
-permissions:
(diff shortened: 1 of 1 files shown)


foolip:
I tested in https://github.com/foolip/web-features/actions/runs/10315510828/job/28555822160 that this fixes the problem. I guess that be specifying `permissions`, only the set permissions are there, and if it's omitted the default permissions apply. I think that permissions to write to the repo was the missing bit, but don't see why we should tinker with permissions at all. It was just a failed attempt to fix the problem.

ddbeck (approved):

foolip:
It's alive! https://github.com/web-platform-dx/web-features/pull/1570

Thank you @tropicadri!

PR #1571, opened 2024-08-09
Make the spec draft updater remove completed specs and work around nulls

foolip:
And update dist, which was overlooked. These changes together will allow
the workflow to send PRs that pass the CI.


ddbeck (approved):

ddbeck:
Prettier is failing, but you're free to merge this once that's fixed.

PR #1577, opened 2024-08-12
Add Web Crypto API

autonome:
- [x] moved files
- [x] removed draft date
- [x] determined name
- [x] added entry point
- [x] verified baselines
- [x] added to a group, or not (see below)
- [x] determined whether the feature should be split up or not
- [ ] am i missing anything else?

## Name

Sheesh:

- Spec: Web Cryptography API
- MDN: Web Crypto API
- CIU: Web Cryptography
- Node: Web Crypto API
- W3C GH: /webcrypto
- Blink: WebCrypto
- Developer: no consistent pattern that i could find

Thoughts:

- I posit "crypto" will confuse people and also make cryptographers mad
- There's a precedent for "Web {x}" in WF already
- There's very little "API" suffix in WF so far, even when specs have it, but some (not nearly all) folks do seem to refer to it that way

Proposal for WF: Web Cryptography

We can use "crypto" as shorthand in URLs and file paths, but should expand to the full name in any other context.

I'd be ok with popping "API" on there if people think it's needed. Not adding it keeps options open for future.

Edit: Updated to use Web Cryptography

## Description

The description feels a bit short, and I like "operations" as Caniuse went with, vs "systems" from MDN which sounds heavyweight. I tried to balance it with the "utilities" bit since there are standalone functions. Thoughts, preferences from anyone?

Edit: Updated to use "cryptographic operations" instead of "systems".

## Group

This feels like it could be its own group, as there's no other group it aligns with, and also it's a general feature area along a specific theme. Could name it `Web Crypto API` per how developers refer to it, or `Cryptography` for a more general bucket for potential future features to be added to which may not be in the same API namespace.

Edit: Leaving group out for now, we can always come back to it

## Splits/Merges

Discussed w/ @ddbeck and deleted `webcrypto-secure-curves` files and merged into here, as they're considered by developers as part of the Web Cryptography feature, not something separate and standalone.


autonome:
Hmm, secure curves needs to be in here too.

autonome:
@ddbeck I'm going to merge https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/webcrypto-secure-curves.yml into the general webcrypto api feature.

Even if new curves are added (likely), they're more like flavors of the same basic APIs, not new APIs in and of themselves. wdyt?

ddbeck:
> @ddbeck I'm going to merge https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/webcrypto-secure-curves.yml into the general webcrypto api feature.
> 
> Even if new curves are added (likely), they're more like flavors of the same basic APIs, not new APIs in and of themselves. wdyt?

Reproducing my answer from chat:

> Yes, that seems like a reasonable thing to do.
> 
> When I encounter something like this, I search Wikipedia, web, Stack Overflow, and caniuse (unchecking the MDN box) for the spec language ("secure curves") and the narrowest names in the list of compat keys ("ed25519" here, though adding "javascript" is needed some places like Stack Overflow). If I can't find evidence that it's a distinct thing that people are writing and asking about, then I look to see what predecessor feature it might be part of.
> 
> So for this feature that approach checks out: Web Crypto is the feature that developers care about but it has some smaller bits tacked on that developers don't recognize as a distinct "thing" (we've been calling these "later additions").

features/webcryptoapi.yml
```diff
@@ -1,7 +1,12 @@
-draft_date: 2024-08-12
-name: Web Cryptography API
-description: TODO
-spec: https://w3c.github.io/webcrypto/
+name: Web Cryptography
```
ddbeck, line 1: Web Crypto seems to be the more widely used name (e.g., as in docs on MDN). I'll suggest an "also known as" in the description, however, since it seems mixed.

```suggestion
name: Web Crypto
```

features/webcryptoapi.yml
```diff
@@ -1,7 +1,12 @@
...
+description: The Web Cryptography API provides utilities and primitives for performing basic cryptographic operations.
```
ddbeck, line 2: I have an allergy to the word "basic" (basic for whom, exactly? Not me!). I'm suggesting "low-level" here as probably capturing the same idea and including some examples to fill it out:

```suggestion
description: The Web Crypto API performs low-level cryptographic operations, such as encryption, decryption, and signature verification. Also known as the Web Cryptography API.
```

features/webcryptoapi.yml
ddbeck, line 1: Also, you'll need to change the file name to `web-crypto`. We try to avoid common suffixes and prefixes, when it still makes sense (e.g., "CSS", "JS", "API" etc.—"Web" is the hardest to avoid, as it's often distinguishing as in "WebUSB").

ddbeck (changes requested):

features/webcryptoapi.yml
ddbeck: @autonome This wasn't accepted and the existing text still stands. Were you going to revisit this yourself?

features/webcryptoapi.yml
autonome: I wrote up a reasoning for this: https://github.com/web-platform-dx/web-features/pull/1577#issue-2461382789

I didn't see a developer preference either way, and I am concerned "crypto" will forever be confusing. Where are you finding this? I want to make sure I'm looking at the same places. (The whole group should be sharing tips here, since the developer voice part is the fuzziest bit.)

The common pattern I saw in materials (outside of MDN) was a shortname of `webcrypto` and an expanded name of `Web Cryptography API`.

I'm fine putting API back in, as I said in the feature notes.

But I'm not sure the MDN choice of "crypto" here aligns with either developer voice or the optimal disambiguation of terms.

autonome: I'd already changed it to webcrypto.yml, which is the "shortname" pretty much everywhere, will that do?

features/webcryptoapi.yml
autonome: Nope, i just missed it and pushed changes after - just updated now, but w/ the aka reversed until we settle on final name.

autonome: and yes, +1 on nobasicness, thanks for pointing that out!

features/webcrypto.yml
```diff
@@ -1,7 +1,11 @@
-draft_date: 2024-08-19
-name: Web Cryptography API
-description: TODO
-spec: https://w3c.github.io/webcrypto/
+name: Web Cryptography
```
ddbeck, line 1: Ah, I got really confused about the name and identifier. Since we're calling this Web Cryptography, let's set the identifier/filename to `web-cryptography.yml`. Then I think this will be ready to merge. Thank you!

ddbeck (changes requested):

features/webcryptoapi.yml
autonome: per comment below, resolving

features/webcryptoapi.yml
autonome: fixed

features/webcryptoapi.yml
autonome: resolved per conversation below, ended up back on `Web Cryptography` for now

ddbeck (approved):

PR #1579, opened 2024-08-13
New writingsuggestions feature

captainbrosset:
See:

* https://chromestatus.com/feature/5153375153029120
* https://html.spec.whatwg.org/multipage/interaction.html#writing-suggestions

The feature lets devs control which fields (or parts of a webpage) should have the browser-provided writing suggestions/completions turned on or off.

I'm adding MDN docs for it at https://github.com/mdn/content/pull/35445.

The BCD daeta for this landed in https://github.com/mdn/browser-compat-data/pull/24121, which we don't yet have here in the repo, so generating a dist file isn't yet possible.


features/writingsuggestions.yml
```diff
@@ -0,0 +1,6 @@
+name: 'writingsuggestions'
+description: 'The `writingsuggestions` HTML attribute allows you to turn browser-provided writing suggestions on or off for editable regions of a web page.'
```
ddbeck, line 2: What _are_ "writing suggestions"? Not only do I not understand what this is from this description, I can't find any explanation—not the spec, not Chrome Status, etc.—that gives me a clue about what this is referring to.

features/writingsuggestions.yml
```diff
@@ -0,0 +1,6 @@
...
+spec: https://html.spec.whatwg.org/multipage/interaction.html#writing-suggestions
+group: forms
+compat_features:
+  - html.global_attributes.writingsuggestions
```
ddbeck, line 6: The spec makes it seem like there should be a DOM attribute too. Is that being contributed to BCD as well, or is it merely unimplemented? It's not in the linked PR, in any case.

features/writingsuggestions.yml
petele: More context on the [writingsuggestions attribute](https://developer.chrome.com/release-notes/124?hl=en#the_writingsuggestions_attribute) and link to [spec](https://html.spec.whatwg.org/multipage/interaction.html#writing-suggestions)

captainbrosset: Writing suggestions can be seen in action in, for example, Microsoft Edge. Start typing some text in a textarea, at some point you should start seeing greyed-out text positioned after the cursor. Pressing tab adds that text.
I think Firefox also started releasing a similar feature.

I'm writing docs on MDN about the attribute: https://github.com/mdn/content/pull/35445

Chromestatus entry: https://chromestatus.com/feature/5153375153029120

features/writingsuggestions.yml
captainbrosset: The corresponding DOM property had been added to BCD before my PR: https://github.com/mdn/browser-compat-data/blob/main/api/HTMLElement.json#L2406

Do you think we need to list here too? I can add it for completeness.

features/writingsuggestions.yml
captainbrosset: Let me know if you can find a better way to explain this feature. Writing suggestions is super vague, I agree, but I can't find a short way of saying what this does without being overly specific, which may be a problem in the future as other implementations take this in different UX directions.

ddbeck: OK, after rooting around a bit, I found https://github.com/whatwg/html/issues/10209 and [this explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/WritingSuggestions/explainer.md) to be illustrative.

I started writing a description like this:

```suggestion
name: writingsuggestions
description: The `writingsuggestions` HTML attribute turns on or off a browsers' writing suggestions, such as text completions or incremental autocompletions. 
```

I don't think this does the central thing though, which is explain how is this not `autocomplete`? Is the main difference that it's limited to on/off and works on any `contenteditable` element, in addition to form fields?

features/writingsuggestions.yml
ddbeck: Yes, please include it. I can't imagine it would belong with any other feature.

features/writingsuggestions.yml
captainbrosset: Writing suggestions (at least for now) do not appear as dropdown lists below an editable region. They only appear as inline text. This is what it looks like in Edge:

image

So, in that sense, it's different from what the `autocomplete` attribute does (which is disallow, or control the type of completions that are offered in a text field's dropdown list).

The other consideration is that `writingsuggestions` is meant to work on inputs, textarea, contenteditable, but I believe also on editable regions that use the EditContext API (so this could be a div or canvas for example).
And also, on any other element that you want. For instance, you could do `<body writingsuggestions="false">` to turn the feature off on your entire page. Or `<form writingsuggestions="false">` to turn it off in just this one form, maybe because it uses another custom way to provide completions.

So maybe:

```suggestion
name: writingsuggestions
description: The `writingsuggestions` HTML attribute turns on or off a browser's writing suggestions on all or part of a page. Supporting browsers display writing suggestions as content that can be inserted by pressing the tab key in editable text regions.
```

features/writingsuggestions.yml
```diff
@@ -0,0 +1,7 @@
+name: writingsuggestions
+description: The `writingsuggestions` HTML attribute turns on or off a browser's writing suggestions, such as text completions or incremental autocompletions.
```
ddbeck, line 2: Continuing from https://github.com/web-platform-dx/web-features/pull/1579#discussion_r1724684775. I like the idea of making this concrete. What about something like this?

```suggestion
description: The `writingsuggestions` HTML attribute turns on or off a browser's writing suggestions. Writing suggestions vary by browser. For example, if turned on then a browser might show inline text completions accepted by pressing <kbd>Tab</kbd>.
```

ddbeck (changes requested):
One more round on the description. Also I think you can merge `main` into this branch now and generate a useful dist.

captainbrosset:
I'm still seeing ` html.global_attributes.writingsuggestions is unindexable at 'writingsuggestions'` after updating from main and running `npm install`. We use BCD 5.5.46, but looks like 5.5.47 is now available, we need to bump to get `html.global_attributes.writingsuggestions`.

ddbeck (approved):
Thank you, Patrick!

PR #1582, opened 2024-08-13
Add `primitive-types` group

petele:
Needed for features like String, Number, BigInt, etc.


groups/primitive-types.yml
```diff
@@ -0,0 +1,2 @@
+name: Primitive types
```
ddbeck, line 1: We don't have descriptions yet for groups, so I'm going to leave this comment here for clarity and future use.

```suggestion
# JavaScript primitive data types, such as strings or numbers, and their corresponding object wrappers, such as `String` and `Number`.
name: Primitive types
```

ddbeck (approved):

PR #1594, opened 2024-08-14
Add IndexedDB

petele:


ddbeck:
I'm sorry, @petele. You've run into one of the more troublesome features and apparently the spec-generated drafts don't detect when a _different_ draft also contains overlapping keys.

Please see the top of https://github.com/web-platform-dx/web-features/blob/main/features/draft/indexeddb.yml and have a think about whether you still want to touch this. 😅

PR #1595, opened 2024-08-14
Add Z-index

jamesnw:


features/z-index.yml
```diff
@@ -0,0 +1,11 @@
...
+description: The `z-index` CSS property allows you to order overlapping elements, with higher values in front of or on top of lower values.
+spec: https://drafts.csswg.org/css2/#z-index
+snapshot: ecmascript-5
+group: css
+compat_features:
+  - css.properties.z-index
+  - css.properties.z-index.auto
+  - css.properties.z-index.negative_values
+# Baseline low date of 2015 seems late, but is due to lack of Edge support until
+# then.
```
ddbeck, line 11: I'm going to suggest deleting this comment otherwise, we're going to putting it _lots_ of places. 😅

Effectively, Baseline is undefined prior to the release of Edge 12 (since it's the youngest browser in the [core browser set](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#core-browser-set). It's possible that someday we'll do some sort of backtracking into IE data but for now this is just something we're living with.

See also: https://github.com/web-platform-dx/web-features/issues/623

features/z-index.yml
```diff
@@ -0,0 +1,11 @@
+name: z-index
+description: The `z-index` CSS property allows you to order overlapping elements, with higher values in front of or on top of lower values.
+spec: https://drafts.csswg.org/css2/#z-index
+snapshot: ecmascript-5
```
ddbeck, line 4: I'm guessing this is just from a mistaken copy and paste.

```suggestion
```

features/z-index.yml
```diff
@@ -0,0 +1,11 @@
+name: z-index
+description: The `z-index` CSS property allows you to order overlapping elements, with higher values in front of or on top of lower values.
```
ddbeck, line 2: Nice description. One suggestion for brevity (lots of phrases like "allows you to…", "is used to…", "enables…" can be reduced to whatever verb immediately follows) and one for clarity:

```suggestion
description: The `z-index` CSS property orders overlapping elements, with higher values appearing in front of or on top of lower values.
```

ddbeck (changes requested):
Nice first feature, @jamesnw! Welcome aboard. I've got a few nitpicky suggestions in line comments. Let me know if you have questions. Thank you!

features/z-index.yml
jamesnw: This was actually intentional- the [caniuse result](https://caniuse.com/?search=z-index) is for ECMAScript 5, and I made some assumptions about what snapshot was for. In this case, should we omit caniuse completely?

ddbeck: That's very strange! I think it came up on caniuse because of some sort of search word splitting. I can't think of any reason `z-index` would be related to ES5.

The `snapshot` field is a bit under-developed (and not published for consumers, at least not yet). It's meant to capture which version of a specification introduced the feature. In this case, `z-index` has no relationship to the ECMAscript 5 spec (as best as I can tell).

features/z-index.yml
jamesnw: Excellent, thanks for the context!

ddbeck (approved):

Issue #1499, opened 2024-07-24
Historical investigations: caniuse and BCD mismatches


ddbeck:
I'm using this issue as a place to log BCD-caniuse discrepancies that are historic (pre-2020) for later investigation and upstream bug reports or fixes. Please feel free to edit this description with additions.

<! -- - [ ] `dotted.path.to.compat.key` (browsers_identifier, browser_identifier) Description text -->

- [ ] `html.elements.progress` (safari_ios) Note says indeterminate not supported. Caniuse says this is a partial implementation and affects only Safari 7.

PR #1596, opened 2024-08-14
Add Cursor style

jamesnw:


features/cursor-style.yml.dist
```diff
@@ -0,0 +1,126 @@
...
+  baseline_low_date: 2021-12-07
+  baseline_high_date: 2024-06-07
+  support:
+    chrome: ≤83
+    chrome_android: "83"
+    edge: "83"
+    firefox: ≤72
+    firefox_android: "95"
+    safari: ≤13.1
+    safari_ios: ≤13.4
```
ddbeck, line 15: Across the board, caniuse says that Safari doesn't support this feature but BCD says it does. See https://github.com/Fyrd/caniuse/issues/7066. I've filed https://github.com/mdn/browser-compat-data/issues/24142 to see if that can/should be fixed. I think this PR is going to be blocked for a little bit until we've figured who's correct.

(I'll wait for triage on BCD to see if it's correctable soon—if not, we can turn this into a draft feature so the PR doesn't have to be held open for forever.)

features/cursor-style.yml.dist
```diff
@@ -0,0 +1,126 @@
...
+  # baseline_high_date: 2024-06-07
+  # support:
+  #   chrome: "5"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "3"
+  #   firefox_android: "95"
+  #   safari: "5"
+  #   safari_ios: "4.2"
+  - css.properties.cursor.none
```
ddbeck, line 85: Up to this point, these are off by one version for Edge when compared to caniuse. I'm not worried about this since it's pre-2020, but I did note it in: https://github.com/web-platform-dx/web-features/issues/1499.

features/cursor-style.yml
```diff
@@ -0,0 +1,47 @@
+name: Cursor Style
+description: The `cursor` CSS property styles the pointer, allowing you to provide hints to the user on how to interact with the hovered element.
+spec: https://drafts.csswg.org/css-ui-4/#cursor
+caniuse:
+  - css3-cursors
+  - css3-cursors-newer
```
ddbeck, line 6: Depending on the results of https://github.com/mdn/browser-compat-data/issues/24142, we may want to include https://caniuse.com/css3-cursors-grab too.

ddbeck (changes requested):

features/cursor-style.yml
```diff
@@ -0,0 +1,47 @@
+name: Cursor Style
```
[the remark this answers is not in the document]
ddbeck: Oh that said, we'll also need to rename the file. I think we _do_ want to use `cursor.yml`, however, giving CSS priority over, for example, SVG attributes.

ddbeck:
Now waiting on https://github.com/mdn/browser-compat-data/pull/24387.

jamesnw:
@ddbeck I updated the dist with the update BCD info, but it removed a lot of baseline support due to Safari iOS being dropped. I picked a compute_from to match Caniuse `css-cursors`, but am now wondering if I should remove the `css-cursors-newer` and `css-cursors-grab` which don't match.

ddbeck:
> @ddbeck I updated the dist with the updated BCD info, but it removed a lot of baseline support due to Safari iOS being dropped.

Yes, I think that's right—you shouldn't expect this all to work on all platforms. Odd but true.

> I picked a compute_from to match Caniuse `css-cursors`, but am now wondering if I should remove the `css-cursors-newer` and `css-cursors-grab` which don't match.

I think we should remove the `compute_from` completely and let it go. If you want to be especially careful, then you could add a TODO comment linking back to https://github.com/web-platform-dx/web-features/issues/1038. This feels like one of those features that you can "just" use without worrying about the implications on mobile and we ought to be able to communicate that fact.

jamesnw:
> > @ddbeck I updated the dist with the updated BCD info, but it removed a lot of baseline support due to Safari iOS being dropped.
> 
> Yes, I think that's right—you shouldn't expect this all to work on all platforms. Odd but true.
> 
> > I picked a compute_from to match Caniuse `css-cursors`, but am now wondering if I should remove the `css-cursors-newer` and `css-cursors-grab` which don't match.
> 
> I think we should remove the `compute_from` completely and let it go. The headline status will be correct with respect to all three caniuse features. If you want to be especially careful, then you could add a TODO comment linking back to #1038. This feels like one of those features that you can "just" use without worrying about the implications on mobile and we ought to be able to communicate that fact.

@ddbeck  Makes sense. I've removed the `compute_from` and added a TODO.

ddbeck (approved):
Thank you! 🎉

PR #1597, opened 2024-08-14
Add caret-color

jamesnw:


features/caret-color.yml
```diff
@@ -0,0 +1,6 @@
+name: Caret Color
```
ddbeck, line 1:
```suggestion
name: caret-color
```

ddbeck (approved):
Nice work here. One style nit (CSS property features are often named directly after the property, when it's not ambiguous), but this is ready to merge as soon as the suggestion is applied.

ddbeck (approved):

PR #1598, opened 2024-08-14
Add resize

jamesnw:
Currently, this is not baseline due to lack of Safari on iOS support. There is a case to be made that this is a feature that is only applicable in non-mobile/non-touch situations- if that's the case, is there a way to calculate baseline without a browser?


features/resize.yml
```diff
@@ -0,0 +1,11 @@
+name: Resize
```
ddbeck, line 1: This is a tough one. I think there's a risk of confusion with the `resize` method on resizable `ArrayBuffer`s, so we ought to do a relatively rare thing and use parenthetical disambiguation text:

```suggestion
name: resize (CSS property)
```

features/resize.yml
```diff
@@ -0,0 +1,11 @@
...
+description: The `resize` CSS property sets whether an element can be resized by the user, and on which axes.
+spec: https://drafts.csswg.org/css-ui-4/#resize
+caniuse: css-resize
+group: css
+# Baseline is false due to no Safari on iOS support, but it is available on all non-mobile browsers.
+compat_features:
```
ddbeck, line 7: Since we're using `caniuse: css-resize`, we have an obligation to set the headline status so we line up (at least roughly) with it. It doesn't make a huge difference in a non-Baseline case like this, but this would line us up with caniuse's Firefox 5:

```suggestion
status:
  compute_from: css.properties.resize.block_level_support
compat_features:
```

features/resize.yml
```diff
@@ -0,0 +1,11 @@
+name: Resize
+description: The `resize` CSS property sets whether an element can be resized by the user, and on which axes.
+spec: https://drafts.csswg.org/css-ui-4/#resize
+caniuse: css-resize
+group: css
+# Baseline is false due to no Safari on iOS support, but it is available on all non-mobile browsers.
```
ddbeck, line 6:
```suggestion
# TODO: Fix when we have a solution for single form factor features. See https://github.com/web-platform-dx/web-features/issues/1038
```

ddbeck (changes requested):
> Currently, this is not baseline due to lack of Safari on iOS support. There is a case to be made that this is a feature that is only applicable in non-mobile/non-touch situations- if that's the case, is there a way to calculate baseline without a browser?

Yes, this is a known issue. I'm suggesting in a line comment linking to the issue that would need to be resolved to fix this.

ddbeck (approved):

PR #1603, opened 2024-08-15
Add Opacity

jamesnw:


features/opacity.yml
```diff
@@ -0,0 +1,12 @@
...
+spec:
+  - https://drafts.csswg.org/css-color-3/#transparency
+  - https://drafts.csswg.org/css-color-3/#transparent
+caniuse: css-opacity
+group: css
+compat_features:
+  - css.properties.opacity
+  - css.properties.opacity.percentages
+  - css.types.filter-function.opacity
+  - css.types.color.transparent
```
captainbrosset, line 12: I would personally remove the `transparent` color name. Although `opacity` and `filter:opacity()` let me tweak the entire opacity of a rendered element, the `transparent` color only lets me make a part of an element (its text, or background, or border color) completely transparent. So, as a developer, I wouldn't consider using them together to achieve a particular task.

captainbrosset: In that vein of thoughts, although `opacity` and the filter `opacity()` function sort of do the same thing, do they complete eachother to form a standalone developer feature?
I know I'd use either one or the other, depending on what I'm doing, but not them together.
So, they might even need to be separated into 2 features. What do you think?

captainbrosset (changes requested):

features/opacity.yml
jamesnw: Now that I have worked through a few more features, I agree! I moved `transparent` to #1619 and include the opacity filter function in a later `filter-function` group of features.

jamesnw:
I removed the `caniuse` link to [CSS 3 Opacity](https://caniuse.com/css-opacity), because it is larger than just the`opacity` property and so is out of sync.

captainbrosset (approved):

PR #1604, opened 2024-08-15
Add SVG Opacity

jamesnw:


features/svg-opacity.yml
```diff
@@ -0,0 +1,15 @@
+name: SVG opacity
+description: The SVG opacity properties and attributes control the transparency of parts of an SVG.
```
autonome, line 2: - Should `propertyize` the property name

- "parts of an SVG" is a bit generic, especially as this feature has a few distinct aspects, eg global vs element scope, CSS vs SVG-specific

features/svg-opacity.yml
```diff
@@ -0,0 +1,15 @@
...
+  - css.properties.flood-opacity
+  - css.properties.stop-opacity
+  - css.properties.stroke-opacity
+  - svg.elements.feFlood.flood-opacity
+  - svg.elements.stop.stop-opacity
+  - svg.global_attributes.fill-opacity
+  - svg.global_attributes.flood-opacity
+  - svg.global_attributes.opacity
+  - svg.global_attributes.stroke-opacity
+group: css
```
autonome, line 15: do we have/need an `svg` group?

autonome (changes requested):

features/svg-opacity.yml
jamesnw: Added in https://github.com/web-platform-dx/web-features/pull/1604/commits/39c1b73dc85faa9f0654d4d64bf3e82495397235

features/svg-opacity.yml
jamesnw: In revisting this, I think the filter and gradient opacity belong in filter/gradient features, so I removed them here, and expanded the description. https://github.com/web-platform-dx/web-features/pull/1604/commits/39c1b73dc85faa9f0654d4d64bf3e82495397235

features/svg-opacity.yml
```diff
@@ -0,0 +1,10 @@
+name: SVG opacity
+description: The SVG `opacity` properties and attributes control the transparency of a stroke, fill, or entire SVG element.
```
ddbeck, line 2: In general, this feature seems to run in the opposite direction of `paint-order` https://github.com/web-platform-dx/web-features/pull/1696/files#r1749729649. What makes this an SVG thing versus a CSS thing (like paint-order)? Not saying you're wrong here, but I am trying to learn the boundaries of CSS and SVG.

More description-specific stuff:

- If it's practical to do so (i.e., we're not talking about a dozen distinct names), name things in series rather than using wildcards. Sometimes we end up having to acknowledge a long list of things ("the Example API", "the example family of CSS properties", etc.) but if it's practical to include the actual characters developers would type, let's do it.
- Prefer to put the most specific thing first (i.e., "the example CSS property" not "the CSS example property"), to aid skimming. Imagine them all in a list and you'll probably see the issue: _everything_ ends up starting with the same-y "The SVG …" text.

```suggestion
description: The `opacity`, `fill-opacity`, and `stroke-opacity` SVG attributes control the transparency of a stroke, fill, or entire SVG element. The `fill-opacity` and `stroke-opacity` CSS properties correspond to the SVG attributes.
```

jamesnw: > What makes this an SVG thing versus a CSS thing (like paint-order)?

Here's an attempt at documenting this-

- CSS properties that can be applied to HTML elements AND SVG elements (`font-family`, `writing-mode`) are generally grouped together. 
- CSS properties that can only be applied to SVG elements (`stop-color`, `vector-effect`) are generally a separate feature.

Some CSS properties for SVG elements also have analogous attributes, but this is orthogonal. The attributes should be follow the same rules, and be grouped with their related properties. 

Given this, I'm moving `svg.global_attributes.opacity` to the `opacity` feature.

features/svg-opacity.yml
```diff
@@ -0,0 +1,9 @@
+name: SVG opacity
```
ddbeck, line 1: Let's rename this file to `opacity-svg.yml` and change the title like so:

We don't have a lot of cases where this happens, but we kinda-sorta have a practice of using parentheses for disambiguation text. See https://github.com/web-platform-dx/web-features/pull/1492 for another example

```suggestion
name: Opacity (SVG)
```

features/svg-opacity.yml
ddbeck: Thank you! It's very helpful to hear how you're thinking through this stuff.

PR #1607, opened 2024-08-16
Contributing docs: add hints on lumping and splitting

ddbeck:
I've been doing a lot explaining on this question: when do I create a new feature (versus adding a key and using a `compute_from` and so on)?

This is my attempt to provide a general answer (which is hard!—the specifics often matter).

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -32,6 +32,31 @@ For example, the `fetch()` API is a feature, the `:has()` CSS pseudo-class funct
 
 For the time being, this repository focuses only on features that are implemented in some browser and that are already documented on MDN.
 
+### Hints for distinguishing features
+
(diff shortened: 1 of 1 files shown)


captainbrosset (approved):

captainbrosset:
I expect that we'll go back and forth on the exact wording, and that others may have slightly different ways of doing this, or additional tips. But this is a great addition and I think we should merge it.

Elchi3 (approved):
Maybe our future selves will come up with more strict/lintable guidelines on how to create distinct web-features but until then these guidelines sound good to me.

ddbeck:
Yes, I hope that we'll find more strict rules about these things. I think there will always be some fundamental fuzziness if we're faithfully reflecting the developer perspective, though I also hope that we'll earn some trust with developers, such that we'll be able to nudge them in helpful directions too.

Merging this now. Thank you both!

Issue #1600, opened 2024-08-15
`dist` doesn't catch errors that `test` does


autonome:
>> hm, worked locally w/ `dist`, so maybe we should have that script fail on this also

> Agreed, it's caught me up a couple of times. I usually will run `npm run dist && npm run test` in the same line just to double check. Can you file an issue for this please?

_Originally posted by @petele in https://github.com/web-platform-dx/web-features/issues/1592#issuecomment-2289589482_
            
The workflow in the docs and from talking to folks is to run `dist` to check the yaml changes for a feature.

However, you can have a perfectly valid entry, as above, which runs fine in `dist` locally, but fails the automated test.

I haven't looked, so not sure what the right solution is. Fix the test to accept valid yaml entries? Cram `test` in to `dist`? Just get everyone to run both per above?

`dist` already takes a bit long, so not sure we want to always run `test` with it.

(Side thought... I'd like a `watch` command on just the given feature I'm working on...)

ddbeck:
Can you point me to a concrete example (e.g., a failing GitHub Actions workflow run) of `dist` succeeding and `test` failing?

I suspect some failures could be improved by better onboarding docs (e.g., recommending turning on Prettier formatting in your editor). But it's also possible there's some actual bugs.

> `dist` already takes a bit long, so not sure we want to always run `test` with it.

`dist` can run faster by specifying feature files as an argument: `npm run dist -- feature/some-file.yml` won't refresh _everything_, just the files you're working on.

> (Side thought... I'd like a `watch` command on just the given feature I'm working on...)

I found this hard to do in a platform-independent way, though a PR would be welcome. You might want to see what options are available in your system and shell. I have [a script](https://github.com/ddbeck/web-features/blob/32d6faae8bb1e0e090e9448473e76e1e54316501/bin/dist-watch.fish) for watching for YAML changes and rerunning dist which you might find instructive though not portable (a fish shell script that depends on a non-standard on macOS command line tool, `fswatch`).

autonome:
> Can you point me to a concrete example (e.g., a failing GitHub Actions workflow run) of `dist` succeeding and `test` failing?

https://github.com/web-platform-dx/web-features/pull/1592#pullrequestreview-2238777344

> `dist` can run faster by specifying feature files as an argument: `npm run dist -- feature/some-file.yml` won't refresh _everything_, just the files you're working on.

yesss, this is great, thank you! i can wire this up to whatever watch system now.

[...]

PR #1608, opened 2024-08-16
Contributing docs: add some first-time contributing hints

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/1600

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -100,6 +100,7 @@ To learn more about the fork and pull request process, see [Fork a repository](h
 When you have the repository cloned locally, set up your local environment:
 
 1. [Install Node.js](https://nodejs.org).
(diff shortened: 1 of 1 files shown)


petele:
I tried adding Prettier to VSCode, and it didn't report a warning I expected, but it was reported when running npm test.

Repro steps: 
Put 2 spaces after `description:`.

Expect: ✅ Error in VS Code reporting extra space,  ✅ `npm run test` reports warning
Actual: ❌  No error in VS Code, ✅  `npm run test` reports warning

ddbeck:
@petele I don't think the Prettier VS Code extension gives warnings. I think it's only a formatter. You can invoke it via the Command Palette "Format Document" or (my recommendation) turn on the `editor.formatOnSave` preference. I also suggest turning on the `prettier.requireConfig` setting so it doesn't start running everywhere, unbidden.

PR #1610, opened 2024-08-16
Add system-color

jamesnw:
https://github.com/Fyrd/caniuse/issues/4179 shows interest as a feature, including links to posts about it.


features/system-color.yml
```diff
@@ -0,0 +1,10 @@
+name: system-color
```
ddbeck, line 1: This seems to be a bit more "natural" in use than the hyphenated form. Also, see my commentary on the description:

```suggestion
name: System colors
```

features/system-color.yml
```diff
@@ -0,0 +1,10 @@
...
+description: The `system-color` CSS keywords allow authors to match default colors from the user agent.
```
ddbeck, line 2: As written, this suggests `system-color` is a keyword you can actually type and get something from. I think that's not the case. I'd suggest something along these lines, using examples to represent the real-world use:

```suggestion
description: The system color CSS keywords, such as `ButtonBorder` or `LinkText`,  allow authors to match default colors from the user agent.
```

features/system-color.yml
```diff
@@ -0,0 +1,10 @@
...
+spec: https://drafts.csswg.org/css-color-4/#css-system-colors
+group: css
+status:
+  compute_from: css.types.color.system-color
+compat_features:
+  - css.types.color.system-color
```
ddbeck, line 8: This is not a problem here, but CSS types in browser compat data (BCD) are quite messy. We'll probably see some cases where we'll have to write support statuses by hand, or have no status at all. I was about to suggest having a note about this, but I think feature grouping will probably do the work here.

ddbeck (changes requested):

jamesnw:
Replaced by #1619

PR #1611, opened 2024-08-16
Add named color

jamesnw:


features/named-color.yml
```diff
@@ -0,0 +1,7 @@
+name: Named colors
+description: The `named-color` CSS values are a standard set of colors that can be referenced by a name, such as `red` or `limegreen`.
+spec: https://drafts.csswg.org/css-color-4/#named-colors
+group: css
+compat_features:
+  - css.types.color.named-color
+  - css.types.color.named-color.rebeccapurple
```
captainbrosset, line 7: It won't change the baseline high status, but using: `compute_from: css.types.color.named-color` would make it seem like named colors have always been supported (earliest versions of browsers), which they were (one color was added later, but probably fine to hide the date details).

features/named-color.yml
```diff
@@ -0,0 +1,7 @@
+name: Named colors
+description: The `named-color` CSS values are a standard set of colors that can be referenced by a name, such as `red` or `limegreen`.
```
captainbrosset, line 2: Using the term `named-color` in a code block makes it seem (at least to me) like this is a keyword you can use when writing CSS.
Possible rephrase:

```suggestion
description: A standard set of colors that can be referenced by names, such as `red` or `limegreen`.
```

captainbrosset (changes requested):

captainbrosset (approved):

features/named-color.yml
```diff
@@ -0,0 +1,9 @@
+name: Named colors
+description: A standard set of colors that can be referenced by names, such as `red` or `limegreen`.
```
ddbeck, line 2: These are generally formed as complete sentences. And this doesn't seem to mention anywhere that we're talking about CSS values.

Very optionally, an opportunity to use an also known as, if we want to reference some ancient history. (This is me mostly demonstrating a convention to you than earnestly advocating for it.)

```suggestion
description: Some CSS color values can be referenced by name, such as `red` or `limegreen`. They stand in for specific RGB color values. Also known as the X11 color set.
```

ddbeck (changes requested):

features/named-color.yml
jamesnw: I'll omit the X11 reference as there are a few minor differences. Although while looking into X11, I finally found the reason why `gray` is darker than `dark gray` 😆

jamesnw:
Replaced by #1619

PR #1615, opened 2024-08-16
Add hsl

jamesnw:


features/hsl.yml
```diff
@@ -0,0 +1,13 @@
+name: HSL
+description: The HSL color model specifies colors using hue, saturation and lightness channels. It can be specified using `hsl()` and `hsla()` functions.
```
ddbeck, line 2: We should make it clear we're talking about CSS here, specifically.

Also, "specifies" is sometimes unclear and there are usually better alternatives (see [the guideline entries relating to this](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#specifies)). I think "pick" is a nice choice (aligning with color pickers, for instance) for colors:

```suggestion
description: The `hsl()` and `hsla()` CSS functions pick colors using hue, saturation and lightness channels.
```

features/hsl.yml
```diff
@@ -0,0 +1,13 @@
...
+spec: https://drafts.csswg.org/css-color-4/#the-hsl-notation
+caniuse: css3-colors
+group: css
+status:
+  compute_from: css.types.color.hsl
+# Editorial override has been applied to ignore mixed_type_parameters.
```
ddbeck, line 8: It's unclear why. Is this the same reason as the Oklab feature? If so, I'd suggest copying the whole comment here (including the key word TODO). Otherwise, it'd be nice to more about this to figure out how to proceed.

features/hsl.yml
```diff
@@ -0,0 +1,13 @@
+name: HSL
+description: The HSL color model specifies colors using hue, saturation and lightness channels. It can be specified using `hsl()` and `hsla()` functions.
+spec: https://drafts.csswg.org/css-color-4/#the-hsl-notation
+caniuse: css3-colors
+group: css
+status:
+  compute_from: css.types.color.hsl
```
ddbeck, line 7: If we're going to mention the `hsla()` version in the feature description, then I think it makes sense to drop the `compute_from`, since the alpha channel is so central to the overall feature.

ddbeck (changes requested):

jamesnw:
Replaced by #1619

PR #1619, opened 2024-08-19
Add Color types

jamesnw:
Adds Color Type Group, and members:

- System color
- Named color
- currentColor
- RGB
- HWB
- HSL

I considered grouping the models from the sRGB space (HWB, HSL), but the HWB support story is significantly different.

Adds existing members:
 - Color function
 - Lab/LCH
 - OkLab/OkLCH


features/hsl.yml
```diff
@@ -0,0 +1,12 @@
+name: HSL
+description: The `hsl()` and `hsla()` CSS functions pick colors using hue, saturation and lightness channels.
+spec: https://drafts.csswg.org/css-color-4/#the-hsl-notation
+caniuse: css3-colors
+group: color-types
+status:
+  compute_from: css.types.color.hsl.alpha_parameter
```
jamesnw, line 7: Some notes from the [initial PR](https://github.com/web-platform-dx/web-features/pull/1615)-

- @ddbeck suggested omitting this `compute_from`, as the alpha component was a key part of the story.  
- `mixed_type_parameters` was initially excluded to match Oklab, but the associated TODO was unrelated to `mixed_type_parameters`. I'm not sure why it was omitted there, and suggest including it here. 

However, that moves this to `baseline: false`, as Safari doesn't support mixed type parameters. So, I opted to set the `compute_from` to the `alpha_parameter` feature.

features/hsl.yml
```diff
@@ -0,0 +1,12 @@
+name: HSL
+description: The `hsl()` and `hsla()` CSS functions pick colors using hue, saturation and lightness channels.
```
ddbeck, line 2: Let's mention the alpha channel too:

```suggestion
description: The `hsl()` and `hsla()` CSS functions pick colors using hue, saturation, lightness, and alpha (transparency) channels.
```

features/hsl.yml
ddbeck: OK that works. Except…

Comparing this with the linked caniuse entry (css3-colors), it appears that caniuse doesn't care about the alpha channel, so we make this feature appear rather newer than caniuse does (e.g., shipping in Edge 12 versus Edge 79). It's not a huge problem, but it is after our "don't worry about minor discrepancies before 2020" cutoff. I've filed https://github.com/mdn/browser-compat-data/issues/24191 on BCD to deal with the problem upstream.

In the mean time, I'd like to add something like this:

```suggestion
  compute_from: css.types.color.hsl.alpha_parameter
# TODO: For alignment with caniuse, add `hsla()` features when https://github.com/mdn/browser-compat-data/issues/24191 is resolved and update compute_from
```

features/hwb.yml
```diff
@@ -0,0 +1,9 @@
+name: HWB
+description: The `hwb()` CSS function picks colors using hue, whiteness and blackness channels. It is used by many color pickers.
```
ddbeck, line 2: Kinda like not mentioning support information in descriptions, I think it's a good idea to avoid other non-durable information like popularity. I don't think we lose too much cutting it out like this:

```suggestion
description: The `hwb()` CSS function picks colors using hue, whiteness and blackness channels.
```

features/rgb.yml
```diff
@@ -0,0 +1,14 @@
+name: RGB
+description: The `rgb()`, rgba()` and hexadecimal notation pick colors using red, green and blue channels.
```
ddbeck, line 2: Nits for serial comma, a typo, and parallelism. I think we also ought to mention the alpha channel, if we're pinning the headline support status to it:

```suggestion
description: The `rgb()`, rgba()`, and hexadecimal (as in `#004488ff`) notations pick colors using red, green, blue, and alpha (transparency) channels.
```

ddbeck (changes requested):
This is really nice—the group here makes a lot of sense. A few small suggestions for you, but this is close to done. Thank you James!

features/hwb.yml
jamesnw: I was wanting to add the primary use case or reason it exists since this seems more narrowly focused, but I'm fine removing it for your reasons and also since the other color types don't have similar info.

features/transparent-color.yml
```diff
@@ -0,0 +1,5 @@
+name: Transparent color
+description: The `transparent` CSS color value is a shorthand for transparent black.
+spec: https://drafts.csswg.org/css-color-3/#transparent
+compat_features:
+  - css.types.color.transparent
```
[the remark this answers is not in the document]
ddbeck: I also sent https://github.com/mdn/browser-compat-data/pull/24209

features/hwb.yml
```diff
@@ -0,0 +1,9 @@
+name: HWB
+description: The `hwb()` CSS function picks colors using hue, whiteness and blackness channels.
```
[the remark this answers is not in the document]
jamesnw: I had one job 10 years ago that didn't use the serial comma, and it's a hard habit to break. 😆 

Would it be worth adding a note to the style guide about the serial comma?

features/named-color.yml
```diff
@@ -0,0 +1,10 @@
+name: Named colors
+description: Some CSS color values can be referenced by name, such as `red` or `limegreen`. They stand in for specific RGB color values.
+spec: https://drafts.csswg.org/css-color-4/#named-colors
+group: color-types
+status:
+  compute_from: css.types.color.named-color
+compat_features:
+  - css.types.color.named-color
+  - css.types.color.named-color.rebeccapurple
+  - css.types.color.transparent
```
jamesnw, line 10: This needs to be updated to `css.types.color.named-color.transparent` pending #1675

features/hwb.yml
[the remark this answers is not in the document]
ddbeck: Not a bad idea, though that led me down a bit of a rabbit hole. I've opened https://github.com/web-platform-dx/web-features/pull/1686.

And I was an English major, so MLA style was cemented in me for life. 😄

features/named-color.yml
ddbeck: I'll merge this as-is and fix it in the Dependabot PR. 👍

ddbeck (approved):
Thank you!

PR #1620, opened 2024-08-19
Add catch for invalid spec URLs

jamesnw:
Previously, a missing or invalid spec URL would throw an error without a message. This change surfaces that error when running `npm run test:specs`.

In addition, the test was reporting the number of specs as the number of features, so now it reports both.

### DIFF
--- a/scripts/specs.ts
+++ b/scripts/specs.ts
@@ -77,7 +77,8 @@ function suggestSpecs(bad: URL): void {
     }
 }
 
-let checked = 0;
+let checkedFeatures = 0;
+let checkedSpecs = 0;
 let errors = 0;
 
 // Ensure every exception in defaultAllowlist is needed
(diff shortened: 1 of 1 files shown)


scripts/specs.ts
```diff
@@ -92,19 +93,25 @@ for (const [allowedUrl, message] of defaultAllowlist) {
 for (const [id, data] of Object.entries(features)) {
     const specs = Array.isArray(data.spec) ? data.spec : [data.spec];
     for (const spec of specs) {
-        const url = new URL(spec);
-        if (!isOK(url)) {
+        try {
+            var url = new URL(spec);
```
ddbeck, line 97:
```suggestion
            const url = new URL(spec);
```

ddbeck (dismissed):
Oh excellent, thank you!

scripts/specs.ts
ddbeck: Ha, so I started to say something about this and forgot about it.

I think `var` is pretty confusing—it's weird that it's possible to reference it outside the scope here. Would you mind changing this a `let url` (where the original `const url = URL(spec);` was), then reassigning on this line?

```suggestion
            url = new URL(spec);
```

ddbeck (changes requested):

scripts/specs.ts
jamesnw: Makes sense to me! Updated.

ddbeck (approved):
Thank you!

PR #1630, opened 2024-08-20
Add keywords to font optical sizing

jamesnw:
The keywords were very likely added simultaneously to the property, so I suspect the `<=` values for the keywords could be adjusted in BCD.

### DIFF
--- a/features/font-optical-sizing.yml
+++ b/features/font-optical-sizing.yml
@@ -2,3 +2,9 @@ name: font-optical-sizing
 description: The `font-optical-sizing` CSS property sets whether text rendering is optimized for viewing at different sizes.
 spec: https://drafts.csswg.org/css-fonts-4/#font-optical-sizing-def
 group: fonts
+status:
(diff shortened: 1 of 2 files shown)


features/font-optical-sizing.yml
```diff
@@ -2,3 +2,9 @@ name: font-optical-sizing
 description: The `font-optical-sizing` CSS property sets whether text rendering is optimized for viewing at different sizes.
 spec: https://drafts.csswg.org/css-fonts-4/#font-optical-sizing-def
 group: fonts
+status:
+  compute_from: css.properties.font-optical-sizing
```
ddbeck, line 6: Yeah, you're right. I've sent https://github.com/mdn/browser-compat-data/pull/24198 to fix this. I think we ought to drop the compute_from and let the next BCD release fix the version information.

```suggestion
```

ddbeck (changes requested):

ddbeck (approved):

PR #1632, opened 2024-08-20
Add font-palette keywords

jamesnw:
Only added `dark`, `light`, and `normal` keywords- the others were coming from the tag in WPF.


ddbeck (approved):
This makes sense. Thank you!

PR #1636, opened 2024-08-21
Add speech API

autonome:
- [x] confirmed not already in progress
- [x] moved files
- [x] updated filename if needed (eg remove "TC39-" or s/webdriver2/webdriver/)
- [x] removed draft date
- [x] determined name
- [x] added entry point
- [x] verified baselines
- [x] checked api list against mdn list
- [x] added to a group or not
- [ ] what else?

## Resources

- Spec: https://wicg.github.io/speech-api/
- MDN: https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API
- CIU: https://caniuse.com/speech-recognition
- Developer
  - SO: https://stackoverflow.com/questions/tagged/whatever
  - NPM: https://www.npmjs.com/search?q=whatever

## Filename

Philip did the synth part already and named it `speech-synthesis.yml`.

The remaining bits are the `SpeechGrammar` and `SpeechRecognition` APIs, so i'll split this up into three files for now, along the `speech-*.yml` pattern.

## Name

Again, will follow the pattern already started w/ the synth api.

## Group

There's already separate groups made, which seems odd. Feels like if we're coalescing 14k pieces of web platform surface area into 2-3k features, all of this would be the `Web Speech API`, which is how developers, spec authors and MDN all seem to think about it. But Philip kicked this one off and is one of the spec authors so, I'll leave it as is for now.

## Splits/Merges

See above, I suspect we group this into a single feature later.


autonome:
@ddbeck any idea why SpeechGrammar is in BCD but not mapping here?

ddbeck:
> @ddbeck any idea why SpeechGrammar is in BCD but not mapping here?

I'm sorry, but I'm not 100% sure what this means. Are you asking why the computed status is unsupported but the compat data shows some kind of support (e.g., https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammar#browser_compatibility) it's because of prefixes and runtime flags, which don't count toward a Baseline status.

autonome:
bah, i had commands that searched `features/draft/spec` but not `features/draft` so missed the file there.

PR on hold indefinitely, if/when speech apis get unprefixed.

ddbeck:
It's OK to have features that aren't Baseline in the feature set (we just need to make sure that it makes sense when we do!), if you still want to move ahead on this.

autonome:
Ok, i think this is ready now. `SpeechGrammar` is all prefixed everywhere , looks like.

jamesnw (approved):
Pre-review, overall looks good!

features/speech-recognition.yml
```diff
@@ -0,0 +1,45 @@
+name: Speech recognition
+description: The `SpeechRecognition` API provides the ability to recognize voice context from an audio input (typically via the device's default speech recognition service) and respond appropriately.
```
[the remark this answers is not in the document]
autonome: thanks! i went with the 2nd one, is simpler. re "appropriately", yeah that's ambiguous, i will update to clarify about how the result can be handled (for both apis).

features/speech-synthesis.yml
```diff
@@ -3,3 +3,46 @@ description: The `SpeechSynthesis` API converts text into audio using a syntheti
 spec: https://wicg.github.io/speech-api/#tts-section
 group: speech
 caniuse: speech-synthesis
+compat_features:
```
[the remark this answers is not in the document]
autonome: thanks for pointing out, adding a "check CIU baseline" to my checklist

autonome:
Thanks for review @jamesnw, fixes made, ready for @ddbeck

features/speech-grammar.yml
```diff
@@ -0,0 +1,14 @@
+name: Speech grammar
+description: The `SpeechGrammar` interface of the Web Speech API represents a set of words or patterns of words for the `SpeechRecognition` service to recognize.
```
ddbeck, line 2: This text (and the MDN docs) strongly imply that you wouldn't use this API alone—you'd only ever use it with `SpeechRecognition`. I'd combine this with SpeechRecognition.

features/speech-recognition.yml
```diff
@@ -0,0 +1,45 @@
+name: Speech recognition
+description: The `SpeechRecognition` API converts audio into text using the device's speech recognition service, providing methods and events for fine-grained control and observation of the interaction.
```
ddbeck, line 2: I think we can go more brief here and dodge having to make observation or interaction more meaningful.

```suggestion
description: The `SpeechRecognition` API converts audio into text using the device's speech recognition service.
```

features/speech-synthesis.yml
```diff
@@ -1,5 +1,50 @@
 name: Speech synthesis
-description: The `SpeechSynthesis` API converts text into audio using a synthetic voice.
+description: The `SpeechSynthesis` API converts text into audio using a synthetic voice, with the ability to control various aspects of the interaction and output, including specifcying custom voices.
```
ddbeck, line 2: Suggestions for brevity and to avoid a circular description

```suggestion
description: The `SpeechSynthesis` API converts text to speech from artificial voices.
```

ddbeck (changes requested):
Getting pretty close. One big suggestion (merging a feature) and some nits.

features/speech-synthesis.yml
jamesnw: I find "to speech from artificial voices" confusing, as the "from" is implicitly "text" already- would "to speech with artificial voices" work?

features/speech-recognition.yml
autonome: I always to include a mention when there are events. Not a blocker and we can come up w/ general approaches to this and modify later if we think is needed.

ddbeck (approved):
Thank you!

Issue #1624, opened 2024-08-20
Cover the features listed in the State Of XYZ surveys


captainbrosset:
Here is another potentially useful burndown list to add features to the repo.
The CSS and HTML features are mentioned, in some way, in either State of CSS or State of HTML (over the past 4 years).
This was made by using the GraphQL API as detailed in https://github.com/Devographics/surveys/issues/247#issuecomment-2296633083.

## CSS (2021/2022/2023/2024)

- [ ] accent_color
- [ ] anchor_positioning
- [ ] aria_attributes
- [ ] aspect_ratio
- [ ] at_container_size
- [ ] at_container_style
- [ ] at_container
- [ ] at_property
- [ ] at_scope
- [ ] backdrop_filter
- [ ] blend_modes
- [ ] break_rules
- [ ] calc
- [ ] cascade_layers
- [ ] clip_path
- [ ] color_contrast
- [ ] color_function
- [ ] color_gamut
- [ ] color_mix
- [ ] color_scheme
- [ ] comparison_functions
- [ ] conic_gradient
- [ ] containment
- [ ] content_visibility
- [ ] css_nesting
- [ ] currentcolor
- [ ] discrete_properties_animations
- [ ] exponential_functions
- [ ] feature_support_queries
- [ ] filter_effects
- [ ] flexbox_gap
- [ ] flexbox
- [ ] focus_visible
- [ ] font_display
- [ ] font_palette
- [ ] font_variant_numeric
- [ ] font_variant
- [ ] forced_colors
- [ ] gradient_color_spaces
- [ ] grid
- [ ] hanging_punctuation
- [ ] has_selector
- [ ] houdini
- [ ] image_function
- [ ] image_set
- [ ] individual_transform_properties
- [ ] initial_letter
- [ ] intrinsic_sizing
- [ ] light_dark
- [ ] line_clamp
- [ ] linear_easing_function
- [ ] logical_properties
- [ ] marker
- [ ] masks
- [ ] media_queries_range_contexts
- [ ] multi_column
- [ ] nth_child_of_s
- [ ] object_fit
- [ ] object_view_box
- [ ] offset_path
- [ ] overflow_anchor
- [ ] overscroll_behavior
- [ ] perspective
- [ ] pointer_events
- [ ] position_sticky
- [ ] prefers_color_scheme
- [ ] prefers_contrast
- [ ] prefers_reduced_data
- [ ] prefers_reduced_motion
- [ ] relative_colors
[...]

PR #1643, opened 2024-08-21
Add initial letter

SondraE:
#1624 Adds features from State of CSS


features/initial-letter.yml
```diff
@@ -0,0 +1,6 @@
+name: initial-letter
+description: The `initial letter` CSS property sets the number of lines the first letter occupies.
```
ddbeck, line 2: Huh, I didn't know this existed! I'm making some suggestions here to make it more clear where this applies and to work in some related typography terms, though they're not strictly required. What do you think?

```suggestion
description: The `initial-letter` CSS property sets the number of lines the first letter of an element occupies. You can use the property to make a raised capital or drop cap.
```

ddbeck (approved):
Thank you, @SondraE! This looks good, though I have optional suggestions on the description for you to consider before we merge this.

PR #1644, opened 2024-08-21
Add Font face features

jamesnw:
- Adds api.CSSFontFaceRule to font-face
- Adds font-display
~- Adds unicode-range~


cklim0815is (approved):

features/unicode-range.yml
```diff
@@ -0,0 +1,7 @@
+name: unicode-range
+description: The `unicode-range` CSS descriptor sets the characters to be used from a font face. If the characters aren't present on the page, the font is not downloaded.
```
autonome, line 2: Non-blocking:
- "sets" doesn't seem right, maybe "specifies"?
- could be more specific about the descriptor contents, eg "from a font face, using Unicode codepoints individually, in a range, or wildcard match"

autonome (approved):
Pre-review before @ddbeck.

Not a blocking comment, just an observation on feature composition that I'm sure we'll figure out over time: From a designer/developer POV, seems like `font-display` could be merged into `font-face` feature, vs disappearing in a sea of `font` group features. Would help with discovery.

features/unicode-range.yml
jamesnw: Thanks, I like these suggestions!

features/font-display.yml
```diff
@@ -0,0 +1,7 @@
+name: font-display
+description: The `font-display` CSS descriptor sets what is displayed while a font face loads.
```
ddbeck, line 2:
```suggestion
description: The `font-display` CSS descriptor sets whether to show a substitute font or nothing while a font face loads. The property manages what's known as a flash of unstyled text or flash of invisible text.
```

What do you think of an elaboration like this? The first bit is to avoid a circular description ("font-display … sets what is displayed"), while the second is address what this is for, though maybe it's being a bit too assertive.

features/unicode-range.yml
```diff
@@ -0,0 +1,7 @@
+name: unicode-range
+description: The `unicode-range` CSS descriptor specifies the characters to be used from a font face, using Unicode codepoints individually, in a range, or with a wildcard match. If the characters aren't present on the page, the font is not downloaded.
```
ddbeck, line 2: cc: @autonome 

I'm going to do something terrible here which is to lay down a lot of text on how to do something, then ask you to ignore it and throw it out—skip to HOWEVER if you find this unbearable. 😄 

So I'm on a bit of a crusade against "specify" (and "determine") in place of "set", "pick", or "choose." They're a bit like "utilize" — they have a lot of bonus syllables that say the same thing as a more direct alternative ("use"). But this crusade is for a good cause: speaking directly to developers, in a direct and confident way.

Anyway, here's my attempt to revise this. I think not loading is the central point, so I tried to move it to the beginning of the description, where you're more likely to read/skim it.

```suggestion
description: The `unicode-range` CSS descriptor controls whether to load a font, if the characters picked by the descriptor appear in the page. You can choose characters by Unicode code point, code point range, or wildcard pattern.
```

HOWEVER—and I only noticed this after writing most of the above, so forgive me—I think this feature is somewhat needless. It's part of `@font-face` and doesn't meaningfully depart from its overall status. I think we should add the `unicode-range` key to `font-face` and the `font-unicode-range` to the `caniuse` array.

ddbeck (changes requested):

features/font-display.yml
jamesnw: I like that, and adding that specific use case makes sense.

features/unicode-range.yml
jamesnw: Not unbearable at all! And I agree with the merge.

ddbeck (approved):

PR #1648, opened 2024-08-22
Add background-blend-mode and mix-blend-mode

jamesnw:
The `lighter-plus` and `darker-plus` did not seem to garner enough community interest to warrant separate features or block baseline.

Also, `css.types.blend-mode` are shared between mix- and background-, so I put them with the earlier feature, but `lighter-plus` and `darker-plus` are only available on mix-.

For both mix- and background-, the Safari versions are slightly different between [Caniuse](https://caniuse.com/css-backgroundblendmode) and BCD- is this a candidate for https://github.com/web-platform-dx/web-features/issues/1499?

Part of #1624


groups/blend-mode.yml
```diff
@@ -0,0 +1,3 @@
+# https://drafts.fxtf.org/compositing-1/
+# Blend modes control how parts of an element are composited, or visually mixed.
+name: Blend mode
```
captainbrosset, line 3: I think this should be a child of CSS (i.e. add `parent: css`).

features/background-blend-mode.yml
```diff
@@ -0,0 +1,8 @@
+name: background-blend-mode
+description: The `background-blend-mode` CSS property sets how an element's background is visually mixed.
```
captainbrosset, line 2:
```suggestion
description: The `background-blend-mode` CSS property blends an element's background image and background color.
```

I've never used this so I don't know whether this is correct, but MDN seems to say it is.

features/mix-blend-mode.yml
```diff
@@ -0,0 +1,12 @@
+name: mix-blend-mode
+description: The `mix-blend-mode` CSS property sets how an element's content is visually mixed.
```
captainbrosset, line 2: How about:

```suggestion
description: The `mix-blend-mode` CSS property blends an element's content with its background or parent elements.
```

captainbrosset (changes requested):

features/mix-blend-mode.yml
jamesnw: I like the "blends" verb, so I accepted the change. I also added example blend modes to convey that there are options.

ddbeck (approved):
I think this is ready to go now. Thank you!

PR #1129, opened 2024-05-22
Add clip-path

foolip:

Issue #1128, opened 2024-05-22
How to represent clip-path?


foolip:
We already have a [masks](https://github.com/web-platform-dx/web-features/blob/main/features/masks.yml) feature, but not one for https://drafts.fxtf.org/css-masking/#the-clip-path from the same spec.

https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path#browser_compatibility is a bit messy and I need advice.

@progers how do you think we should represent this? Should we do like https://caniuse.com/css-clip-path and have a single feature that's still not fully supported? Or can it be split into smaller parts? If so, what are those smaller parts?

foolip:
This situation would be easier to deal with if we had https://github.com/web-platform-dx/web-features/issues/915.

captainbrosset:
My two cents is that we should have some kind of initial support feature for clip-path. I've used it with basic circle and polygon shapes for ages and would find it weird if the feature was marked as non baseline.

Maybe we call it something like "basic shape clip path".

progers:
Fredrik (fs@opera.com) is actively working on https://crbug.com/40134477 which is the bug tracking Chromium's implementation of external resources for clip-path/filter/stroke/etc. The +1 count on that bug indicates it is an important feature for developers. External mask-image references were part of the css-masking area of interop-2023, which is another datapoint that developers care about this. The lack of external resources for filter/stroke/etc is not reflected on caniuse for other properties (e.g., https://caniuse.com/css-filters). However this is handled, we should be consistent about the external resource issue for clip-path/filter/mask/etc.

@foolip , do you have enough information to make a call on this?

foolip:
Thank you @progers! It sounds like external resources is the only bugs/limitations that we should either treat as separate follow-on features, or track with notes and possibly partial support. And we should probably make the same call for https://caniuse.com/css-clip-path and https://caniuse.com/css-filters.

Since we currently don't have partial support or notes the only option in the immediate term is to align with https://caniuse.com/css-filters and make no mention at all about external resources.

But since there is no urgency to add this feature right now, I think we should block on https://github.com/web-platform-dx/web-features/issues/915.

mirisuzanne:
To me this seems like a case where the sub-feature is consistently supported across browsers, and the un-supported aspect can be treated as a new feature if/when it gains traction. I agree with @captainbrosset here that is seems strange to hold up features that have broad support and usage, just because there is more functionality expected down the road.

tidoust:
Closing as addressed.

PR #1649, opened 2024-08-22
Add clipping, shapes and masking

jamesnw:
Targeting the [CSS4 "Clipping, shapes & masking"](https://github.com/CSS-Next/css-next/blob/main/list-of-css-features.md) group.

- [x] Clip-path (needed for #1624)
- [x] Shapes (needed for #1624)
- [x] Add group to Masks
- [x] Merge with masks-
  - [x] mask-composite
  - [x] mask-mode
- [x] New mask stories
  - [x] mask-border (different compat story, use case)
  - [x] mask-type (only applies to SVG <mask>, discrete feature)


groups/clipping-shapes-masking.yml
```diff
@@ -0,0 +1,2 @@
+name: Clipping, shapes and masking
+parent: css
```
captainbrosset, line 2: The potential issue with this group is that it now contains features that apply to both CSS and SVG.
Your features make sense to me, and I'm fine with making them apply to both languages, if others are. But in that case, we need to make this group belong to both CSS and SVG (which is not possible today).

jamesnw: Good point- I've been thinking we will need to address the overlap between CSS and SVG more thoroughly at some point. Allowing for multiple parents for a group would be complex, so I'll remove the parent group here for now.

features/path-shape.yml.dist
jamesnw, line 1: This is a bit of a surprising result- these features do have some implementation, and I'm not sure if I'm misreading the BCD data, or if there is a bug in generating this.

jamesnw:
I just saw #1129, which is blocked with more conversation in #1128.

features/clip-path-animatable.yml
```diff
@@ -0,0 +1,6 @@
+name: Animatable clipping paths
+description: Clipping paths in CSS are animated using transitions and CSS animations.
```
mirisuzanne, line 2:
```suggestion
description: Clipping paths in CSS are animated using CSS transitions and animations.
```

features/clip-path-animatable.yml
```diff
@@ -0,0 +1,6 @@
...
+spec: https://drafts.fxtf.org/css-masking-1/#the-clip-path
```
mirisuzanne, line 3: The 'current' spec:
```suggestion
spec: https://www.w3.org/TR/css-masking/#the-clip-path
```

I would update the other spec links to match.

features/clip-path-boxes.yml
```diff
@@ -0,0 +1,8 @@
+name: Clip path boxes
+description: The `fill-box`, `stroke-box`, and `view-box` values for the `clip-path` set the part of the shape that will be used to clip the element.
```
mirisuzanne, line 2:
```suggestion
description: The `fill-box`, `stroke-box`, and `view-box` values for `clip-path` set an edge of the elements box to use as the clipping shape.
```

mirisuzanne (commented):
I like this way of breaking things out to clarify the support for various parts. Left a few comments on things that I think could be clarified.

features/clip-path-animatable.yml
jamesnw: I'll defer this until resolution of https://github.com/web-platform-dx/web-features/issues/1785

features/clip-path-boxes.yml
jamesnw: I added an apostrophe in `element's` but wanted to verify that was the intent.

mirisuzanne: the intent was possessive, but I'm dyslexic. The element has a box. :)

jamesnw: I just wanted to confirm it wasn't like `attorneys general` or something 😆

HiddenCedarsInvestmentGroup (approved):
😊

HiddenCedarsInvestmentGroup (approved):
😊

features/path-shape.yml
```diff
@@ -0,0 +1,10 @@
+name: "`path()` shape"
```
ddbeck, line 1: Let's do this feature in a separate PR (from synchronous discussion)

features/clip-path-animatable.yml
```diff
@@ -0,0 +1,6 @@
+name: Animatable clipping paths
+description: Clipping paths in CSS are animated using CSS transitions and animations.
```
ddbeck, line 2: Might be nice to make this more concrete:

```suggestion
description: The `clip-path` property can be animated using CSS transitions and animations.
```

features/clip-path.yml
```diff
@@ -0,0 +1,19 @@
+name: clip-path
+description: The `clip-path` CSS property and SVG attribute sets an area of an element that should be displayed. Everything outside the area will be hidden.
```
ddbeck, line 2: For parallelism with the second sentence (and avoid "display" as a verb):

```suggestion
description: The `clip-path` CSS property and SVG attribute sets the visible area of an element. Everything outside the area will be hidden.
```

features/mask-type.yml
```diff
@@ -0,0 +1,8 @@
+name: mask-type
+description: The `mask-type` CSS property on an SVG `<mask>` sets whether the mask is a `luminance` or `alpha` mask.
```
ddbeck, line 2:
```suggestion
description: The `mask-type` CSS property on an SVG `<mask>` element sets whether the mask is a `luminance` or `alpha` mask.
```

features/path-shape.yml
ddbeck: To elaborate here: the BCD data is very messy. It has subfeatures to describe support in `d`, `shape-outside`, and `offset-path`, but _also_ represents that information as partial support on `path`. It ought to do one or the other. We'll need to fix this in BCD.

(Please @ me on that PR, when you open it.)

ddbeck (changes requested):
A couple small requests and elaborations on previous requests

features/path-shape.yml
jamesnw: Moved to https://github.com/web-platform-dx/web-features/pull/1799

features/clip-path.yml
```diff
@@ -0,0 +1,19 @@
+name: clip-path
+description: The `clip-path` CSS property and SVG attribute sets the visible area of an element. Everything outside the area will be hidden.
```
ddbeck, line 2:
```suggestion
description: The `clip-path` CSS property and SVG attribute set the visible area of an element. Everything outside the area will be hidden.
```

ddbeck (approved):
There's a typo to fix, otherwise this is ready to go. Thank you!

PR #1650, opened 2024-08-23
Publishing docs: apply miscellaneous edits

ddbeck:
I just did a web-features release and made a few improvements to the publishing docs as I went. Hopefully this stuff will help me complete it a little faster next time:

- Explain where we're checking for unescaped HTML
- Provide Markdown literals for pasting into the release description
- Add follow up step for "version required" labels
- Add a regex hint for creating releases
- Use full stops consistently
- Number the steps

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -7,7 +7,7 @@
 
 These are the steps to publish a regular release on NPM and as a GitHub release:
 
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks Daniel! Maybe someone other than you should at release the package to see if the docs are useful enough. This reads good to me, though!

ddbeck:
Yes, I agree that would be good to do—though I have some more changes to follow up to this. Luckily, Philip wrote the original text and I learned to release from it, so the foundations are solid.

PR #1651, opened 2024-08-23
Add forced-colors

jamesnw:
Related to #1624


features/forced-colors.yml
```diff
@@ -0,0 +1,14 @@
+name: Forced colors
+description: The `forced-colors` CSS @media rule and `forced-color-adjust` CSS property control styles when a user has chosen a limited palette of colors in forced colors mode.
```
captainbrosset, line 2:
```suggestion
description: The `forced-colors` CSS @media rule detects when a user has chosen to use a forced colors mode, also known as high-contrast mode, and the `forced-color-adjust` CSS property sets whether force colors apply to an element.
```

Not sure if you like this better or not, but I feel like at least mentioning the term "high contrast" is necessary here. And splitting the 2 entry points into 2 sentences allows to give more details.

features/forced-colors.yml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  - https://drafts.csswg.org/mediaqueries-5/#forced-colors
+  - https://drafts.csswg.org/css-color-adjust-1/#forced-color-adjust-prop
+group: css
+status:
+  compute_from: css.at-rules.media.forced-colors
```
captainbrosset, line 8: Can you use the feature without `forced-color-adjust`? It seems that merely detecting forced colors mode without being able to change it wouldn't be useful.
I wonder if we should just remove the `compute_from` rule here.

captainbrosset (changes requested):

features/forced-colors.yml
jamesnw: You can use it without `forced-color-adjust`. For example, you can add a border to a button if it otherwise gets styles from box-shadow, which is removed in forced colors mode.

captainbrosset (approved):

features/forced-colors.yml
```diff
@@ -0,0 +1,14 @@
+name: Forced colors
+description: The `forced-colors` CSS @media rule detects when a user has chosen to use a forced colors mode, also known as high-contrast mode, and the `forced-color-adjust` CSS property sets whether force colors apply to an element.
```
ddbeck, line 2: I'll apply this fix myself and merge.

```suggestion
description: The `forced-colors` CSS @media rule detects when a user has chosen to use a forced colors mode, also known as high-contrast mode, and the `forced-color-adjust` CSS property sets whether forced colors apply to an element.
```

ddbeck (approved):
Interesting. It might turn out that we split these features up later, but for now I'd like to see how it goes lumping them.

There's a typo to fix but I'll apply that and merge.

Thank you!

ddbeck:
Oh wait, I can't because push to the PR branch. Oops! — @jamesnw if you'd apply the fix, we can merge soon after. Thank you!

PR #1652, opened 2024-08-23
Add prefers- media queries

jamesnw:
- Adds a few keys to `prefers-color-scheme`
- Adds `prefers-contrast` (for #1624)
- Adds `prefers-reduced-data` (for #1624)
- Adds `prefers-reduced-motion` (for #1624)
- Adds `prefers-reduced-transparency`

I included the corresponding HTTP headers, although a case could be made for a ["User Preference Media Features Client Hints Headers"](https://wicg.github.io/user-preference-media-features-headers/#sec-ch-prefers-contrast) feature instead. 

BCD is missing the `http.headers.Sec-CH-Prefers-Contrast` that I would otherwise use on the `prefers-contrast` feature, as well as `http.headers.Sec-CH-Forced-Colors` that I would use on #1651.


jamesnw:
Combined with #1673

PR #1667, opened 2024-08-26
Contributing docs: explain `compat_features` versus tags

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/1633


docs/CONTRIBUTING.md
```diff
@@ -78,7 +78,7 @@ This table lists the fields that can be found in a feature file, and provides a
 | `spec` | One or more specification URLs for this feature. | String, or array of strings | Yes |
 | `group` | An optional group, or list of groups that this feature belongs to. See the definition of groups under [Create a new feature from scratch](#create-a-new-feature-from-scratch). | String, or array of strings | No |
 | `caniuse` | The feature's ID on the [Can I Use](https://caniuse.com/) website. | String | No |
-| `compat_features` | The list of browser-compat-data entries that make up this feature. In many cases, the corresponding browser-compat-data entries already map to the feature ID, and this is therefore not needed. | Array of strings | No |
+| `compat_features` | @mdn/browser-compat-data (BCD) entry keys (e.g., `css.properties.background-color`) that make up this feature. If there is no `compat_features` key and there exists a matching `web-features:<feature-id>` tag in BCD, then this key is automatically populated from the tag. | Array of strings | No |
```
jamesnw, line 81:
```suggestion
| `compat_features` | @mdn/browser-compat-data (BCD) entry keys (e.g., `css.properties.background-color`) that make up this feature. If `compat_features` is not set in `<feature-id>.yml`, the `dist` script will populate `compat_features` in `<feature-id>.yml.dist` with BCD entry keys tagged with `web-features:<feature-id>` in BCD, if any exist. | Array of strings | No |
```

I tried to clarify the distinction between the `yml` and `yml.dist` file here, but I'll defer to you on if this is a helpful distinction to make here.

jamesnw (approved):

docs/CONTRIBUTING.md
ddbeck: Good idea. I like making this more concrete. 👍

autonome:
This removes the explanation of why this is optional, which should stay in some form:

> In many cases, the corresponding browser-compat-data entries already map to the feature ID, and this is therefore not needed.

Also, this does not convey the intricacies of your answer in #1633:

> The gist is that the contents of the authored YAML file (i.e., <feature-id>.yml) overrules other possible sources of data. This goes for status as well. If you provide a compat_features value, then the BCD tag is completely ignored. So either provide a complete compat_features list or none at all.

But maybe that's a lot for the table entry? Perhaps there's a better home for it, since this bit helped me understand the flow of the system as a whole better.

ddbeck:
> But maybe that's a lot for the table entry?

Yeah, I think it's too much for the table. Really, we need a standalone fields reference. Maybe the `guidelines.md` file needs to become that.

That said, I tried to bring my comment into the section that follows, about dist files. This effect now gets at least three mentions:

- In the table ("If `compat_features` is not set in…")
- In the _Feature dist files_ section ("The contents of the authored YAML file overrules all other implicit sources of data.")
- In the _Associate BCD keys with your feature_ section ("In this case, you don't need to explicitly list the BCD keys that your feature depends on.")

What do y'all think?

jamesnw (approved):
LGTM!

Elchi3 (approved):

Alex666oguH (approved):

PR #1673, opened 2024-08-27
Add Media queries

jamesnw:
- Adds Media Query base feature for pre-2015 supported features
  - Includes `color-index`, which is old but never supported in Firefox. It doesn't warrant its own feature.
- Adds Media Query group
- Adds Prefer features 
  - Adds a few keys to `prefers-color-scheme` (I did not rename feature to match the other queries, as that would be a breaking change).
  - Adds `prefers-contrast` (for #1624)
  - Adds `prefers-reduced-data` (for #1624)
  - Adds `prefers-reduced-motion` (for #1624)
  - Adds `prefers-reduced-transparency`
  - Note: I included the corresponding HTTP headers, although a case could be made for a ["User Preference Media Features Client Hints Headers"](https://wicg.github.io/user-preference-media-features-headers/#sec-ch-prefers-contrast) feature instead. 
  - Note: BCD is missing the `http.headers.Sec-CH-Prefers-Contrast` that I would otherwise use on the `prefers-contrast` feature, as well as `http.headers.Sec-CH-Forced-Colors` that I would use on #1651.
- Add Interaction Media Queries
- Add Resolution media query
  - The data on this one is a mess. I'm not seeing a match with caniuse, and there are multiple BCD issues for Safari support- https://github.com/mdn/browser-compat-data/pull/18706, https://github.com/mdn/browser-compat-data/issues/20635, https://github.com/mdn/browser-compat-data/issues/24224
- Adds `color-gamut` media feature
- Adds `display-mode` media feature
- Adds `dynamic-range` media feature
- Adds `video-dynamic-range` media feature
- Adds `inverted-colors` media feature
- Adds Overflow media features
- Adds `scripting` media feature
- Adds `update` media feature


features/media-queries-update.yml
```diff
@@ -0,0 +1,6 @@
+name: Update frequency media feature
```
jamesnw, line 1: I was unsure about name here- the key is `update` but I think `Update frequency" is more useful as a description.

features/media-queries-interaction.yml
```diff
@@ -0,0 +1,11 @@
+name: Interaction media queries
+description: The interaction CSS media queries set styles based on a user's ability to hover using `hover` or `any-hover` or use a pointer using `pointer` and `any-pointer`. For example, a smartphone would match `@media (hover:none) and (pointer:coarse)`.
```
captainbrosset, line 2:
```suggestion
description: The `pointer`, `any-pointer`, `hover`, and `any-hover` CSS media features set styles based on the presence of pointing devices and their ability to hover over elements. For example, a smartphone device would match the `(hover: none) and (pointer: coarse)` media feature.
```

features/media-queries-inverted-colors.yml
```diff
@@ -0,0 +1,6 @@
+name: inverted-colors media feature
+description: The `inverted-colors` CSS media feature sets styles based on whether the user has inverted all colors.
```
captainbrosset, line 2: This description screams for a "Such as when ...", in order to make it easier to understand when/why/how a user might do that. Is this an OS setting similar to high contrast modes?

features/media-queries-overflow.yml
```diff
@@ -0,0 +1,9 @@
+name: Overflow media features
+description: The `overflow-block` and `overflow-inline` CSS media features set styles based on how a user's device will show content that is larger than the screen. Options are `none` (for instance, a billboard), `scroll` (for instance, a laptop), or `paged` (for instance, a printer).
```
captainbrosset, line 2:
```suggestion
description: The `overflow-block` and `overflow-inline` CSS media features set styles based on the way a device displays content that's larger than its container element. For example, a laptop lets users scroll to reveal content, while a printer displays the overflowing content on additional pages.
```

Since this can be used for printers, I don't think saying user's device helps here. Also removing "screen" for the same reason.

I'm not sure about the second sentence with the different values. The billboard example confuses me a bit. When would that happen? So I tried to rephrase above.

features/media-queries-prefers-contrast.yml
```diff
@@ -0,0 +1,6 @@
+name: prefers-contrast
+description: The `prefers-contrast` CSS media feature detects the requested contrast, more or less.
```
captainbrosset, line 2:
```suggestion
description: The `prefers-contrast` CSS media feature sets styles based on whether the user prefers more or less contrast.
```

I like that all previous features use the "sets styles based on" phrase. So I'm changing it here for consistency.

features/media-queries-prefers-reduced-data.yml
```diff
@@ -0,0 +1,6 @@
+name: prefers-reduced-data
+description: The `prefers-reduced-data` CSS media feature detects whether the user has a preference for using less internet traffic.
```
captainbrosset, line 2:
```suggestion
description: The `prefers-reduced-data` CSS media feature sets styles based on whether the user prefers to consume less internet traffic.
```

It would be good to  give a short example of when a web developer might want to use this. MDN has a nice example about using a system font rather than loading a custom web font. Maybe add "For example, you can use this media feature to avoid loading large font files and use a system font instead."

features/media-queries-prefers-reduced-motion.yml
```diff
@@ -0,0 +1,12 @@
+name: prefers-reduced-motion
+description: The `prefers-reduced-motion` CSS media feature detects whether the user requested less non-essential animation.
```
captainbrosset, line 2:
```suggestion
description: The `prefers-reduced-motion` CSS media feature sets styles based on whether the user prefers to minimize the amount of non-essential animations on the device.
```

features/media-queries-prefers-reduced-transparency.yml
```diff
@@ -0,0 +1,9 @@
+name: prefers-reduced-transparency
+description: The `prefers-reduced-transparency` CSS media feature detects whether the user requested less transparent effects, which can improve contrast and readability.
```
captainbrosset, line 2:
```suggestion
description: The `prefers-reduced-transparency` CSS media feature sets styles based on whether the user prefers to reduce the amount of transparent effects on their device, for example to improve contrast and readability.
```

features/media-queries-prefers-contrast.yml
```diff
@@ -0,0 +1,6 @@
+name: prefers-contrast
```
captainbrosset, line 1:
```suggestion
name: prefers-contrast media feature
```

For consistency here too.
Same comment for other files below.

features/media-queries-update.yml
```diff
@@ -0,0 +1,6 @@
...
+description: The `update` CSS media feature sets styles based on whether and how fast the user's device can modify display after it has been rendered. Values are `none`, `slow`, and `fast`.
```
captainbrosset, line 2:
```suggestion
description: The `update` CSS media feature sets styles based on whether and how fast the user's device can modify display after it has been rendered.
```

I feel like enumerating the values isn't needed here. The first sentence already encapsulates the information that this can be used to detect devices that don't refresh at all, and devices that refresh slow or fast.

features/media-queries.yml
```diff
@@ -0,0 +1,18 @@
...
+  compute_from: css.at-rules.media
+compat_features:
+  - css.at-rules.media
+  - css.at-rules.media.color
+  - css.at-rules.media.grid
+  - css.at-rules.media.height
+  - css.at-rules.media.width
+  - css.at-rules.media.monochrome
+  - css.at-rules.media.orientation
+  - css.at-rules.media.aspect-ratio
```
captainbrosset, line 16: Should we capture these 2 media features, and maybe others, as separate features?
I think keeping height and width here is fine, this is what the vast majority of media query usage happens. But the other features feel sufficiently specific to be in their own web-features files.

features/prefers-color-scheme.yml
```diff
@@ -1,3 +1,10 @@
 name: prefers-color-scheme
 description: The `prefers-color-scheme` CSS media query detects the requested color scheme, light or dark.
```
captainbrosset, line 2:
```suggestion
description: The `prefers-color-scheme` CSS media feature sets styles based on the requested color scheme, light or dark.
```

captainbrosset (changes requested):

features/media-queries.yml
jamesnw: I opted not to create separate features for pre-2015 media queries, thinking it would be easy to split if users requested it, but I could go either way.

jamesnw (commented):
Thanks for the review! I have a few comments/questions, and resolved all suggestions I accepted.

features/media-queries-color-gamut.yml
```diff
@@ -0,0 +1,6 @@
+name: color-gamut media feature
```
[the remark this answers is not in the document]
mirisuzanne: I think that's a good clarification to the `media-queries` feature, and still I would consider naming the individual features e.g. `color-gamut media query` as a more common way for developers to talk about these features. 

I don't have a strong sense on IDs, but looking at what's there, I think we're past the point of naming things in a way that is nicely sorted. Folders by group would be more helpful on that front.

features/media-queries-color-gamut.yml
```diff
@@ -0,0 +1,6 @@
...
+description: The `color-gamut` media feature sets styles based on the colors a device can display. Values are `srgb` if the device can display at least sRGB, `p3` if the device can display at least Display P3, and `rec2020` if the device can display at least REC. 2020.
```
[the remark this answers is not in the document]
mirisuzanne: Since the color() function now supports a number of additional spaces with distinct gamuts, I do wonder if there will be a push for consistency?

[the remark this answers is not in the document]
jamesnw: Good point, I'll remove the values.

features/interaction.yml
```diff
@@ -0,0 +1,12 @@
+name: Interaction media queries
+description: "The `pointer`, `any-pointer`, `hover`, and `any-hover` CSS media queries set styles based on the presence of pointing devices and their ability to hover over elements. For example, a smartphone device would match the `(hover: none) and (pointer: coarse)` media query."
```
ddbeck, line 2: I think "would" makes this more abstract that it needs to be. What do you think of this?

```suggestion
description: "The `pointer`, `any-pointer`, `hover`, and `any-hover` CSS media queries set styles based on the presence of pointing devices and their ability to hover over elements. For example, most smartphones match the `(hover: none) and (pointer: coarse)` media query."
```

features/prefers-contrast.yml
```diff
@@ -0,0 +1,6 @@
+name: prefers-contrast media query
+description: The `prefers-contrast` CSS media query sets styles based on whether the user prefers more or less contrast.
```
ddbeck, line 2: A little circular. The minimalist rewrite:

```suggestion
description: The `prefers-contrast` CSS media query sets styles based on whether the user prefers more or less color contrast.
```

And the maximalist one:

```suggestion
description: The `prefers-contrast` CSS media query sets styles based on whether the user prefers more or less contrast, the difference between foreground and background colors.
```

I'm not sure which I like better, so you can decide. 😄

features/prefers-reduced-data.yml
```diff
@@ -0,0 +1,6 @@
+name: prefers-reduced-data media query
+description: The `prefers-reduced-data` CSS media query detects whether the user has a preference for using less internet traffic. For example, you can use this media query to avoid loading large font files and use a system font instead.
```
ddbeck, line 2: me dodging having to make a decision about Internet or internet:

```suggestion
description: The `prefers-reduced-data` CSS media query detects whether the user has a preference for using less network traffic. For example, you can use this media query to avoid loading large font files and use a system font instead.
```

features/prefers-reduced-motion.yml
```diff
@@ -0,0 +1,12 @@
+name: prefers-reduced-motion media query
+description: The `prefers-reduced-motion` CSS media query sets styles based on whether the user prefers to minimize the amount of non-essential animations on the device.
```
ddbeck, line 2:
```suggestion
description: The `prefers-reduced-motion` CSS media query sets styles based on whether the user prefers to minimize the amount of non-essential animations on the device, such as scrolling, panning, zooming, and strobing.
```

features/prefers-reduced-motion.yml
```diff
@@ -0,0 +1,12 @@
...
+spec:
+  - https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion
+  - https://wicg.github.io/user-preference-media-features-headers/#sec-ch-prefers-reduced-motion
```
ddbeck, line 5: I don't think the header spec really sheds any new light on this feature. Let's omit it.

```suggestion
spec: https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion
```

features/prefers-reduced-transparency.yml
```diff
@@ -0,0 +1,9 @@
+name: prefers-reduced-transparency media query
+description: The `prefers-reduced-transparency` CSS media query sets styles based on whether the user prefers to reduce the amount of transparent effects on their device, for example to improve contrast and readability.
```
ddbeck, line 2: nit, needless pedantry:

```suggestion
description: The `prefers-reduced-transparency` CSS media query sets styles based on whether the user prefers to reduce the amount of transparent effects on their device, for example to improve contrast and legibility.
```

features/prefers-reduced-transparency.yml
```diff
@@ -0,0 +1,9 @@
...
+spec:
+  - https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-transparency
+  - https://wicg.github.io/user-preference-media-features-headers/#sec-ch-prefers-reduced-transparency
```
ddbeck, line 5:
```suggestion
spec: https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-transparency
```

features/resolution.yml.dist
```diff
@@ -0,0 +1,53 @@
...
+  # baseline_low_date: 2022-09-12
+  # support:
+  #   chrome: "29"
+  #   chrome_android: "29"
+  #   edge: "12"
+  #   firefox: "8"
+  #   firefox_android: "8"
+  #   safari: "16"
+  #   safari_ios: "16"
+  - css.at-rules.media.resolution
```
ddbeck, line 23: I'm very suspicious of this data. MDN says a unit is required. I think this whole feature is actually busted in Safari, or at very least it hasn't been tested recently. See https://bugs.webkit.org/show_bug.cgi?id=78087

I suspect we'll want to land the rest of this, but split this feature into its own PR while we fix BCD.

features/scripting.yml
```diff
@@ -0,0 +1,5 @@
+name: scripting media query
+description: The `scripting` CSS media query sets styles based on whether scripting such as JavaScript is available. Values are `enabled` if scripting is available, `initial-only` if scripting is only available on page load (for example, printed or server-rendered content), or `none`.
```
ddbeck, line 2: The server-rendered bit has me scratching my head a little. Is this a real thing that people do?

features/scripting.yml
```diff
@@ -0,0 +1,5 @@
...
+spec: https://drafts.csswg.org/mediaqueries-5/#scripting
```
ddbeck, line 3:
```suggestion
spec: https://drafts.csswg.org/mediaqueries-5/#scripting
group: media-queries
```

features/update.yml
```diff
@@ -0,0 +1,6 @@
+name: Update frequency media query
+description: The `update` CSS media query sets styles based on whether and how fast the user's device can modify display after it has been rendered.
```
ddbeck, line 2: Can we say something here about animations? I feel like that's the main thing this is going to be used for: avoiding animations on devices that can't/won't show them smoothly.

features/scripting.yml
jamesnw: This is a use case from the spec, and an example would be Opera Mini, which prerenders on a server, and then has limited JS functionality afterwards. But it does seem like a less practical example, so I removed it.

features/resolution.yml.dist
jamesnw: Moved to https://github.com/web-platform-dx/web-features/pull/1798

Issue #973, opened 2024-04-26
Fine-grained notifications support


foolip:
https://github.com/web-platform-dx/web-features/pull/960 revealed that support for notifications is complicated:

https://github.com/web-platform-dx/web-features/blob/313249804f39d05a8beea254ab1f0b29a9e52984/feature-group-definitions/notifications.yml#L5-L29

Let's sort that out. Perhaps we should define persistent notifications and non-persistent notifications as separate features, but that seems unlikely to be enough.

autonome:
Some thoughts on feature composition for this:

- MDN has Chrome Android as "partial" and Safari iOS as "full", with no baseline banner
- CIU has the reverse, but gives baseline low
- Neither treat SW as separate from the core Notifications feature

We could break out SW into separate feature as suggested above. But that makes Notifications baseline false, then in conflict w/ CIU and seems like that's against the spirit of Baseline's theory of availability.

Additionally, the bookmark-to-homescreen constraint on Safari iOS seems far more restrictive to availability than Chrome Android's constraint of ServiceWorker.

Given a goal of communicating general availability to developers, I recommend:

- Keep this as one feature
- Compute baseline from `api.ServiceWorkerRegistration.showNotification` to get a value of `low`

This results in an outcome that is consistent with both MDN and CIU, and aligns with the goal of communicating availability to developers in a reasonable way: The feature is available but may have platform-specific constraints.


Platform variation/constraints likely need a more consistent + programmatically-detectable way of being communicated, perhaps as change in BCD.

ddbeck:
This one is really gnarly—it took me a while to get my head around it.

## Summary 

The way this feature is reported on caniuse—using two very different entry points to qualify support for the Notifications API—it's very, very difficult to compute a status that aligns with caniuse, even if we do some reasonable changes in BCD.

I think option 2 (or 1+2), described below, is the way to go—drop `caniuse: notifications`, define more distinct notifications features, then ask Alexis to consider aligning to one of our features—though I'd feel better about if we talked to Alexis first.

Detailed discussion follows.

## The status quo

The status quo is this:

- `notifications` has no compat keys. We set a status manually, to match caniuse.
- The BCD entries conflict with caniuse in ways that make reconciling caniuse and a status computed from compat data impossible:
[...]

PR #1678, opened 2024-08-27
Add Notifications API

autonome:
- [x] reviewed prior work, or confirmed not already in progress
- [x] moved files
- [x] updated filename if needed (eg remove "TC39-" or s/webdriver2/webdriver/)
- [x] removed draft date
- [x] determined name
- [x] added entry point
- [x] verified baselines
- [x] checked api list against mdn list
- [x] added to a group or not
- [ ] what else?

## See also

#960 and #973


autonome:
Wrote up the feature composition proposal here: https://github.com/web-platform-dx/web-features/issues/973#issuecomment-2360576590

We can consider a group once we get to push notifications.

ddbeck:
I had a look at this and ended up poring over everything to do with notifications. It's very messy—primarily because caniuse describes this feature in a peculiar way—so we need to tread carefully as long as we're setting the `caniuse` key on this feature. See https://github.com/web-platform-dx/web-features/issues/973#issuecomment-2364007085 for a much, much more detailed take.

features/notifications.yml
```diff
@@ -1,38 +1,86 @@
...
 #
-# Another complication is the two entry points to this API:
-# - new Notification() for non-persistent notifications.
-# - showNotification() in service workers for persistent notifications.
+# BCD notes:
 #
-# Finally, BCD and caniuse.com disagree about whether it was Safari 6 or 7 that
-# first supported notifications, and BCD originally agreed with caniuse.com:
-# https://github.com/mdn/browser-compat-data/pull/7949
+# https://github.com/mdn/browser-compat-data/blob/0398a9fa847b3a616fa7968f79aadfc6cba4718f/api/Notification.json#L18
```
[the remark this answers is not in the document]
captainbrosset: Just tested and, wow, it really does seem like you can't create notifications on Chrome for Android from a "normal" webpage.

From a developer's stand point, using a service worker vs. not using one, to create notifications, seems a little more than an implementation detail to me.
I'm almost tempted to create 2 features: "Service worker notifications" and "Notifications". But, curious to hear your thoughts on this.

features/notifications.yml.dist
```diff
@@ -1,4 +1,223 @@
...
+  # baseline: false
+  # support:
+  #   chrome: "33"
+  #   edge: "14"
+  #   firefox: "26"
+  #   firefox_android: "26"
+  #   safari: "11"
+  #   safari_ios: "16.4"
+  - api.Notification.body
+  - api.Notification.icon
```
captainbrosset, line 94: Headsup that https://github.com/mdn/browser-compat-data/pull/24537 is going to make `icon` (and `tag`) be `false` on Safari.

features/notifications.yml.dist
```diff
@@ -1,4 +1,223 @@
...
+  - api.Notification.show_event
+
+  # baseline: false
+  # support:
+  #   chrome: "20"
+  #   edge: "14"
+  #   firefox: "26"
+  #   firefox_android: "26"
+  #   safari: "7"
+  - api.Notification.tag
```
captainbrosset, line 147: See https://github.com/mdn/browser-compat-data/pull/24537

features/notifications.yml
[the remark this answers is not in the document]
autonome: Crazier than Apple requiring that the page first be bookmarked to homescreen? 🤪

I agree that these are both *major* differences in how the feature is accessed at a platform level.

However, I think separating into multiple features muddies the waters for developers. The spec doesn't do it. Caniuse doesn't do it. MDN doesn't do it. And developers still seem to consider this feature as one thing - notifications.

My recommendation is to communicate the feature as a single thing.

Caniuse is a model of how the *consumers* of Web Features can communicate what constraints come along with specific platforms.

features/notifications.yml
```diff
@@ -1,38 +1,52 @@
...
-# features could be added to represent specific things that work.
-#
-# https://caniuse.com/notifications marks iOS support as partial with this note:
-# "Requires website to first be added to the Home Screen."
+# This feature is the general Notifications API, and does not anchor to the
+# Caniuse feature.
 #
-# BCD doesn't use partial_implementation for safari_ios, but has this note:
-# "Notifications are supported in web apps saved to the home screen."
+# See `features/notifications-pwa.yml` for the Notifications feature for mobile
```
captainbrosset, line 5:
```suggestion
# See `features/notifications-apps.yml` for the Notifications feature for mobile
```

features/notifications.yml
```diff
@@ -1,38 +1,52 @@
...
+# browsers that have platform-specific requirements and aligns with Caniuse.
+description: The Notifications API displays messages to the user outside of the browser, typically using the notification system of the host operating system. This feature is solely client-side, and does not integrate with server-originating notifications.
```
captainbrosset, line 7:
```suggestion
description: The Notifications API displays messages to the user, typically using the notification system of the host operating system.
```

Trying to make this shorter, in order to reuse it almost as is in the other feature. My reasoning is that it may be hard for users to understand why there are 2 related features. If we make the descriptions the same, but just add a bit about service worker/install in the other one, then that might help.

features/notifications-apps.yml
```diff
@@ -0,0 +1,19 @@
+name: Notifications from service workers and installed apps
+# This feature is for the platform-specific nature of notifications support
+# in Chrome Android and Safari iOS, and anchors to the Caniuse feature.
+#
+# See `features/notifications.yml` for the full Notifications feature, which
+# does not align with Caniuse.
+description: "The Notifications API displays messages to the user, with platform-specific requirements in some cases. Chrome Android supports notifications from installed service workers, and Safari iOS supports notifications from websites that have been bookmarked on the homescreen."
```
captainbrosset, line 7:
```suggestion
description: The Notifications API, when used from a service worker or home screen app, displays messages to the user, typically using the notification system of the host operating system. Chrome Android supports notifications from installed service workers, and Safari iOS supports notifications from websites that have been added to the home screen."
```

See my other comment. I used the exact same description that was in the general feature, but added the "when used from ..." caveat. And I kept the follow-up about the chrome and safari specifics.

Minor: I changed "bookmarked" with "added" for Safari, because Safari iOS uses the verb "add" in its UI: "Add to Home Screen".

features/notifications.yml
```diff
@@ -1,38 +1,52 @@
...
 #
-# Another complication is the two entry points to this API:
-# - new Notification() for non-persistent notifications.
-# - showNotification() in service workers for persistent notifications.
+# BCD notes:
 #
-# Finally, BCD and caniuse.com disagree about whether it was Safari 6 or 7 that
-# first supported notifications, and BCD originally agreed with caniuse.com:
-# https://github.com/mdn/browser-compat-data/pull/7949
+# https://github.com/mdn/browser-compat-data/blob/0398a9fa847b3a616fa7968f79aadfc6cba4718f/api/Notification.json#L18
```
captainbrosset, line 17: These comments don't make sense anymore in this file.

features/notifications.yml
```diff
@@ -1,38 +1,52 @@
...
-status:
-  baseline: false
-  support:
-    chrome: "20"
-    chrome_android: "42"
-    edge: "14"
-    firefox: "22"
-    firefox_android: "22"
-    safari: "6"
+compat_features:
```
captainbrosset, line 19: One problem we now have with this is that there isn't a BCD key that we can use to compute the browser support. There isn't a BCD key that says "supported in the main thread of regular websites". This means that the status now depends on all of the keys below, which have widely different browser support, because they represent later additions to the spec.
Should we add a new BCD key to simplify things?

captainbrosset (changes requested):
Thank you for splitting. Seeing it like this makes sense to me. I do think there's a bit more work needed for the more general feature though.

features/notifications-apps.yml
autonome: Hm, need to shave 38 chars, i'll remix.

autonome: Oh, or not... passed tests locally.

autonome: Nope, I just hadn't pulled. Pushed a shorter version. Not great, but has the same bits there.

features/notifications.yml
autonome: I'm not familiar with how BCD keys are birthed. Is that a typical approach to this kind of conundrum?

autonome:
Re-requested review from @captainbrosset for the description shortening in `notifications-apps.yml`.

features/notifications.yml
captainbrosset: BCD keys usually represent properties, methods, events, interfaces, etc. So, things that you can actually use as a developer.
But, there's also a small number of BCD keys that represent "special cases". Like when a method can be used with a new type of parameter which couldn't be used before.
When a feature evolves in a substantial new way, even if this evolution doesn't come with a new method or property, we tend to create a new BCD key to track this evolution across browsers.

In the present case, I would suggest creating a new BCD key in [Notification.json](https://github.com/mdn/browser-compat-data/blob/main/api/Notification.json) like `api.Notification.website_main_thread_support`.

For Chrome Android, this key would have the following support data:

```json
"chrome_android": {
  "version_added": false,
  "notes": "On Chrome for Android, notifications are only available through service workers."
}
```

And for Safari iOS:

```json
"safari_ios": {
  "version_added": false,
  "notes": "On Safari for iOS, notifications are only supported in web apps saved to the home screen."
}
```

I know that BCD already has notes for these caveats, but there's no BCD key that you can look at to know exactly which browsers support sending notifications from a webpage's main thread.

I'll create a PR in BCD in a minute that get that discussion going.

captainbrosset: See https://github.com/mdn/browser-compat-data/pull/24705#issuecomment-2425873063.

autonome:
This now has the BCD changes from https://github.com/mdn/browser-compat-data/pull/24705 and implements the approach recommended in #973.

features/notifications-apps.yml
```diff
@@ -0,0 +1,19 @@
+name: Notifications from service workers and installed apps
+# This feature is for the platform-specific nature of notifications support
+# in Chrome Android and Safari iOS, and anchors to the Caniuse feature.
+#
+# See `features/notifications.yml` for the full Notifications feature, which
+# does not align with Caniuse.
+description: The Notifications API, when used from a service worker for Chrome Android or home screen app for Safari iOS, displays messages to the user, typically using the notification system of the host operating system.
```
captainbrosset, line 7:
```suggestion
description: The Notifications API, when used from a service worker (for example on Chrome for Android) or a home screen installed webapp (for example on Safari for iOS), displays messages to the user, typically using the notification system of the host operating system.
```

Sorry for suggesting yet another description change here, and for making this one pretty long, but the current phrasing makes it sound like the feature is **only** supported on Chrome Android and Safari iOS. I added "for example" to make it clearer that the API is actually supported everywhere.

captainbrosset (approved):

features/notifications-apps.yml
```diff
@@ -0,0 +1,19 @@
...
+spec: https://notifications.spec.whatwg.org/
+caniuse: notifications
+status:
+  baseline: low
```
captainbrosset, line 11: Oh, I'm just realizing that we need a `baseline_low_date` field here.
I think it needs to be 2023-03-27, which is Safari iOS 16.4's date.

features/notifications-apps.yml
ddbeck: Maybe something like this, to show the entry point?

```suggestion
description: Notifications via service worker registration's `showNotification()` method, sent from installed web applications (for example, saved to the Home Screen on Safari for iOS), show a message to the user, typically using the notification system of the host operating system.
```

features/notifications.yml
```diff
@@ -1,38 +1,42 @@
...
-  support:
-    chrome: "20"
-    chrome_android: "42"
-    edge: "14"
-    firefox: "22"
-    firefox_android: "22"
-    safari: "6"
+# See `features/notifications-apps.yml` for the Notifications feature for mobile
+# browsers that have platform-specific requirements and aligns with Caniuse.
+description: The Notifications API displays messages to the user, typically using the notification system of the host operating system.
```
ddbeck, line 7:
```suggestion
description: The `Notification()` constructor shows a message to the user, typically using the notification system of the host operating system.
```

features/notifications.yml
```diff
@@ -1,38 +1,42 @@
...
+spec: https://notifications.spec.whatwg.org/
```
ddbeck, line 8:
```suggestion
spec: https://notifications.spec.whatwg.org/
status:
  compute_from: api.Notification.Notification
```

features/notifications-apps.yml
autonome: Took @ddbeck's change, as it incorporate's @captainbrosset's callout but w/ fewer parens.

features/notifications-apps.yml
ddbeck: (Creating an issue backlink here, no action)

Some day we should handle this from version numbers, see https://github.com/web-platform-dx/web-features/issues/2438

ddbeck (approved):
Let's go! Thank you @autonome and @captainbrosset for making this happen—a ton of work, but a huge improvement to the BCD and really exercising what we can do feature construction. 👏

PR #1683, opened 2024-08-27
Add alt text in content

jamesnw:
New baseline feature, as of 2024-07-09. I was on the fence about whether this warranted a feature, but found https://www.stefanjudis.com/today-i-learned/css-content-accepts-alternative-text/, with a widget using BCD data.


features/alt-text-in-content.yml
```diff
@@ -0,0 +1,6 @@
+name: Alt text in content
```
ddbeck, line 1: I misunderstood this name at first. I parsed it as "alt text in [arbitrary] content" as if there was some new global `alt` attribute. Instead, what we have is alt text for the `content` property. I'm thinking there are two routes on the name here (since we don't have a lot of developer-originating language to go on):

- Take a nod from the spec

   ```suggestion
   name: Alt text for generated content
   ```

- Make the content property a disambiguation:

     ```suggestion
   name: Alt text (content property)
   ```

features/alt-text-in-content.yml
```diff
@@ -0,0 +1,6 @@
...
+description: "Add alternate text in the CSS `content` property after a `/`. For example, `content: url('upvote.svg') / 'Upvote'` will have the accesible text 'Upvote'."
```
ddbeck, line 2: Trying to work this into our usual "X does Y" construction, rather than an imperative and future tense.

Also some tweaks to make more explicit what's going on with the example.

```suggestion
description: "The `/` notation in `content` CSS property values adds alternative text to generated content. For example, the `content: url('upvote-icon.svg') / 'Upvote'` declaration adds accessible "Upvote" text to the SVG icon. Also known as alternate text."
```

ddbeck (changes requested):

features/alt-text-in-content.yml
jamesnw: Good point- I went with "generated content" based on wording in a recent Mastodon thread on this.

features/alt-text-in-content.yml
```diff
@@ -0,0 +1,6 @@
+name: Alt text for generated content
```
ddbeck, line 1: Sorry, I forgot to also ask to match the ID to the new name, whatever you picked.

So this file probably needs to become something like `alt-text-generated-content.yml`

ddbeck (changes requested):

features/alt-text-in-content.yml
jamesnw: Done!

ddbeck (approved):
Thank you!

PR #1684, opened 2024-08-27
Add calc.color_component to relative-color

jamesnw:
Currently in [BCD](https://github.com/mdn/browser-compat-data/blob/6d239656b244f6ccb6a2886bfacaedbdf49cdc71/css/types/calc.json#L81-L85), `css.types.calc.color_component `is tagged as `web-features:calc`. It should be included in `relative-color` instead.


features/relative-color.yml
```diff
@@ -20,3 +20,4 @@ compat_features:
   # Seemingly incorrect notes, see
   # https://github.com/mdn/browser-compat-data/pull/22960
   # - css.types.color.rgb.relative_syntax
+  - css.types.calc.color_component
```
ddbeck, line 23: If you look above this, there are a bunch of comments and links to BCD issues. This feature also predates `compute_from`. It's a mess.

Maybe we should add `color_component` as a comment, then in a follow up PR figure out the status of each of the linked issues, an uncomment/update/`compute_from` as appropriate. What do you think?

jamesnw: Sounds good- I added a comment, commented this out, and will tackle this next!

ddbeck (approved):

PR #1687, opened 2024-08-28
Invert .prettierignore (use opt-out, not opt-in)

queengooborg:
This PR inverts the direction of the `.prettierignore` file, so that it lists the files and folders to ignore (as the file name implies), rather than all the files and folders to _only_ format.

## Motivation

In another PR (my first), the [first review comment](https://github.com/web-platform-dx/web-features/pull/1622#discussion_r1723110227) had requested that the new script was added to the `.prettierignore` file as an opt-in.  However, it is not standard to use a `.ignore` file for opting into formatting, so this was very confusing to me.  Additionally, since I'm assuming the plan is to eventually format all files using Prettier, having the `.prettierignore` file use an opt-out (or using proper terminology, _ignore_) structure, it serves as a to-do list for what still needs to be formatted.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,24 +1,19 @@
-# Exclude all files (but not directories)
-/**/*.*
+# Exclude any generated files/folders
+/**/*.dist
+/**/dist
 
-# Opt-ins
-!GOVERNANCE.md
-!eslint.config.js
-
+# Files/folders that have not yet been formatted
+# XXX Format all these files
+README.md
+index.ts
+types.ts
+2022-backgrounder.md
+towards-features.md
+/.github/**
+/docs
 !/docs/publishing.md
-
-!/features/*.yml
-/features/*.yml.dist
-!/groups/*.yml
-!/snapshots/*.yml
-
-!/packages/compute-baseline/**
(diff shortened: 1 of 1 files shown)


.prettierignore
```diff
@@ -1,24 +1,19 @@
...
-
-# Opt-ins
-!GOVERNANCE.md
-!eslint.config.js
+# Exclude any generated files/folders
+/**/*.dist
+/**/dist
 
+# Files/folders that have not yet been formatted
+# XXX Format all these files
```
ddbeck, line 6: We use `TODO:` everywhere else, so to make sure this doesn't become unfindable:

```suggestion
# TODO: Format all these files
```

ddbeck (approved):
Thank you, @queengooborg. This is a good fix. When we originally started using prettier it was _only_ for the YAML files, so `.prettierignore` existed only to prevent editors from automatically formatting everything else. It makes sense now to flip it around.

One nit in a line comment, but otherwise ready to go.

queengooborg:
Formatting of all the remaining files will soon follow as well!

ddbeck (approved):

PR #1693, opened 2024-08-28
Add Writing mode, Vertical form controls

jamesnw:


autonome (commented):
Looks good to me. Interesting situation as there's no actual property changes, just implementer behavior change (IIUC). The baseline differences do seem to justify a separate feature broken out for just form controls in this case, based on how we've been doing this so far.

However, it does poke at the feature composition question again, since in {n}yrs we may end up with a seemingly-bizarre set of features, which might only make sense in ancient historical context.

features/writing-mode.yml.dist
```diff
@@ -0,0 +1,52 @@
...
+  #   chrome: "48"
+  #   chrome_android: "48"
+  #   edge: "79"
+  #   firefox: "43"
+  #   firefox_android: "43"
+  #   safari: "10.1"
+  #   safari_ios: "10.3"
+  - css.properties.writing-mode.horizontal-tb
+  - css.properties.writing-mode.vertical-lr
+  - css.properties.writing-mode.vertical-rl
```
ddbeck, line 45: Not a blocker but I've sent https://github.com/mdn/browser-compat-data/pull/24517 to fix this.

features/vertical-form-controls.yml
```diff
@@ -0,0 +1,6 @@
+name: Vertical form controls
+description: Vertical form controls display elements in a form, such as radios, progress bars, or selects, vertically when the writing mode is `vertical-lr` or `vertical-rl`
```
ddbeck, line 2: In the original phrasing, the cause-and-effect was unclear to me. I think this makes it more concrete. Towards Dietrich's commentary, I think it will also help us understand what this is part of, when it's time to merge it into  `writing-mode`.

```suggestion
description: The `writing-mode` property orients form elements (such as radio buttons, progress bars, or select menus) vertically when the writing mode is `vertical-lr` or `vertical-rl`.
```

ddbeck (approved):
One optional suggestion on a description, but this is pretty good looking already.

ddbeck (approved):

PR #1695, opened 2024-08-28
Add transform-box, add missing transform properties

jamesnw:


features/individual-transforms.yml
```diff
@@ -2,3 +2,10 @@ name: Individual transform properties
 description: Transform elements with separate `translate`, `rotate`, and `scale` CSS properties.
```
autonome, line 2: non-blocking suggestion: The description could be more explicit about how these allow individual transformation vs the grouped property, giving context to why the feature is named as such. Eg, I like the "simple transforms independently" in the spec language.

features/transform-box.yml
```diff
@@ -0,0 +1,10 @@
+name: transform-box
+description: The `transform-box` CSS property sets the reference box for transformations.
```
autonome, line 2: Ditto, I like the spec language, which gives just a bit more colour: "relative to the position and dimensions of the reference box". (The MDN description I found oddly phrased.)

autonome (commented):
Pre-review, a couple of description suggestions.

features/transform-box.yml
jamesnw: "Oddly phrased" seems to be the standard when looking at descriptions of this property. I updated it, and hope this is one standard I didn't match, but the phrasing is tricky.

features/transform-box.yml
```diff
@@ -0,0 +1,10 @@
+name: transform-box
+description: The `transform-box` CSS property sets the position and dimensions, or reference box, relative to which transformations are calculated.
```
captainbrosset, line 2:
```suggestion
description: The `transform-box` CSS property sets the position and dimensions of the reference box relative to which an element's individual transformations are calculated.
```

I was having trouble parsing the initial sentence. Maybe I misunderstood what you meant.
I added the word "individual" in there in an attempt to make it clear that this only applies for individual transform properties (if, indeed, this is how this works).

captainbrosset (approved):
Approving with a comment on the description. But it feels good enough for me not to have to block/review again.

features/transform-box.yml
jamesnw: I liked the phrasing, but removed the word `individual`- it applies to the individual transform properties (scale, translate, rotate) as well as `transform`.

ddbeck (approved):
Happy enough to merge. Thank you!

Issue #1631, opened 2024-08-20
Clarifying the `caniuse` key


autonome:
The contribution guide says the `caniuse` is optional, but nothing else really.

We should explain:

- why it's there
- what it is (or could be) used for
- reasons to add it or not

ddbeck:
I need to turn this into a PR but…

- The `caniuse` key says that this feature is (approximately) the same as the corresponding caniuse feature.
- Use the `caniuse` key when:
  - You want the headline status for the feature to appear on caniuse.
  - The high-level feature is roughly equivalent to the caniuse feature(s) or a strict superset of the caniuse feature(s), subject to the caveats below.
- Don't use the `caniuse` key when:
  - The caniuse key is merely related (e.g., in `grid.yml`, don't set `caniuse: css-subgrid`) or not at all related.
  - The headline status (not the overall, not per-key status) would not be accurate with respect to the table on caniuse (e.g., caniuse says a feature is unsupported in a core browser set browser, but the web-features status is Baseline).
  - The status's first-supported version number differs from caniuse's by more than one year before 2020 or at all from 2020. This means there's a major disagreement—and a likely error—in BCD or caniuse's data.
- Additional notes:
  - Use `compute_from` to improve the correspondence of a feature's headline status with respect to the caniuse feature.
  
    But don't forget to use your judgement! Caniuse isn't perfect. Don't use `compute_from` in a way that would not make sense if the corresponding caniuse key didn't exist (e.g., pins support to _before_ support for some essential key in the feature). In such situations, it's better to comment out the caniuse and make a TODO comment or open an issue about why you did it.
  - If you see a <1 year pre-2020 discrepancy between caniuse and a computed status, make a note of it in https://github.com/web-platform-dx/web-features/issues/1499.

PR #1697, opened 2024-08-28
fill in more of AbortSignal

autonome:


jamesnw:
Is it intentional that you're removing the AbortController items? The `api.AbortSignal.abort_event` and `api.AbortSignal.aborted` items also were removed.

If you define `compat_features` in `AbortSignal.yml`, it will not combine it with the existing tagged BCD keys, so you need to add those as well.

autonome:
> Is it intentional that you're removing the AbortController items? The `api.AbortSignal.abort_event` and `api.AbortSignal.aborted` items also were removed.
> 
> If you define `compat_features` in `AbortSignal.yml`, it will not combine it with the existing tagged BCD keys, so you need to add those as well.

Meant to mark as draft, was just pushing as a placeholder. Definitely not planning to wipe out the existing keys!

features/aborting.yml
```diff
@@ -2,3 +2,12 @@ name: AbortController and AbortSignal
...
 caniuse: abortcontroller
+compat_features:
+  - api.AbortSignal
+  - api.AbortSignal.abort_event
+  - api.AbortSignal.abort_static
+  - api.AbortSignal.abort_static.reason_parameter
+  - api.AbortSignal.aborted
+  - api.AbortSignal.reason
+  - api.AbortSignal.throwIfAborted
+  - api.AbortSignal.timeout_static
```
ddbeck, line 13: This is (still?) wiping out a bunch of keys from BCD and explicitly mentioned in the description. What's the story here?

autonome: Ugh, I thought I'd added the difference back in but missed the AbortController section. Added back.

features/aborting.yml
```diff
@@ -2,3 +2,16 @@ name: AbortController and AbortSignal
 description: The `AbortController` and `AbortSignal` APIs allow you to cancel an ongoing operation, such as a `fetch()` request.
 spec: https://dom.spec.whatwg.org/#aborting-ongoing-activities
 caniuse: abortcontroller
+compat_features:
```
ddbeck, line 5: I looked at the additions to this and they make sense to me as later additions. However, the computed status now departs from Can I Use significantly (eventually we'll have docs for this, but see https://github.com/web-platform-dx/web-features/issues/1631#issuecomment-2310669265 for now). This needs a `compute_from` and a dist refresh:

```suggestion
status:
  compute_from: api.AbortController
compat_features:
```

ddbeck (changes requested):

features/aborting.yml
autonome: hmmm yeah i need to add that to my checklist

ddbeck (approved):
Thank you!

PR #1707, opened 2024-08-29
Add ImageBitmap and related APIs

autonome:
First take at splitting features out of html, lmk if there's any group we want for that high level.


features/imagebitmap.yml
```diff
@@ -0,0 +1,23 @@
+name: ImageBitmap
+description: The `ImageBitmap` interface represents a bitmap image which can be created from a variety of source objects. It can be drawn to a <canvas> without undue latency, and provides an asynchronous and resource efficient pathway to prepare textures for rendering in WebGL.
+spec: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap
+status:
+  compute_from: api.ImageBitmap
+compat_features:
+  - api.createImageBitmap
+  #  - api.createImageBitmap.options_parameter
```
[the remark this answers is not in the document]
autonome: hm, is only three of them actually:

#- api.createImageBitmap.options_parameter
#- api.createImageBitmap.resizeWidth_resizeHeight_resizeQuality
#- api.createImageBitmap.worker_support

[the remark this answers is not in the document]
Elchi3: So, I'm not sure where these keys come (maybe from an older BCD version?). Currently BCD offers these:

- api.createImageBitmap
- api.createImageBitmap.options_colorSpaceConversion_parameter 
- api.createImageBitmap.options_imageOrientation_parameter 
- api.createImageBitmap.options_premultiplyAlpha_parameter 
- api.createImageBitmap.options_resizeHeight_parameter 
- api.createImageBitmap.options_resizeQuality_parameter 
- api.createImageBitmap.options_resizeWidth_parameter
- api.createImageBitmap.svgimageelement_as_source_image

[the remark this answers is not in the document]
autonome: thanks, removed the erroring keys and added the svg one in. i searched one of the erroring ones and yeah i think i was looking at an old BCD release.

features/imagebitmap.yml
```diff
@@ -0,0 +1,21 @@
+name: ImageBitmap
+description: The `ImageBitmap` interface represents a bitmap image which can be created from a variety of source objects. It can be drawn to a <canvas> without undue latency, and provides an asynchronous and resource efficient pathway to prepare textures for rendering in WebGL.
+spec: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap
+status:
+  compute_from: api.ImageBitmap
```
autonome, line 5: While `createImageBitmap` is not baseline, `ImageBitmap` however does appear like it should be.

But neither Caniuse or MDN show it as baseline.

@ddbeck wdyt the right move is here?

autonome:
@captainbrosset thanks for the feedback. The high level developer need which the these APIs meet seems mostly around performant/ergonomic transformation/transfer of content.

Image processing, images as sprites, performant duplication/patterns, texture generation, and capturing/moving/modifying frames for anything video/animation.

Example: Say I want to grab some frames from a currently playing video and manipulate them, and/or stash them, then later push to a canvas collage to use as a preview image or whatever.

There's loads of examples of how much fun... er, useful, this is.

The later addition of a global function to create ImageBitmap maybe seems to speak to this being something to file under "Image Processing" or something like that?

From a granularity perspective though, Image Processing seems more like a group than a feature, where ImageBitmap and friends seem more like a feature.

WDYT?

autonome:
@captainbrosset prompted a conversation on feature composition at the biweekly meeting, which was great! There was no conclusion, but general agreement that the boundaries of a feature are not clear.

Until we get there, I'll instead provide explicit rationale, so that the thinking behind the proposal of a given feature is clear. Perhaps out of this practice, useful criteria may emerge. Maybe we need a concept of a "feature thesis"? 🤔

So in the spirit of EIBTI, writing out the thought process I went through when proposing this as a standalone feature:

- Developer usage: I searched Codepen and Github and a general web search, and it doesn't seem to get lumped into other features. It gets called out by name in a few standalone demos dedicated to it. Code references are often in heavy graphics toolkits (games, 3d, xr) and working with canvas (images in/out).
- No obvious other home: It's an interstitial feature, often used *between* distinct features such as media elements, canvas. It is most related to Image, but not strictly a subset or different version/flavor of it.
- Specs: The spec authors decided not to hang these APIs off Image, and instead made separate globals.
- Feature prior art: The size and uniqueness of the functionality map reasonably to something groups well together - few methods/properties that stand out as belonging to something else. There are definitely features with far fewer components, and which are related to bigger features.
- Group guidance?: There's no `image` or related groups, and `media-elements` isn't right, based on the name and comments there. Other pre-existing groups are a stetch, or not relevant.

Very much open to better ideas, this just seemed sufficient in my first look.

I'm sure we'll come across more "glue" features when breaking up the mega-specs, so more conversation is better since it'll inform those future decisions.

autonome:
Another option could be to do something around `transferables`, `Image Transfers`, `rasterization`? Something in that direction is a bigger tent to pull the various svg/canvas/media element APIs into, and a name oriented more around the language of the use-case.

autonome:
We discussed in person last week and there were two conclusion-ish things that the assembled group agreed on:

- If an API was created solely as an interchange between other APIs, it does not necessarily mean the API should be a web feature
- If an API has a global constructor, it likely should be a web feature

On the former point, we are still not left with a clear path. Eg API X exists *only* as way for Feature Y and Z to interact bidirectionally. Does it belong to Feature Y or to Z? 🤷🏽

The latter point is helpful in this particular case, and in our conversation we concluded `ImageBitmap` should be a feature.

@ddbeck feel free to provide more colour or corrections as needed!

ddbeck:
OK, forgive me because I don't think I made my point last week very clearly. On the matter of constructors:

The existence of a global constructor strongly implies that there exists a feature, though the interface itself might not be a feature. For example, the `MouseEvent()` constructor is a clue that there is an event or events feature lurking nearby, but the ability to construct an `InputEvent` object is probably not, in an of itself, feature-ish. And in such a case, the constructor lives with feature representing the events that can fire.

So my "theory" about interstitial APIs is that they probably came into existence _for_ some concrete feature, then other features co-opted or glommed onto that interface. I have yet to see a counter example yet, though I would be interested to be proven wrong.

So, my chief question is: what was the first API to use an `ImageBitmap` object? Or did this really come into existence alongside changes to multiple other APIs?

autonome:
> So, my chief question is: what was the first API to use an `ImageBitmap` object? Or did this really come into existence alongside changes to multiple other APIs?

Probably `canvas`, as it's the target declared in this language in the spec (even though other targets are valid use-cases also):

> An [ImageBitmap](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap) object represents a bitmap image that can be painted to a canvas without undue latency. 

Happy to put this in `canvas` somewhere if that's what you and @captainbrosset think is best, or elsewhere if you have a suggestion.

features/imagebitmap.yml
```diff
@@ -0,0 +1,21 @@
+name: ImageBitmap
+description: The `ImageBitmap` interface represents a bitmap image which can be created from a variety of source objects. It can be drawn to a <canvas> without undue latency, and provides an asynchronous and resource efficient pathway to prepare textures for rendering in WebGL.
```
ddbeck, line 2: OK, making a big rewrite suggestion here, aiming for a few things:

- `createImageBitmap()` is the entry point we ought to name first
- "variety of source objects" — let's use some examples instead
- "undue latency" — this sounds kinda lawyerly and I think we can do better by comparing it to higher latency things (I know this point and the one above is from the docs/specs, not assigning blame here)

That said, I wasn't able to preserve the stuff about WebGL. 

```suggestion
description: The `createImageBitmap()` global method creates an `ImageBitmap` object from a source such as an image, SVG, blob, or canvas. An `ImageBitmap` object represents pixel data that can be drawn to a canvas with lower latency than other types, such as `ImageData`.
```

features/imagebitmap.yml
ddbeck: OK, I misunderstood this question originally.

**The only change we need to make here is to drop the `compute_from` entirely.**

Here's the background:
 
The `createImageBitmap()` method itself _is_ well-supported, but there are some details that that make it look rougher than it is.

On caniuse, Safari is shown as unsupported due to a bug that was fixed over a year ago, relating to the `premultiply` option. Meanwhile, BCD doesn't reflect this bug or the fix. (This needs PRs/issues filed in both places.)

Meanwhile, caniuse says that this depends on the `resizeQuality` option, which Firefox doesn't support. BCD is consistent on this, though I think it's a little bit of a bummer to keep this from being Baseline over it. Oh well.

To recap, there's a number fixes needed for this, but the only one that happens _here_ is that we need to not use `compute_from` here.

features/imagebitmap.yml.dist
```diff
@@ -0,0 +1,144 @@
...
+  # baseline_high_date: 2022-07-15
+  # support:
+  #   chrome: "66"
+  #   chrome_android: "66"
+  #   edge: "79"
+  #   firefox: "46"
+  #   firefox_android: "46"
+  #   safari: "11.1"
+  #   safari_ios: "11.3"
+  - api.ImageBitmapRenderingContext
```
ddbeck, line 28: I'm skeptical this shipped in Safari 11, before `ImageBitmap` itself in Safari 15. This is probably not correct. Noted in https://github.com/web-platform-dx/web-features/issues/1499

features/imagebitmap.yml.dist
```diff
@@ -0,0 +1,144 @@
...
+  - api.createImageBitmap.svgimageelement_as_source_image
+
+  # baseline: false
+  # support:
+  #   chrome: "54"
+  #   chrome_android: "54"
+  #   edge: "79"
+  #   safari: "15"
+  #   safari_ios: "15"
+  - api.createImageBitmap.options_resizeQuality_parameter
```
ddbeck, line 144: The rest of this aligns with caniuse.

ddbeck (changes requested):
OK, I dug into this better today. I read the documentation and some specs more closely and trolled through some spec history.

I got my wires crossed and thought this was some sort API interchange type with spotty support, but clearly it's not—it's providing a specific thing back onto canvas, so it doesn't make sense as some kind of later addition. It's its own thing.

I will now be going back on my previous review. 😅

features/imagebitmap.yml
```diff
@@ -0,0 +1,21 @@
...
+spec: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap
```
ddbeck, line 3:
```suggestion
caniuse: createimagebitmap
spec: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap
```

features/imagebitmap.yml
ddbeck: Opened https://github.com/mdn/browser-compat-data/pull/24496 and https://github.com/Fyrd/caniuse/pull/7160 to capture the premultiply stuff.

autonome: ack, dropped the `compute_from`, thanks for filing the related issues!

features/imagebitmap.yml
ddbeck, line 1: One last thing: let's rename the file to `createimagebitmap`.

(I hate that I keep forgetting this. I'm thinking that we'll want some tooling for setting new identifiers. I'll file an issue about this later.)

ddbeck (approved):
Approved, pending a file rename.

PR #1713, opened 2024-08-30
File system access API

captainbrosset:
Adding a new entry for the File System Access API.

Spec: https://wicg.github.io/file-system-access/
Blog: https://developer.chrome.com/docs/capabilities/web-apis/file-system-access
MDN: https://developer.mozilla.org/en-US/docs/Web/API/File_System_API

Taking this opportunity to create a group to bundle this with OPFS.

Pinging @tomayac to take a look at the description I wrote and at the browser compat data, to double check that they seem correct.


features/file-system-access.yml
```diff
@@ -0,0 +1,12 @@
+name: File system access
+description: The `showOpenFilePicker()`, `showDirectoryPicker()`, and `showSaveFilePicker()` functions, and the `FileSystemHandle` interface access files and directories on the user's device and allow you to read and write to them.
```
tomayac, line 2:
```suggestion
description: The `showOpenFilePicker()`, `showDirectoryPicker()`, and `showSaveFilePicker()` functions, and the `FileSystemHandle` interface allow for getting access to files and directories on the user's device and read and (if the user allows) write to them.
```

Not sure if it's worth mentioning that `FileSystemHandle` and its children are defined in the actual WHATWG FS standard, whereas the picker methods are still only in WICG. They share the `FileSystemHandle`.

tomayac (commented):
Thanks for working on this! Left a comment with a suggestion.

features/file-system-access.yml
ddbeck: I agree that we don't need to mention the handle interface, especially since it doesn't look as if developers will frequently call that API by name. It looks likely they'll use the objects from the `show` methods instead.

For brevity then, suggesting:

```suggestion
description: The `showOpenFilePicker()`, `showDirectoryPicker()`, and `showSaveFilePicker()` methods request access to files and directories on the user's device and returns a handle for reading and writing to them.
```

features/file-system-access.yml
```diff
@@ -0,0 +1,12 @@
...
+spec: https://wicg.github.io/file-system-access/
+caniuse: native-filesystem-api
+group: file-system
+compat_features:
+  - api.DataTransferItem.getAsFileSystemHandle
```
ddbeck, line 7: I think this might be part of a distinct thing, like https://caniuse.com/dragndrop. What do we lose by leaving it out of this feature?

tomayac: It's part of file system access, albeit of course it's also part of drag and drop. FWIW, we document it in the context of file system access at Google:
- https://developer.chrome.com/docs/capabilities/web-apis/file-system-access#drag_and_drop_integration
- https://web.dev/patterns/files/drag-and-drop-files?hl=en

ddbeck: OK @captainbrosset can you add a comment to this file saying something like the following?

api.DataTransferItem.getAsFileSystemHandle is likely to be part of a drag and drop feature, but it can't be in both features at once. See https://github.com/web-platform-dx/web-features/issues/1173 for details

captainbrosset: Leaving it out doesn't change the overall feature support data.

It's worth thinking about why it's here though. This method is an extension to the dragndrop feature which allows you to turn a dropped file/directory into a `FileSystemHandle`, which then allows you to write to it, just like if you had opened it using `showOpenFilePicker` instead.

Overall, I can think of 4 different ways to get access to a `FileSystemHandle`, and using one of the show picker methods is just one of them:

- Using `showOpenFilePicker` or `showDirectoryPicker`.
- Dropping the file on the browser window, and using `getAsFileSystemHandle`.
- Registering the app as a file handler, via the web app manifest, and using `window.launchQueue.setConsumer()` to retrieve the file when the installed PWA is launched after double-clicking an associated file.
- Retrieving the handle from IndexedDB storage.

I think this feature should not only cover the show methods, but also contain the various methods/interfaces that let you play with FileSystemHandle objects in code.

tomayac: IDB requires having gone through either D&D or the picker methods. File handling could be an independent feature (OS integration), but I can also agree to it being part of overall file system.

ddbeck: I'm not objecting to its inclusion here. I just really want the `https://github.com/web-platform-dx/web-features/issues/1173` text in the YAML file somewhere, because I think these file APIs are going to be rather central to figuring out BCD keys shared across multiple features. When it comes time to tackle that problem, I want breadcrumbs to the places where that's significant.

ddbeck (changes requested):

features/file-system-access.yml
captainbrosset: Sounds good. I'll add a comment about this in here.
@tomayac I agree that File Handling should be its own separate feature.

ddbeck (approved):

PR #1714, opened 2024-08-30
Add dictionary support for hyphens

jamesnw:
Adds language support for hyphens. Pins support on the `auto` value. For more discussion, see https://github.com/web-platform-dx/web-features/pull/1714#discussion_r1740502046.

<details><summary>Initial description</summary>This naming and split seems odd, but the support story here is that `hyphens` support for English was added initially, and now support for other languages is quite spotty. Only 13 of the 60 non-English languages are baseline low. 

Also adds `auto` and `language_english` to `hyphens`, which also better brings it inline with Caniuse.

We should ensure that this does not communicate that English is the default language, but rather that English was the only supported language initially.</description>


features/hyphens.yml
```diff
@@ -2,3 +2,8 @@ name: Hyphenation
 description: The `hyphens` CSS property controls when long words are broken by line wrapping. Although called `hyphens`, the property applies to word-splitting behavior across languages, such as customary spelling changes or the use of other characters to mark an intraword line break.
```
captainbrosset, line 2: Can we mention in here somewhere that the rules used for breaking up words only work for English? This way, it makes more sense why there's also a non-english hyphens feature.

features/hyphens.yml.dist
```diff
@@ -5,12 +5,36 @@ status:
   baseline: low
   baseline_low_date: 2023-09-18
   support:
-    chrome: "55"
+    chrome: "88"
     chrome_android: "55"
-    edge: "79"
+    edge: "88"
```
captainbrosset, line 10: The only weird thing here is that caniuse shows Edge as fully supported only since version 105. It says this about Edge 88 to 104: "Only supported on Android & Mac platforms". I think that might be a bug on caniuse. What do you think?

features/non-english-hyphens.yml
```diff
@@ -0,0 +1,67 @@
+name: Non-English hyphens
+description: The `hyphens` CSS property uses language-specific rules to determine when long words are broken by line wrapping. The content language is set with the HTML `lang` attribute.
```
captainbrosset, line 2: I feel like this description should make it clearer that this is for non-English and that readers are supposed to have read the description for `hyphens` first. 
I know it's in the name, but I feel like there needs to be more of a connection between the two features somehow.

Having a `see_also` field would be useful for this kind of feature. Or, perhaps, we can put them together in a new group, maybe called `typography` (which would be under the CSS group).

As for the description, how about something like this: "When used with a non-English HTML `lang` attribute value, the `hyphens` CSS property uses language-specific rules to determine when long words are broken by line wrapping."

captainbrosset (dismissed):

features/non-english-hyphens.yml
jamesnw: > Having a `see_also` field would be useful for this kind of feature.

I agree, that would be useful. 

For the description, I want to stay away from the idea that English is the default- per the spec, you should need to also specify your language as English in order to get English hyphens.  As I think through this, the reason I split them apart was to tell a fuller support story, but unless we split each language into a feature, it won't be possible for someone to differentiate between French support (baseline low) and Esperanto support (none). 

Because of this, I'm proposing that I combine the two into a single `hyphens` feature, and adding a note that "support for non-English languages varies". While that violates the "don't mention support in the description" rule from the style guide, it does seem important to let people know. Thoughts?

features/hyphens.yml.dist
jamesnw: Hmm, it looks like Caniuse is correct - https://github.com/Fyrd/caniuse/pull/5793, but I can't find any mention of the issue in BCD. I think this would be a good candidate for further investigation in https://github.com/web-platform-dx/web-features/issues/1499.

features/non-english-hyphens.yml
jamesnw: I've joined them in https://github.com/web-platform-dx/web-features/pull/1714/commits/e730703740535f1f76e51077955eb7e4f308b5d6

ddbeck (approved):

PR #1726, opened 2024-09-03
Add Capture stream

autonome:
Closes #1456

@ddbeck (or anyone else), I basically fixed the TODOs in the descriptions and no other changes yet.

But I'm iffy on the name and grouping.

Seems like if we're popping these capture APIs into a discrete feature then maybe we'd group them together and give the feature a name like `Capture stream` or some approach like that, rather than split them by element type?


andrew:
Quick test comment, feel free to delete

features/capture-stream-audio-video.yml
```diff
@@ -0,0 +1,6 @@
+name: captureStream() for <audio> and <video>
+description: The `captureStream()` method for `<audio>` and `<video>` elements returns a MediaStream object which is streaming a real-time capture of the content being rendered in the media element.
```
ddbeck, line 2: I'm not really clear on what this does. Does this capture it? (pun intended)

```suggestion
description: The `captureStream()` method for `<audio>` and `<video>` elements returns a `MediaStream` for the media element's content. You can use this to record media or send it elsewhere, such as a canvas or WebRTC connection.
```

features/capture-stream-canvas.yml
```diff
@@ -0,0 +1,10 @@
+name: captureStream() for <canvas>
+description: The `captureStream()` method for `<canvas>` elements returns a `MediaStream` which includes a `CanvasCaptureMediaStreamTrack` containing a real-time video capture of the contents of the canvas. Firefox doesn't support `CanvasCaptureMediaStreamTrack`, but has `CanvasCaptureMediaStream` which is equivalent.
```
ddbeck, line 2: Extending my previous suggestion:

```suggestion
description: The `captureStream()` method for `<canvas>` elements returns a `MediaStream` which includes a `CanvasCaptureMediaStreamTrack` representing real-time video of the canvas image. You can use this to record the canvas, or send it elsewhere, such as another canvas or WebRTC connection.
```

One thing I want to note here is that we should endeavor to avoid putting support notes into the description—they should stand the test of time, as browsers resolve their differences. Instead, I think we should deal with this in some sort of notes/partial implementation embellishment. Maybe add a comment like this (wrapped appropriately)?

```
# TODO: Firefox implements `CanvasCaptureMediaStream` equivalent to `CanvasCaptureMediaStreamTrack`. Add a note when https://github.com/web-platform-dx/web-features/issues/915 is resolved. See also: https://bugzil.la/1231131
```

ddbeck (changes requested):
This looks plausible to me and I'd be happy to see it land, after fixing the canvas description.

The one thing that gives me pause is that caniuse uses a single feature to represent all of this: https://caniuse.com/mediacapture-fromelement. However, I don't know if that's the right move (or at least one we should take before dealing with https://github.com/web-platform-dx/web-features/issues/971)—there's a Baseline high canvas feature hidden alongside the media elements stuff.

I think it makes sense to make these two features, then do some kind of composition to align with caniuse later.

features/capture-stream-canvas.yml
autonome: sgtm, adding comment

autonome:
> I think it makes sense to make these two features, then do some kind of composition to align with caniuse later.

@ddbeck do you mean the current split? or splitting out the baseline high feature and joining the other parts?

I think latter approach makes sense from a "features help developers make compat decisions" POV.

features/capture-stream-canvas.yml
```diff
@@ -0,0 +1,14 @@
...
+# Add a note when https://github.com/web-platform-dx/web-features/issues/915 is resolved.
+#
+# See also: https://bugzil.la/1231131
+spec: https://w3c.github.io/mediacapture-fromelement/#html-canvas-element-media-capture-extensions
+group: canvas
+compat_features:
+  - api.CanvasCaptureMediaStreamTrack
+  - api.CanvasCaptureMediaStreamTrack.canvas
+  - api.CanvasCaptureMediaStreamTrack.requestFrame
+  - api.HTMLCanvasElement.captureStream
```
ddbeck, line 14: From our call today: let's make this key the `compute_from`—this is usable, it's just a bit of an oddity when you work with the resulting tracks, but that's probably more on the side of "supported with a caveat" not "unsupported with a circuitous workaround."

ddbeck:
> > I think it makes sense to make these two features, then do some kind of composition to align with caniuse later.
> 
> @ddbeck do you mean the current split? or splitting out the baseline high feature and joining the other parts?

Yes, what you've done here is good. We should proceed with it (provided you've made the change we discussed on a call today, in a comment above).

autonome:
> Yes, what you've done here is good. We should proceed with it (provided you've made the change we discussed on a call today, in a comment above).

I'm just pushing @foolip's fine work in #1456 to the finish line. Made the `compute_from` change, ready for review.

ddbeck (approved):

ddbeck:
Thank you!

PR #1727, opened 2024-09-03
Add a feature for SVG support in the Async Clipboard API

captainbrosset:
See https://blogs.windows.com/msedgedev/2024/07/11/seamless-svg-copy-paste-on-the-web/ for some background on this.


autonome:
We added this in #1703.

(I didn't realize how new this is, will fix a regular web papercut for me!)

ddbeck:
I'm going to close this as a duplicate of #1703. If you still think this should be a standalone feature, please merge main into your branch and reopen for discussion/review. Thank you!

captainbrosset:
I don't think web devs know that the clipboard API doesn't support all formats by default. I don't think many of them know that it only supports text, html, and png out of the box. And, on top of this, that browsers sanitize the html and png data to avoid security problems, so you don't even get to read the same HTML code that you wrote in the clipboard earlier.

Because of this fact, and because the addition of SVG, and web custom formats happened much later than the bulk of the clipboard API, I would still prefer to track these as separate features. I think splitting these gives a better chance for devs to understand the limits of the clipboard API, and the additional features they need to achieve their particular use case (for example, a Figma-like app would want to check the Baseline status for both a clipboard, and a clipboard-svg feature. Or, an excel-like app would want to check the status of a clipboard-web-custom-format feature, to know whether they can pass their custom spreadsheet+macro code, unsanitized, in the clipboard, between a native and a web app).

ddbeck:
OK, please reopen after you've merged in main and resolved the errors. The existing PR will fail on merge, due to duplicate keys.

features/clipboard-custom-format.yml
```diff
@@ -0,0 +1,6 @@
+name: Async clipboard support for custom formats
+description: Support for custom formats in the `navigator.clipboard` API allows you to read and write unsanitized custom data in the system clipboard. Custom formats have a mime-types that starts with the `web ` prefix.
```
autonome, line 2:
```suggestion
description: Support for custom formats in the `navigator.clipboard` API allows you to read and write unsanitized custom data in the system clipboard. Custom formats have mime-types that start with the `web ` prefix.
```

autonome: (or unpluralize it all)

features/clipboard-unsanitized-formats.yml
```diff
@@ -1,5 +1,5 @@
-name: Read unsanitized clipboard data
-description: The optional `formats.unsanitized` parameter of the `navigator.clipboard.read()` method reads unsanitized data from the system clipboard.
+name: Read unsanitized HTML clipboard data
```
autonome, line 1: Non-blocking comment: Seems like the spec goes out of its way to make this a list, even though the list only has HTML right now. Not making this specific to HTML future-proofs it to changes.

autonome (approved):

features/clipboard-unsanitized-formats.yml
```diff
@@ -1,5 +1,5 @@
...
+description: The optional `formats.unsanitized` parameter of the `navigator.clipboard.read()` method reads unsanitized text/html data from the system clipboard.
```
ddbeck, line 2:
```suggestion
description: The optional `formats.unsanitized` parameter of the `navigator.clipboard.read()` method reads unsanitized `text/html` data from the system clipboard.
```

features/clipboard-svg.yml
```diff
@@ -0,0 +1,6 @@
+name: Async clipboard support for SVG
+description: Support for the `image/svg+xml` mime-type in the `navigator.clipboard` API allows you to read and write SVG data in the system clipboard.
```
ddbeck, line 2: Avoiding "support for" because that's the kind of phrasing we could put in every single feature, if we let it become a habit. 😄 

```suggestion
name: SVG clipboard items
description: The `image/svg+xml` data type (also known as the MIME type) for `ClipboardItem` objects represent SVG data written to or read from the clipboard.
```

features/clipboard-custom-format.yml
```diff
@@ -0,0 +1,6 @@
+name: Async clipboard support for custom formats
+description: Support for custom formats in the `navigator.clipboard` API allows you to read and write unsanitized custom data in the system clipboard. Custom formats have mime-types that start with the `web ` prefix.
```
ddbeck, line 2: "Custom formats" and the `web ` prefix seem like the most important information here. I'm suggesting front-loading this information. I also think we can be more specific about which interface this actually applies to, which should help relate this information to documentation.

(Also I feel like MIME types are a bit old school but I could be convinced otherwise)

```suggestion
name: Custom formats for clipboard items
description: The `web ` prefix for `ClipboardItem` data types (also known as MIME types) allows reading and writing ununsanitized custom data from the device clipboard.
```

ddbeck (changes requested):
Some nitpicking on the names and descriptions, but this looks like a good forward. Thank you, Patrick!

features/clipboard-unsanitized-formats.yml
captainbrosset:
```suggestion
name: Read unsanitized clipboard data
```

features/clipboard-svg.yml
captainbrosset:
```suggestion
name: SVG clipboard items
description: The `image/svg+xml` data type (also known as the MIME type) for `ClipboardItem` objects represents SVG data written to or read from the clipboard.
```

ddbeck (approved):

PR #1732, opened 2024-09-04
Remove long-animation-frames.yaml

jamesnw:
I came across this file and believe it is there by accident. Renaming it to `.yml` and running dist does not add any keys, so the tag doesn't appear to exist in BCD. Because it is a .yaml file, it is not included in data.json.

It was originally added in https://github.com/web-platform-dx/web-features/pull/645. @Elchi3 is there a reason this file should be kept?

### DIFF
--- a/features/long-animation-frames.yaml
+++ b/features/long-animation-frames.yaml
@@ -1,2 +1,0 @@
-name: Long animation frame timing
-spec: https://w3c.github.io/long-animation-frames/


Elchi3 (approved):
uh oh, this is from long time ago and dist files weren't even a thing then. I think removing is fine, I'm happy to work on a proper feature for this one.

PR #1736, opened 2024-09-05
Logical properties

jamesnw:
An alternate feature set from this would be something like-

- Logical border (perhaps split into -color, -style, -width, etc)
- Logical clear
- Logical float
- Logical inset
- Logical min and max size
- Logical padding

And then have physical counterparts for each. 

In looking at how people are talking about this feature, it seemed that people were more interested in Logical properties all together, in knowing that they could use `-block-start` instead of `-top` (in LTR, TTB writing mode) across properties.


features/logical-properties.yml
```diff
@@ -0,0 +1,73 @@
+name: Logical properties
+description: The logical CSS properties control borders, size, margin, and padding with directions and dimensions relative to the writing mode. For example, in a left to right, top to bottom writing mode, `block-end` refers to the bottom. Also known as flow relative.
```
autonome, line 2:
```suggestion
description: CSS logical properties control borders, size, margin and padding with directions and dimensions relative to the writing mode. For example, in a left to right, top to bottom writing mode, `block-end` refers to the bottom. Also known as flow relative.
```

autonome: Minor grammatical disambiguation fix.

features/logical-properties.yml
```diff
@@ -0,0 +1,73 @@
...
+spec: https://drafts.csswg.org/css-logical-1/
+caniuse: css-logical-props
+group: css
+# Picked to match caniuse- clear and float came 2 years later, but don't warrant separate features.
```
autonome, line 6:
```suggestion
# Picked to match Caniuse. Clear and float came 2 years later, but don't warrant separate features.
```

autonome (approved):
+1, logical properties are documented as a single feature across many guides, blogs, etc.

features/logical-properties.yml
jamesnw: Thanks for the review!

ddbeck (approved):
This is great. Thank you!

Issue #1716, opened 2024-09-02
`update-drafts` shouldn't modify `draft_date` when there are no other changes


ddbeck:
Updating `draft_date` produces lots of noise in the diff on updates. See https://github.com/web-platform-dx/web-features/pull/1715 for an example. In that case, there are no real changes to the feature files, just updates to the dates.

Instead, the script ought to check if there are any differences _except_ the date before writing a file. If the only difference is the date, then the file ought to be skipped. This will also make it easier to find areas we haven't touched recently.

autonome:
This also fixes a dev workflow papercut when merging, b/c these always show as conflicts.

PR #1742, opened 2024-09-07
`update-drafts` script: don't make date-only modifications

ddbeck:
~~Blocked by https://github.com/web-platform-dx/web-features/pull/1679, then a rebase is needed.~~

This skips writing changes to draft features when the only change is the `draft_date` value. This prevents needlessly large diffs when updating these files and makes it possible to find out which areas we've touched least recently.
    
Fixes https://github.com/web-platform-dx/web-features/issues/1716.


scripts/update-drafts.ts
```diff
@@ -178,7 +196,43 @@ async function main() {
...
+
+    // If there's a file for this spec already, write updates only if something
+    // other than the `draft_date` changed. Because changes can be comments, we
+    // have to check a diff rather than parsing and comparing values.
+    const changes = diff.diffLines(originalFile, proposedFile);
+    const noChanges =
+      changes.length === 1 && !changes[0].added && !changes[0].removed;
+    const onlyDateChanges =
+      !noChanges &&
+      changes.length === 3 &&
```
autonome, line 224: document the magic 1 and 3 numbers above

autonome (approved):
looks good to me, just a note on some inline docs that'd be helpful to future readers

scripts/update-drafts.ts
ddbeck: Good idea! I won't even remember that in a week. 😅 Fixed with 13e94dbc.

Elchi3 (approved):
Works as expected for me 👍

PR #1746, opened 2024-09-09
Extend migrate-to-bcd CLI

ddbeck:
This adds environment variables for setting the path to BCD and a verbosity option.

Follow up from the discussion at https://github.com/web-platform-dx/web-features/pull/1622#discussion_r1724863422.

cc: @queengooborg Hopefully this accommodates everyone's config preferences (implicit, explicit, and environment variables).

### DIFF
--- a/scripts/migrate-to-bcd.ts
+++ b/scripts/migrate-to-bcd.ts
@@ -7,7 +7,42 @@ import yargs from "yargs";
 
 import { features } from "../index.js";
 
+const argv = yargs(process.argv.slice(2))
+  .scriptName("migrate-to-bcd")
+  .usage("$0 [bcd-path]", "Migrate compat_features keys to BCD tags")
+  .env("MIGRATE_TO_BCD")
+  .option("bcd-path", {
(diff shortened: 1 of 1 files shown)


scripts/migrate-to-bcd.ts
```diff
@@ -1,15 +1,48 @@
...
+    alias: "v",
+    describe: "Show more information",
+    type: "count",
+    default: 0,
+    defaultDescription: "warn",
+  })
+  .middleware((argv) => {
+    // Read `--bcd-path` option from BCD_PATH if set, for compatibility with BCD
+    // Collector's config style
+    if (process.env.BCD_PATH) {
```
queengooborg, line 29: I'm assuming this line would override the `bcd-path` argument, so we'd need the following, right?

```suggestion
    if (process.env.BCD_PATH && !argv.bcdPath) {
```

ddbeck: Yes, good catch!

queengooborg (approved):

PR #1750, opened 2024-09-09
Update `checkVisibility`

autonome:
- added some missing properties
- added compute_from so that baseline low versions are correct for the core feature


jamesnw (approved):
Pre-review- this change looks good to me. I confirmed it's not on caniuse, and that there aren't other related keys.

ddbeck (approved):

Issue #445, opened 2023-11-30
`motion-path`: create a group for later additions


ddbeck:
https://github.com/web-platform-dx/web-features/pull/333 covered only the features that aligned with the caniuse motion path feature. There are a number of later additions with a much more complicated compatibility story. We ought to group them and relate them to `motion-path` in a sensible way.

Affected features:

- css.properties.offset-path.basic-shape
- css.properties.offset-path.coord-box
- css.properties.offset-path.ray
- css.properties.offset-path.url
- css.properties.offset-position
- css.types.ray
- css.types.ray.position
- css.types.ray.size

PR #1754, opened 2024-09-09
Add keywords to motion-path

jamesnw:
`motion-path` feature was built with tags in BCD, and was missing many keywords.

I used `offset` for the status- previously it as `offset-path`. This pushes the date later, but it is more in sync with [Caniuse](https://caniuse.com/css-motion-paths). Everything is the same to Caniuse, except for Chrome, and I'm not able to explain the difference here.

Fixes https://github.com/web-platform-dx/web-features/issues/445


jamesnw:
@ddbeck I opened https://github.com/Fyrd/caniuse/issues/7154.

Issue #1388, opened 2024-07-12
Guideline for entrypoints that are too long to spell out


foolip:
In https://github.com/web-platform-dx/web-features/pull/1363#discussion_r1674276165 @captainbrosset wrote:

> We usually try to name the entry point of the feature in the description. In this case it would be ServiceWorkerRegistration.sync. Although it's a bit long, and devs don't actually write ServiceWorkerRegistration in their code, but instead get a registration object by doing `await navigator.serviceWorker.ready`.  So I'm not too sure what to do here.

I don't think we've dealt with this before, but it will happen quite a lot. This would be terrible:

> The `(await navigator.serviceWorker.ready).sync` API ...

Because it's extremely unlikely anyone would write the code like that.

This specific case is complicated by the fact that there are many ways to get a `ServiceWorkerRegistration` instance, see the backreferences from https://w3c.github.io/ServiceWorker/#get-the-service-worker-registration-object

Another case is properties on methods on elements. Some precedent from that:

> The `showPicker()` method for `<input>` elements shows the user interface for picking a value. For example, for `<input type="date">` it shows the interface for picking a date.

> The `preservesPitch` property for `<audio>` or `<video>` adjusts the pitch of audio to sound more natural when the playback rate is faster or slower than the default.

What should a guideline say?

ddbeck:
I think the principle is something like this:

To aid search, include literal text that a web developer would inevitably type if they were to invoke the feature. If there's no essential entry point to the feature, then include only concise snippets of essential literal code instead. Never use made-up or idiosyncratic example code.

- 👍 Recommended: `display: flex`, `fetch()`, etc.
- 👎 Not recommended: `exampleButton.addEventLIstener("newevent", /* … */)`

foolip:
@ddbeck that sounds good to me, and I don't think we have anything like `exampleButton.addEventListener("newevent", /* … */)` yet. We don't seem to even have cases like `video.someMethod()` where `video` stands in for a `<video>` element instance.

captainbrosset:
Your proposal sounds good to me. Feel free to open up a PR.

PR #1755, opened 2024-09-09
Add environment variables group

jamesnw:
There's a case to be made to merge `safe-area-inset` with `environment-variables`, but I think there's enough interest to have them separate.

I think the "CSS" in  "CSS environment variables" is appropriate here- even though other web tech doesn't have environment variables, it's more commonly used in build and backend tools, so I think differentiation is helpful.


features/environment-variables.yml
```diff
@@ -0,0 +1,7 @@
+name: CSS environment variables
```
captainbrosset, line 1: The concern I have with having this separate from the safe-area-inset feature is that you can't use `env()` on its own. It was introduced for safe-area-insets, and nothing else at the time.

features/titlebar-area.yml
```diff
@@ -0,0 +1,9 @@
+name: Titlebar area environment variables
+description: The `titlebar-area-*` CSS environment variables represent the header area of a Progressive Web App that is not hidden by the system's buttons. For example, ` height:\ env(titlebar-area-height)` sets the element's height to the height of the space reserved for the titlebar.
```
captainbrosset, line 2: I think we need to mention the phrase "Window Controls Overlay" somewhere in there, for searchability. This feature is only useful if the PWA is a desktop one, only if it is actually installed, only if the manifest defines the `window-controls-overlay` display-mode, and only if the user actually opted-in to that display mode after having installed the app.
Only then do the `titlebare-area-*` variables have non-0 values.

We clearly can't say all of this in there, how about this:

```suggestion
description: The `titlebar-area-*` CSS environment variables represent the area that the titlebar of a PWA that uses the `window-controls-overlay` display-mode normally occupies. For example, `height:\ env(titlebar-area-height)` sets the element's height to the height of the space normally reserved for the titlebar.
```

That said, how about this different direction:

- We keep `env()` and `safe-area-inset` as one feature.
- We create a second feature called `window-controls-overlay`, which includes the new `titlebar-area-*` variables + the other WCO-related bits (Navigator.windowControlsOverlay, WindowControlsOverlay, WindowControlsOverlayGeometryChangeEvent, ...)

features/safe-area-inset.yml
```diff
@@ -0,0 +1,8 @@
+name: Safe area inset environment variables
+description: The `safe-area-inset-*` CSS environment variables represent an area of a non-rectangular screen that will not be cut off. For example, `top:\ env(safe-area-inset-top);` positions the top of an element below the notch on an iPhone.
```
ddbeck, line 2: There's some discussion on how exactly to formulate it into a guideline in https://github.com/web-platform-dx/web-features/issues/1388, but in general, we want avoid putting things into code text that a developer wouldn't actually type, so I'm suggesting not using the wild card asterisk here and let the example do the work.

Also, reading the spec and seeing the mention of a rectangle within a non-rectangle really helped me understand what was going on here. What do you think of this amendment?

```suggestion
description: The `safe-area-inset-` CSS environment variables represent a rectangle that cannot cut off content within a non-rectangular screen. For example, `top:\ env(safe-area-inset-top);` positions the top of an element below the notch on an iPhone.
```

features/titlebar-area.yml
jamesnw: The separate WCO feature is in #1762

features/environment-variables.yml
jamesnw: In talking with @mirisuzanne, `env()` variables come up often in spec discussions, but seldom ends up in the final spec. I merged the `environment-variables` feature with `safe-area-insets`, and think the `environment-variables` group will be useful here. WCO is split into a separate PR.

mirisuzanne: Yeah, I tend to agree with @captainbrosset here that `env()` is only a syntax structure for delivering other features. It's like asking when browsers added support for 'css built-in functions' as a concept, instead of asking about support of each individual function. It makes some sense to teach broadly about how function syntax works, but you would almost never ask when that concept gained support, apart from specific features that use it.

captainbrosset (approved):

ddbeck (approved):

PR #1762, opened 2024-09-10
Add Window controls overlay

jamesnw:
Split off from #1755


captainbrosset (approved):

captainbrosset:
In case other reviewers want to learn more about WCO before reviewing:

- https://blogs.windows.com/msedgedev/2022/09/27/closing-pixel-gap-native-web-window-controls-overlay/
- https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API
- https://alistapart.com/article/breaking-out-of-the-box/

features/window-controls-overlay.yml
```diff
@@ -0,0 +1,18 @@
+name: Window Controls Overlay
```
ddbeck, line 1: I hate it when we have to capitalize features, but this one is so weirdly named I don't think we have a reasonable alternative.

features/window-controls-overlay.yml
```diff
@@ -0,0 +1,18 @@
...
+  - api.WindowControlsOverlay.getTitlebarAreaRect
+  - api.WindowControlsOverlay.visible
+  - api.WindowControlsOverlayGeometryChangeEvent
+  - api.WindowControlsOverlayGeometryChangeEvent.WindowControlsOverlayGeometryChangeEvent
+  - api.WindowControlsOverlayGeometryChangeEvent.titlebarAreaRect
+  - api.WindowControlsOverlayGeometryChangeEvent.visible
+  - css.properties.custom-property.env.titlebar-area-height
+  - css.properties.custom-property.env.titlebar-area-width
+  - css.properties.custom-property.env.titlebar-area-x
+  - css.properties.custom-property.env.titlebar-area-y
```
ddbeck, line 18:
```suggestion
  - css.properties.custom-property.env.titlebar-area-y
  - html.manifest.display_override.window-controls-overlay
```

features/window-controls-overlay.yml
```diff
@@ -0,0 +1,18 @@
+name: Window Controls Overlay
+description: The Window Controls Overlay API displays content in the title bar area of a progressive web app that is installed on desktop and has `display_override:\ window-controls-overlay` in the manifest.
```
ddbeck, line 2: This contains almost everything we need, but does require some clean ups:

- If we're mentioning the main entry point, we should lead with it like we do for HTML elements, CSS properties, and some APIs.
- Every other example of `display_override` I could find had an array of strings. I don't think you'd ever literally type `display_override: window-controls-overlay`, but I don't know this feature well and could be corrected.
- "on desktop" sounds clipped, so I went with "desktop device" ([see also](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#device))
- Due to tech writer esoterica, replacing avoiding "displays" as a verb.

```suggestion
description: The `display_override: ["window-controls-overlay"]` web application manifest member shows content in the title bar area of a progressive web app that is installed on a desktop device.
```

ddbeck (changes requested):

features/window-controls-overlay.yml
captainbrosset: `display_override: ["window-controls-overlay"]` is the correct syntax, sorry I missed it in my previous review. This manifest member provides a list of display-mode fallbacks. When using the `display-mode` member, you don't get to specify a sequence of fallbacks, only the mode you prefer, and you then let the browser choose how to fallback in case a mode isn't possible.

features/window-controls-overlay.yml
jamesnw: Good catch!

ddbeck (approved):

PR #1763, opened 2024-09-10
Add CredentialsManagement public key options and other missing bits

autonome:
Date- and spec-wise this seems to all jive w/ the original feature, no baseline changes.

The only weird bit is that Caniuse seems incorrect here, unless they're only counting "support" as also supporting some of the optional parameters (unlikely, and would be strange): https://caniuse.com/credential-management. But that doesn't change this PR, just a side note.


autonome:
In the spirit of "small patches for higher velocity", not going to blow this up into larger Credentials patch. Instead will do it in series of smaller chunks+refactors.

autonome:
Hrm, I did not search BCD keyspace first... yeah this set of features is sprinkled all over. Moved these to the right home at first glance, but need to look more, flipping back to draft.

autonome:
Ok, checked spec, and reviewed the various versions these additions shipped in - these do not justify a separate feature, and don't change baseline status of either webauthn or fedcm. Requesting review from @nsatragno as well, per the original feature addition.

features/fedcm.yml
```diff
@@ -4,5 +4,8 @@ spec: https://w3c-fedid.github.io/FedCM/
 group: credential-management
 compat_features:
   - api.CredentialsContainer.get.identity_option
+  - api.CredentialsContainer.get.identity_option.context_option
+  - api.CredentialsContainer.get.identity_option.error_api
```
ddbeck, line 8: Let's leave this key out (but maybe mention it in a comment above the `compat_features:` line). BCD marked as non-standard because it's not formally part of the spec yet (per https://github.com/w3c-fedid/FedCM/pull/498).

ddbeck (approved):
I have one minor suggestion. It'd be nice to get Nina's subject-matter expert review on it, but I have pretty high confidence in this one if she doesn't get to it by the end of the week.

PR #1764, opened 2024-09-10
Add counters and list styles

jamesnw:
`css.properties.list-style-type.asterisks` and `css.properties.list-style-type.footnotes` appear to not be part of spec, but are available in Firefox.


features/list-style.yml.dist
```diff
@@ -0,0 +1,306 @@
...
+
+  # baseline: false
+  # support:
+  #   chrome: "91"
+  #   chrome_android: "91"
+  #   edge: "91"
+  #   safari: "5.1"
+  #   safari_ios: "5"
+  - css.properties.list-style-type.asterisks
+  - css.properties.list-style-type.footnotes
```
jamesnw, line 299: `asterisks` and `footnotes` do not appear in the spec- should they be present here?

features/list-style.yml.dist
```diff
@@ -0,0 +1,306 @@
...
+
+  # baseline: false
+  # support:
+  #   firefox: "35"
+  #   firefox_android: "35"
+  - css.properties.list-style-type.symbols
+  - css.properties.list-style.symbols
```
jamesnw, line 306: These could be a separate feature, potentially, but the similar `css.at-rules.counter-style.symbols` was included with the `counter-style` feature.

features/list-style.yml
```diff
@@ -0,0 +1,116 @@
+name: List style
+description: The `list-style` CSS and the `list-style-image`, `list-style-position`, and `list-style-type` longhands style the marker for a list item which can be selected using the `::marker` pseudoelement.
```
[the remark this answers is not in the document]
captainbrosset: Thinking more about this, I remembered that you can use `list-style` outside of `::marker` (duh, of course, I've been doing this forever). With `::marker` being quite new (and not baseline), it feels odd to me to add it to this feature, also because it allows devs to do things that `list-style` doesn't.

My understanding is that there are 2 levels of marker customizations in CSS.
First level: `list-style`, which lets you change the type of marker, or even use an image.
Second level: if that's not enough, `::marker` lets you change the color or size of markers, independent of the list item content. And, if that's still not enough, it allows you to use your own text content as the marker.

Thoughts about splitting these features?

captainbrosset (changes requested):

features/counters.yml.dist
```diff
@@ -0,0 +1,96 @@
...
+  # support:
+  #   firefox: "82"
+  #   firefox_android: "82"
+  - css.properties.counter-reset.reset_does_not_affect_siblings
+
+  # baseline: false
+  # support:
+  #   firefox: "96"
+  #   firefox_android: "96"
+  - css.properties.counter-reset.reversed
```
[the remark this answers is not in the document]
jamesnw: `reset_does_not_affect_siblings` seems mostly like bringing the implementation into compliance with the spec in an edge case, and I don't think it deserves a feature.

`reversed` is an interesting thread to pull at. Thinking through what that feature would be, I think the feature would be something like "Reversed lists", and we would likely want to pull `html.elements.ol.reversed` out of the `list-items` feature, which has a different support story.

There are blogs and questions about how to reverse lists, but I'm not sure if it raises to the level of feature.

features/marker.yml
```diff
@@ -0,0 +1,8 @@
+name: ::marker
+description: The `::marker` CSS pseudoelement styles the font, text, and color of a list item marker.
```
captainbrosset, line 2:
```suggestion
description: The `::marker` CSS pseudo-element styles the font, text, and color of a list item marker.
```

features/counters.yml.dist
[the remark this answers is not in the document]
captainbrosset: I'd love to hear a third opinion on this. I could go both ways, but slight preference for splitting reversed lists into its own feature.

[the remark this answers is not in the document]
jamesnw: @ddbeck  or @mirisuzanne for a third opinion - do you think "Reversed lists" should be a separate feature? It would include `css.properties.counter-reset.reversed` (only supported in Firefox) and `html.elements.ol.reversed` (Baseline High), and probably be computed from `html.elements.ol.reversed`.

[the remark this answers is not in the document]
ddbeck: In an ideal world, we'd have a "reversed list" feature alongside the property and element features. We can't do that just yet (see https://github.com/web-platform-dx/web-features/issues/1173), so I'm suggesting the following:

- Split `counter-reset: reversed` CSS declaration into it's own feature (you can take inspiration from one the `contain-` features, for descriptions)
- Put a comment in that feature mentioning that, after https://github.com/web-platform-dx/web-features/issues/1173 is resolved, that a "reversed list" feature would be a good addition, crossing the CSS and HTML boundary)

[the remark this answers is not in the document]
mirisuzanne: I think splitting it out is also the best option for making the support data visible - since the support story for `counter-reset.reversed` doesn't match either of the other groupings. To me, at least, it seems like placing an unsupported key in with supported keys only acts to 'hide' the browser support that this project is meant to surface.

features/counter-reset-reversed.yml
```diff
@@ -0,0 +1,6 @@
+name: Reversed counter-reset
+description: "The `reversed()` CSS function is used with `counter-reset` to create a counter that counts down."
```
ddbeck, line 2: Avoiding "is used to" 

```suggestion
description: "The `reversed()` CSS function for `counter-reset` creates a counter that counts down."
```

features/counters.yml
```diff
@@ -0,0 +1,17 @@
+name: Counters (CSS)
+description: The `counter()` and `counters()` CSS functions, used with `counter-reset` and `counter-increment`, automatically number headings or ordered list items.
```
ddbeck, line 2: The "used with" phrasing here really confused me, as if it'd be `counter-reset: counter(…)`. I think listing things without more detail might be in our interest here:

```suggestion
description: The `counter-reset` and `counter-increment` CSS properties and the `counter()` and `counters()` functions automatically number headings or ordered list items.
```

features/counters.yml
```diff
@@ -0,0 +1,17 @@
...
+spec: https://drafts.csswg.org/css-lists-3/#auto-numbering
+caniuse: css-counters
+group: counters
+status:
+  compute_from: css.types.counter
```
ddbeck, line 7: We ought to avoid pinning the status to something that predates an explicit item in the description. Can you pin to the properties instead?

features/list-style.yml
```diff
@@ -0,0 +1,114 @@
+name: List style
+description: The `list-style` CSS and the `list-style-image`, `list-style-position`, and `list-style-type` longhands style the marker for a list item.
```
ddbeck, line 2:
```suggestion
description: The `list-style` shorthand CSS property and the `list-style-image`, `list-style-position`, and `list-style-type` longhand properties set the position and appearance of a list item's marker.
```

Oh also here's an interesting style question that I don't think we've talked about before "longhands." MDN uses "constituent properties" — I wonder if we ought to do the same? Curious to have your thoughts on it, though I don't think we should resolve the question for everything in this PR.

features/list-style.yml.dist
ddbeck: Both are marked as `standard_track: true` in BCD. If we drop them, then we ought to contribute an issue/PR upstream to fix that.

features/marker.yml
```diff
@@ -0,0 +1,8 @@
+name: ::marker
+description: The `::marker` CSS pseudo-element styles the font, text, and color of a list item marker.
```
ddbeck, line 2: For selectors and combinators, I think it's really nice to include "select" somewhere in the description text:

```suggestion
description: The `::marker` CSS pseudo-element selects the font, text, and color of a list item marker.
```

ddbeck (changes requested):
One issue with a `compute_from` to resolve here, plus some bonus fussing with descriptions while I'm here.

features/list-style.yml.dist
jamesnw: This was a can of worms. There are 43 items here not included in the spec, due to a quirk in the collector. I've opened https://github.com/mdn/browser-compat-data/issues/24545 to address that. In the meantime, I'm thinking I should remove all the keys here that are not in the spec, correct?

ddbeck: Yes, let's set aside the keys that are not in the spec and return to them later.

features/marker.yml
```diff
@@ -0,0 +1,8 @@
+name: ::marker
+description: The `::marker` CSS pseudo-element selects the font, text, and color of a list item marker.
```
captainbrosset, line 2: "selects" feels like the wrong verb here. CSS selectors select _elements_.
Maybe "styles".

captainbrosset (approved):

features/marker.yml
ddbeck: Sorry this is my fault, I had an incomplete thought on this in https://github.com/web-platform-dx/web-features/pull/1764#discussion_r1777341651. We probably want to say something closer to this:

```suggestion
description: The `::marker` CSS pseudo-element selects a list item marker for styling numbers or bullets.
```

jamesnw: Good timing! I just opened up a [PR](https://github.com/web-platform-dx/web-features/pull/1869) with more pseudo elements.

ddbeck (approved):
Great, let's merge this now. @captainbrosset if there's more work on the description, please @ me or open a PR—we'll straighten it out.

features/list-style.yml
jamesnw: @ddbeck  Wanted to circle back on this- in talking with Miriam and doing some research, I would like to stick with longhands and shorthands over constituent properties. They seem to be the more common phrase- searching for CSS constituent properties, for instance, doesn't return useful information, but CSS shorthands do. Should this go in a styleguide?

ddbeck: @jamesnw Yes, please. You can file an issue or open a PR to cover longhand/shorthand terminology. 👍

jamesnw: PR open in https://github.com/web-platform-dx/web-features/pull/1891

PR #1769, opened 2024-09-11
Increment minor version to v1.3.0

ddbeck:
JSON diff: https://gist.github.com/ddbeck/c7446b91e0b129b24c5dbea19bdbabfe


Elchi3 (approved):
"A minor version is required for releases that contain only additions, such as new feature or new properties on existing types."

That seems to be the case from looking at the diff, so LGTM.

PR #1770, opened 2024-09-11
Remove unused release script code

ddbeck:
Remove some unused code from a previous (and premature) attempt to automate the release process.

For https://github.com/web-platform-dx/web-features/pull/1769, I tried using `scripts/release.ts`, which was incomplete and predates our current release process, documented in [`publishing.md`](https://github.com/web-platform-dx/web-features/blob/1ef1f93dbf33d12f0e5a2f7fb095508675ad3ef5/docs/publishing.md#publishing).

It kinda works—I want to open a PR to fix what's broken—but there's also a lot of unused code that I don't foresee using, at least not in the same way. Some it might return in some form later, but I think deleting this unusued/untested procedure in favor of incrementally automating the documented process is a better path forward.

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -65,40 +65,6 @@ yargs(process.argv.slice(2))
         .demandOption("semverlevel", "You must provide a semver level");
     },
     handler: init,
-  })
-  .command({
-    command: "update <pr>",
-    describe: "Update an existing release pull request",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Agreed, let's only bring this back if we need to.

PR #1771, opened 2024-09-11
Add CSS and SVG animations, fill missing Web Animations

jamesnw:
This is mostly backfilling older functionality. It adds a new Animation Group as well.

Low/false baseline items did not seem particularly feature-worthy.


features/css-animations.yml
```diff
@@ -0,0 +1,51 @@
+name: CSS Animations
+description: The `animation` CSS property applies `@keyframes` animation sequences to transform elements.
```
captainbrosset, line 2:
```suggestion
description: The `animation` CSS property animates an element's style over time, using keyframes described in `@keyframes` rules.
```

Wanted to propose an alt description because I wasn't super keen on using the word "transform", since it could be interpreted as CSS transforms. And because I thought it was important to mention the time component of CSS animations, to separate this feature from the [scroll-driven animations feature](https://web-platform-dx.github.io/web-features-explorer/features/scroll-driven-animations/).

features/svg-animations.yml
```diff
@@ -0,0 +1,97 @@
+name: SVG animations
+description: The `<animate>`, `<animateMotion>`, and `<animateTransform>` SVG elements declaratively animate SVG elements. Also known as SMIL.
```
captainbrosset, line 2:
```suggestion
description: The `<animate>`, `<animateMotion>`, and `<animateTransform>` SVG elements animate an SVG element's attributes over time. Also known as SMIL.
```

Here I also think we should mention the over time concept, and maybe also mention that it's the attributes that get animated (if I'm not mistaken)

features/svg-animations.yml
```diff
@@ -0,0 +1,97 @@
...
+spec: https://svgwg.org/specs/animations/
+caniuse: svg-smil
+group:
+  - animation
+  - svg
+status:
+  compute_from: svg.global_attributes.class.animation_support
```
captainbrosset, line 9: Looking at the dist file, I see 2 main groups: one which became baseline low in 2015, and the other which became baseline low in 2020.
Do you think SMIL was completely usable in 2015? Any idea what the 2020 group added? Is it worth thinking about it in terms of a separate feature? Probably not because it's all baseline high by now.
But if not, then I'd add a comment on this line which explains why we chose this particular compute_from key, even if a lot of SMIL features landed at least 5 years prior to that.

features/web-animations.yml
```diff
@@ -1,6 +1,70 @@
 name: Web animations
-description: The web animations API allows you to animate and synchronize the animations of DOM elements.
+description: The web animations JavaScript API allows you to animate and synchronize the animations of DOM elements.
```
captainbrosset, line 2:
```suggestion
description: The web animations API programmatically animates elements over time, and allows you to synchronize the animations of multiple elements.
```

We don't really prefix API with JavaScript in other features, even when they are, indeed, JS-based. I know that you probably felt it was important to add it here, to make it clear that this was separate from CSS and SVG animations. What about using the word "programmatically", to differentiate from "declaratively"?
I'm not super sold on it. Let me know your thoughts.

I also slightly rephrased because the existing sentence felt a bit hard to parse to me (I was probably the one who wrote it originally 😄).

And I, again, added the time concept here. I remember, years ago, when the WAAPI was supposed to grow the ability to support other types of timelines, such as scroll position (I still see some verbiage about this in the spec). I guess this never happened, and we now have scroll-driven animations purely in CSS defined in a separate spec, and not doable with WAAPI. So it's probably a good thing to say that these are time-based only.

captainbrosset (changes requested):

features/svg-animations.yml
jamesnw: It looks like almost all keys are quite old, but didn't have Edge support until Edge adopted Chromium. Perhaps there was earlier Edge support? Might be worth adding to #1499.

features/svg-animations.yml
jamesnw: Can you explain the significance of differentiating between the element and the element's attributes? Perhaps a poor analogy, but I think of it as "You press the gas pedal to turn the wheels" vs "You press the gas pedal to move the car"- the intended outcome is moving the car, and the wheels are somewhat incidental.

features/web-animations.yml
jamesnw: Looking at this again, "Web Animations API" is a proper noun, and probably should be capitalized. I do think that JavaScript is an important word to include here, as that is both how this is used, and the key differentiator with other animation techniques. Those do mention their language, and we shouldn't assume JavaScript as the default language. 

How does this look? I agree the time factor is important to include.

```suggestion
description: The Web Animations API programmatically animates elements over time, and allows you to synchronize the animations of multiple elements in JavaScript.
```

features/svg-animations.yml
captainbrosset: Oh ok, sounds good then.

features/svg-animations.yml
captainbrosset: I think what prompted me to talk about attributes is that we describe CSS animations as animating an element's _styles_.
So, I felt compelled to mention exactly _what_ were SVG animations animating.

Animating an element feels quite abstract to me, but perhaps it's just me. I won't block on this. If you prefer "animate SVG elements", go for it.

captainbrosset (approved):

features/svg-animations.yml
ddbeck: I'm going to weigh in here and say that I prefer _elements_ over _attributes_, since it covers the whole feature. This is because `<animateMotion>` is the odd one out, in that it doesn't obviously animate an element's attribute.

features/svg-animations.yml
```diff
@@ -0,0 +1,99 @@
+name: SVG animations
+description: The `<animate>`, `<animateMotion>`, and `<animateTransform>` SVG elements declaratively animate SVG elements. Also known as SMIL.
+spec: https://svgwg.org/specs/animations/
+caniuse: svg-smil
+group:
+  - animation
+  - svg
+# Most features are quite old, but only became baseline when Edge adopted Chromium.
+status:
+  compute_from: api.SVGAnimateElement
```
ddbeck, line 10: Can we pin to `svg.elements.{animate,animateMotion,animateTransform}` instead? That'll align more closely with caniuse and guarantee to not under-report the entrypoints mentioned in the description.

features/web-animations.yml
```diff
@@ -1,6 +1,70 @@
 name: Web animations
-description: The web animations API allows you to animate and synchronize the animations of DOM elements.
+description: The Web Animations API programmatically animates elements over time, and allows you to synchronize the animations of multiple elements in JavaScript.
```
ddbeck, line 2: Maybe I'm off base here, but does this work as an entry point—it's the `compute_from` key, after all? Then we don't need the explicit note about scripting either.

```suggestion
description: The `animate()` method of `Element` objects programmatically animates elements over time and can synchronize the animations of multiple elements.
```

ddbeck (changes requested):

features/svg-animations.yml
```diff
@@ -0,0 +1,99 @@
+name: SVG animations
```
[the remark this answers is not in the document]
captainbrosset: For what it's worth:

image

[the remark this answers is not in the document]
captainbrosset: "SVG animation" (singular) shows a bit more searches than the plural form, but still way below smil.

features/web-animations.yml
captainbrosset: Yes, I think this work.

features/css-animations.yml
```diff
@@ -0,0 +1,51 @@
+name: CSS Animations
```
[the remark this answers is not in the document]
jamesnw: Thanks for the thought process on this! I went with `Animations (CSS)` because I agree the disambiguation will be needed.

features/scroll-driven-animations.yml
```diff
@@ -1,7 +1,9 @@
 name: Scroll-driven animations
 description: CSS scroll-driven animations are a type of CSS animations that don't run over time, but are instead driven by the user's scroll position.
 spec: https://drafts.csswg.org/scroll-animations-1/
-group: scrolling
+group:
+  - scrolling
+  - animation
 compat_features:
```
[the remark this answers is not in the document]
jamesnw: I uncommented both and set the compute from to avoid a second pass if needed.

features/svg-animations.yml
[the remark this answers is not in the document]
jamesnw: I decided to go with `SMIL SVG animations`. In looking at Stackoverflow,  there were a lot of people talking about animating in SVG with `<animate*` but not talking about it as SMIL, so I think it is more understandable with both SMIL and SVG.

features/smil-svg-animations.yml.dist
```diff
@@ -1,18 +1,18 @@
-# Generated from: svg-animations.yml
+# Generated from: smil-svg-animations.yml
 # Do not edit this file by hand. Edit the source file instead!
 
 status:
   baseline: high
-  baseline_low_date: 2020-01-15
-  baseline_high_date: 2022-07-15
+  baseline_low_date: ≤2020-01-15
+  baseline_high_date: ≤2022-07-15
```
ddbeck, line 7: I sent the (non-blocking) https://github.com/mdn/browser-compat-data/pull/24607 to remove the ranges. It shouldn't change the overall date, just remove the uncertainty.

ddbeck (approved):

PR #1773, opened 2024-09-12
Add Storage APIs

autonome:
Caniuse:
- has a feature for `Web Storage - name/value pairs` at 97% which points to the storage spec
- has a feature for `Storage` at 95% which also points to the storage spec
- the rest are pulls from MDN

MDN
- rolls most everything up under `Storage API` and points at the various flavours
- `Web Storage API` and `Storage` interface (local/session storage)
- `Shared Storage`
- `IndexedDB`
- Storage quotas/eviction maybe needs a separate feature...
- `Cookies`
- `Cache API`

Web Features
- Feature`Storage buckets` from #555 
- Feature `Storage access` from #1032
- no storage-related groups yet
- Drafts generated from specs
  - features/draft/spec/storage.yml - `Storage` globals and `StorageManager` (but not the storage implementations)
  - features/draft/spec/shared-storage.yml - Shared Storage API
  - features/draft/spec/saa-non-cookie-storage.yml - Extending Storage Access API (SAA) to non-cookie storage -
  - features/draft/spec/requeststorageaccessfor.yml - the `requestStorageAccessFor` API
  - features/draft/spec/indexeddb-3.yml - IndexedDB
  - features/draft/spec/html.yml - has the `Storage` implementation globals and their APIs
  - Cache API is in the serviceworkers spec
  - Cookies in a various places

Plan:

Ignore Cookie and Cache APIs for now. Start w/ the Storage interface and its implementations, as that'll set the direction and be reference point for storage approaches that followed it. Add the "late additions" and recent privacy storage bits as separate features. The rest is mostly non-controversial standalone features. Add a group called "storage" and put everything in it.

- Feature: `Storage API` - implementations of the `Storage` interface and their globals, and includes `Storage manager`
- Feature: `Shared storage`
- Feature: `Shared storage access for non-cookie storage` (late addition, explicit difference)
- Group: `storage`

Not addessed:
- Cookies
- Cache
- IndexedDB


features/storage-manager.yml
```diff
@@ -1,9 +1,11 @@
...
+# Name
+#
+#
+name: Storage Manager
 description: TODO
 spec: https://storage.spec.whatwg.org/
+group: storage
 compat_features:
-  - api.Navigator.storage
   - api.StorageManager
```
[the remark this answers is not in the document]
autonome: 👍🏼 Moved this all back into `Storage`.

features/indexeddb.yml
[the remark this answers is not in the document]
autonome: Removed IndexedDB for now.

autonome:
Ok integrated earlier review comments, simplified overall approach and took out `IndexedDB`.

features/shared-storage-non-cookie.yml
```diff
@@ -1,7 +1,7 @@
-draft_date: 2024-09-09
 name: Extending Storage Access API (SAA) to non-cookie storage
```
ddbeck, line 1: Not much ink has been spilt on this yet. All I've got is this: https://developers.google.com/privacy-sandbox/blog/saa-non-cookie-storage

```suggestion
name: Non-cookie storage access
```

(rename the file to match)

features/shared-storage-non-cookie.yml
```diff
@@ -1,7 +1,7 @@
...
-description: TODO
+description: "This extension of the Storage Access API enables content in cross-site iframes to request access to first-party data beyond cookies."
```
ddbeck, line 2:
```suggestion
description: The first parameter to `document.requestStorageAccess()` requests reading and storing site data from sources other than cookies, such as local storage, session storage, or IndexedDB.
```

features/shared-storage-non-cookie.yml
```diff
@@ -1,7 +1,7 @@
...
 spec: https://privacycg.github.io/saa-non-cookie-storage/
+group: storage
 compat_features:
   - api.Document.hasUnpartitionedCookieAccess
```
ddbeck, line 6: Move this key to `storage-access`. It's an alias to `document.hasStorageAccess()`.

groups/storage.yml
```diff
@@ -0,0 +1 @@
+name: Storage
```
ddbeck, line 1: At some point we'll need a description for these. Would you mind leaving a comment with a draft description, so we know what you were thinking when we come back to this?

ddbeck (dismissed):

features/storage-access.yml
```diff
@@ -1,3 +1,12 @@
...
 description: The `document.requestStorageAccess()` method allows content in iframes to request storing and reading cookies and other site data, while the `document.hasStorageAccess()` method checks if such access is granted.
 spec: https://privacycg.github.io/storage-access/
+group: storage
+# Gets this to match on baseline status on MDN
+status:
+  compute_from: api.Document.requestStorageAccess
+compat_features:
+  - api.Document.hasStorageAccess
+  - api.Document.requestStorageAccess
+  - api.Document.requestStorageAccessFor
```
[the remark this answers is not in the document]
autonome: Yes, I read all about it before adding it. My initial thought was to break it as a feature.

But I've learned that reviews tend to fold these features back in, so now I just wait it out 😂

I'd rather not just leave it hanging. How about either I do the feature now, or leave here commented-out for future work?

features/storage.yml
```diff
@@ -0,0 +1,31 @@
+name: Storage
+description: "`Storage` is for access and manipulation of a particular domain's stored data either for the duration of a session, or persistent across sessions, using name/value pairs."
```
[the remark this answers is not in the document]
autonome: Me too, I figured it wouldn't fly though!

features/shared-storage-non-cookie.yml
autonome: omg so much better

groups/storage.yml
autonome: Added description

autonome:
Ok I think all review comments addressed, just figure out a way forward for `requestStorageAccessFor`.

features/localstorage.yml
```diff
@@ -0,0 +1,31 @@
...
+  - api.StorageEvent
+  - api.StorageEvent.key
+  - api.StorageEvent.newValue
+  - api.StorageEvent.oldValue
+  - api.StorageEvent.storageArea
+  - api.StorageEvent.url
+  - api.StorageManager
+  - api.StorageManager.estimate
+  - api.StorageManager.persist
+  - api.StorageManager.persisted
```
captainbrosset, line 27: Note that these BCD keys (but also `navigator.storage` and `storage_event`, I think) are not only useful for local/sessionStore APIs, but apply to other forms of storage that an origin can use too.

For example, a developer can use the `navigator.storage.estimate()` promise to discover how much their origin currently uses across IndexedDB, Cache API, and File System API.

I wrote https://patrickbrosset.com/articles/2023-01-17-web-storage/ and https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria, so I have fairly up to date understanding of this area. And I think we should split the StorageManager stuff into its own feature.

In fact, it's not clear to me that the StorageManager actually applies to local/sessionStorage at all. Because these browser-storage mechanisms (as well as cookie) predate the StorageManager API, browsers didn't fold them into the way the StorageManager API handles origin data from the very start. They do intend to manage all types of storage the same way eventually, but I'm not certain this has been done.

features/shared-storage.yml
```diff
@@ -1,7 +1,7 @@
-draft_date: 2024-09-09
-name: Shared Storage API
-description: TODO
+name: Shared storage
+description: The `sharedStorage` API stores data across sites. Unlike `localStorage`, third-parties can write to shared storage, which you can read from the origin browsing context.
```
captainbrosset, line 2: This one is tricky. It's quite a complex API. Devs sort of need to understand privacy implications of ads before being able to use it. I think the description could do a bit of better job at making the main use case explicit. The use of "site", "third-parties", "origin", and "browsing context" assume knowledge here. Let me try something:

```suggestion
description: The `sharedStorage` API stores data to a shared space where the data can then be processed without the ability to track users across the different sites they visit. A common use case is measuring the reach of third-party ads without using user-tracking cookies."
```

captainbrosset (changes requested):

features/storage-access.yml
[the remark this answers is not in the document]
ddbeck: Eh, let's leave it in for now. I can see it both ways now and I think doing something "wrong" is more likely to get resolved better than setting aside a tiny API surface sitting on top of a conceptually massive feature (that's not terribly likely to get implemented, given vendor positions, at least as defined).

features/localstorage.yml
autonome: I had it split out initially as the usage and use-cases are distinct enough from the act of using storage that it seemed like a standalone feature. So whether it does or does not apply to local/sessionStorage, +1 to a separate feature. @ddbeck thoughts?

ddbeck: Yeah, sorry for not catching this on initial review myself. StoragManager _does_ seem like a feature, with either `navigator.storage` or `navigator.storage.estimate()` as the entry point for the description. Probably something like, "The `navigator.storage` API provides information about the availability and persistence of data stored by the browser."

(I believe `navigator.storage.getDirectory()` is already part of another feature, for OPFS.)

autonome: Split out StorageManager again. Is baseline low because of `estimate`, but we could compute from the main feature if that seems like an addition that shouldn't hold the feature back. I erred on the conservative side here because it's a good chunk of the functionality of the overall feature.

captainbrosset: Yeah, I tend to agree on keeping `estimate` part of the equation here. Even if it actually lies to you (for security reasons), it's one of the key features of the whole API.

features/storage-manager.yml
```diff
@@ -1,11 +1,11 @@
-draft_date: 2024-09-09
-name: Storage Standard
-description: TODO
+name: Storage manager
+description: "The `navigator.storage` API provides information about the availability and persistence of data stored by the browser."
```
captainbrosset, line 2:
```suggestion
description: "The `navigator.storage` API provides information about the availability and persistence of the data that a site stores on the device, by using APIs such as the Cache API or the IndexedDB API."
```

I feel like this is a bit clearer this way. I'm not certain about which APIs to mention though, as this has been a moving target. Eventually, _all_ client-side stored data will be managed through by the Storage Manager, but I don't think this is the case currently.

captainbrosset (approved):
Approving because I think this is the right way to create those features. But, please take a look at my remaining comments.

features/storage-buckets.yml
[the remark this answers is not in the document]
autonome: Not a new feature, dist already in the tree.

features/non-cookie-storage-access.yml
```diff
@@ -1,9 +1,8 @@
-draft_date: 2024-09-09
-name: Extending Storage Access API (SAA) to non-cookie storage
-description: TODO
+name: Non-cookie storage access
+description: "This extension of the Storage Access API enables content in cross-site iframes to request access to first-party data beyond cookies."
```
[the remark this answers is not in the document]
autonome: Thanks, yes this description needed work - and in fixing it up I also found a key in the wrong place! Both updated now.

autonome:
Addressed all review comments but GH still showing blocked for some reason, so removing @ddbeck review to unblock merging, can follow up w/ any remaining issues afterward. Thanks everyone for helping w/ this one!

PR #1779, opened 2024-09-13
Sorting the CSS-Cascade-6 keys

mirisuzanne:
While `revert-layer` is a new global value for explicit defaulting (like `initial`, etc) - it's generally considered part of the Cascade Layers feature.

Those other keys here are a problem. 😅 There are several names we would use for them in spec circles…

- What they are: the 'global keywords' or 'global values'
- What they do: 'explicit defaulting'

There are a number of resources that list them without any name, or only refer to them individually (including CanIUse), and a couple that refer to them as part of 'inheritance' (which is pretty vague). MDN has them separate but cross-referenced, and then also listed as values for the `all` property. CanIUse also shows the BCD keys individually, but each named `global keywords: (keyword)`. 

Since they are the only values allowed on all CSS properties, they are also the only values allowed on `all`. The `all` property is useless without these values - but they are very useful without `all`.

(side note: several of the top articles describe the feature wrong)

I find 'explicit defaulting' to be the most _useful_ term here, for both the values and the `all` shorthand. But it also feels the most technical, and I can't argue that it's a _popular_ term (I guess I have a new mission). I think the most accurate reflection of developer language is likely to list them as separate features. 

We could consider adding them to a group (named what?) for better cross-reference.


features/initial-value.yml
```diff
@@ -0,0 +1,6 @@
+name: initial
+description: The `initial` keyword can be applied to any CSS property, resetting that property to its initial value as defined by the specification. For example, the `initial` value of the `display` property is `inline`, regardless of the element being styled. This is often confused with `revert` which will provide the user or browser default style, specific to the element.
```
jamesnw, line 2:
```suggestion
description: The `initial` keyword resets any CSS property to its initial value as defined by the spec. For example, the `initial` value of the `display` property is `inline`, regardless of the element being styled. This is often confused with `revert` which will provide the user or browser default style.
```

features/revert-value.yml
```diff
@@ -0,0 +1,6 @@
+name: revert
+description: The `revert` keyword can be applied to any CSS property, resetting that property to the previous default, provided by either the user or browser stylesheets. For example, setting `display:revert` on a `<div>` will likely result in `display:block`, the browser default style. This is often confused with `initial` which provides the specified initial value of a property, regardless of the element.
```
jamesnw, line 2: This is too long- you'll likely need to pick either the example or the comparison with initial.

```suggestion
description: The `revert` CSS keyword resets any property to its previous default, provided by either the user or browser stylesheets. For example, setting `display:/ revert` on a `<div>` will likely result in `display:block`, the browser default style. This is often confused with `initial`.
```

features/all.yml
```diff
@@ -0,0 +1,6 @@
+name: all
+description: The `all` CSS property is a shorthand for resetting all CSS properties, except for `direction` and `unicode-bidi`. Only the global CSS keywords for explicit defaulting (`initial`, `inherit`, etc.) are allowed here, since they are the only values supported on all CSS properties.
```
jamesnw, line 2:
```suggestion
description: The `all` CSS property resets all CSS properties, except for `direction` and `unicode-bidi`. Only the global CSS keywords for explicit defaulting (`initial`, `inherit`, etc.) are allowed here, since they are the only values supported on all CSS properties.
```

jamesnw (commented):
I think the feature composition makes sense. I made a few suggestions for description length, and also to favor the `The <property> <sets> the <noun>` [format](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#descriptions) that moves the most important parts to the front.

features/unset-value.yml
```diff
@@ -0,0 +1,6 @@
+name: unset
+description: The `unset` keyword can be applied to any CSS property, removing any values previously declared by the author, user, or browser. For inherited properties (like `color`) the property will inherit the parent value, while non-inherited properties (like `display`) will default to the specification-defined initial value for that property (`inline`). In order to fallback to the browser default styles instead, authors can use `revert` or `revert-layer`.
```
ddbeck, line 2: This will exceed the length limit (300 characters). Here's my take on shortening this:

```suggestion
description: The `unset` keyword resets any CSS property, removing any values previously set by the author, user, or browser. Inherited properties (like `color`) reset to the parent element's computed value. Non-inherited properties (like `display`) reset to the specified initial value (`inline`).
```

ddbeck (changes requested):
I like this. I have some suggestions for consistency in descriptions or to satisfy the test scripts. The only other thought I have here is that "explicit defaulting" is a really helpful lens. I don't think we can get away with defining a feature with that name (or at least not yet), but I think a _group_ for explicit defaulting makes a ton of sense, if you'd like to put all of these in such a group.

features/all.yml
```diff
@@ -0,0 +1,6 @@
+name: all
+description: The `all` CSS property resets all CSS properties, except for `direction` and `unicode-bidi`. Only the global CSS keywords for explicit defaulting (`initial`, `inherit`, etc.) are allowed here, since they are the only values supported on all CSS properties.
```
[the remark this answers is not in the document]
mirisuzanne:
```suggestion
description: The `all` CSS property is a shorthand for all CSS properties, except for `direction` and `unicode-bidi`. It accepts only the keywords for explicit defaulting (such as `initial` and `inherit`), since they are the only values supported on all CSS properties.
```

features/cascade-layers.yml
```diff
@@ -1,5 +1,14 @@
 name: Cascade layers
 description: The `@layer` CSS at-rule avoids specificity conflicts by providing priority levels for different groups of CSS rules, such as low-priority styles like resets, and high-priority styles like UI components.
-spec: https://drafts.csswg.org/css-cascade-5/#layering
+spec: https://www.w3.org/TR/css-cascade-5/#layering
```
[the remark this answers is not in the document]
mirisuzanne:
```suggestion
spec: https://drafts.csswg.org/css-cascade-5/
```

features/inherit-value.yml
```diff
@@ -0,0 +1,6 @@
+name: inherit
+description: The `inherit` keyword can be applied to any CSS property, to 'inherit' the computed value of that property from the parent element.
```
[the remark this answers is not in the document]
mirisuzanne: Fixing the copy-paste typo:

```suggestion
description: The `inherit` keyword resets any CSS property to the computed value of that property from the parent element.
```

mirisuzanne:
@jamesnw @ddbeck thanks for the feedback! I think I've incorporated your notes, and added the `explicit-defaults` group, but check my work.

Ideally, `revert-layer` is both part of the Cascade Layers feature, and also the `explicit-defaults` group. But I don't think that's possible, unless we put all of Cascade Layers into that group? We could call it 'explicit cascading' instead…

jamesnw:
> Ideally, `revert-layer` is both part of the Cascade Layers feature, and also the `explicit-defaults` group. But I don't think that's possible, unless we put all of Cascade Layers into that group? We could call it 'explicit cascading' instead…

Features can't be part of another feature, but we could add a `Cascade Layers` group, and then `explicit defaults` could either be a child of that larger group, or `revert-value` could be in both groups.

jamesnw (approved):

ddbeck:
@jamesnw's suggestion is the only way we could handle that now. We need to resolve https://github.com/web-platform-dx/web-features/issues/1173 to be able to put the key in multiple features (or at a higher level, have composite features that can be made up smaller atoms, see https://github.com/web-platform-dx/web-features/issues/971).

I think this looks good for now, but it would be nice to leave a comment on the cascade layers file, mentioning the possibility of doing that later?

ddbeck (approved):

mirisuzanne:
@ddbeck looks like it needs to be re-approved with that change.

ddbeck:
We're so close now. Prettier needs to be run on this and then we can merge.

jamesnw:
> We're so close now. Prettier needs to be run on this and then we can merge.

@ddbeck  Done!

PR #1786, opened 2024-09-16
Adds widows, orphans, and box-decoration-break

vwallen:
Adds features:

- widows and orphans
- box-decoration-break

`caniuse` keys omitted for discrepancies between BCD and:

- https://caniuse.com/css-widows-orphans
- https://caniuse.com/css-boxdecorationbreak


features/box-decoration-break.yml
```diff
@@ -0,0 +1,8 @@
+name: box-decoration-break
+description: The `box-decoration-break` CSS property sets whether the box decoration of an element divided by page, column, or region breaks wraps each fragment or is split across the break.
```
autonome, line 2: "breaks wraps" together sounds funky

features/box-decoration-break.yml.dist
```diff
@@ -0,0 +1,12 @@
+# Generated from: box-decoration-break.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
```
autonome, line 5: this can't be right, the main property is supported everywhere. maybe add a `compute_from` for it and see if the resulting dist file changes? very strange that the other browsers don't show up at all though.

https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break#browser_compatibility

features/widows-orphans.yml
```diff
@@ -0,0 +1,7 @@
+name: widows and orphans
```
autonome, line 1: Names are sentence-case, so should capitalize `widows`.

However, I realized that's not clearly articulated in the docs, just something I picked up in reviews. Filed #1804 for that.

autonome (changes requested):

features/box-decoration-break.yml.dist
jamesnw: The others aren't showing up because `box-decoration-break` is prefixed everywhere but Firefox.

features/widows-orphans.yml
```diff
@@ -0,0 +1,7 @@
+name: Widows and orphans
+description: The `widows` and `orphans` CSS properties set the minimum lines included in a text fragment created by page, column, or region breaks.
+spec: https://drafts.csswg.org/css-break-4/#widows-orphans
```
ddbeck, line 3:
```suggestion
spec: https://drafts.csswg.org/css-break-4/#widows-orphans
caniuse: css-widows-orphans
```

features/box-decoration-break.yml
```diff
@@ -0,0 +1,8 @@
+name: box-decoration-break
+description: The `box-decoration-break` CSS property sets whether the box decoration of an element divided across a page, column, or region wraps each fragment or is split across the break.
```
ddbeck, line 2: suggestions to include an example of a box decoration and to improve parallelism ("wraps or… splits" instead of "wraps… or is split"):

```suggestion
description: The `box-decoration-break` CSS property sets whether box decorations, such as borders or backgrounds, of an element divided across a page, column, or region wraps each fragment or splits across the break.
```

ddbeck (changes requested):
Two small changes suggested here, otherwise this is very tidy. Thank you!

jamesnw (approved):
LGTM!

ddbeck (approved):

PR #1794, opened 2024-09-17
Add border, border-radius and outline

jamesnw:
The caniuse for [outline](https://caniuse.com/outline) was just a bit too far off, with none of the features lining up with what I am seeing there.


features/border-radius.yml
```diff
@@ -0,0 +1,22 @@
+name: border-radius
+description: The `border-radius` CSS property rounds the corners of the border drawn around an element.
+spec: https://drafts.csswg.org/css-backgrounds-3/#corners
+caniuse: border-radius
```
ddbeck, line 4: Let's `compute_from: css.properties.border-radius` to better align with caniuse.

features/outlines.yml
```diff
@@ -0,0 +1,22 @@
+name: Outlines
+description: The `outline` CSS property sets the color, style, and width of a line around an element, outside of the border. The `outline-offset` CSS property sets the distance from the element's edge to the outline.
+spec: https://drafts.csswg.org/css-ui-4/#outline
+status:
+  # The `outline` shorthand was introduced at a later date.
+  compute_from: css.properties.outline-style
```
ddbeck, line 6: I think it's a bit weird for us to name `outline` as the entry point to this feature, yet say that it's supported before the thing we name in the description exists. I think we have three options here (in order of my uninformed feeling about it):

1. `compute_from` on the constituent properties (`outline-color`, `-offset`, etc.) and change the description to match.
2. Break out `outline` into its own feature.
3. Drop the `compute_from` and say that outlines are a recent feature.

features/outlines.yml
```diff
@@ -0,0 +1,22 @@
+name: Outlines
+description: The `outline` CSS property sets the color, style, and width of a line around an element, outside of the border. The `outline-offset` CSS property sets the distance from the element's edge to the outline.
+spec: https://drafts.csswg.org/css-ui-4/#outline
```
ddbeck, line 3: It'd be nice to leave a `TODO` comment here saying we _could_ reference https://caniuse.com/outline but the differences in support need reconciliation.

ddbeck (changes requested):
I have some thoughts on outlines, but I like the rest of this as it is.

RByers:
> The caniuse for [outline](https://caniuse.com/outline) was just a bit too far off, with none of the features lining up with what I am seeing there.

FWIW I think caniuse was actually right here and BCD is wrong about outline. Filed [this bug to correct BCD](https://github.com/mdn/browser-compat-data/issues/26045)

PR #1799, opened 2024-09-18
Add path shape

jamesnw:
Split from #1649. BCD data needs investigation- see https://github.com/web-platform-dx/web-features/pull/1649#discussion_r1765227595


jamesnw:
@ddbeck Can you take a look at the BCD data?

PR #1800, opened 2024-09-18
Add Baseline status lookup page to site

petele:
Adds a simple Baseline status lookup page to the site using the `<baseline-status>` universal widget. 

This provides a super quick, easy way to look up the status of a feature, and get a visual element you can either screenshot or reference. It's not meant to be exhaustive, but instead a quick way to verify something, and get the status via the widget.


captainbrosset:
I haven't yet given thoughts about this, but I do agree that aligning the baseline landing page and the explorer, under one common umbrella, seems like the right thing to do.
Both sites use eleventy now, both sites are hosted on our GH org. Makes sense to merge and have one common experience I think.

petele:
I'm on board with a common experience. I wanted something that folks could see and search with the widget quickly. Happy to discuss what this might look like moving forward!

captainbrosset:
Thanks for proposing this Pete. We discussed this today at the WebDX call during our old issues/PRs triage and decided to close it. There was still a bit of uncertainty around where to land this code, and no one stepped up to do it. If you're still interested, feel free to re-open, but seeing as there hasn't been any changes since last year, let's close this for now.

PR #1801, opened 2024-09-18
Adds page-setup and page-selectors for @page rules

vwallen:
I named the @page rule feature "page setup" because that is the common usage for grouping those traits together (page dimensions, orientation, and margins). Can I Use uses the label "paged media" which follows the specification terminology, so would be a suitable alternative.

Used the phrasing "an @page rule" but "a @page rule" is more appropriate depending on how the at-rules are spoken aloud. Is there a style guideline for this?


features/page-setup.yml
```diff
@@ -0,0 +1,13 @@
+name: Page setup
+description: The `@page` rule sets the page-specific dimensions, orientation, and margins for content such as printed documents, ebooks, or slides.
+spec: https://drafts.csswg.org/css-page-3/#at-page-rule
+group:
+  - layout
+  - print
+caniuse: css-paged-media
+compat_features:
+  - css.at-rules.page
+  - css.at-rules.page.page-orientation
```
jamesnw, line 10: I think I'd recommend moving `css.at-rules.page.page-orientation` into its own feature, and then pinning baseline on `css.at-rules.page.size`.

It looks like support for @page is incorrect in Safari. I opened https://github.com/mdn/browser-compat-data/issues/24566, and I think that will block this.

ddbeck: I sent https://github.com/mdn/browser-compat-data/pull/24583 to fix this.

features/page-orientation.yml
```diff
@@ -0,0 +1,8 @@
+name: page-orientation
+description: The `page-orientation` CSS property sets the rotation of a page after it has been layed out. This allows pages with the same `size` declaration to display at different orientations.
```
jamesnw, line 2:
```suggestion
description: The `page-orientation` CSS property sets the rotation of a page after layout for print. This allows pages with the same `size` declaration to display with different orientations.
```

I wasn't sure about `layed out` vs `laid out` so I avoided the question 😆 .

I also thought mentioning 'print' was a good idea here, but I'm not sure my suggestion is the ideal way.

features/page-selectors.yml
```diff
@@ -0,0 +1,11 @@
+name: Page selectors
+description: The `page` CSS property matches a page defined by an `@page` rule. The `:first`, `:left`, and `:right` pseudo-classes match pages based on their position in sequence after pagination.
+spec: https://drafts.csswg.org/css-page-3/#page-selectors
+group:
+  - css
```
jamesnw, line 5:
```suggestion
  - selectors
```

features/page-setup.yml
```diff
@@ -0,0 +1,14 @@
+name: Page setup
+description: The `@page` rule sets the page-specific dimensions, and margins for content such as printed documents, ebooks, or slides.
```
jamesnw, line 2:
```suggestion
description: The `@page` rule sets the page-specific dimensions and margins for content such as printed documents, ebooks, or slides.
```

jamesnw (approved):
A few wording changes, but the feature set looks good!

features/page-orientation.yml
vwallen: Alternate suggestion:

```suggestion
description: The `page-orientation` CSS property sets the rotation of a page after pagination. This allows pages with the same `size` declaration to display with different orientations.
```

jamesnw: This looks good to me!

features/page-setup.yml
```diff
@@ -0,0 +1,14 @@
+name: Page setup
+description: The `@page` rule sets the page-specific dimensions and margins for content such as printed documents, ebooks, or slides.
```
ddbeck, line 2: For consistency with other descriptions, to mention CSS.

```suggestion
description: The `@page` CSS at-rule sets the page-specific dimensions and margins for content such as printed documents, ebooks, or slides.
```

ddbeck (commented):
Sorry, I only just now got to the style question. See my line comments. Thank you!

PR #1802, opened 2024-09-18
Adds additional compat keys to page-breaks and column-breaks

vwallen:


features/page-breaks.yml.dist
```diff
@@ -56,10 +62,86 @@ compat_features:
...
   #   edge: "12"
   - css.properties.page-break-after.avoid
   - css.properties.page-break-before.avoid
+
+  # baseline: false
+  # support:
+  #   firefox: ≤72
+  #   firefox_android: "79"
+  - css.properties.break-after.always
+  - css.properties.break-before.always
```
jamesnw, line 147: These two are not in the spec. Can you remove these here, and then open an issue on BCD flagging that they should not be marked as standard_track?

jamesnw (changes requested):
I think if we were creating these new features here, I would recommend pinning baseline to something, as most of the feature is usable. But I would recommend not changing the support status at all, which will be the case after the non-spec keys are removed.

features/page-breaks.yml.dist
jamesnw: Victor flagged elsewhere that these keys are in css-breaks-4, so I retract this comment.

features/page-breaks.yml.dist
```diff
@@ -2,11 +2,17 @@
...
-    chrome: "50"
-    chrome_android: "50"
-    edge: "12"
+    chrome: ≤80
+    chrome_android: "80"
+    edge: ≤80
+    firefox: "92"
+    firefox_android: "92"
+    safari: ≤13.1
+    safari_ios: ≤13.4
```
ddbeck, line 15: Unfortunately, this is going to be a little messy to align to caniuse's `css-page-break`. See https://github.com/web-platform-dx/web-features/pull/331#discussion_r1580845902 for background on how this pair of features came to be.

What needs to happen now: pick a `compute_from` that aligns us to caniuse. It's so messy, so I won't expect version-perfect results here, but we need to report that Chrome and Edge support this, while Safari and Firefox don't. I _think_ this means pinning to `avoid` values, but I'm not 100% sure.

ddbeck (changes requested):

features/page-breaks.yml.dist
ddbeck: This also provides more context: https://github.com/Fyrd/caniuse/issues/7045

vwallen: Updated to compute from `css.properties.page-break-after.avoid` which appears to be the closest match

ddbeck: OK, I've opened https://github.com/web-platform-dx/web-features/issues/1878 to fix this for good.

features/page-breaks.yml
```diff
@@ -1,5 +1,54 @@
 name: Page breaks
 description: The `break-after`, `break-before`, `break-inside` CSS properties (along with `page-break-` aliases) control where printed pages start and end. Also known as pagination or page breaking.
-spec: https://drafts.csswg.org/css-break-3/#breaking-controls
+spec: https://drafts.csswg.org/css-break-4/#breaking-controls
 group: layout
 caniuse: css-page-break
+status:
+  compute_from: css.properties.page-break-after.avoid
```
ddbeck, line 7: Can we add this before merging?

```suggestion
status:
  # TODO: review this when https://github.com/web-platform-dx/web-features/issues/1878 is resolved
  compute_from: css.properties.page-break-after.avoid
```

ddbeck (approved):
One request for a comment before merging this. Thank you!

Issue #1789, opened 2024-09-17
dist should ignore non-.yml files


autonome:
> /Users/dietrich/misc/web-features/scripts/dist.ts:378
        throw new Error(
              ^
>
> Error: Cannot generate dist for features/draft/spec/.shared-storage.yml.swp, only YAML input is supported
    at <anonymous> (/Users/dietrich/misc/web-features/scripts/dist.ts:378:15)
    at Array.map (<anonymous>)
    at main (/Users/dietrich/misc/web-features/scripts/dist.ts:370:15)
    at <anonymous> (/Users/dietrich/misc/web-features/scripts/dist.ts:421:3)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:120:12)

PR #1806, opened 2024-09-19
`dist` script: be more permissive with file extensions

ddbeck:
This lets more files remain in the `features` directory—such as editor `.swp` files—without throwing an exception. 

Fixes https://github.com/web-platform-dx/web-features/issues/1789.

### DIFF
--- a/scripts/dist.ts
+++ b/scripts/dist.ts
@@ -379,9 +379,7 @@ function main() {
         .crawl(fileOrDirectory)
         .sync();
     }
-    if (isDistOrDistable(fileOrDirectory)) {
-      return fileOrDirectory;
-    }
+    return isDistOrDistable(fileOrDirectory) ? fileOrDirectory : [];
   });
 
   // Map from .yml to .yml.dist to filter out duplicates.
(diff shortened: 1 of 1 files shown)


scripts/dist.ts
```diff
@@ -353,16 +353,35 @@ const tagsToFeatures: Map<string, Feature[]> = (() => {
...
     if (fs.statSync(fileOrDirectory).isDirectory()) {
       return new fdir()
         .withBasePath()
-        .filter((fp) => !(fp.endsWith(".md") || fp.endsWith(".DS_Store")))
+        .filter(isDistOrDistable)
         .crawl(fileOrDirectory)
         .sync();
     }
-    return fileOrDirectory;
+    if (isDistOrDistable(fileOrDirectory)) {
```
autonome, line 382: this change results in new possibility of no return value for the function, is this ok?

autonome (approved):
looks good if the no retval is expected/handled

scripts/dist.ts
ddbeck: No, that's no good. Fixed with ff5cd6d1

autonome: fix looks good 👍🏼

ddbeck:
Since this change is specifically for contributors and doesn't impact what we publish, I'm going to take @autonome's approval as sufficient under the [owner-delegate rule](https://github.com/web-platform-dx/web-features/blob/main/GOVERNANCE.md#owner-delegate-rule).

PR #1783, opened 2024-09-16
`update-drafts`: delete completed specs

ddbeck:
`update-drafts` ought to clean up specs that are finished—that is, specs that have no more keys to assign to features. This PR fixes that deficiency.

Although a change made in https://github.com/web-platform-dx/web-features/pull/1694 satisfies this, it counteracts the benefit of https://github.com/web-platform-dx/web-features/pull/1742, which was to avoid needless diffs (see https://github.com/web-platform-dx/web-features/pull/1782 for an example).

PR #1807, opened 2024-09-19
`update-drafts`: increase frequency of updates

ddbeck:
With the lighter-weight diffs introduced in https://github.com/web-platform-dx/web-features/pull/1783 it now makes sense to run this more frequently, so contributors are less likely to see already-assigned keys.


Elchi3 (approved):
Also https://github.com/web-platform-dx/web-features/pull/1876 makes this less annoying, too.

PR #1811, opened 2024-09-19
Update baseline-in-the-wild.md

meyerweb:
Adding BCD Watch

### DIFF
--- a/gh-pages/src/baseline-in-the-wild.md
+++ b/gh-pages/src/baseline-in-the-wild.md
@@ -13,6 +13,7 @@ You can find Baseline in the following places. If you spot somewhere else using 
 ## Featured projects and tools
 
 - [`<baseline-status>` web component](https://github.com/web-platform-dx/baseline-status)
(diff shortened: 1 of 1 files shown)


gh-pages/src/baseline-in-the-wild.md
```diff
@@ -13,6 +13,7 @@ You can find Baseline in the following places. If you spot somewhere else using
 ## Featured projects and tools
 
 - [`<baseline-status>` web component](https://github.com/web-platform-dx/baseline-status)
+- [BCD Watch](https://bcd-watch.igalia.com/)
```
captainbrosset, line 16: Oh I thought bcd-watch only used BCD. Can you say more about how it uses web-features?

petele: I did, it's a helpful tool, though agree it's not exactly `web-features`.

captainbrosset: I think Eric's comment makes sense. We should add it to the list.

captainbrosset (approved):

PR #1819, opened 2024-09-20
Adds multi-column layout

vwallen:
Following `grid` and `flexbox`, groups the setup properties (`column-count`, `column-width`) with the rendering properties (`column-gap`, `column-rule`) together into a single feature called "Multi-column Layout" which aligns with how these features show up in posts/documentation. The `align-content` property follows previous features that include it with the context in which it applies. 

Splits out `column-fill` and `column-span` into their own features because they apply to the content rather than the layout structure of the columns and/or they are properties of the child elements rather than the parent with the columns declaration. 

Adds `multi-column` group


features/multi-column.yml
```diff
@@ -0,0 +1,20 @@
+name: Multi-column layout
+description: Multi-column layout flows an element's content across one or more content areas in a single row, without affecting the `display` properties of its children.
```
captainbrosset, line 2: I feel like "content areas" should be replaced with "columns". That would make it easier to understand I think.

The "without affecting the display" last part feels a bit obscure to me. I would suggest either clarifying it or removing it.

captainbrosset (approved):

features/multi-column.yml
vwallen: I used "content areas" to avoid a recursive description like "Multi-column layout flows content into columns", but maybe this is a gray area?

"Without affecting the `display` properties" is a distinction from `flexbox` and `grid` that is called out in both the spec and posts comparing these layout options. It would probably be clearer to state the "`display` property" (singular), as that the key point: this property of child elements is unaffected using multi-column on the parent

captainbrosset: Re "display property", I think I get it. Grid and Flex blockify their items, so inline items become blocks. But Multi-column doesn't do that. If that's the main thing to convey, then what about this: "without affecting the `display` property of its children. For example, inline elements remain inline, unlike in Grid or Flexbox layouts, where grid and flex items become block elements".
Might be too much details, and too long. But I thought giving a concrete example would help make this part of the description easier to understand (i.e. less spec-talk). Up to you.

features/column-fill.yml.dist
```diff
@@ -0,0 +1,47 @@
...
+  # support:
+  #   chrome: "50"
+  #   chrome_android: "50"
+  #   edge: "12"
+  #   firefox: ≤72
+  #   firefox_android: "79"
+  #   safari: "9"
+  #   safari_ios: "9"
+  - css.properties.column-fill.auto
+  - css.properties.column-fill.balance
```
ddbeck, line 43: The Firefox data is dubious. I sent https://github.com/mdn/browser-compat-data/pull/24585 to fix it.

features/multi-column.yml
```diff
@@ -0,0 +1,20 @@
+name: Multi-column layout
+description: Multi-column layout flows an element's content across one or more columns in a single row, without affecting the `display` property of its children.
```
ddbeck, line 2: An idea here, to show a likely entrypoint into this feature:

```suggestion
description: The `columns` property creates a multi-column layout that flows an element's content across one or more columns in a single row, without affecting the `display` property of its children.
```

Also, can you tell me a little more about the "without affecting" clause? I get the impression that this supposed to be making a comparison to something else, but I'm not sure what.

vwallen: It's a distinction from `flexbox` and `grid` that alter the display of child elements and can be a factor in choosing multi-column over these alternatives

vwallen: I chose not to open with `columns` to maintain some consistency with the flexbox and grid layout systems, but those are not distinct properties, so I'm not strongly tied to that

ddbeck: Ah, that makes sense. I think ignoring this comment is fine. 👍

features/column-span.yml.dist
```diff
@@ -0,0 +1,43 @@
...
+  # support:
+  #   chrome: "50"
+  #   chrome_android: "50"
+  #   edge: "79"
+  #   firefox: "71"
+  #   firefox_android: "79"
+  #   safari: "9"
+  #   safari_ios: "9"
+  - css.properties.column-span.all
+  - css.properties.column-span.none
```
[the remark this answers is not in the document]
ddbeck: This is now merged in BCD.

features/column-fill.yml.dist
ddbeck: This is now merged in BCD. We'll pick this up in web-features on the next BCD release and upgrade.

ddbeck (approved):
@vwallen If you merge main into the branch and refresh the dist here, this will merge cleanly. Thank you!

PR #1149, opened 2024-05-28
Fix typos everywhere

szepeviktor:
@foolip I've taken the liberty of fixing spelling mistakes.

PR #1825, opened 2024-09-23
Check for misspellings in CI

szepeviktor:


jamesnw:
I'm impressed by how few typos our process has let through so far!

My thought is that spelling checks are really helpful, but I'd prefer them as a comment in a PR, and not as a blocker that requires overrides in order to pass. That provides authors the info they need when they need it, but is a much lower long term cost.

szepeviktor:
I see.
You may want to see #1149 too.

`|| true` done in the latest commit.

szepeviktor:
@ddbeck Hello 👋🏻

I could rebase this PR every week and report typos to you.

szepeviktor:
@ddbeck new findings:
https://github.com/web-platform-dx/web-features/blob/999db5a1b93cd4a93d91f24cedf7dbfe698e85e7/features/font-variant-position.yml#L7
"glpyhs"

szepeviktor:
new finding: "signficantly"

szepeviktor:
new findings: "relase"

.github/workflows/pull_request.yml
```diff
@@ -29,6 +29,25 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          set -o pipefail
+          mkdir -p "${{ runner.temp }}/typos"
+          RELEASE_ASSET_URL="$(
+            gh api /repos/crate-ci/typos/releases/latest \
+              --jq '."assets"[] | select(."name" | test("^typos-.+-x86_64-unknown-linux-musl\\.tar\\.gz$")) | ."browser_download_url"'
+          )"
```
ddbeck, line 42: I have a question here: ci/typos seems to be [available as a GitHub Action](https://github.com/crate-ci/typos/blob/master/docs/github-action.md) (which I imagine we could pin and upgrade with Dependabot). How does this approach compare to using the Action?

Also, have you seen this used in other open projects with a lot of churn on files? I'd be curious to know what a real-world PR looks like following a workflow like this (I'm curious how contributors respond to the feedback given).

.github/workflows/pull_request.yml
```diff
@@ -29,6 +29,25 @@ jobs:
...
+          wget --quiet --output-document=- "${RELEASE_ASSET_URL}" \
+            | tar -xz -C "${{ runner.temp }}/typos" ./typos
+          git grep --files-with-matches --null -I -e '.' \
+            | xargs --null --verbose -- "${{ runner.temp }}/typos/typos" --format json \
+            | jq --raw-output '"::warning file=\(.path),line=\(.line_num),col=\(.byte_offset)::\"\(.typo)\" should be \"" + (.corrections // [] | join("\" or \"") + "\".")' \
+            || true
```
ddbeck, line 48: This is rather complicated! Is there any way this can be broken up or commented to make each step better explained? Or maybe an example showing the transformation going on here? I'm just not sure what this is doing.

ddbeck (commented):
Hi @szepeviktor, thanks again for opening this PR.

This PR came up at the WebDX CG meeting today. I'm reviewing this now, but I've also asked that others have a look too. I'm not familiar with this tool, so I don't have a ton of confidence in my review of it—you'll find that I'm mostly asking questions at this point.

.github/workflows/pull_request.yml
szepeviktor: Actually it is enumerate files -> run typos -> convert output to GitHub check annotations.
All this code is hiding inside the GitHub Action of typos.
We talk about that typos should not stop a PR that why I've copied the code from the Action and modified errors to warnings.

I've sent around ~500 PR-s advertising typos: https://github.com/pulls?q=is%3Apr+author%3Aszepeviktor+archived%3Afalse+Fix+typos
~5% of the people hate it.

.github/workflows/pull_request.yml
szepeviktor: I do not recommend running typos in "passive" mode as a typos in code could be detrimental.

E.g. Go language encourages programmers to use 2 and 3 letter variable names. They usually look like English misspellings. That is all I know.

szepeviktor: typos is just a super useful software - I need an LLM to argue.

captainbrosset:
Discussed on the WebDX CG call now. For a long time we didn't have a way to fix typo problems. Now we do. So this seems less important. Also, the proposed workflow seems complicated to the group. It also feels like it could add friction for legitimate things that are not dictionary words.

PR #1829, opened 2024-09-24
Revise `individual-transforms` description

ddbeck:
This creates a stronger link to the "as opposed to" clause.

That said, another way to do this would be:

> The `translate`, `rotate`, and `scale` CSS properties apply simple transformations independently, as opposed to applying compound transformations with the `transform` CSS property.


jamesnw (approved):
I agree that `single` is more accurate than `simple` here.

captainbrosset (approved):

PR #1831, opened 2024-09-24
Add object-fit and object-position

jamesnw:
They both use `caniuse: object-fit`, but I'm not sure if multiple features can share a feature key. 

Also adds missing `object-view-box` key.


autonome (approved):
Yes, seems not right to have >1 Caniuse mapping, I posted to the channel for broader discussion and input.

The only other comment I have is around the `replaced element` terminology, which seems more on the spec language end of the spectrum than developer/designer POV (esp new learners). If you're using that everywhere already, then maybe not an issue (or a broader one), but doesn't block merging this either way.

ddbeck:
On the caniuse question @autonome highlighted:

Narrowly, caniuse mappings should be one-to-one or one-or-many. As far as I know, we have no features yet that are many-to-one. It would be unhelpful to show multiple statuses for a single caniuse feature, so I think many-to-one also doesn't make sense. I think we ought to pick one of these to have the caniuse mapping and drop it from the other.

In the future, when we have feature composition nailed down, a feature combining `object-fit` and `object-position` is a natural follow on and a place to move the caniuse mapping. Honestly, I see why this is is a single feature on caniuse. Imagine if there was an `object` shorthand property; this would be easy to put down as a single feature.

Long-term, I'd like it if web-features could take any reasonable address/identifier (e.g., caniuse's `object-fit`) and resolve a feature (if we have one) or an anonymous slice of a feature (if we don't).

features/object-position.yml
```diff
@@ -0,0 +1,7 @@
+name: object-position
+description: The `object-position` CSS property places images, videos, and other replaced elements within their container.
+spec: https://drafts.csswg.org/css-images-3/#the-object-position
+# Caniuse includes object-position with object-fit
+caniuse: object-fit
```
ddbeck, line 5:
```suggestion
# object-fit on caniuse includes both object-fit and object-position.
# `caniuse: object-fit` is set in object-fit.yml
```

features/object-position.yml
```diff
@@ -0,0 +1,7 @@
+name: object-position
+description: The `object-position` CSS property places images, videos, and other replaced elements within their container.
```
ddbeck, line 2: I worry a little that "container" implies "containing element" and not the element itself. We've used the term "box" elsewhere I think it might do OK here.

```suggestion
description: The `object-position` CSS property places images, videos, and other replaced elements within their boxes.
```

Or we could spell it out in more words:

```suggestion
description: The `object-position` CSS property places an image, video, or other replaced element within the element's content boundaries.
```

ddbeck (changes requested):

ddbeck (approved):
Cheers!

PR #1832, opened 2024-09-24
Add gradients feature

jamesnw:
I was on the fence about combining linear and radial gradients, but this mirrors Can I Use and MDN, and there isn't a separate support story.


vwallen:
Splitting `linear`/`repeating-linear` and `radial`/`repeating-radial` is more consistent with the `conic-gradients` feature, and makes sense within the `gradients` group

features/gradients.yml
```diff
@@ -0,0 +1,31 @@
+name: Gradients
+description: The `linear-gradient()`, `radial-gradient()` CSS functions and their repeating counterparts create background that progress smoothly between multiple colors.
```
ddbeck, line 2: nits

```suggestion
description: The `linear-gradient()` and `radial-gradient()` CSS functions and their repeating counterparts create backgrounds that progress smoothly between multiple colors.
```

features/gradients.yml
```diff
@@ -0,0 +1,31 @@
...
+spec: https://drafts.csswg.org/css-images-3/#gradients
+# Todo: caniuse support needs to be reconciled for Safari and Firefox.
```
ddbeck, line 4:
```suggestion
# TODO: caniuse support needs to be reconciled for Safari and Firefox.
```

ddbeck (approved):
This is good, but I have a couple extremely minor textual issues to fix.

ddbeck:
Oh, on the question of splitting: I'm agnostic. If you want to split it, I'll re-review. But I suspect we'll want a way to address both and a `TODO` saying something like "we should have separate linear and radial grandients when https://github.com/web-platform-dx/web-features/issues/971 is resolved" would also be fine.

PR #1834, opened 2024-09-24
Add overscroll-behavior

jamesnw:
BCD Data for Edge is incorrect- see [issue on BCD.](https://github.com/mdn/browser-compat-data/issues/24511)

I'm unsure if that is a blocker on this feature.


features/overscroll-behavior.yml
```diff
@@ -0,0 +1,29 @@
+name: overscroll-behavior
+description: The `overscroll-behavior` CSS property sets the behavior when the edges of a scrolling area are reached. The `contain` keyword prevents scroll chaining, and the `none` keyword prevents scroll chaining, bouncing effects, and device specific behaviors like scroll to refresh.
```
vwallen, line 2: It might be better to avoid references to concepts such as scroll chaining or specific boundary actions and concentrate on the intention to disable these behaviors

```suggestion
description: The `overscroll-behavior` CSS property allows disabling default scrolling behaviors when the edges of a scrolling area are reached.
```

features/overscroll-behavior.yml
```diff
@@ -0,0 +1,29 @@
+name: overscroll-behavior
+description: The `overscroll-behavior` CSS property allows disabling default scrolling behaviors when the edges of a scrolling area are reached.
+spec: https://drafts.csswg.org/css-overscroll-1/
+caniuse: css-overscroll-behavior
+group: scrolling
+# BCD has incorrect Edge support data. This can be removed when that is resolved.
```
captainbrosset, line 6: If we keep that in, we should link to the BCD bug. Our future selves will likely only read this and act on it a long time after the bug has been resolved.

captainbrosset (dismissed):

features/overscroll-behavior.yml
ddbeck: A big +1 to this. Any time we can do `# TODO: <action> when <url> resolves` is a winner to me.

I've sent https://github.com/mdn/browser-compat-data/pull/24598 to do the actual fix in BCD.

features/overscroll-behavior.yml
```diff
@@ -0,0 +1,29 @@
+name: overscroll-behavior
+description: The `overscroll-behavior` CSS property allows disabling default scrolling behaviors when the edges of a scrolling area are reached.
```
ddbeck, line 2: Since there's another comment on this that needs action…

https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#allows

```suggestion
description: The `overscroll-behavior` CSS property disables default scrolling behaviors when the edges of a scrolling area are reached.
```

ddbeck (changes requested):

ddbeck (approved):

PR #1838, opened 2024-09-25
Add feature for CSS Font Loading API

queengooborg:
This PR adds a feature for the CSS Font Loading API.


features/font-loading.yml
```diff
@@ -0,0 +1,45 @@
+name: CSS Font Loading
```
jamesnw, line 1:
```suggestion
name: Font loading
```
The common name here is "CSS Font Loading API", which makes me grumpy as it's a JS API for loading fonts, which are used in CSS. I think we should either remove "CSS" or keep the whole name, including API.

jamesnw (changes requested):
Nice feature! A few small suggestions.

jamesnw (approved):
This is looking good to me!

features/font-loading.yml
```diff
@@ -0,0 +1,47 @@
+name: Font loading
+description: The CSS Font Loading API dynamically loads custom fonts and tracks when fonts are loading or ready to use.
```
ddbeck, line 2: I'm going to apply this to be able to merge today, but for posterity:

Prefer an essential entry point to an API over a spec name. It's much more concrete and directly connects with what a developer does with this feature.

```suggestion
description: The `document.fonts` API dynamically loads custom fonts and tracks when fonts are loading or ready to use.
```

ddbeck (approved):

PR #1840, opened 2024-09-25
Add feature for the Battery Status API

queengooborg:
This PR adds a feature for the Battery Status API.


features/battery.yml
```diff
@@ -0,0 +1,16 @@
+name: Battery Status
```
captainbrosset, line 1:
```suggestion
name: Battery status
```

features/battery.yml
```diff
@@ -0,0 +1,16 @@
...
+description: The Battery Status API allows web developers to obtain information about the device's battery, including charge level and status.
```
captainbrosset, line 2:
```suggestion
description: The `navigator.getBattery()` API obtains information about the device's battery, including charge level and status. Also known as the Battery API.
```

captainbrosset (dismissed):

features/battery.yml
```diff
@@ -0,0 +1,16 @@
+name: Battery status
+description: The `navigator.getBattery()` API obtains information about the device's battery, including charge level and status. Also known as the Battery API.
+spec: https://w3c.github.io/battery/
```
ddbeck, line 3:
```suggestion
spec: https://w3c.github.io/battery/
caniuse: battery-status
```

features/battery.yml
```diff
@@ -0,0 +1,16 @@
...
+status:
+  compute_from: api.BatteryManager
```
ddbeck, line 5: This is not wrong exactly but, in general, prefer `compute_from` values that are the most-essential entry point(s) to the feature, what a developer would actually have to type to use it. Here that's the `navigator` API, not the `BatteryManager` interface.

(The exception is when you need to align with caniuse—where you'd use the minimal set of keys to achieve that, assuming the entry point is not sufficient—but that's not needed here.)

```suggestion
  compute_from: api.Navigator.getBattery
```

ddbeck (changes requested):

ddbeck (approved):

PR #1846, opened 2024-09-25
Add background group

jamesnw:
- Adds background group
- Adds background-attachment


vwallen:
Looks good and will help for the remainder `css-backgrounds-3`

features/background-attachment.yml
```diff
@@ -0,0 +1,10 @@
+name: background-attachment
+description: The `background-attachment` CSS property sets whether the background layers of an element move as the element scrolls.
```
ddbeck, line 2: I'm not super comfortable with "background layer" (I'm thinking of writing defensively due to `@layer`), so I'm going to try suggesting some changes here. Consider all of the "layer"-only suggestions optional—I'm happy to open a follow up PR if you prefer.

```suggestion
description: The `background-attachment` CSS property sets whether an element's background image or gradient moves as the element scrolls.
```

features/background-color.yml
```diff
@@ -0,0 +1,6 @@
+name: background-color
+description: The `background-color` CSS property sets the fill color of an element, behind any content and background layers.
```
ddbeck, line 2:
```suggestion
description: The `background-color` CSS property sets the fill color of an element, behind any content and background images or gradients.
```

features/background.yml
```diff
@@ -0,0 +1,6 @@
+name: background
+description: The `background` CSS property is a shorthand that sets the properties of the background layers of an element.
```
ddbeck, line 2:
```suggestion
description: The `background` CSS property is a shorthand that sets several background properties at once.
```

ddbeck (approved):
OK, I've left some commentary on the descriptions (I thought I had a more substantive issue to raise, but it didn't survive closer inspection). Everything is strictly optional, so pick and choose what you'd like to do here and merge—I can open a follow up PR if there's anything that requires further discussion.

features/background-attachment.yml.dist
```diff
@@ -0,0 +1,56 @@
...
+  # baseline_high_date: 2024-09-14
+  # support:
+  #   chrome: "1"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "25"
+  #   firefox_android: "25"
+  #   safari: "15.4"
+  #   safari_ios: "15.4"
+  - css.properties.background-attachment.local
```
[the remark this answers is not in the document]
jamesnw: Opened https://github.com/Fyrd/caniuse/issues/7178

features/background-attachment.yml
jamesnw: This is a good point, and I don't think your rewrites introduced ambiguity, they look good to me!

PR #1854, opened 2024-09-26
Add missing grid keys

jamesnw:
None of these keys seem to warrant new features. The `grid-auto-flow` keys have incorrect Safari data, but will likely have the same support as the other added keys once that is resolved on [BCD](https://github.com/mdn/browser-compat-data/issues/24535).


jamesnw:
@mirisuzanne Can you verify that none of the new keys in the [dist](https://github.com/web-platform-dx/web-features/pull/1854/files#diff-ed22abd54ebf6f8a71d619382c7aee0ece312736de8fa61f9a6577241cb989f3) are features that are separate from what's already in the grid feature?

features/grid.yml.dist
```diff
@@ -157,6 +179,21 @@ compat_features:
...
+  #   chrome: "57"
+  #   chrome_android: "57"
+  #   edge: "79"
+  #   firefox: "52"
+  #   firefox_android: "52"
+  #   safari: ≤13.1
+  #   safari_ios: ≤13.4
+  - css.properties.grid-auto-flow.column
+  - css.properties.grid-auto-flow.dense
+  - css.properties.grid-auto-flow.row
```
ddbeck, line 195: These version ranges make no sense. I sent https://github.com/mdn/browser-compat-data/pull/24539 to fix this so it matches the other Safari 10.1 items.

mirisuzanne (approved):
@jamesnw There are some potential overlapping features here – alignment, gap – that work in various layout models, and the list changes over time. The `grid_context` qualifier makes those keys specific to the grid feature, and I think that's probably a fine choice. 

Gap in flexbox context arrived late (after flexbox), and allignment in block context is fairly new (despite blocks being old) - so it won't always work to combine these things in the other places. But I think that's also probably fine?

This is a place where it might be useful at some point to have overlapping features.

The confusingly named `flex` type is correctly associated with Grid and not Flexbox. 🤷🏻‍♀️

features/grid.yml.dist
jamesnw: @ddbeck  I've updated the dist with the BCD changes here, and it looks much better.

ddbeck (approved):

PR #1859, opened 2024-09-27
Take Ink Enhancement feature out of draft

queengooborg:
This PR brings the Ink Enhancement API out of draft.


features/ink-enhancement.yml
ddbeck, line 1: This is likely to be renamed to `ink.yml` (see next comment)

features/ink-enhancement.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-09-09
-name: Ink API
-description: TODO
+name: Ink enhancement
```
ddbeck, line 1: It's not clear to me where this is coming from. On first look, "Ink (API)" seems to be much more common.

(Though it's possible I'm looking in the wrong places—I'd be happy to be proven wrong with examples of developers referring to this as "ink enhancement".)

```suggestion
name: Ink
```

features/ink-enhancement.yml
```diff
@@ -1,6 +1,5 @@
...
+description: The Ink API allows browsers to directly make use of available OS-level compositors when drawing pen strokes in an inking app feature, thereby reducing latency and increasing performance.
```
ddbeck, line 2: A heads up: if you're going to copy and paste from MDN, you're pretty much guaranteeing at least one round of revisions. Perversely, MDN has a hard time briefly describing web APIs in a developer-oriented way. You might want to read through our existing descriptions and [the guidelines](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#descriptions) to get a better handle on what these things need to say.

OK, onward. What I'm aiming to do here:

- Name the entry point
- Actively describe what happens—what the developer gets—by invoking this API

```suggestion
description: The `navigator.ink` API draws between pointer events, reducing the delay between moving a stylus or cursor and a pen stroke appearing on screen.
```

ddbeck (changes requested):

features/ink-enhancement.yml
captainbrosset: The word enhancement comes from the fact that this API was born out of the need for a faster way to draw ink trails on a canvas. The usual way involves listening for pointer events and drawing shapes on the canvas as a result of them. However, the time it takes for the pointer events to be dispatched to and handled by the webapp is sometimes noticeable to users when drawing quickly. It's visible in the form of your drawing lagging slightly behind your pencil or finger, which can feel weird.
The Ink API solves this by allowing you to bypass the entire process and plug directly into the compositor. You still draw your shapes as before, but now you have a way to tell the device "please draw faster, as the pointer moves, and then I'll handle the events as they get dispatched".

In that sense, it _is_ an enhancement. But it's not an enhancement of an existing _API_. I agree with calling this just **Ink**.

features/ink.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-09-09
-name: Ink API
-description: TODO
+name: Ink
+description: The `navigator.ink` API draws between pointer events, reducing the delay between moving a stylus or cursor and a pen stroke appearing on screen.
```
captainbrosset, line 2: I think "draws between pointer events" is not entirely correct. I think it's more something like "draw before pointer events are dispatched to the webpage". How about:

```suggestion
description: The `navigator.ink` API uses the system compositor to draw to a `<canvas>` element ahead of pointer events, reducing the delay between moving a stylus or cursor and a pen stroke appearing on screen.
```

captainbrosset (approved):

features/ink.yml
ddbeck: I like this. It makes more sense this way. 👍 I see now that my version implied that drawing only happens with events, which is also wrong.

Elchi3 (approved):

PR #1869, opened 2024-09-30
Add Pseudo selectors

jamesnw:
Per https://github.com/web-platform-dx/web-features/pull/1764#discussion_r1780946519, I used `select` as the verb.


features/placeholder-shown.yml
```diff
@@ -0,0 +1,11 @@
+name: :placeholder-shown
+description: The `:placeholder-shown` CSS pseudo-element selects `<input>` and `<textarea>` elements when no value is set and placeholder text is shown.
```
ddbeck, line 2: Two ideas here:

- Avoid circularity with "…is shown"
- Distinguish with similarly-named pseudo?

```suggestion
description: The `:placeholder-shown` CSS pseudo-element selects `<input>` and `<textarea>` elements when no value is set and the element's `placeholder` attribute is not empty. Not to be confused with `::placeholder`, which selects the placeholder text itself.
```

features/placeholder.yml
```diff
@@ -0,0 +1,7 @@
+name: ::placeholder
+description: The `::placeholder` CSS pseudo-element selects text in `<input>` and `<textarea>` elements when no value is set.
```
ddbeck, line 2: Random idea: hint at the use case here?

```suggestion
description: The `::placeholder` CSS pseudo-element selects help text in `<input>` and `<textarea>` elements when no value is set.
```

features/placeholder.yml
```diff
@@ -0,0 +1,7 @@
...
+spec: https://drafts.csswg.org/css-pseudo-4/#placeholder-pseudo
+caniuse: css-placeholder
+group: selectors
```
ddbeck, line 5: I wondered whether this should live with the placeholder attribute. I think they're distinct because they have dramatically different support histories. That said:

```suggestion
group: selectors
# TODO: create some kind of "input placeholders" feature when https://github.com/web-platform-dx/web-features/issues/971 is resolved, combining placeholder selectors and the element attr
```

ddbeck (approved):
This one's really close. I have suggestions exclusively on descriptions and comments. Please resolve each as you see fit and merge, or @ me if you have questions. Thank you!

PR #1870, opened 2024-09-30
Print destination path on feature init

jamesnw:
Tiny QOL change- the first thing I want to do after running `feature-init` is open the newly generated file. This prints the path to the console, so I can click on it and open it.

### DIFF
--- a/scripts/feature-init.ts
+++ b/scripts/feature-init.ts
@@ -81,7 +81,7 @@ async function main() {
     process.exit(0);
   }
   await fs.writeFile(destination, formatted);
-  console.log(`Wrote ${destination}`);
+  console.log(destination);
 }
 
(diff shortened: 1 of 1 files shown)


scripts/feature-init.ts
```diff
@@ -81,6 +81,7 @@ async function main() {
     process.exit(0);
   }
   await fs.writeFile(destination, formatted);
+  console.log(`Wrote ${destination}`);
```
ddbeck, line 84: If you write out just the file name, then it's possible to do something like this: `my-editor $(npm run --silent feature-int -- …)` or pipe it to another command, right?

```suggestion
  console.log(destination);
```

ddbeck (approved):
One optional suggestion for you. Feel free to merge with or without it.

PR #1876, opened 2024-10-01
Turn off drafts updater on forks

ddbeck:
I assume this is not very helpful for forks and it produces a lot of needless branches. I think we should make this opt-in for forks.


autonome (approved):
So bizarre that Github still hasn't fixed this.

Elchi3 (approved):
Was wondering about this, thanks Daniel! 👍

PR #1879, opened 2024-10-01
Feature for element(s)FromPoint

captainbrosset:
These two methods allow devs to retrieve elements at a given point in the viewport. This can be very useful for graphics or text editing apps for example.
This is baseline high, and even though `elementFromPoint()` (singular) appeared way before `elementsFromPoint()` (plural), I don't think it's useful to split them up.


features/element-from-point.yml
```diff
@@ -0,0 +1,9 @@
+name: document.elementFromPoint() and document.elementsFromPoint()
+description: The `document.elementFromPoint()` and `document.elementsFromPoint()` find the top-most element at a given point in the viewport, or all elements at a given point in the viewport, respectively.
```
ddbeck, line 2: Little wording fix:

```suggestion
description: The `document.elementFromPoint()` and `document.elementsFromPoint()` methods find the top-most element at a given point in the viewport, or all elements at a given point in the viewport, respectively.
```

ddbeck (approved):
I'm happy for this to merge, if you apply the description suggestion. Thank you!

PR #1881, opened 2024-10-01
Add missing cjk-ideographic key

jamesnw:
This was accidentally removed in https://github.com/web-platform-dx/web-features/pull/1764, but it should be included, as it is part of the spec.


ddbeck (approved):
Thank you!

PR #1884, opened 2024-10-01
Adds positioning properties and schemes

vwallen:


features/absolute-positioning.yml
```diff
@@ -0,0 +1,6 @@
+name: Absolute positioning
+description: "The `position: absolute` CSS declaration removes an element from the normal flow and positions it relative to its containing block or closest positioned ancestor."
```
jamesnw, line 2: `containing block` is a term with very little developer understanding, but I could not find a better one while working with anchor positioning. Flagging in case anyone else has one.

```suggestion
description: "The `position: absolute` CSS declaration removes an element from the normal flow and positions it relative to its containing block, which is often the root element, or closest positioned ancestor."
```

features/backdrop.yml
```diff
@@ -0,0 +1,6 @@
+name: "::backdrop"
+description: The `::backdrop` CSS pseudo-element is a box underneath an element, such as a `<dialog>`, that displays in the top layer. It can be used to create obscuring effects distinguishing a modal element from the elements underneath.
```
jamesnw, line 2:
```suggestion
description: The `::backdrop` CSS pseudo-element is a box underneath an element in the top layer, such as a `<dialog>`. It can be used to create obscuring effects distinguishing a modal element from the elements underneath.
```

jamesnw (approved):

jamesnw:
The feature and group set makes sense to me!

PR #1885, opened 2024-10-02
Add atob/btoa based on CIU approach

autonome:


features/base64encodedecode.yml
```diff
@@ -0,0 +1,7 @@
+name: Base64 encoding and decoding
+description: Utility functions for encoding and decoding strings to and from base 64.
+spec: https://html.spec.whatwg.org/multipage/webappapis.html#atob
+caniuse: atob-btoa
+compat_features:
+  - api.atob
+  - api.btoa
```
ddbeck, line 7: For posterity: https://stackoverflow.com/questions/33854103/why-were-javascript-atob-and-btoa-named-like-that

features/base64encodedecode.yml
```diff
@@ -0,0 +1,7 @@
+name: Base64 encoding and decoding
+description: Utility functions for encoding and decoding strings to and from base 64.
```
ddbeck, line 2: We've got the room here, so we can name both "entrypoints" to this feature and make it a complete sentence.

```suggestion
description: The `atob()` and `btoa()` global functions encode and decode strings to and from base 64, respectively.
```

ddbeck (changes requested):

features/base64encodedecode.yml
autonome: THANKS BRENDAN

ddbeck (approved):

PR #1886, opened 2024-10-02
Add closewatcher feature

autonome:


features/closewatcher.yml
```diff
@@ -0,0 +1,11 @@
+name: CloseWatcher
+description: "The CloseWatcher API listens and responds to close requests for developer-implemented user interface components such as sidebars or custom keyboards."
+spec: https://html.spec.whatwg.org/multipage/interaction.html#closewatcher
```
ddbeck, line 3: I think this link gives better context:

```suggestion
spec: https://html.spec.whatwg.org/multipage/interaction.html#close-requests-and-close-watchers
```

features/closewatcher.yml
```diff
@@ -0,0 +1,11 @@
+name: CloseWatcher
+description: "The CloseWatcher API listens and responds to close requests for developer-implemented user interface components such as sidebars or custom keyboards."
```
ddbeck, line 2: The examples here blur the lines a bit between browser/system UI and things happening inside the confines of the viewport. What do you think of this instead?

```suggestion
description: "The `CloseWatcher` API listens and responds to close requests on in-page components, such as modals, popovers, and sidebars."
```

ddbeck (changes requested):
This is nice! A couple suggestions on the description and spec URL. Thank you!

PR #1888, opened 2024-10-02
New Selection API features

captainbrosset:
The Selection API has been evolving over the years, so I've spent some time here making the data match caniuse as best I could.
I think the support data makes sense.

I also really wanted to split out `getComposedRanges` because this is much newer, and Chromium is working on an implementation currently, and it unlocks a developer use case that was previously impossible.


features/selection-api-shadow.yml
captainbrosset, line 1: I don't like this feature ID at all. Let me know of better ideas if you have any.

features/selection-api.yml
captainbrosset, line 1: I couldn't name this just selection.yml, because that's already taken by the ::selection pseudo.

autonome: Matches both the spec title and the Caniuse key, so seems fine to me to keep `API` here, and there's precedent for doing so.

features/selection-api-shadow.yml
autonome: The primary aspect of feature seems to be in the API name itself, and the shadow tree boundary crossing is after the fact, as the MDN description notes:

> The Selection.getComposedRanges() method returns an array of [StaticRange](https://developer.mozilla.org/en-US/docs/Web/API/StaticRange) objects representing the current selection ranges, and can return ranges that potentially cross shadow boundaries.

Calling it by the API name in both the key and title might align more with what a developer would search for.

features/selection-api.yml
```diff
@@ -0,0 +1,58 @@
...
+  - api.Selection.removeAllRanges
+  - api.Selection.selectAllChildren
+  - api.Selection.setBaseAndExtent
+  - api.Selection.setPosition
+  - api.Selection.toString
+  - api.Selection.type
+  - api.Window.getSelection
+
+  # The selectionchange events on <input> and <textarea> are not really needed.
+  # The event is fired on document too and his is what developers use.
```
jamesnw, line 39:
```suggestion
  # The event is fired on document too and this is what developers use.
```

features/selection-api-shadow.yml
```diff
@@ -0,0 +1,6 @@
+name: Selections across shadow tree boundaries
+description: The `window.getSelection().getComposedRanges()` method returns ranges that represent the current user selection, even if the selection spans across shadow tree boundaries.
+spec: https://w3c.github.io/selection-api/#dom-selection-getcomposedranges
+group: selection
```
jamesnw, line 4: Should this also go in a Shadow Dom group?

jamesnw (approved):
A few nits, but overall looks good!

features/selection-api-shadow.yml
captainbrosset: There's no shadow dom group right now. There's a web components one though, so I'll add it to that one.

features/selection-api-shadow.yml
captainbrosset: Done, thanks.

PR #1890, opened 2024-10-02
Add BarProp

autonome:
Weird one.

The meaning of the `visible` property changed, but hasn't been implemented everywhere.

The old meaning is still implemented by Apple, so the implementation is not consistent.

Given that incompatibility, leaving this as baseline false provides the best guidance for developers making decisions today.

Cainuse has all green however. If we merge this feature as baseline false, I'll go file an issue there about it.


features/barprop.yml
```diff
@@ -0,0 +1,7 @@
+name: BarProp
+description: "BarProp is implemented by browser UI such as the address bar or scrollbars, historically to represent whether the user interface component was visible. For privacy reasons, the spec was changed to make `visible` represent whether is a popup or not. Apple has not yet implemented the change."
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#barprop
+compat_features:
+  - api.BarProp
+  - api.BarProp.visible
+  - api.BarProp.visible.returns_popup
```
ddbeck, line 7: This should include the APIs that actually return this interface (because you can construct a `BarProp` and you wouldn't even if you could). This is going to be `api.Window.locationbar`, etc. as seen in non-normative note in the spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#browser-interface-elements

features/barprop.yml
```diff
@@ -0,0 +1,7 @@
+name: BarProp
+description: "BarProp is implemented by browser UI such as the address bar or scrollbars, historically to represent whether the user interface component was visible. For privacy reasons, the spec was changed to make `visible` represent whether is a popup or not. Apple has not yet implemented the change."
```
ddbeck, line 2: Two things here:

- Following from my suggestion to widen the set of keys, we have _just_ enough room to just name everything rather than refer to the interface.
- Never mention support or standards status in the description. There's an obscure Greek god that regards this as a kind of hubris, who makes documentation go out of date just after you stop thinking about it.

```suggestion
description: The `visible` property of the `window.locationbar`, `window.menubar`, `window.personalbar`, `window.scrollbars`, `window.statusbar`, and `window.toolbar` APIs historically represented the visibility of a browser user interface. For privacy reasons, it now represents whether the window is a pop-up.
```

features/barprop.yml
```diff
@@ -0,0 +1,7 @@
...
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#barprop
```
ddbeck, line 3: I think it's nicer to link to the prose before the IDL:

```suggestion
spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#browser-interface-elements
```

ddbeck (changes requested):
Huh, this is _so_ weird. Good find. Some suggestions about keys (I think we can sweep in some more keys) and the description. Thank you!

ddbeck:
> Cainuse has all green however. If we merge this feature as baseline false, I'll go file an issue there about it.

The caniuse entries are just generated from BCD, so only https://caniuse.com/mdn-api_barprop_visible_returns_popup shows the Safari difference. This is normal; no action required.

ddbeck (approved):
Thank you!

PR #1891, opened 2024-10-02
Shorthand and longhand styleguide entry

jamesnw:
Based on conversation in https://github.com/web-platform-dx/web-features/pull/1764#discussion_r1777335770


docs/guidelines.md
```diff
@@ -195,6 +195,13 @@ Omit "is used" where there's no loss in meaning.
 For example, prefer "The feature reads…" over "The feature is used to read…"
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635981))
 
+#### longhands and shorthands
+
+In CSS, use the terms "shorthand" and "longhand" to describe the relationship
+between CSS properties that combine multiple properties into a single
+declaration and the individual properties.
```
ddbeck, line 202: Contrasting with what not to do would be cool, here I think.

Also, I'd been using [sembr](https://sembr.org/) in this file, or at least one-sentence per line. I don't feel strongly on that point though.

One thing is left unanswered: should I use "longhand properties" or "longhands". I think I prefer the former, but I'd like to know whether I am alone in this. 😅 

```suggestion
Prefer the terms "shorthand" and "longhand" to describe the relationship between CSS properties that combine multiple properties into a single declaration and the individual properties.

Avoid the phrase "constituent properties" for longhand properties, even though this is common on MDN Web Docs.
```

jamesnw: > One thing is left unanswered: should I use "longhand properties" or "longhands". I think I prefer the former, but I'd like to know whether I am alone in this. 😅

I added a bit of clarification here-

Avoid using these terms without the word "property", so prefer "The `text-wrap` CSS property is a shorthand" over "The `text-wrap` CSS shorthand", and "It is a longhand property of..." over "It is a longhand of".

I think what matters is that shorthand and longhand don't replace the word "property" but rather clarify the type of property, but we don't need to say "*hand property" in that order, necessarily.

ddbeck (approved):
I really like this version! I'm going to request reviews from the active folks around here, so it gets some extra visibility ahead of merging.

Elchi3 (approved):

docs/guidelines.md
```diff
@@ -195,6 +195,15 @@ Omit "is used" where there's no loss in meaning.
 For example, prefer "The feature reads…" over "The feature is used to read…"
 ([#727](https://github.com/web-platform-dx/web-features/pull/727#discussion_r1537635981))
 
+#### longhands and shorthands
+
+Prefer the terms "shorthand" and "longhand" to describe the relationship between CSS properties that combine multiple properties into a single declaration and the individual properties. 
+
+Avoid using these terms without the word "property", so prefer "The `text-wrap` CSS property is a shorthand" over "The `text-wrap` CSS shorthand", and "It is a longhand property of..." over "It is a longhand of".
```
captainbrosset, line 202: Some formatting helps parse this sentence:

```suggestion
Avoid using "shorthand" and "longhand" without the word "property":

* Prefer "The `text-wrap` CSS property is a shorthand" over "The `text-wrap` CSS shorthand".
* Prefer "It is a longhand property of" over "It is a longhand of".
```

captainbrosset (approved):

autonome (approved):

PR #1892, opened 2024-10-02
Adds new overflow properties. Adds keys to overflow and logical properties

vwallen:


features/overflow-clip-margin.yml
```diff
@@ -0,0 +1,9 @@
+name: Overflow clip margin
```
ddbeck, line 1: If we're reporting on just one property and there's no well-known concept its referring to, the feature name is probably the property itself (this is specific to CSS and is more art than science, sorry).

```suggestion
name: overflow-clip-margin
```

features/overflow-clip-margin.yml
```diff
@@ -0,0 +1,9 @@
...
+description: "The `overflow-clip-margin` CSS property sets how far overflow content will display outside the bounds of an element when clipped by effects such as `overflow: clip`."
```
ddbeck, line 2: Obscure style nit: avoid the future tense

```suggestion
description: "The `overflow-clip-margin` CSS property sets how far overflow content may appear outside the bounds of an element before it's clipped by effects such as `overflow: clip`."
```

features/scroll-behavior.yml
```diff
@@ -0,0 +1,9 @@
+name: scroll-behavior
+description: The `scroll-behavior` CSS property allows smooth scrolling for scroll actions not performed by the user, such as those triggered by navigation.
```
ddbeck, line 2: This definition works, but only if we were covering a specific declaration (`scroll-behavior: smooth`). Since this feature is a little more general, we ought to generalize to the full range of things it can do:

See also: https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#allows

```suggestion
description: The `scroll-behavior` CSS property controls whether scrolling is smooth or snaps, for scroll actions not performed by the user such as those triggered by navigation.
```

features/text-overflow.yml
```diff
@@ -0,0 +1,10 @@
+name: Text overflow
+description: The `text-overflow` CSS property sets how hidden overflow content is signalled to users. It can be clipped, display an ellipsis (…), or display a custom string.
```
ddbeck, line 2: "is signaled to" is a bit indirect, so I'm suggesting "appears to" instead. Also "it" was a unclear to me (the property or content?) and I attempted to resolve the ambiguity.

```suggestion
description: The `text-overflow` CSS property sets how hidden overflow content appears to users. The property can clip content, truncate content with an ellipsis (…), or truncate with a custom string.
```

ddbeck (changes requested):
This all really nice. All I have is a few comments on the descriptions.

ddbeck (approved):

PR #1895, opened 2024-10-02
Add textwrap group and feature

jamesnw:
There are some odd parts to this-

`text-wrap-style: balance` has less support than `text-wrap: balance`.

`text-wrap: wrap` is in the `text-wrap: nowrap` feature, and `text-wrap-mode: nowrap` is in the `text-wrap-mode` feature. Maybe those should get sorted differently?


jamesnw:
> Looks like is missing some dist files?

No- the files without dists in the diff are ones where I just changed the groups, so there were no changes to the dist.

features/text-wrap-stable.yml
```diff
@@ -0,0 +1,7 @@
+name: "text-wrap: stable"
+description: "The `text-wrap: stable` CSS declaration wraps text as normal, but does not change wrapping on earlier lines while the text is edited."
```
autonome, line 2: - "as normal" - would be more correct to use "default" instead of "normal"
- "is edited" -  reads like missing a word, does it mean while the text is *being* edited?

autonome:
> `text-wrap-style: balance` has less support than `text-wrap: balance`.

looks like doesn't impact status though?

> `text-wrap: wrap` is in the `text-wrap: nowrap` feature, 

Yeah, this is counterintuitive. Seems like `wrap` would be a better feature, with `nowrap` in it, if we had to choose.

> and `text-wrap-mode: nowrap` is in the `text-wrap-mode` feature. Maybe those should get sorted differently?

CSS seems full of these mode switchers and shorthands... if having the wrap `mode` standalone w/ variable values aligns w/ how you've been handling these so far, then this seems fine?

jamesnw:
> > `text-wrap-style: balance` has less support than `text-wrap: balance`.
> 
> looks like doesn't impact status though?
> 
> >

It does- `text-wrap-style: balance` is `baseline: false`, since it doesn't have Chrome support.


I ended up re-sorting the keys, and created a `text-wrap-style` feature, and made it more of a support-based feature set.

PR #1897, opened 2024-10-03
Add beforeunload

autonome:
~~Hrm, I just saw the deprecated keys conversation in chat, and it seems like this feature should've given a warning or failure? Maybe I didn't see it, if just a warning.~~

~~Here we've got a non-deprecated interface where the implementation key is deprecated.~~

~~MDN doesn't mention deprecation, just recommends using other approaches.~~

~~Caniuse highlights the deprecation in the window implementation, but not for the interface.~~

~~Ok, just a sub-key is deprecated. Neither the interface or implementation are.~~

~~Will include it, commented out, with an explanation.~~

Edit again again:

- added 4 subkeys from BCD
- 3 are deprecated
- 1 is not, but makes the feature baseline false b/c Safari iOS

And I think that's seems right. Using this feature in the modern way is not baseline at this time.

Having a better explicit way to park deprecations would be good to figure out.


jamesnw:
> Using this feature in the modern way is not baseline at this time.

I agree with this assessment. This might be a good candidate for Interop... 🤔

features/beforeunload.yml
```diff
@@ -0,0 +1,11 @@
+name: Beforeunload
+description: "The `beforeunload` event is fired when the current window is about to be unloaded. Typically this is done to display a confirmation dialog that asks users to confirm if they really want to leave the page when they try to close or reload it, or navigate somewhere else."
```
jamesnw, line 2:
```suggestion
description: "The `beforeunload` event is fired when the current window is about to be unloaded. Typically this is used to display a dialog to confirm if users really want to leave the page when there is unsaved data that would be lost."
```

A bit of tightening, but also mentioning data loss as it shouldn't be used just to prevent users from leaving.

jamesnw (approved):

autonome:
Thanks for review @jamesnw. Merging, but also poking @ddbeck for his thoughts on how this handled the deprecated keys. Having them in-place adjacent to their fellow keys is good for visibility for future WF contributors. I'd prefer something explicit that allows programmatic checks, eg a new YAML top level `deprecated` section or something like that. But seems like a comment is good enough for now.

PR #1898, opened 2024-10-03
Adds ruby annotation layout properties and group

vwallen:


jamesnw (approved):
LGTM!

autonome (changes requested):
Can you remove these keys from `features/draft/spec/css-ruby-1.yml`? Hmm, unless y'all haven't been doing that for CSS keys?

jamesnw:
> Can you remove these keys from `features/draft/spec/css-ruby-1.yml`? Hmm, unless y'all haven't been doing that for CSS keys?

We haven't been removing them- for some reason I thought we were avoiding that to avoid merge conflicts. I'm fine either way?

autonome (approved):

autonome:
> > Can you remove these keys from `features/draft/spec/css-ruby-1.yml`? Hmm, unless y'all haven't been doing that for CSS keys?
> 
> We haven't been removing them- for some reason I thought we were avoiding that to avoid merge conflicts. I'm fine either way?

Never mind in that case! I have been, likely because early on was told to do so. But since nothing actually tests the aggregate set of keys in the drafts folders against the completed keys, it doesn't really matter. Worst that happens is someone picks out keys in drafts to work on, and they've already been done. But I typically check keys first before starting work on anything, regardless.

PR #1899, opened 2024-10-03
Add speak and speak-as

jamesnw:
`speak-as` is a weird property. What has wide-ish support and any sort of [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style/speak-as) is `speak-as` inside an `@counter-style` rule to describe how markers in a list should be read, and that is part of the [@counter-style](https://web-platform-dx.github.io/web-features-explorer/features/counter-style/) feature.

This feature is NOT that. This feature applies to any element, and confusingly has a different set of valid properties. Should I disambiguate here somehow?

@autonome - I added this to the `speech` group, but the name of that group is "Web Speech API". Could that group be broadened to include these hints for the speech synthesis? I think just renaming it "Speech" would suffice.


autonome:
> This feature is NOT that. This feature applies to any element, and confusingly has a different set of valid properties. Should I disambiguate here somehow?

If this feature is truly distinct from the `speak-as` inside a `@counter-style`, then is worth adding a comment in the yml file. I haven't seen us making this kind of disambiguation inside descriptions, which aim to be standalone.

> @autonome - I added this to the `speech` group, but the name of that group is "Web Speech API". Could that group be broadened to include these hints for the speech synthesis? I think just renaming it "Speech" would suffice.

Sounds good to me. I am all for making groups generally useful as a discovery mechanism, whether for WF contributors or developers generally, and this would achieve that for either audience.

autonome (approved):

features/speak-as.yml
```diff
@@ -0,0 +1,14 @@
+name: speak-as
+description: The `speak-as` CSS property sets how any element's content is spoken.
+# Note: This feature is distinct from `speak-as` in a @counter-style rule.
```
ddbeck, line 3: Let's tell developers directly that we too can be confused by this. 😅 

```suggestion
description: The `speak-as` CSS property sets how any element's content is spoken. Not to be confused with the `speak-as` descriptor of `@counter-style` at-rules.
```

ddbeck (approved):
A suggestion for the description, to handle the point of confusion you noted. Free to merge with or without the suggestion though.

PR #1902, opened 2024-10-04
Add History API

autonome:


features/history.yml.dist
```diff
@@ -0,0 +1,73 @@
...
+  # baseline_high_date: 2018-01-29
+  # support:
+  #   chrome: "5"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "4"
+  #   firefox_android: "4"
+  #   safari: "5"
+  #   safari_ios: "4"
+  - api.History.pushState
```
captainbrosset, line 45: This is the support data that matches caniuse.com. So, it seems like everything up to this point is really necessary to use the API, but the 2 following keys are nice-to-haves (`state` and `scrollRestoration`). Reading their documentation, it does seem like you can use the API without them.
So I would be tempted to use `compute_from` on `api.History.pushState` to match caniuse.com.

features/history.yml
```diff
@@ -0,0 +1,14 @@
+name: History
+description: "The History API allows manipulation of the browser session history, from navigations to state management, in the tab or frame that the current page is loaded in."
```
captainbrosset, line 2:
```suggestion
description: "The History API manipulates the browser session history, from navigations to state management, in the tab or frame that the current page is loaded in."
```

captainbrosset (changes requested):

jamesnw (approved):
Approved pending changes from @captainbrosset

features/history.yml
autonome: the curse of the passive voice!

autonome:
Implemented @captainbrosset's changes, and dismissing review for merge. Github review flow isn't great for the "r+ w/ changes addressed" approach.

PR #1904, opened 2024-10-04
Add ImageData

autonome:


features/imagedata.yml
```diff
@@ -0,0 +1,15 @@
...
+# baseline high for a long time.
+status:
+  compute_from: api.ImageData
+compat_features:
+  - api.ImageData
+  - api.ImageData.ImageData
+  - api.ImageData.colorSpace
+  - api.ImageData.data
+  - api.ImageData.height
+  - api.ImageData.width
```
ddbeck, line 15: What would it look like if we parked all of this in `canvas-2d.yml` instead? It seems like this is the oldest, most important place where one would get or use an `ImageData` object.

ddbeck (approved):

ddbeck:
@autonome This needs a merge conflict resolved, then we can merge it. Thank you!

PR #1905, opened 2024-10-07
Add Location

autonome:


features/location.yml
```diff
@@ -0,0 +1,22 @@
+name: Location
+description: "The `Location` global object represents the current address of the web page loaded in the browser. It makes all component parts of the address accessible and can be manipulated to initiate navigations."
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-location-interface
```
ddbeck, line 3: This is… the most intense warning I've ever seen in a web spec. 😅

features/location.yml
```diff
@@ -0,0 +1,22 @@
...
+# `ancestorOrigins` makes this baseline false but is not required for usage of the feature.
+status:
+  compute_from: api.Location
+compat_features:
```
ddbeck, line 7: What would this look like with `api.Window.location` and `api.Document.location`? I _think_ those are going to be the most interesting uses of the `Location` API and the impetus for this API to even exist.

features/location.yml
autonome: I know, right?!

features/location.yml
autonome: Argh, I forgot the actual implementation. Again. Adding a full search on feature strings and substrings to my pre-commit checklist.

autonome: Ok baseline remains same

features/location.yml
```diff
@@ -0,0 +1,24 @@
+name: Location
+description: "The `Location` global object represents the current address of the web page loaded in the browser. It makes all component parts of the address accessible and can be manipulated to initiate navigations."
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-location-interface
+# `ancestorOrigins` makes this baseline false but is not required for usage of the feature.
+status:
+  compute_from: api.Location
```
ddbeck, line 6: OK, now that it works with `window.location`, I think we ought to make a few changes here:

- Use `location` as the entry point in the description (though making this explicitly `window.location` would also be fine) and status computation
- Make the description a little more direct and active ("get the parts" versus "mak[ing] accessible", "navigate" versus initiate navigations).
- Include the term "URL" for findability

```suggestion
description: The `location` global object represents the current page's address. You can use it to get the parts of the address (such as `location.hostname` or `location.pathname`) or navigate to another URL.
spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-location-interface
# `ancestorOrigins` makes this baseline false but is not required for usage of the feature.
status:
  compute_from: api.Window.location
```

ddbeck (changes requested):

ddbeck (approved):
🎉

PR #1906, opened 2024-10-07
Add Domparser

autonome:
- Caniuse groups `DOMParser`, `XMLSerializer` and other misc bits into a single feature: https://caniuse.com/xml-serializer
- We already have a feature for `XMLSerializer` from #1444

This PR:
- adds `DOMParser` as a standalone feature
- adds a `Parsing and Serialization` group with both features

If we merged them, it looks like it wouldn't change baseline, but I prefer this explicit approach over Caniuse's kitchen sink approach because developer searches can match on the API better. For example, the Caniuse feature was at the bottom of their own search results for the API name.


groups/parsing-and-serialization.yml
```diff
@@ -0,0 +1,3 @@
+# Caniuse lumps all parsing/serialization of html and xml into one feature.
+# For Web Features, they are separate features in a group.
+name: Parsing and Serialization
```
captainbrosset, line 3:
```suggestion
name: Parsing and serialization
```

captainbrosset (approved):

PR #1908, opened 2024-10-07
Add HashChangeEvent

autonome:
I matched the name with MDN and the spec, instead of `Hashchange event` which Caniuse has, but is a hybrid of readable language and the spec correctness that doesn't actually align fully with either.


features/hashchange.yml
```diff
@@ -0,0 +1,9 @@
+name: HashChangeEvent
```
ddbeck, line 1: We have a few other event features (e.g., scrollend) and we've so far used the event name (as in `window.addEventListener("hashchange", () => …`), not the interface. I think this would be a good fit here too:

```suggestion
name: hashchange
```

features/hashchange.yml
```diff
@@ -0,0 +1,9 @@
...
+description: "The `HashChangeEvent` is emitted when the fragment identifier of the current page URL has changed."
```
ddbeck, line 2: Mostly for consistency with other event feature descriptions, but also to explain to folks who don't remember what fragment refers to:

```suggestion
description: "The `hashchange` event fires when the URL fragment identifier (the part of the URL starting with `#`) of the current page has changed."
```

features/hashchange.yml
```diff
@@ -0,0 +1,9 @@
...
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-hashchangeevent-interface
+caniuse: hashchange
+compat_features:
+  - api.HashChangeEvent
+  - api.HashChangeEvent.HashChangeEvent
+  - api.HashChangeEvent.newURL
+  - api.HashChangeEvent.oldURL
```
ddbeck, line 9:
```suggestion
  - api.HashChangeEvent.oldURL
  - api.Window.hashchange_event
```

features/hashchange.yml
autonome: Aghgh I went back and forth on this one

ddbeck (approved):

PR #1913, opened 2024-10-07
Add messaging APIs and their infra

autonome:
Everything in this area is grouped in the `Communications` section of the HTML spec. I added a group called `Messaging` instead of going the spec route, since it aligns more with how developers refer to these features.

I joined `MessageChannel` and `MessagePort` into `Channel Messaging`, aligning with how the spec, MDN and Caniuse do it.

`MessageEvent` is implemented in numerous places, but doesn't seem like a standalone feature... but `messageerror` was already done, so I followed that precedent here, adding the various implementation keys as well.

A couple of notes on `postMessage`:
- Name: Caniuse calls it `Cross-document messaging` - which is how the spec describes it too.  I dug around and few references by developers to `Cross-document messaging`, but a zillion to `postMessage`, so I'm erring on the side of usage-in-the-wild. Sometimes tutorials just called it `Web Messaging` also, but they pre-dated all the newer messaging stuff.
- Composition: I pulled in the implementations and put them in a feature together. There's some trade-off against discoverability of the API in the implementation itself. Really a limitation of the "one key, one home" approach in situations like this.

This whole feature group has this same challenge: Do you put the key in the implementation (eg, putting `postMessage` in each of `Window`, `ServiceWorker`, etc) or put the implementation keys in the feature. Might be worth a deeper discussion with the group. I'll bring this up at the meeting this week.

Features previously done in this area:
- `BroadcastChannel`
- `messageerror`

Features added here:
- `Channel Messaging` - includes `MessageChannel` and `MessagePort`
- `postMessage`
- `EventSource`
- `MessageEvent`

Group added here:
- `Messaging`


features/messageevent.yml
```diff
@@ -0,0 +1,24 @@
...
+  # Revisit once the "keys in impl or feature" question is answered.
+  #- api.BroadcastChannel.message_event
+  - api.DedicatedWorkerGlobalScope.message_event
+  - api.MessageEvent
+  - api.MessageEvent.MessageEvent
+  - api.MessageEvent.data
+  - api.MessageEvent.lastEventId
+  - api.MessageEvent.origin
+  - api.MessageEvent.ports
+  - api.MessageEvent.source
```
ddbeck, line 22: I don't think `MessageEvent` is really a feature. Ideally, we'd site this with _every_ message feature that uses. We can't do that yet, so until then, I think this should live with the oldest feature that uses it. That would be the `window.postMessage` feature below.

features/messageevent.yml
```diff
@@ -0,0 +1,24 @@
...
+  # Already generated in Broadcast channel.
+  # Revisit once the "keys in impl or feature" question is answered.
+  #- api.BroadcastChannel.message_event
+  - api.DedicatedWorkerGlobalScope.message_event
+  - api.MessageEvent
+  - api.MessageEvent.MessageEvent
+  - api.MessageEvent.data
+  - api.MessageEvent.lastEventId
+  - api.MessageEvent.origin
+  - api.MessageEvent.ports
```
ddbeck, line 21: I think the `ports` property probably belongs with channel messaging

features/postmessage.yml
```diff
@@ -0,0 +1,16 @@
+name: PostMessage
```
ddbeck, line 1: I don't see this name being widely used. I think it's actually called this:

```suggestion
name: Cross-document messaging
```

(I think the file name should change to match, too.)

ddbeck (changes requested):

features/eventsource.yml
```diff
@@ -0,0 +1,17 @@
+name: Server-sent events
+description: "The `EventSource` API creates a persistent one-way connection to the originating server for receiving streams of events."
```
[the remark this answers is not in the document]
autonome: It's important that developers understand that it's 1) a persistent and 2) one-way connection and that 3) it's initiated by the client.

How's this remix:

> description: "The `EventSource` API creates a persistent connection which listens to a stream of events from a server.

features/postmessage.yml
autonome: I noted this in the PR comment - I found that mostly just documentors and implementers call it `Cross-document messaging`.

In my experience, developers even *verb* `postMessage`, which I kinda love.

Happy to disagree and commit if you think we should go with the spec name, can always iterate later if we decide it's not working.

features/messageevent.yml
autonome: See note on this at top. Discussed at standup. Lmk if you think we should keep the `messageerror` route as I've done here, or fold into `postMessage`, until we have compound features.

features/server-sent-events.yml
```diff
@@ -0,0 +1,17 @@
+name: Server-sent events
+description: "The `EventSource` API creates a persistent one-way connection to the originating server for receiving streams of events."
```
ddbeck, line 2: Continuing from https://github.com/web-platform-dx/web-features/pull/1913#discussion_r1792707456

```suggestion
description: The `EventSource` API creates a connection to a server and listens to a stream of events sent by the server.
```

features/postmessage.yml
```diff
@@ -0,0 +1,16 @@
...
+description: "The `postMessage()` method safely enables cross-origin communication for windows and workers, including popups and iframes."
```
ddbeck, line 2: Suggestions here: putting this into a more active voice (see: [enables](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#enables)) and adding an _also known as_.

Also (_gravelly action move voice_), no one is safe.

```suggestion
description: "The `postMessage()` global method sends cross-origin messages to windows and workers, including popups and iframes. Also known as cross-document messaging."
```

features/postmessage.yml
ddbeck: OK, I'm willing to go along here, but a few things need to happen:

- Link to an example of a developer verbing of `postMessage` — I would like to have an example, I think it'd make a good illustration for the docs.
- Set the case of this name to the case of the API

  ```suggestion
  name: postMessage
  ```

- _Also known as_ the "cross-document messaging" (see suggestion below).

features/postmessage.yml
```diff
@@ -0,0 +1,16 @@
...
+status:
+  compute_from: api.Window.postMessage
+compat_features:
+  # Already generated in Broadcast channel.
+  # Revisit once the "keys in impl or feature" question is answered.
+  #- api.BroadcastChannel.postMessage
+  - api.DedicatedWorkerGlobalScope.postMessage
+  - api.MessagePort.postMessage
+  - api.Window.postMessage
+  - api.Worker.postMessage
```
ddbeck, line 16: I think the `BroadcastChannel` and `MessagePort` equivalents are rather distinct, albeit confusingly named. `BroadcastChannel` is explicitly same origin. `MessagePort`'s `postMessage()` is explicitly non-global. I think I'd take this approach instead:

- Move `api.MessagePort.postMessage` to `channel-messaging.yml`
- Change the comment to say:

```suggestion
  # Other features, such as channel messaging and broadcast channels have
  # `postMessage` methods too, but are not covered by this feature.
  - api.DedicatedWorkerGlobalScope.postMessage
  - api.Window.postMessage
  - api.Worker.postMessage
```

features/messageevent.yml
```diff
@@ -0,0 +1,17 @@
...
+  #- api.BroadcastChannel.message_event
+  - api.DedicatedWorkerGlobalScope.message_event
+  - api.MessageEvent
+  - api.MessageEvent.MessageEvent
+  - api.MessageEvent.data
+  - api.MessageEvent.lastEventId
+  - api.MessageEvent.origin
+  - api.MessageEvent.source
+  - api.Window.message_event
+  - api.Worker.message_event
```
ddbeck, line 17: Continuing from https://github.com/web-platform-dx/web-features/pull/1913#discussion_r1792714432.

OK, thank you for bringing this up at our standup.

> `MessageEvent` is implemented in numerous places, but doesn't seem like a standalone feature... but `messageerror` was already done, so I followed that precedent here, adding the various implementation keys as well.

I think it's a bad precedent and we shouldn't honor it. 😅 

Truthfully, I think there's a little something to `messageerror` being a feature, but it's definitely a transitional feature. Someday, those keys ought to fold into their parent interfaces and they'll be an unremarked upon convention around message-sending APIs. This is where fixing https://github.com/web-platform-dx/web-features/issues/1173 would be helpful.

I think the actual task here is to move the keys to `postmessage.yml`.

ddbeck (changes requested):

features/eventsource.yml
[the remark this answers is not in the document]
ddbeck: Continued at https://github.com/web-platform-dx/web-features/pull/1913#discussion_r1793415198

features/messageevent.yml
ddbeck: Continued at https://github.com/web-platform-dx/web-features/pull/1913#discussion_r1793469067

features/postmessage.yml
autonome: https://www.startpage.com/sp/search?query=%22postmessaging%22

Sometimes it's like "postmessaging between iframes", things like that.

> Check out the MDN web documentation on [iframes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) and [postmessaging](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to familiarize yourself with these concepts.

from https://cloud.google.com/looker/docs/embed-getting-started?hl=en

features/postmessage.yml
```diff
@@ -1,16 +1,27 @@
...
+  # Other features, such as channel messaging and broadcast channels have
+  # `postMessage` methods too, but are not covered by this feature.
   - api.DedicatedWorkerGlobalScope.postMessage
-  - api.MessagePort.postMessage
   - api.Window.postMessage
   - api.Worker.postMessage
+
+  # Already generated in Broadcast channel.
+  # Revisit once the "keys in impl or feature" question is answered.
+  #- api.BroadcastChannel.message_event
```
ddbeck, line 18: This comment is unclear, more so now that we have the context-setting comment above. I think we can drop the whole thing.

```suggestion
```

ddbeck (changes requested):
Very close now. Two small fixes to make and then we can land this. Thank you!

features/postmessage.yml
autonome: yeh, already had done locally

ddbeck (approved):

PR #1914, opened 2024-10-07
Adds view transitions features for cross-document transitions and transition selection, and adds a new group

vwallen:
New Features:
- Active view transition (selectors)
- Cross-document view transitions
- View transition class

Updated Features:
- View transitions

Groups:
- View Transitions


autonome (changes requested):
Should make the new group a child of the CSS group.

ddbeck (approved):
I think this makes sense overall, but I do have one question (definitely not knowledgeable enough to make a suggestion here): why are the pseudos and property separate features, instead of late additions to the "original" view transitions feature?

(I don't think view-transitions group needs to be a child of CSS—if I understand correctly, the API is required for use.)

vwallen:
@ddbeck the rationales for the splits were
- The at-rule, view-class, and pseudos are part of a different spec (level 2 vs 1) than the original and have a different and incompatible support story from the original (and each other). All things being equal this could also have been handled with `compute_from`
- The at-rule and related keys are strictly cross-document which feels right as a separate feature even though it uses common mechanisms otherwise
- The classes work in a way unusual and distinct enough (added CSS classes to a non-DOM entity) that a separate feature seemed warranted

I'm not strongly committed to this division. The first commit in this branch actually included everything new in the original feature if you want to see that in context: 4f59b14

ddbeck:
OK, cool. I'm happy with this. Let's merge this.

PR #1916, opened 2024-10-07
Add Length Units

SondraE:


autonome (commented):
Did a pass for the basic feature bits like files, names, descriptions etc and all looks good to me. But it's a lot of features and areas I'm not super familiar with the area, nor how Caniuse addresses units, so will leave for James or Daniel to do another pass.

jamesnw (approved):

PR #1919, opened 2024-10-07
Adds CSS transitions feature

vwallen:


autonome (changes requested):
James is better for review of feature composition here, but early feedback on the group is that should remove `-css` and instead make this a child of the `css` group, see example: https://github.com/web-platform-dx/web-features/blob/main/groups/transforms.yml

features/transitions-css.yml
ddbeck, line 1: I think this is ready to land with a rename from `transitions-css.yml` to `transitions.yml` (and the refresh based on that)

jamesnw (approved):
🚀

PR #1927, opened 2024-10-08
SVG core feature

jamesnw:
This adds an SVG catchall. It started similar to the JavaScript initial support feature- https://github.com/web-platform-dx/web-features/pull/1670, but ended up being most of the feature set, as SVG has not had much active implementation recently. The spec leads the implementation significantly, which has led to lots of partial implementation.

I also added well as a `context-fill-stroke` for the only newish feature that seemed to warrant a feature. 

I also added a draft with ~175 SVG-related keys that are deprecated.


autonome:
Do you plan break this up into smaller features or leave as one big one?

jamesnw:
> Do you plan break this up into smaller features or leave as one big one?

My plan is to leave as one big one, and as the need arises, break off smaller pieces. For instance, if SVG starts getting attention from implementers and new features start getting added, we may want to split things out.

ddbeck:
I couldn't figure out why there was no headline status in `svg.dist.yml` (no comment with `⬇️`). For some reason, this feature is not producing any status at all and is _also_ not failing the build (this should fail the schema validation). I'm investigating.

jamesnw:
> I couldn't figure out why there was no headline status in `svg.dist.yml` (no comment with `⬇️`). For some reason, this feature is not producing any status at all and is _also_ not failing the build (this should fail the schema validation). I'm investigating.

I had `status.compat_features` set instead of `status.compute_from`.

features/svg.yml
```diff
@@ -0,0 +1,901 @@
...
+  - svg.elements.feSpotLight.y
+  - svg.elements.feSpotLight.z
+  - svg.elements.feTile
+  - svg.elements.feTile.in
+  - svg.elements.feTurbulence
+  - svg.elements.feTurbulence.baseFrequency
+  - svg.elements.feTurbulence.numOctaves
+  - svg.elements.feTurbulence.seed
+  - svg.elements.feTurbulence.stitchTiles
+  - svg.elements.feTurbulence.type
```
ddbeck, line 631: Caniuse has a feature SVG filters https://caniuse.com/svg-filters (these lines plus the corresponding API stuff above).

It's a fairly old feature on caniuse (it was added no later than 2012), so if you think it's sufficiently historic to ignore for now that's fine, but I wanted to raise the possibility now, just in case.

features/svg.yml
```diff
@@ -0,0 +1,901 @@
+name: SVG
+description: The SVG image format creates two-dimensional vector graphics with declarative or scripted interaction and animation.
```
ddbeck, line 2:
```suggestion
description: The SVG image format, represented by the `<svg>` element, creates two-dimensional vector graphics with declarative or scripted interaction and animation.
```

ddbeck (approved):
Two suggestions here. Provisionally approved, though another round of review might be needed depending on your take on my question about a caniuse correspondence.

features/svg.yml
jamesnw: I like this idea, and seeing as it comprises 1/3rd of the keys, it seems like a reasonable split. e

ddbeck (approved):
Let's go! 🚀

PR #1941, opened 2024-10-10
Add additional keys to `paint`

jamesnw:


autonome (approved):

features/paint.yml
```diff
@@ -3,3 +3,56 @@ description: The `paint()` CSS function creates a custom image, drawn using a pa
 spec: https://drafts.css-houdini.org/css-paint-api-1/
 group: css
 caniuse: css-paint-api
+status:
+  compute_from: css.types.image.paint
+compat_features:
+  - api.CSS.paintWorklet_static
+  - api.PaintWorkletGlobalScope
+  - css.types.image.paint
+  - css.types.image.paint.additional_parameters
```
captainbrosset, line 12: Maybe add a comment here saying something like:
"At the time of writing, additional parameters are not supported anywhere. When support is added, reflect on whether this should be split into a separate feature or not".

I feel like this might unlock specific developer use cases.

PR #1942, opened 2024-10-10
Update scroll-snap keys

jamesnw:
Also moves from explicit status to use compute_from. No change in status, just in how it's derived.


autonome (changes requested):
Should pull in the other pieces like events from `features/draft/spec/css-scroll-snap-2.yml`.

jamesnw:
> Should pull in the other pieces like events from `features/draft/spec/css-scroll-snap-2.yml`.

@autonome Those are experimental and only in Chromium, so they'll have a different support story. I'm planning on making that a separate feature.

autonome (approved):

PR #1945, opened 2024-10-11
Add drag and drop

autonome:


features/file-system-access.yml
```diff
@@ -4,8 +4,6 @@ spec: https://wicg.github.io/file-system-access/
 caniuse: native-filesystem-api
 group: file-system
 compat_features:
-  # api.DataTransferItem.getAsFileSystemHandle is likely to be part of a drag and drop feature, but it can't be in both features at once. See https://github.com/web-platform-dx/web-features/issues/1173 for details
```
jamesnw, line 7: I agree it makes more sense in Drag and drop, but should we move this comment there?

features/draganddrop.yml
```diff
@@ -0,0 +1,41 @@
+name: Drag and Drop
+description: "The Drag and Drop API lets web applications add dragging and dropping of elements and external files such as images onto web pages. Developers can customize which elements can become draggable, the type of feedback the draggable elements produce, and the droppable elements."
```
jamesnw, line 2:
```suggestion
description: "The Drag and Drop API lets users drag and drop elements and external files such as images onto web pages. Developers can customize which elements can become draggable, the type of feedback the draggable elements produce, and the droppable elements."
```

I don't think we can or need to try to escape the circular definition here. I noted MDN uses `drag-and-drop`, but that doesn't seem to be the preferred term per Google Trends. https://trends.google.com/trends/explore?date=today%205-y&q=drag-and-drop,drag%20and%20drop&hl=en

jamesnw (approved):
The feature set looks good to me!

features/file-system-access.yml
autonome: Good call, I moved it over and reworded.

PR #1949, opened 2024-10-11
Add Xpath, XSLT and an xml group

autonome:


autonome:
Could add `XMLHTTPRequest` to the group if anyone thinks it should be there.

Need to find a home for `XMLDocument`, see comment in the XPath feature.

dontcallmedom:
-1 to add XMLHttpRequest - despite its name, it has nothing to do with XML

features/xpath.yml
```diff
@@ -0,0 +1,25 @@
+name: XPath
+description: "XML Path Language (`XPath`) uses a syntax akin to file paths to provide a flexible way of pointing to or selecting different parts of an XML document. It can also be used to test addressed nodes within a document to determine whether they match a pattern or not."
```
ddbeck, line 2: So the platform feature here is the API used to execute XPath queries, not so much the language itself (XPath in the browsers is relatively antiquated with respect to XPath as it is specified today). I think this might be a good way to describe it (kinda like we do for [JSON](https://github.com/web-platform-dx/web-features/blob/45bbdebd61a32ec8242bdbe6eb26b14dbb9704a3/features/json.yml)):

```suggestion
description: "The `document.evaluate()` method selects elements in an HTML or XML document based on an expression of XPath, a domain specific language for querying XML documents. Also known as XML Path Language."
```

features/xpath.yml
```diff
@@ -0,0 +1,25 @@
...
+spec: https://dom.spec.whatwg.org/#xpath
+group: xml
+compat_features:
+  - api.Document.evaluate
+  # Doesn't really fit well here, maybe should stand alone,
+  # or be part of DOM?
+  #- api.XMLDocument
```
ddbeck, line 9: I agree, this should be part of the core DOM feature(s). I _think_ you get an `XMLDocument` if you have an XHTML doctype, so it's still "just" a document for the purposes of DOM.

features/xslt.yml
```diff
@@ -0,0 +1,15 @@
+name: XSLT
+description: "Extensible Stylesheet Language Transformations (`XSLT`) is an XML-based language used for the transformation of XML documents into new distinct XML documents. XSLT is often used to convert data between different XML schemas or to convert XML data into web pages or PDF documents."
```
ddbeck, line 2: Like JSON and XPath:

```suggestion
description: The `XSLTProcessor` API transforms XML documents into new XML or HTML documents, using XSLT stylesheets. You can use XSLT to convert data between different XML schemas or to convert XML data into web pages or PDF documents. Also known as Extensible Stylesheet Language Transformations.
```

groups/xml.yml
```diff
@@ -0,0 +1,3 @@
+# Group comprised of XML features such as parsing/serializing, querying and more.
+# It does not include XMLHTTPRequest, since it's not specific to XML.
```
ddbeck, line 2: Good shout on XMLHTTPRequest here.

```suggestion
# Group comprises XML features such as parsing/serializing, querying and more.
# It does not include XMLHTTPRequest, since it's not specific to XML.
```

features/xpath.yml
```diff
@@ -0,0 +1,25 @@
+name: XPath
+description: "XML Path Language (`XPath`) uses a syntax akin to file paths to provide a flexible way of pointing to or selecting different parts of an XML document. It can also be used to test addressed nodes within a document to determine whether they match a pattern or not."
+spec: https://dom.spec.whatwg.org/#xpath
+group: xml
```
ddbeck, line 4:
```suggestion
group: xml
caniuse: document-evaluate-xpath
```

ddbeck (changes requested):
This is good. Some little edits here, to tidy descriptions/comments here and link to a caniuse feature, but otherwise looking good.

ddbeck (approved):

PR #1952, opened 2024-10-11
Add bidirectional-override

jamesnw:


features/bidirectional-override.yml
```diff
@@ -0,0 +1,18 @@
+name: Bidirectional override
```
autonome, line 1: The title seems ambiguous, maybe should include some combo of `Unicode` or `layout` or `direction` or something like that, which would disambiguate from people thinking it's the `bdo` tag (separate but related feature).

features/bidirectional-override.yml
```diff
@@ -0,0 +1,18 @@
...
+description: The `unicode-bidi` and `direction` CSS properties override the Unicode layout algorithm in non-HTML documents. Authors should use the `dir` global HTML attribute instead.
```
autonome, line 2: "in non-HTML documents" doesn't seem right? Do they only apply in that case?

autonome (commented):
Some questions

features/bidirectional-override.yml
jamesnw: Good point, I renamed "Layout direction override"

features/bidirectional-override.yml
jamesnw: They are applied in HTML documents, but use is discouraged. Stylesheets could be disabled, so text direction should be set in the HTML and not in CSS. I updated the description to clarify.

features/layout-direction-override.yml
```diff
@@ -0,0 +1,17 @@
+name: Layout direction override
+description: The `unicode-bidi` and `direction` CSS properties override the Unicode layout algorithm. For HTML documents, authors should use the `dir` global HTML attribute and `<bdo>` HTML element instead.
```
ddbeck, line 2: As written, this sounds like an anti-feature. Is it possible you could link to some history about this feature? I'm wondering who it's _for_, if it's not deprecated.

jamesnw: Good point- I'm not seeing historical posts about this, but both [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi) and [WebReference](https://webreference.com/css/properties/unicode-bidi/) say it is explicitly for DTD designers, so I updated the description.

features/layout-direction-override.yml
```diff
@@ -0,0 +1,17 @@
+name: Layout direction override
+description: The `unicode-bidi` and `direction` CSS properties override the Unicode layout algorithm, and is intended for Document Type Definition (DTD) designers. For HTML documents, authors should use the `dir` global HTML attribute and `<bdo>` HTML element instead.
```
autonome, line 2:
```suggestion
description: The `unicode-bidi` and `direction` CSS properties override the Unicode layout algorithm, and are intended for Document Type Definition (DTD) designers. For HTML documents, authors should use the `dir` global HTML attribute and `<bdo>` HTML element instead.
```

autonome (approved):

ddbeck (approved):
Optional description suggestions for you. In any case, ready to go. Thank you!

Issue #915, opened 2024-04-19
Partial support and notes


foolip:
Both BCD and caniuse have a concept of partial support. Examples:
https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock#browser_compatibility
https://caniuse.com/webp

We have already accumulated a number of features that could benefit from the same:

https://github.com/web-platform-dx/web-features/blob/477f11f33e2966cbd15575ac733178332ee51e51/feature-group-definitions/avif.yml#L6

https://github.com/web-platform-dx/web-features/blob/477f11f33e2966cbd15575ac733178332ee51e51/feature-group-definitions/js-modules.yml#L12-L14

https://github.com/web-platform-dx/web-features/blob/477f11f33e2966cbd15575ac733178332ee51e51/feature-group-definitions/webp.yml#L6

https://github.com/web-platform-dx/web-features/pull/912 is another case currently in review.

In BCD, `partial_implementation` must come with a note, and we should probably do the same.

atopal:
What's the implication for the Baseline status of those items? Is that something we need to assess on a case by case basis or does it make sense to consider partial implementations generally not baseline?

foolip:
I think partial implementation should mean not Baseline, so we for things we consider partially supported we should currently mark them as not supported and add TODOs. This data wouldn't be for more accurate Baseline data, but to capture nuance accurately if web-features is imported to caniuse.com.

foolip:
Note that there are things that BCD considers `partial_implementation` that I think we should still override and say it's not important enough for the whole feature to be considered partial. A missing `onfoo` event handler property is the most obvious such case, where it makes sense to point out at the most granular level in BCD, but it's at most worthy of a note for the overall feature.

ddbeck:
> I think partial implementation should mean not Baseline,

I just wanted to chime in to say that I think this is correct. I think "partial" ought to mean "`baseline: false` with annotations" not a "half Baseline" or something.

mirisuzanne:
Short term, I think it's more helpful to say 'this sub-feature is fully supported' when browsers are consistent in what part they support. If/when browsers do extend the feature, they will discuss support for the new sub-feature separately.

[...]

PR #1966, opened 2024-10-14
Adds display: contents

vwallen:
This would replace an existing draft PR that appears inactive: https://github.com/web-platform-dx/web-features/pull/1131

The support story here is messy: BCD shows high support, but Can I Use shows partial support across the board due to buggy implementations or significant accessibility problems (for example, removing `<button>` elements from focus navigation). 

The `css.properties.display.contents.contents_unusual` refers to default behavior for elements where `display: contents` would result in incomplete, non-semantic, or non-nonsensical behavior. This part of the spec doesn't provide any guidance as to how this should affect accessibility.

Closes #1131.


features/display-contents.yml
```diff
@@ -0,0 +1,8 @@
+name: "display: contents"
+description: "The `display: contents` CSS declaration sets an element to display only its contents, not itself. The element does not generate a display box and may lose semantic or accesibility features."
+spec: https://drafts.csswg.org/css-display-3/#box-generation
+# caniuse: css-display-contents
```
vwallen, line 4: Disagreement between Can I Use and BCD regarding level of support (versions match)

vwallen:
@captainbrosset I saw your comment on the other PR (https://github.com/web-platform-dx/web-features/pull/1131).

How does this look to you?

features/display-contents.yml
captainbrosset, line 1: The accessibility part is interesting here. I did some research.

My understanding is that the spec didn't originally made it clear to implementers what should happen, and so, what ended up happening is that elements that should be focusable were no longer focusable when set to display:contents. That's because these elements would no longer generate a box in the layout for themselves, and therefore, the browser didn't have anywhere to put the focus on.

This became an accessibility issue, reported [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1791648), and [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1366037).
So much, that web industry leaders told devs not to use display:contents at all.

Because not everyone knew about this, an [interop focus area](https://github.com/web-platform-tests/interop/issues/568) was proposed last year to fix this, and got accepted in the Interop 24 cycle.
Unfortunately, it seems like that focus area got scoped down, and [the description of the focus area](https://github.com/web-platform-tests/interop/blob/main/2024/README.md#accessibility) only says that a new test should be added, but doesn't require implementers to pass the test. [The test now exists](https://wpt.fyi/results/css/css-display/focus).

Turns out this test now passes on chromium browsers, and that's thanks to [David Baron implementing the fix in January 2024](https://chromium-review.googlesource.com/c/chromium/src/+/3910374).

The status is the following now:

- On Chromium-based browsers, elements that have a display:contents property **are** now focusable (e.g. tabbing through a page will go through them, `document.activeElement` is updated accordingly, and calling `el.focus()` does work).
  - However, note that no focus ring appear yet. This is considered as a separate issue that still needs discussion.
- On Safari and Firefox, those elements **are not** focusable, so the accessibility problem remains.

My recommendation:

We decided early on in the project that accessibility was not (yet) going to impact baseline status. Also, we don't have the concept of partial support like caniuse (which shows the feature as supported, but with a note).

So I would recommend making this a baseline:high feature, computed from `css.properties.display.contents`, because that's the property devs can use today, and it does the job, and my understanding is that it doesn't impact accessibility negatively if you don't use it on focusable elements.
But also, keep that "may lose semantic or accessibility features" part in your description.

The other option would be to add a new BCD key named something like `css.properties.display.contents.focusable` which would be true for chromium, and false for others. And then, we'd potentially create two features:

1. `Inaccessible display:contents`, computed from `css.properties.display.contents`, baseline high.
2. `display:contents`, computed from `css.properties.display.contents.focusable`, not baseline.

@ddbeck to weight in on this. The latter option might seem weird, but because this is a current problem which devs are asked to think about, so having a feature about it isn't unreasonable. And, over time, those 2 features could be merged.

captainbrosset (changes requested):

features/display-contents.yml
```diff
@@ -0,0 +1,8 @@
...
+group: css
+compat_features:
+  - css.properties.display.contents
+  - css.properties.display.contents.contents_unusual
```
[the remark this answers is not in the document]
vwallen: Updated

features/display-contents.yml
ddbeck: > We decided early on in the project that accessibility was not (yet) going to impact baseline status. Also, we don't have the concept of partial support like caniuse (which shows the feature as supported, but with a note).

This isn't exactly right. We modified the Baseline doc (see https://github.com/web-platform-dx/web-features/pull/519) to say that assistive technology _outside the browser_ is out of scope. Making elements unfocusable and not showing focus states (which are stylable now)—all out of spec—seems in-bounds for Baseline (and BCD). We can and should capture this information.

> So I would recommend making this a baseline:high feature, computed from `css.properties.display.contents`, because that's the property devs can use today, and it does the job, and my understanding is that it doesn't impact accessibility negatively if you don't use it on focusable elements.

I don't agree that we should have two features for this (at least without https://github.com/web-platform-dx/web-features/issues/915). The current, surprising behavior of `display: contents` feels distinctly broken. It's a "feature" that, if used, breaks other features we describe (or will shortly) as widely available (such as `tabindex`).

Instead, I think we should (as you suggested) create a behaviorial BCD entry like `css.properties.display.contents.focusable_descendants` then set the support data based on the results of the WPT case.

Once that feature lands in BCD, we can then publish a single `display: contents` feature that depends on the new BCD key. We ought to omit the "…lose semantic" bit from the description, as that is flatly _not_ an intended or desirable behavior of the feature.

Right now, I think that means blocking this PR (or moving it draft status) or merging it into the `drafts` folder and picking this up again when the BCD work is done.

(As an aside, I only somewhat recently learned about `display: contents`, used it in a personal project and now—thanks to this discussion—I'm removing it. It silently broke things and that makes me sad.)

captainbrosset: +1 from me on going the BCD route for this, and marking this PR as draft in the meantime.

ddbeck: I've sent https://github.com/mdn/browser-compat-data/pull/24766. I'll update here if/when it merges in BCD.

features/display-contents.yml
```diff
@@ -0,0 +1,10 @@
+name: "display: contents"
+description: "The `display: contents` CSS declaration sets an element to display only its contents, not itself. The element does not generate a display box and may lose semantic or accessibility features."
```
ddbeck, line 2: Continuing from https://github.com/web-platform-dx/web-features/pull/1966/files#r1803434536

```suggestion
description: "The `display: contents` CSS declaration sets an element to display only its contents, not itself."
```

features/display-contents.yml
```diff
@@ -0,0 +1,10 @@
...
+spec: https://drafts.csswg.org/css-display-3/#box-generation
+# caniuse: css-display-contents
+group: css
+status:
+  compute_from: css.properties.display.contents
```
ddbeck, line 7: Continuing from https://github.com/web-platform-dx/web-features/pull/1966/files#r1803434536

Let's remove this:

```suggestion
```

features/display-contents.yml
```diff
@@ -0,0 +1,10 @@
...
+compat_features:
+  - css.properties.display.contents
+  - css.properties.display.contents.contents_unusual
```
ddbeck, line 10: Continuing from https://github.com/web-platform-dx/web-features/pull/1966/files#r1803434536

```suggestion
  - css.properties.display.contents.contents_unusual
  - css.properties.display.contents.focusable_elements
```

features/display-contents.yml
ddbeck: [css.properties.display.contents.focusable_elements](https://github.com/mdn/browser-compat-data/pull/24766) has merged, so we'll get it in the next BCD release (probably Friday or sooner). I'll add some comments suggesting what to do with this, specifically.

ddbeck (changes requested):
I've made some suggestions to handle the accessibility issue. This is still blocked until the next BCD release, even if these suggestions are resolved.

ddbeck (approved):

PR #1967, opened 2024-10-14
Adds display CSS property and common declarations

vwallen:


features/display-flow-root.yml
```diff
@@ -0,0 +1,7 @@
+name: "display: flow-root"
+description: "The `display: flow-root` CSS declaration sets the element as the root element of a new flow layout for its children, preventing margin collapse with sibling elements."
```
jamesnw, line 2:
```suggestion
description: "The `display: flow-root` CSS declaration sets an element as the root element of a new flow layout for its children, preventing margin collapse with sibling elements."
```

Alternatively, 
```suggestion
description: "The `display: flow-root` CSS declaration sets the selected element as the root element of a new flow layout for its children, preventing margin collapse with sibling elements."
```

features/display.yml
```diff
@@ -0,0 +1,10 @@
+name: Display
+description: The `display` CSS property sets the display behavior of an element's box within its layout and sets the layout behavior for its child elements.
+spec: https://drafts.csswg.org/css-display-3/#the-display-properties
+group: css
+compat_features:
+  - css.properties.display
+  - css.properties.display.none
```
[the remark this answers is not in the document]
vwallen: Added

ddbeck (approved):

features/display-ruby.yml
```diff
@@ -0,0 +1,12 @@
...
+spec: https://drafts.csswg.org/css-display-3/#layout-specific-display
+group:
+  - css
+  - ruby
+compat_features:
+  - css.properties.display.ruby
+  - css.properties.display.ruby-base
+  - css.properties.display.ruby-base-container
+  - css.properties.display.ruby-text
+  - css.properties.display.ruby-text-container
```
jamesnw, line 12: Documenting for posterity-

The HTML spec marks `<ruby-base>` and `<ruby-text-container>` as obsolete so I was looking into whether the CSS values should be deprecated (first in the CSS spec, then in BCD).

However, there is an open [issue](https://github.com/whatwg/html/issues/1771) to remove them from the list of obselete features. At some point this may see more implementation and need updates, but this is good as is.

PR #1968, opened 2024-10-14
Adds shadow parts

vwallen:


features/shadow-parts.yml
```diff
@@ -0,0 +1,11 @@
+name: Shadow Parts
+description: The `part` and `exportparts` HTML attributes expose elements of a shadow DOM as named parts. The `::part()` CSS pseudo-element selects named parts for styling.
```
jamesnw, line 2:
```suggestion
description: The `part` and `exportparts` HTML attributes expose elements of a shadow DOM as named parts, which can be selected by the `::part()` CSS pseudo-element for styling.
```
I'm uncertain if this is an improvement- I'll leave it up to you.

features/shadow-parts.yml
```diff
@@ -0,0 +1,11 @@
+name: Shadow Parts
```
jamesnw, line 1:
```suggestion
name: Shadow parts
```

features/shadow-parts.yml
vwallen: I had a similar phrasing originally, so 👍

PR #1972, opened 2024-10-15
Add keys from BCD 5.6.6 release to existing features

jamesnw:
New keys introduced by BCD 5.6.6:

- `css.properties.white-space-collapse.preserve-spaces` (added compute_from to avoid baseline impact)
- `css.properties.column-width.auto` (no impact to baseline)
- `css.properties.object-position.applies_to_iframe_elements` (added compute_from to avoid baseline impact)


features/multi-column.yml.dist
```diff
@@ -73,6 +73,19 @@ compat_features:
...
+  # baseline_high_date: 2022-07-15
+  # support:
+  #   chrome: "50"
+  #   chrome_android: "50"
+  #   edge: "79"
+  #   firefox: "50"
+  #   firefox_android: "50"
+  #   safari: "9"
+  #   safari_ios: "9"
+  - css.properties.column-width.auto
```
captainbrosset, line 87: I'm confused here. The browser versions for the `auto` value on MDN doesn't match what we have here:
image

Checking BCD confirms what's on MDN:
image

Any idea why our code returns these older browser versions for this key?

features/object-position.yml
```diff
@@ -3,5 +3,8 @@ description: The `object-position` CSS property places images, videos, and other
 spec: https://drafts.csswg.org/css-images-3/#the-object-position
 # object-fit on caniuse includes both object-fit and object-position.
 # `caniuse: object-fit` is set in object-fit.yml
+status:
+  compute_from: css.properties.object-position
```
captainbrosset, line 7: The description says "images, videos, and other replaced elements", but now we're ignoring iframes, which is also a replaced element. So I think we should do something like this: call out the iframe exception in the description, and add a TODO comment about removing the callout once iframes are taken into account.

features/white-space-collapse.yml
```diff
@@ -2,9 +2,12 @@ name: white-space-collapse
 description: The `white-space-collapse` CSS property sets whether new line characters are shown as line breaks, and whether multiple consecutive spaces are all displayed or combined.
 spec: https://drafts.csswg.org/css-text-4/#white-space-collapsing
 group: white-space
+status:
+  compute_from: css.properties.white-space-collapse
```
captainbrosset, line 6: Maybe add a comment to document the fact that we're computing from this property in order to ignore the `preserve-spaces` value. And that we're doing this because `preserve-spaces` is experimental and might change (according to BCD). This will help our future selves maintain/evolve that feature over time.

captainbrosset (changes requested):

features/multi-column.yml.dist
ddbeck: That's right. `compute-baseline` checks ancestor compat features (we turned this on in https://github.com/web-platform-dx/web-features/pull/1485). This is an oddity of BCD's data model. It can't directly express the idea that a feature is, for example, behind a prefix when some parent of that feature is behind a parent.

features/object-position.yml
ddbeck: I really dislike the way this information has been represented in BCD. I think the addition of the `applies_to_iframe_elements` key ought to be reverted (and left a comment saying so). Until https://github.com/mdn/browser-compat-data/issues/23586 is resolved, I'd suggest:

- Undoing all the changes in this feature so far
- Adding a comment mentioning `css.properties.object-position.applies_to_iframe_elements` and https://github.com/mdn/browser-compat-data/issues/23586 and saying that the compat data for which elements `object-position` works with is unclear and needs to be sorted out upstream.

jamesnw: Updated in e95bb194

ddbeck (approved):

ddbeck:
I'm going to move head with this. Patrick, if you still have suggestions or concerns, would you open a follow up PR or issue? Thank you!

PR #1974, opened 2024-10-15
Adds host scoping pseudo-element

vwallen:


features/host.yml
```diff
@@ -0,0 +1,10 @@
+name: Host
+description: The `::host` CSS pseudo-element selects the containing element of the shadow tree in which it is used.
```
vwallen, line 2: Unclear if the description should be extended to include `::host()` and `::host-context()` explicitly. I left them off for brevity

jamesnw: I think `:host-context()` should be a separate feature, based  on support, and `:host()` should be mentioned in the description here.

jamesnw, line 2:
```suggestion
description: The `:host` CSS pseudo-class selects the containing element of the shadow tree in which it is used.
```

features/host.yml
```diff
@@ -0,0 +1,10 @@
...
+spec: https://drafts.csswg.org/css-scoping-1/
```
jamesnw, line 3:
```suggestion
https://drafts.csswg.org/css-scoping-1/#host-selector
```

jamesnw (changes requested):

jamesnw (approved):

PR #1976, opened 2024-10-15
Adds @namespace rule

vwallen:


features/namespace.yml
```diff
@@ -0,0 +1,13 @@
+name: "`@namespace`"
+description: The `@namespace` CSS rule sets a default namespace or namespace prefix. Namespace prefixes allow CSS selectors to distinguish elements with the same name but different document types, such as the HTML `<a>` element and the SVG `<a>` element.
+spec: https://drafts.csswg.org/css-namespaces-3/#declaration
+caniuse: css-namespaces
```
jamesnw, line 4: I think we'd need to compute from `css.at-rules.namespace` to match caniuse here, which I think makes sense here.

jamesnw (approved):
Compute from change, but besides that, looks good!

PR #1977, opened 2024-10-15
Update EyeDropper files to reference as "eye-dropper"

DanielRyanSmith:
This was proposed when [adding a new WebDXFeature use counter](https://chromium-review.googlesource.com/c/chromium/src/+/5838617/comments/00743d4b_bb08ba5c).

PR to change browser-compat-data tag name: https://github.com/mdn/browser-compat-data/pull/24730


captainbrosset:
@ddbeck seems like this is the right thing to do, but headsup, this would be a breaking change because of the proposed ID change.
What if we kept the ID unchanged though? This way, no need for the BCD PR, no breaking change.

ddbeck:
I don't really understand why the name or ID is changing here. Looking back at https://github.com/web-platform-dx/web-features/pull/1179, I think the considerations given to "eyedropper" versus "eye dropper" versus "EyeDropper" still make sense today.

For the name, the feature is a getting the color via an _eyedropper_. While it's nice when an interface is named the same as the feature it provides, this is often not the case (or only incompletely). It's also been my hope that names and descriptions would someday be translated, when it makes sense to do so (this might be _Kleurendruppelaar_ in Dutch).

Looking to the ID, I think `eyedropper` is going to stand the test of time. It's nice when IDs have some permanence. And English tends to discard the space or hyphen in compounds (e.g., _e-mail_ is now widely accepted as _email_), so I'd be worried that we'd find ourselves wishing to undo the change before too long anyway.

ddbeck:
Sorry, reading over myself here I worry that I'm sounding a bit harsh. I'm grateful for the prompt to think about this more, even though I don't think this is the right move in this case.

DanielRyanSmith:
Closing this and moving forward with the existing naming convention

PR #1983, opened 2024-10-16
Add document, document fragment, document type

autonome:
kitchen sink at this point, pulled from various drafts.

Ok, whittling down to keys which match the following criteria:
- superkeys like `api.Document`
- keys which refer directly to the document itself or change it, eg `document.lastModified` and `document.
- document-level collections, eg `document.images` and `document.forms` (yes, there may be "primary" features for the thing being collected but the `document` shortcut is document-level)

Need to determine whether to put other interface impls in the feature or in the interface - eg, `document.append` is impl of `Element`... so do we add all the `append` keys to their feature or add them all to an `Element` feature? TBD will ask in chat/meeting.

Will see what's left after that culling and classify the remainders.


autonome:
Ready for a first pass. There's not any real standalone feature candidates in the key list for Document. There's a couple notes on Caniuse keys that are related.

Open question is what approach for keys like `createTreeWalker` - there'll be a `TreeWalker` feature coming out of the DOM PR I've got going. Things like Ranges also will be feature. I assumed it'd go with the specific feature vs leaving it here, but worth asking if there's different opinion there.

ddbeck:
I think a lot of this stuff is going to get swept into a monolithic DOM feature. I'd appreciate if other reviewers also had a look through https://github.com/web-platform-dx/web-features/pull/2007/files#r1813040012.

autonome:
Closing to fold into #2007

PR #1994, opened 2024-10-17
`update-drafts` uses parent spec if one is present

jamesnw:
Before, if a BCD key didn't have a spec_url, it did not show up in the draft features. With this change, if a BCD key doesn't have a spec_url, it will use the closest ancestor's spec.

Apologies for the diff and for resurrecting deleted drafts, but a quick glance looks like most will be simple additions to existing specs. 

<details><summary>These don't have a spec or an ancestor with a spec</summary>

- api.CommandEvent
- api.CommandEvent.CommandEvent
- api.CommandEvent.command
- api.CommandEvent.source
- api.EXT_disjoint_timer_query_webgl2
- api.EXT_disjoint_timer_query_webgl2.queryCounterEXT
- api.FeaturePolicy
- api.FeaturePolicy.allowedFeatures
- api.FeaturePolicy.allowsFeature
- api.FeaturePolicy.features
- api.FeaturePolicy.getAllowlistForFeature
- api.WebSocketError
- api.WebSocketError.WebSocketError
- api.WebSocketError.closeCode
- api.WebSocketError.reason
- api.WebSocketStream
- api.WebSocketStream.WebSocketStream
- api.WebSocketStream.close
- api.WebSocketStream.closed
- api.WebSocketStream.opened
- api.WebSocketStream.url
- css.properties.-webkit-mask-box-image-outset
- css.properties.-webkit-mask-box-image-repeat
- css.properties.-webkit-mask-box-image-slice
- css.properties.-webkit-mask-box-image-source
- css.properties.-webkit-mask-box-image-width
- css.selectors.-webkit-slider-runnable-track
- css.selectors.-webkit-slider-thumb
- css.types.global_keywords
- http.headers.No-Vary-Search
- http.headers.Sec-CH-Save-Data
- javascript.grammar.shorthand_object_literals
- svg.global_attributes.requiredExtensions
- svg.global_attributes.requiredExtensions.mathml
- webassembly.jsStringBuiltins

Notes: for `-webkit-mask-box-image-*`, these will be removed pending https://github.com/mdn/browser-compat-data/issues/22727.

</details>


scripts/update-drafts.ts
```diff
@@ -137,12 +138,29 @@ async function main() {
       continue;
     }
 
-    const spec_url = feature.data.__compat.spec_url;
-    if (!spec_url) {
-      continue;
+    let parent_url = null;
```
ddbeck, line 141: I think some types here makes following what comes later a little easier to follow. I'm suggesting this (and some other types, where it helped me follow the logic):

```suggestion
    let parent_url: string | string[] = null;
```

scripts/update-drafts.ts
```diff
@@ -137,12 +138,29 @@ async function main() {
...
+    if (!feature.data.__compat.spec_url) {
```
ddbeck, line 142: The `spec_url` getter will always give you an array, which is a little tidier:

```suggestion
    if (!feature.spec_url.length) {
```

scripts/update-drafts.ts
```diff
@@ -137,12 +138,29 @@ async function main() {
...
+      const parent_urls = [];
```
ddbeck, line 143:
```suggestion
      const parent_urls: (string | string[])[] = [];
```

scripts/update-drafts.ts
```diff
@@ -137,12 +138,29 @@ async function main() {
...
+      const path = feature.id.split(".");
+      let parentFeature = compat.data as Identifier;
+      while (path.length > 1) {
+        parentFeature = parentFeature[path.shift()];
+        const parent_spec = parentFeature.__compat?.spec_url;
```
ddbeck, line 148: Similarly, you can use `compat.features` to get an object with a `spec_url` getter (probably):

```suggestion
      while (path.length > 1) {
        const parent_spec = compat.features.get(path.shift())?.spec_url ?? [];
```

scripts/update-drafts.ts
```diff
@@ -11,6 +11,7 @@ import { Document } from "yaml";
 import yargs from "yargs";
 
 import { features } from "../index.js";
+import { Identifier } from "@mdn/browser-compat-data";
```
ddbeck, line 14: If you decide to take my suggestions below, this is unneeded.

ddbeck (commented):
Some suggestions on the code itself, but this seems like a good idea.

scripts/update-drafts.ts
jamesnw: This didn't quite work- it was calling `compat.features.get('properties')` instead of `compat.features.get('css.properties')`. I did switch to building a path up and using `compat.features.get()` instead of drilling through the raw data.

features/webgpu.yml
ddbeck, line 1: Were the changes to `webgpu` done by hand or caused by the script? This is the only thing that worries me—I want to preserve the mostly hands-off use of the update-drafts Actions workflow.

ddbeck (commented):
OK, I'm happy with this, but I have a question about webgpu.

Also, I merged the update-drafts workflow earlier today—you might need to merge main into this branch and rerun the script to avoid conflicts.

features/webgpu.yml
jamesnw: Good question on `webgpu`. #1948 moved the draft to the features folder, and because this PR was updating the draft, merging that move in added the keys to the non-draft feature. 

This PR doesn't increase the likelihood that that would happen in the future, but this PR is where it's most likely to happen. It's slightly mitigated by the fact that the action [only adds the draft folder](https://github.com/web-platform-dx/web-features/blob/main/.github/workflows/update_draft_features_weekly.yml#L26), but a simultaneous move could theoretically add keys to a non-draft feature.

I reset `features/webgpu.yml` to match `main`, and updated the drafts, so now the new keys are in `features/draft/spec/webgpu.yml`.

ddbeck (approved):
Looks good. Let's go!

PR #1997, opened 2024-10-18
Adds Text underline position & text shadow

SondraE:


features/text-underline-position.yml
```diff
@@ -0,0 +1,15 @@
+name: Text underline position
+description: "The `text-underline-position` and `text-underline-offset` CSS properties set the position and distance from initial position of text underlines on text with `text-decoration: underline`."
+spec: https://drafts.csswg.org/css-text-decor-4/#line-decoration
+group: text
+status:
+  compute_from: css.properties.text-underline-position
```
autonome, line 6: If grouping these, then should probably be conservative and compute from the later of the headline features (`css.properties.text-underline-offset`) rather than the earlier, so that any developer decisions based on this availability mean that all of it works vs some things not working. cc @ddbeck for thoughts

autonome (changes requested):

features/text-underline-position.yml
jamesnw: That makes sense to me!

jamesnw (approved):

PR #1999, opened 2024-10-18
Add CSSOM and related features

jamesnw:
`api.CSSImportRule.supportsText` remains in the draft, per [supports.yml](https://github.com/web-platform-dx/web-features/blob/main/features/supports.yml#L9-L10).


features/css-object-model.yml
```diff
@@ -0,0 +1,83 @@
+name: CSS Object Model
+description: The CSS Object Model allows users to create and change CSS styles using JavaScript. Also known as CSSOM.
```
ddbeck, line 2: A few things here:

- Ugh, this has so many plausible entry points, which makes describing this difficult.
- I don't think we should mention JavaScript, since it's a browser API, not a JS built-in.
- We should probably have a style entry for "users" to only use it to mean something like "the users of the pages/applications that developers make" and not "developers." Or I have badly misunderstood what CSSOM is for.

```suggestion
description: The CSS object model API reads, creates, and modifies CSS stylesheets and inline styles. Also known as CSSOM.
```

features/get-computed-style.yml
```diff
@@ -0,0 +1,8 @@
+name: getComputedStyle()
+description: The `getComputedStyle()` method returns the values of all CSS properties applied to an element.
```
ddbeck, line 2: As written, I think it implies that it returns something like an array of values. What do you think of something like this?

```suggestion
description: The `getComputedStyle()` global method returns an `CSSStyleDeclaration` object that represents all CSS declarations applied to a given element.
```

groups/cssom.yml
```diff
@@ -0,0 +1,3 @@
+name: CSS Object Models
```
ddbeck, line 1:
```suggestion
name: CSS object models
```

ddbeck (approved):
I like the way this looks. Suggesting some name and description tweaks, but I think this is ready to go once you've had a look over them.

features/css-escape.yml
```diff
@@ -0,0 +1,6 @@
+name: CSS.escape()
+description: The `CSS.escape()` method escapes a string so that it can be used as a valid CSS selector.
```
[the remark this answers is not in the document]
jamesnw: Good catch- because it would escape a `.` or `#`, for example, you can't use it to escape the whole selector, just the identifier.

jamesnw:
> When this (and/or #2000) lands, it'd be nice to file an issue to revisit this feature grouping after we've got full coverage of BCD. It'd be nice to know whether we need/want another model for this information (e.g., sharing keys, showing "dependent" feature relationships, etc.).

Opened https://github.com/web-platform-dx/web-features/issues/2094

Issue #1704, opened 2024-08-29
Document versioning policy


ddbeck:
As of v1.0 we're now obliged to make [SemVer MAJOR releases](https://semver.org/#spec-item-8) for "breaking" changes. We should clearly document what events constitute a breaking change, both to help consumers of the package understand what is needed each time they upgrade and to help maintainers prepare useful release notes.

To that end, we have at least two "obvious" paths:

1. Breaking changes are backwards-incompatible changes to the JSON Schema (e.g., renaming an existing key). This is roughly how `@mdn/browser-compat-data` does versioning, with some documented exceptions.
2. Breaking changes are backwards-incompatible changes to the JSON Schema (as above) and changes to feature IDs (e.g., renaming a feature, such as `css-grid` to `grid`). This is roughly how `webref` and `browser-specs` handle it.

This also has implications for minor and patch releases (e.g., under route 1, new features constitute a SemVer PATCH release, but under route 2 those are MINOR releases).

dontcallmedom:
I think (2) would help deal with #91 and related discussions around evolving features; it allows data consumers to rely on a clear signal well integrated in dependency management tool to determine how much a particular release needs their attention.

captainbrosset:
I agree that (2) seems better in our case.

I would also note that we should think of an additional communication method for consumers who do not use our package. WPT, for example, tags tests with web-features IDs, but does not, as far as I know, consume the package. How would WPT (and others like it) know when an ID changes?

ddbeck:
I can't tell how WPT is getting the web-features JSON (I found some code where the process it, but not where the actual dependency to web-features lives). I'd be interested to know what's going on there in more detail.

I imagine that fixing https://github.com/web-platform-dx/web-features/issues/1186 would at least put non-npm consumers on an equal footing—they'd be able to see that we made a breaking change via version number.

I'd like to get more consumer input before going down this path, but another possibility would be to bake the feature IDs into the schema files. That is, every new, renamed, or removed feature would result in a change to the JSON Schema, while feature churn (e.g., new browser releases) wouldn't.

Longer term, resolving https://github.com/web-platform-dx/web-features/issues/91 within the schema and with some utilities would be the best thing (e.g., tombstone removed/moved features for a time, then do periodic breaking changes on some published schedule).

ddbeck:
[...]

PR #2006, opened 2024-10-20
Publishing docs: add informing consumers to the release steps

ddbeck:
This is primarily a reminder to myself, since James asked for this previously.

I expect we'll eventually get rid of this step, if/when we version the schema itself. See https://github.com/web-platform-dx/web-features/issues/1704


jamesnw (approved):
LGTM!

PR #2008, opened 2024-10-21
Add CanvasPattern

autonome:
No Caniuse. Treats same as other sub-parts of Canvas2d, rolling it up into main feature, eg CanvasGradient. I suspect someday we break these kinds of things out as features, but today is not that day.


Elchi3 (approved):
👍

PR #2012, opened 2024-10-21
Add page visibility, transition and lifecycle features

autonome:


autonome:
Ok the features w/ Caniuse entries now match more closely to the tables there.

features/page-visibility.yml
```diff
@@ -0,0 +1,10 @@
+name: Page Visibility
+description: "The Page Visibility API tells you the current visibility state of the page, and fires events you can watch for to know when it becomes visible or hidden."
```
ddbeck, line 2: Naming the entry points here would be nice, as would explaining what visibility actually is. I also think we shouldn't mention the event, since we're pinning to caniuse, which doesn't acknowledge it.

```suggestion
description: "The `document.visibilityState` and `document.hidden` properties tell you whether the page is visible to the user (for example, it's not minimized or in a background tab)."
```

features/page-visibility.yml
```diff
@@ -0,0 +1,10 @@
+name: Page Visibility
```
ddbeck, line 1:
```suggestion
name: Page visibility
```

features/page-transition-events.yml
```diff
@@ -0,0 +1,12 @@
+name: Page Transition Events
+description: "The page transition events `pageshow` and `pagehide` are fired when a document is being loaded or unloaded due to a navigation, such as clicking on a link on a page, or the back button in a browser."
```
ddbeck, line 2: consistency, brevity nits

```suggestion
description: "The `pageshow` and `pagehide` transition events fire when a document loads or unloads due to a navigation, such as clicking on a link on a page or the back button in a browser."
```

features/page-transition-events.yml
```diff
@@ -0,0 +1,12 @@
+name: Page Transition Events
```
ddbeck, line 1:
```suggestion
name: Page transition events
```

features/page-lifecycle.yml
```diff
@@ -0,0 +1,7 @@
+name: Page Lifecycle
```
ddbeck, line 1:
```suggestion
name: Page lifecycle
```

features/page-lifecycle.yml
```diff
@@ -0,0 +1,7 @@
...
+description: "The Page Lifecycle API lets pages safely handle browser interventions such as page suspensions or discards from memory."
```
ddbeck, line 2: I think we have room enough to describe this rather completely:

```suggestion
description: "The page lifecycle API helps you safely handle page suspensions or discards from memory. The `freeze` and `resume` events fire when the browser suspends or resumes a page from memory while the `document.wasDiscarded` property reports whether the page was reloaded after being dropped."
```

ddbeck (approved):
Looking good. Name and description suggestions only; this is ready to merge after those are tended to. Thank you!

PR #2013, opened 2024-10-21
Fix decodedBodySize and encodedBodySize property names in resource-size description

captainbrosset:
**Before:**

> The decodedSize, encodedSize, and transferSize properties of the PerformanceResourceTiming API reports the size of resources loaded

**After:**

> The decodedBodySize, encodedBodySize, and transferSize properties of the PerformanceResourceTiming API reports the size of resources loaded

I know nothing about this feature, but this is what's documented at MDN.


Elchi3 (approved):
Good catch!

(I have a PR for all the Performance API related features in https://github.com/web-platform-dx/web-features/pull/1934)

Issue #2004, opened 2024-10-18
animation-timeline


SachaG:
### Specification

https://www.w3.org/TR/css-animations-2/#propdef-animation-timeline

### Description

The animation-timeline [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) property specifies the timeline that is used to control the progress of a CSS animation.

### Browser support

https://caniuse.com/mdn-css_properties_animation-timeline

ddbeck:
Thanks for opening this issue. I think this is covered by https://github.com/web-platform-dx/web-features/blob/main/features/scroll-driven-animations.yml. Do we need an "also known as" in the description? Or am I confused about what this is covering?

ddbeck:
Ah, I see now that it looks like we created a single feature, `scroll-driven-animations` that encompasses:

- https://github.com/web-platform-dx/web-features/issues/2002
- https://github.com/web-platform-dx/web-features/issues/2003
- https://github.com/web-platform-dx/web-features/issues/2004 (this issue)

Do I have that right? I don't know a lot about the feature(s), so I'm not sure if it makes sense for these properties to be understood as separate features. I'd appreciate hearing your opinion.

SachaG:
I'm also not sure to be honest, basically I just search the page for `timeline` and couldn't fine these features.

captainbrosset:
My preference would be to keep these together for now. Sure, they are independent properties, but they get used together by developer to achieve one thing only: animating on scroll. 

But maybe @bramus can help shed some light here. Do you think the properties mentioned above are seen as individual things by developers, or only as a group where they use everything at the same time?

bramus:
Technically speaking, `animation-timeline` is bucketed under CSS Animations / Web Animations, as it allows you to change the timeline of an animation.

Practically speaking there are only the `DocumentTimeline` and the `ScrollTimeline`s which you can use with it, so it kinda makes sense to keep it grouped with scroll-driven animations.

In the future more types of timelines might come into existence, but today that’s not the case. All that to say that I’m with Patrick on this: keep it grouped (for now).

ddbeck:
It sounds like the problem is that the existing feature is not searchable. I've opened https://github.com/web-platform-dx/web-features/pull/2020 to attempt to fix this (and the related issues).

Issue #2003, opened 2024-10-18
view-timeline


SachaG:
### Specification

https://drafts.csswg.org/scroll-animations-1/#view-timelines

### Description

The view-timeline [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) is used to define a named view progress timeline, which is progressed through based on the change in visibility of an element (known as the subject) inside a scrollable element (scroller). view-timeline is set on the subject.

### Browser support

https://caniuse.com/mdn-css_properties_view-timeline

Issue #2002, opened 2024-10-18
scroll-timeline


SachaG:
### Specification

https://drafts.csswg.org/scroll-animations-1/#scroll-timeline-shorthand

### Description

The scroll-timeline [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) is used to define a named scroll progress timeline, which is progressed through by scrolling a scrollable element (scroller) between top and bottom (or left and right).

### Browser support

https://caniuse.com/mdn-css_properties_scroll-timeline

PR #2020, opened 2024-10-21
`scroll-driven-animations`: revise description for search

ddbeck:
This is intended to improve the findability of this feature via its most-visible properties.

Fixes https://github.com/web-platform-dx/web-features/issues/2004.
Fixes https://github.com/web-platform-dx/web-features/issues/2003.
Fixes https://github.com/web-platform-dx/web-features/issues/2002.

### DIFF
--- a/features/scroll-driven-animations.yml
+++ b/features/scroll-driven-animations.yml
@@ -1,5 +1,5 @@
 name: Scroll-driven animations
-description: The `animation-timeline`, `scroll-timeline`, and `view-timeline` CSS properties advance animations based on the user's scroll position, instead of the time since the document loaded.
+description: The `animation-timeline`, `scroll-timeline`, and `view-timeline` CSS properties advance animations based on the user's scroll position.
 spec: https://drafts.csswg.org/scroll-animations-1/
 group:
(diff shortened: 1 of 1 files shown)


features/scroll-driven-animations.yml
```diff
@@ -1,5 +1,5 @@
 name: Scroll-driven animations
-description: CSS scroll-driven animations are a type of CSS animations that don't run over time, but are instead driven by the user's scroll position.
+description: The `animation-timeline`, `scroll-timeline`, and `view-timeline` CSS properties advance animations based on the user's scroll position, instead of the time since the document loaded.
```
autonome, line 2: > , instead of the time since the document loaded.

Seems unnecessarily narrow, could be removed.

autonome (approved):

features/scroll-driven-animations.yml
captainbrosset: +1, the time since the document loaded doesn't always play a role in time-based animations.

ddbeck: Thank you both!

PR #2037, opened 2024-10-22
Add additional cursor keys

jamesnw:
No impact to baseline. (Note that baseline seems odd because most cursors aren't supported on iOS).


autonome (approved):
iirc the drafts update themselves no need to do it manually (i wonder if the more-often update landed tho...)

PR #2042, opened 2024-10-22
`release.ts`: fix diff buffer full

ddbeck:
We've added so much data that Node.js errors out when it fills the buffer with the diff between releases. This PR pipes the diff to disk instead of accumulating it in a string.


scripts/release.ts
```diff
@@ -167,10 +167,25 @@ function readPackageJSON(packageDir) {
   );
 }
 
-function diffJson(from: string = "latest", to?: string): string {
+function toPrettyJson(sourceFp: string, destFp: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const jqProcess = spawn("jq", [".", sourceFp]);
```
autonome, line 172: any reason not to use `node-jq`?

scripts/release.ts
```diff
@@ -182,27 +197,20 @@ function diffJson(from: string = "latest", to?: string): string {
...
-      const preparedJson = join(packages["web-features"], "data.json");
-      const prettyPreparedJson = execSync(`jq . "${preparedJson}"`, {
-        encoding: "utf-8",
-      });
-      const fp = join(temporaryDir, "data.HEAD.json");
-      writeFileSync(fp, prettyPreparedJson);
-      return fp;
+      return toPrettyJson(
+        join(packages["web-features"], "data.json"),
+        join(temporaryDir, "data.HEAD.json"),
```
autonome, line 212: (stylistic comments, feel free to ignore)

I liked the less-compact version b/c was more explicit, and easier to read.

My preference is to put all file paths/names/parts in consts at top of file, or at least function.

scripts/release.ts
autonome: ah maybe streams?

autonome (approved):

scripts/release.ts
ddbeck: Actually, this whole thing was silly: the original reason for using `jq` is no longer needed. I can just read, parse, and stringify to get the same result. 🤦

scripts/release.ts
ddbeck: (I ended up reverting this and starting over. Marking it as resolved.)

scripts/release.ts
```diff
@@ -182,9 +192,8 @@ function diffJson(from: string = "latest", to?: string): string {
       "web-features",
       "data.json",
     );
-    const prettyJson = execSync(`jq . "${pkgJson}"`, {
-      encoding: "utf-8",
-    });
+    // fn this prettyJSON
```
ddbeck, line 195:
```suggestion
```

jamesnw (approved):
This change makes sense and looks good to me!

Elchi3 (approved):

PR #2044, opened 2024-10-23
Input (continued)

autonome:
Migrating the work from #1441 

The Caniuse entries seem to line up, so @petele already did that 🎉

I fixed the descriptions to be clearer when they contain a list of `type` values. I don't like the inconsistency w/ the single-type version, but is probably ok like this.

The idea of variants of a specific attribute value for a specific element being standalone features seems fine, but I don't know if that fits into how we've been thinking about the `html-elements` group. Do we expect that group to be each element, once? Or is this fine?

Fixes #1441


features/input-email-tel-url.yml
```diff
@@ -0,0 +1,14 @@
+name: <input> with types email, tel, and url
+description: The `<input>` element allows the user to input information to a form intended for email addresses, telephone numbers or URLs. Particularly useful in combination with form validation.
```
ddbeck, line 2: We don't do cross references in descriptions like caniuse (rarely) does.

```suggestion
description: The `<input type="email">`, `<input type="tel">`, and `<input type="url">` HTML elements represent email address, telephone number, and URL fields in a form.
```

features/input-event.yml
```diff
@@ -1,4 +1,4 @@
-name: input (event)
+name: Input event
```
ddbeck, line 1: This came about in https://github.com/web-platform-dx/web-features/pull/1026 and https://github.com/web-platform-dx/web-features/pull/1492. Let's leave this for now, then do a full audit of event features when we've mapped everything.

```suggestion
name: input (event)
```

features/input-number.yml
```diff
@@ -0,0 +1,12 @@
+name: '<input type="number">'
+description: The `<input>` element allows the user to input numeric information to a form.
```
ddbeck, line 2:
```suggestion
description: The `<input type="number">` HTML element allows the user to input numeric information to a form.
```

features/input-range.yml
```diff
@@ -0,0 +1,16 @@
...
+caniuse: input-range
+compat_features:
+  # Included in input-number.yml
+  # - html.elements.input.max
+  # - html.elements.input.min
+  # - html.elements.input.step
+  - html.elements.input.type_range
+  - html.elements.input.type_range.vertical_orientation
+  # Excluded because of a non-real value
+  # - html.elements.input.type_range.tick_marks
```
ddbeck, line 16: This seems to be fixed now too. Try again with with this uncommented?

features/input-range.yml
```diff
@@ -0,0 +1,16 @@
+name: '<input type="range">'
+description: The `<input>` element allows the user to select a value using a slider widget.
```
ddbeck, line 2: (pet peeve: select is one of those deeply overloaded words when talking about UI. Going abstract here helps a little.)

```suggestion
description: The `<input type="range">` element allows the user to choose a value using a slider widget.
```

features/input-search.yml
ddbeck, line 1: Can we undo this and file an issue instead? This would be a breaking change and I'd prefer to do several of those together, if we're going to do them.

features/input.yml
```diff
@@ -1,8 +1,11 @@
-draft_date: 2024-06-05
 name: <input>
-description: The `<input>` element TODO.
+description: The `<input>` element allows the user to input information to a form; a number of types of input data and control widgets are available, depending on the device and browser.
```
ddbeck, line 2:
```suggestion
description: The `<input type="text">` HTML element allows the user to input information into a form. By default, an `<input>` element is a text input.
```

features/input.yml
ddbeck, line 1: Yeah, now that I've gotten this far, I'm certain we should treat each `type` as its own `input-` "element." That said, since `text` is the default, that's the one that should be represented here.

features/input.yml
```diff
@@ -1,8 +1,11 @@
...
 spec: https://html.spec.whatwg.org/multipage/input.html#the-input-element
-group: forms
+group:
+  - forms
+  - html-elements
+status:
+  compute_from: html.elements.input
 compat_features:
   - api.HTMLInputElement
```
ddbeck, line 10: Sorry, it's hard to annotate this in the diff, but some of this API need to be split out into their respective features, such as `step`, `stepUp`, `stepDown` (number, since I think that's the oldest of the bunch that use this) or `valueAsDate` (input-date-time). Could you have a look through each `api` item, to double check this list?

features/input.yml
```diff
@@ -77,31 +78,14 @@ compat_features:
   # Excluded because of a non-real value
   # - html.elements.input.size
```
ddbeck, line 79: I think all the non-reals are fixed. Uncomment (or move to appropriate features) and refresh?

features/input.yml
```diff
@@ -77,31 +78,14 @@ compat_features:
...
-  - html.elements.input.type_number
   - html.elements.input.type_password
   - html.elements.input.type_password.insecure_login_handling
   - html.elements.input.type_radio
-  - html.elements.input.type_range
-  # Excluded because of a non-real value
-  # - html.elements.input.type_range.tick_marks
-  # - html.elements.input.type_range.vertical_orientation
   - html.elements.input.type_reset
   - html.elements.input.type_submit
```
ddbeck, line 90: I think all the remaining `type_` features can be broken out:

- input-button
- input-checkbox
- input-file
- input-hidden
- input-image
- input-password
- input-radio
- input-reset-submit (I'd group these together, but feel free to separate if you disagree)

`type_text` should remain here.

ddbeck (changes requested):
Several suggestions here, mostly about descriptions. The big things are:

- Let's treat almost all input types as if they were distinct elements (though a few can be grouped together). See line comments for details about this.
- Lots of non-reals have been fixed. Uncommenting them and refreshing will probably fix nearly all of those comments.
- Changing a feature's ID is a breaking change. Let's set that aside for a minute.

features/input-color.yml
```diff
@@ -0,0 +1,9 @@
+name: '<input type="color">'
+description: The `<input>` HTML element with the `type="color"` attribute represents a color picker, which might be styled differently by the browser.
```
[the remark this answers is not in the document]
autonome: Sorry I didn't realize this wasn't marked as draft. I haven't done a full pass on the orig PR yet. Will incorporate your feedback here and then do a full pass.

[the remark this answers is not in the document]
autonome: Oh nice, past me marked it ready for review 😐 My apologies, clearly not ready now that I've read through all your comments!

features/input.yml
autonome: Per https://github.com/openwebdocs/project/issues/206 it looks like the keys in this PR are all fixed.

features/input.yml
autonome: wdyt about also pulling in all the related keys for these as well?

eg for ` html.elements.input.type_radio` there's `api.RadioNodeList` and `css.properties.appearance.radio` etc etc

features/input.yml
autonome: i kept reset and submit separate since they do different things, but also because reset is destructive to local data. not sure if we've given warnings about behavior like that in the work so far?

features/input.yml
autonome: Ok, letting sleeping keys lie!

features/input-color.yml.dist
```diff
@@ -0,0 +1,14 @@
+# Generated from: input-color.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
```
autonome, line 5: BCD says Firefox Android is only partial b/c only allows preset colors not custom ones. Weird, but ok.

autonome:
ok ready for another review pass. not stoked on the "represents" language from the original PR, so lmk if you think needs reworking.

ddbeck (approved):
This is pretty great. I'm going to land this now, then revise all the input descriptions together (including ones not part of this PR) at once. Thank you, Dietrich!

PR #2043, opened 2024-10-22
Add checklist for handling BCD upgrades

ddbeck:
This workflow adds a checklist comment to dependabot PRs that have `@mdn/browser-compat-data` in the diff. This should help make the work around these PRs more consistent.

PR #2073, opened 2024-10-24
Fix checklist for BCD upgrades workflow

ddbeck:
A follow up to https://github.com/web-platform-dx/web-features/pull/2043.

This workflow on the first attempt: https://github.com/web-platform-dx/web-features/actions/runs/11496579249/job/31998472971?pr=2071#logs

Problems:

- No write permissions for the PR
- No checkout to run the diff against (I tested this bit and I have no idea why this worked on my own repo but not here, even though it obviously SHOULD have failed 🙄)

To fix this, I did these things:

- Switched to using https://github.com/dependabot/fetch-metadata instead (to avoid a slow checkout)

  Unfortunately, this makes the PR basically untestable—you can't force Dependabot to do anything. We'll find out if it works tomorrow, I guess.

- Set `continue-on-error` for both steps, since failures of this workflow shouldn't mark the PR as failing

### DIFF
--- a/.github/workflows/bcd_upgrade_checklist.yml
+++ b/.github/workflows/bcd_upgrade_checklist.yml
@@ -3,19 +3,21 @@ name: BCD upgrade
 on:
   pull_request:
     types: [opened]
+permissions:
+  pull-requests: write
 jobs:
   post-comment:
     name: "Post checklist"
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ddbeck:
@Elchi3 if you'd have another look at this, I'd appreciate it. 🙏

Elchi3 (approved):
This looks a lot cleaner to me. Fingers crossed this time :)

PR #2078, opened 2024-10-24
New web app manifest features

captainbrosset:
This isn't a complete list, but it's a good start.


features/app-file-handling.yml
```diff
@@ -0,0 +1,11 @@
+name: File handling
+description: The `file_handlers` web app manifest member registers an installed web app as a handler for files with specific file extensions or MIME types.
+spec: https://wicg.github.io/manifest-incubations/#file_handlers-member
+group: progressive-web-app
+compat_features:
+  - html.manifest.file_handlers
+  - api.LaunchParams
+  - api.LaunchParams.files
+  - api.LaunchQueue
+  - api.LaunchQueue.setConsumer
```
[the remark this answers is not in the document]
captainbrosset: Yes, potentially, but I don't think it's implemented yet, although I might be wrong. My understanding is that, over time, the `window.launchQueue` will become useful in more cases that PWAs can be launched (e.g., when used as a share target, or when following a link).

Perhaps what we should do here is just keep the `api.LaunchParams.files` BCD key, and leave the other launch params and launch queue out for now?
Although I don't really know where launch queue would go ... It's made as a generic way to handle what was given to the app when it launched.

features/manifest.yml.dist
```diff
@@ -0,0 +1,110 @@
...
+  # support:
+  #   chrome: "39"
+  #   chrome_android: "39"
+  #   edge: "79"
+  #   firefox_android: "79"
+  #   safari: "17"
+  #   safari_ios: "11.3"
+  - html.manifest.name
+  - html.manifest.short_name
+  - html.manifest.start_url
```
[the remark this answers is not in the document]
captainbrosset: I'd prefer to split it in two instead. One general manifest feature that captures only the more widely available things, knowing that it doesn't unlock all that much dev scenarios, and knowing that it won't be baseline unless Firefox desktop decides to care about PWAs again.
And, another feature that captures the chrome/edge-only things. I'll push a commit to do this.

features/app-file-handling.yml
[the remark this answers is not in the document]
ddbeck: Yeah, if we don't want to deal with it yet, you could have just `api.LaunchParams.files` and we can deal with the API later. It does look like `"launch_handler"` is supported, so maybe that's the base feature?

features/manifest.yml.dist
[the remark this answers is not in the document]
ddbeck: Splitting it up makes sense. Re-request a review when it's done?

features/app-file-handling.yml
```diff
@@ -0,0 +1,11 @@
+name: File handling
```
ddbeck, line 1: Now that we have pluralized all the others, let's make this one consistent.

```suggestion
name: File handlers
```

features/app-file-handling.yml
ddbeck, line 1: To keep the discussions together, I suggest waiting until we've resolved all the other questions first, but after that:

Since we've changed some of the names in this whole PR, I'd be nice to go through and align all the IDs with the file names (e.g., `app-file-handlers`, app-protocol-handlers`, etc.).

features/manifest.yml.dist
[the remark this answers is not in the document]
captainbrosset: Doing more research on this. Here's my analysis so far:

* The browsers we care about here are Chrome, Chrome for Android, Edge, Firefox for Android, Safari, and Safari for iOS.

  * Firefox on desktop is not in this list because it doesn't support web app manifests at all as far as I know. This is expected.

* These keys are supported in all the above browsers, so should be kept in the main manifest feature:

  * `html.manifest.name`
  * `html.manifest.short_name`
  * `html.manifest.start_url`
  * `html.manifest.icons`
  * `html.manifest.theme_color`
  * `html.manifest.scope`

* `html.elements.link.rel.manifest` should absolutely be in the main manifest feature, and supported in the above browsers. It's missing in Safari according to BCD.

  * I think is a mistake, which I'm fixing in https://github.com/mdn/browser-compat-data/pull/24971

* `html.manifest.display` should also be supported in the same browsers, but because a couple of display types are missing in Safari, the key is marked as partial in BCD, which is a problem for us.

  * Trying to fix it in https://github.com/mdn/browser-compat-data/pull/24970

* `html.manifest.id` is missing on Firefox Android, and my vague recollections of past `id`-related discussions tell me that `id` isn't yet a scenario that developers care about.

  * We should probably ignore this one for now.

* `html.manifest.orientation` and `html.manifest.background_color` are not in Safari or Safari iOS.

  * I also tend to think we should ignore these. `background_color` is a feature that's used to give a PWA window some background color before the stylesheet has loaded, therefore avoiding flashes of white while an app is loading. `orientation` lets you define the default orientation of your app when it launches. While nice features, we could separate them later if needed.

* `html.manifest.description` (and `html.manifest.screenshots`, but this one is not in BCD) are supplemental members that are not made for user agents necessarily, but rather for stores that want to list PWAs. The only thing that uses them right now is the "rich install dialog" in Chrome.

  * I think we should ignore for now.

* `html.manifest.serviceworker.*` is, apparently for web payment.

  * This belongs in a separate feature, and I don't have enough subject matter expertise to work on it, so I'll ignore it for now.

captainbrosset:
@ddbeck I think this is now about ready. The BCD changes have landed and we've got them here. There's a few more manifest-related features that can be created, but those can always be added later.

features/manifest.yml
```diff
@@ -0,0 +1,51 @@
...
+  # Ignored from compute_from for now.
+  - html.manifest.description
+
+  # Used to install a web payment service worker on the fly, which is
+  # a separate feature.
+  # Ignored from compute_from for now.
+  - html.manifest.serviceworker
+  - html.manifest.serviceworker.scope
+  - html.manifest.serviceworker.src
+  - html.manifest.serviceworker.use_cache
```
ddbeck, line 51: Should these move to `payment-handler.yml`? If the answer is no, then you can go ahead and merge with this as-is.

features/app-share-target.yml
ddbeck, line 1: Rename → `app-share-targets.yml` (and fix up the dist too), for consistency?

ddbeck (approved):
OK, I'm happy with this too. I have one question and one suggestion, but this is super close to landing. You're welcome to merge this after you've considered (and/or acted on) the suggestions. Thank you!

features/manifest.yml
captainbrosset: I really dont know enough about this part of the manifest to feel confident moving these to another feature. I'd prefer to keep them here for now, until someone with more experience can take a look.

captainbrosset: Captured this in https://github.com/web-platform-dx/web-features/issues/2302

PR #2079, opened 2024-10-24
Fix checklist for BCD upgrades workflow (round 3)

ddbeck:
Apparently the search term needs to be single-quoted.

A follow up to https://github.com/web-platform-dx/web-features/pull/2077.


Elchi3 (approved):
😅

ddbeck:
I think this one is going to fail too. 😅

PR #2102, opened 2024-10-29
Add additional overflow keys

jamesnw:
I set a `compute_from` where needed to maintain the existing baseline, as the new keys are non-impactful.


features/overflow-shorthand.yml
```diff
@@ -25,3 +27,4 @@ compat_features:
   - css.properties.overflow.visible
   - css.types.overflow
   - css.types.overflow.clip
+  - css.types.overflow.overlay
```
captainbrosset, line 30: MDN has this to say about `overlay`:

> Note: The keyword value overlay is a legacy value alias for auto. With overlay, the scroll bars are drawn on top of the content instead of taking up space.

Do you think the value should be marked as deprecated in BCD?

features/text-overflow.yml
```diff
@@ -3,8 +3,12 @@ description: The `text-overflow` CSS property sets how hidden overflow content a
...
 group: css
 caniuse: text-overflow
+status:
+  compute_from: css.properties.text-overflow
 compat_features:
   - css.properties.text-overflow
   - css.properties.text-overflow.clip
   - css.properties.text-overflow.ellipsis
   - svg.global_attributes.text-overflow
+  - css.properties.text-overflow.string
```
captainbrosset, line 13: This particular key seems interesting as it allows you to define your own ellipsis (if I understand correctly).
I'm tempted to make a separate feature for this. What do you think?

features/overflow-shorthand.yml
jamesnw: Good call. I removed and opened https://github.com/mdn/browser-compat-data/pull/24877

features/text-overflow.yml
jamesnw: I added a `Custom ellipsis` feature.

captainbrosset (approved):

features/custom-ellipsis.yml
```diff
@@ -0,0 +1,5 @@
+name: Custom ellipsis
```
ddbeck, line 1: tiniest nit: I think it'd be nice to make this (and the identifier) plural? Since you're (presumably) using this to elide text generally, not for individual texts (otherwise, you'd just manually elide the text).

```suggestion
name: Custom ellipses
```

ddbeck (approved):
One optional suggestion for you, but feel free to merge either way.

PR #2107, opened 2024-10-29
Updates transition features with additional keys

vwallen:
There's a BCD issue related to the `transition-behavior` key: https://github.com/mdn/browser-compat-data/pull/24868


features/transition-behavior.yml
```diff
@@ -2,3 +2,6 @@ name: transition-behavior
 description: "The `transition-behavior: allow-discrete` CSS declaration allows transitions for properties whose animation behavior is discrete. Such properties can't be interpolated and swap from their start value to the end value at 50%."
 spec: https://drafts.csswg.org/css-transitions-2/#transition-behavior-property
 group: transitions
+compat_features:
+  - css.properties.transition-behavior
+  - css.properties.transition.transition-behavior
```
captainbrosset, line 7: Adding this currently demotes the feature status from baseline:low to baseline:false.
Is your plan to wait for the BCD bug you mentioned to get resolved before merging this? (So that the baseline status remains the same).

features/transitions.yml
```diff
@@ -22,5 +22,8 @@ compat_features:
   - css.properties.transition-duration
   - css.properties.transition-property
   - css.properties.transition-property.all
+  - css.properties.transition-property.IDENT_value
   - css.properties.transition-property.none
   - css.properties.transition-timing-function
+  - css.properties.transition-timing-function.jump
+  - css.properties.transition.gradients_can_animate
```
captainbrosset, line 29: Being able to transition gradients seems like a big deal to me.

The only type of gradient animation I'm aware of is moving the background-position of a gradient.
Another hack uses pseudo-elements and opacity, as described here: https://keithjgrant.com/posts/2017/07/transitioning-gradients/.

Apparently it's also possible to use the `@property` rule to animate gradients, see https://stackoverflow.com/questions/6542212/use-css3-transitions-with-gradient-backgrounds/63848864#63848864

However, true interpolation of gradients doesn't seem to be entirely spec'd out yet: https://drafts.csswg.org/css-images-4/#interpolating-gradients

I tend to think that we shouldn't add this key in here, but define a separate feature instead, specifically focused on animating and transitioning gradients background-images.

jamesnw: The concept of transitioning gradients is a big deal, but I don't think this key backs that concept fully. This key is just about whether this would be supported:

```css
div {
  background: linear-gradient(white, black);
  transition: all 1s;
}
div:hover {
  background: linear-gradient(green, red);
}
```

As you say, there's potential workarounds, but those don't have keys, so I'm not sure we have the basis to build a feature here. 

I'm wondering about going the other way- this behavior isn't spec'd, has been removed from the only browser that supported it (and the Safari "preview" value is suspect). Should this be marked as not "standard_track" in BCD?

jamesnw: > So, what does the key correspond to exactly?

The `transitions.gradients_can_animate` key is specifically the code example above, that using the `transition` property on gradients will work. 

> How about we create a feature called something like "gradient animation" with no BCD keys, and no support.

I'd recommend that as a separate PR- I have questions about whether it makes sense to have empty features (when there is no known Interop or other implementer signals), as well as whether this would be a "technique" rather than a feature.

captainbrosset: +1 to doing this in another PR. It can even be an issue instead of a PR. We don't have to do it now. But remembering this discussion would be useful.

As for the question about the feature being empty, I don't think this should be seen as a blocker because we're likely to want to do this more often going forward (once we'll have fully caught up with the existing features). As new features get spec'd, there will be cases where it will make sense to "reserve" a feature ID. Similarly, as intent-to-prototype emails get sent, we'll likely also want to reserve feature IDs, even if no BCD keys exist yet.

Regarding this being a technique rather than a feature, I'm not sure. I think whatever a developer can use to achieve a task, and that they can reasonably talk about/describe in isolation to other things, is a feature to me.

jamesnw: Sounds good. I opened https://github.com/web-platform-dx/web-features/issues/2128 for further conversation.

features/transition-behavior.yml
ddbeck: I merged v5.6.12 today.

jamesnw (approved):
This looks good to me- @captainbrosset do you want to take a second look as well?

captainbrosset (approved):

PR #2108, opened 2024-10-29
Adds additional key to cubic bezier easing

vwallen:
Individual easing functions each have their own features, but the root `css.types.easing-function` has not been included yet. My initial thought was to create a feature for that, but instead have opted to include it with the `cubic-bezier` feature as the default value `ease` is an alias for this function.

See this BCD issue for including various `cubic-bezier` aliases: https://github.com/mdn/browser-compat-data/issues/24682


ddbeck (approved):
Yes, as long as https://github.com/mdn/browser-compat-data/issues/24682 is unresolved, then this is probably the only thing that makes sense with the existing data.

PR #2109, opened 2024-10-29
Adds attr features

vwallen:
This is an unusual case as the feature has high support, but only as part of `contents` declarations. General usage per the spec is very limited. 

I chose to split the supported declaration into its own feature. The `contents: attr()` feature would be reasonable to link to `caniuse: css-gencontent` but that is already in use in `before-after`. if duplicate caniuse values are acceptable, then this should be added.


features/contents-attr.yml
```diff
@@ -0,0 +1,6 @@
+name: "contents: attr()"
```
captainbrosset, line 1:
```suggestion
name: "content: attr()"
```

features/contents-attr.yml
```diff
@@ -0,0 +1,6 @@
...
+description: "The `contents: attr()` CSS declaration sets an element's contents to the value of the specified attribute."
```
captainbrosset, line 2:
```suggestion
description: "The `content: attr()` CSS declaration sets a `::before` or `::after` pseudo-element's content to the value of the specified HTML attribute."
```

features/attr.yml
```diff
@@ -0,0 +1,19 @@
+name: attr
+description: The `attr()` CSS function returns the value of the specified attribute of an element. A default value may be provided if the attribute is missing or does not parse as the required type.
```
captainbrosset, line 2: Being able to use `attr()` for things other than string values and in properties other than `content` has been a highly requested feature for years.

See https://issues.chromium.org/issues/40320391 and https://github.com/web-platform-tests/interop/issues/521 for example.

Because of this, I'd love it if we changed this feature's description to make it obvious to devs that the feature does cover this highly requested use case. Adding an example might do it:

```suggestion
description: The `attr()` CSS function returns the value of an HTML element's attribute, which can then be used as a typed value in a CSS declaration (types include length, number, angle, color, …). For example `width: attr(length);`. A default value may be provided if the attribute is missing or invalid.
```

captainbrosset (changes requested):

captainbrosset (approved):

features/attr.yml
```diff
@@ -0,0 +1,19 @@
+name: attr
```
ddbeck, line 1: For consistency with other CSS notation features like this:

```suggestion
name: attr()
```

features/attr.yml
```diff
@@ -0,0 +1,19 @@
...
+description: "The `attr()` CSS function returns the value of an HTML element's attribute, which can then be used as a typed value in a CSS declaration (types include length, number, angle, color, etc.). For example `width: attr(length);`. A default value may be provided if the attribute is missing or invalid."
```
ddbeck, line 2: A few ideas here:

- Distinguish this from the content-only form of the feature
- Avoiding non-literal code samples (`width: attr(length);` — length is not, to the best of my knowledge, a defined attribute)
- Active voice on the fallback behavior

```suggestion
description: "The `attr()` CSS function, in the context of any property, returns the value of an attribute of an HTML element, with the option to return it as a specific type or unit. You can set a default value for missing or invalid attributes."
```

features/contents-attr.yml
```diff
@@ -0,0 +1,6 @@
+name: "content: attr()"
```
ddbeck, line 1: I'm suggesting this so that this feature sorts together with its broader counterpart:

```suggestion
name: attr() (content only)
```

features/contents-attr.yml
```diff
@@ -0,0 +1,6 @@
...
+description: "The `content: attr()` CSS declaration sets a `::before` or `::after` pseudo-element's content to the value of the specified HTML attribute."
```
ddbeck, line 2: For consistency with the other description:

```suggestion
description: "The `attr()` CSS function, in `content` property declarations, sets a `::before` or `::after` pseudo-element's content to the value of the specified HTML attribute."
```

features/contents-attr.yml
ddbeck, line 1: Let's rename this to `attr-contents`

ddbeck (commented):
Some ideas for you. Thank you!

captainbrosset (approved):

ddbeck (approved):

PR #2110, opened 2024-10-29
Adds interpolate-size feature

vwallen:


features/interpolate-size.yml
```diff
@@ -0,0 +1,10 @@
+name: interpolate-size
+description: The `interpolate-size` CSS property allows animations and transitions to interpolate between a numeric value and keyword values such as `auto` or `fit-content`.
```
ddbeck, line 2: Since you can explicitly _not_ allow this behavior, I think we have to say that this thing is where the author chooses a behavior:

```suggestion
description: The `interpolate-size` CSS property sets whether animations and transitions interpolate between a numeric value and a keyword value, such as from a fixed length to `auto` or `fit-content`.
```

ddbeck (approved):
Optional description suggestion, but otherwise looking good. Thank you!

PR #2111, opened 2024-10-29
Adds numeric, string, position, and url CSS types

vwallen:
Adding a number of fundamental and domain specific CSS types. Also creates a `types` group under `css`.

A few include `-type` as part of the name to disambiguate from similarly named objects.

The top level dimension type is included with numeric types, but it could also reasonably be its own feature (with or with out including sub-types such as `angel`). As included here, it's included with other numeric types and each sub-type has its own feature.


features/url-type.yml
```diff
@@ -0,0 +1,6 @@
+name: url (CSS type)
+description: The `url` CSS type represents an external resource.
+spec: https://drafts.csswg.org/css-values-4/#string-value
```
jamesnw, line 3:
```suggestion
spec: https://drafts.csswg.org/css-values-4/#urls
```

jamesnw (approved):
This feature set looks like a good way of communicating these keys, but I think I'd like at least one other opinion on that.

ddbeck (commented):
OK, so I'm a little bit worried about CSS types. My concern is that a type, by itself, doesn't really _do_ anything. Is an abstract type, that has no API of its own, a feature? I'm inclined to say no.

We have a few CSS types (e.g., conic gradients) already but, as far as I know, they are pretty closely circumscribed, at least with the description. "The `conic-gradient()` and `repeating-conic-gradient()` CSS functions **create backgrounds**…" They're "types" yes, but implicitly they're types for the `background` property etc.

This implies to me that some of these types, such as they are, ought to live with actual properties. Some of this is an upstream problem of BCD (i.e., we ought to de-mixin types and situate type support with specific properties), but we're stuck with confronting it here. And ideally we'd be able to put keys into multiple features. But without the option to do that, I think it makes sense to put them with the oldest properties they correspond to (if they're foundational) or reference them (if they're distinct later additions).

That said, this is a worry, not an edict. I'm also no CSS expert. I'm going to mark this up with some thoughts, but I'm not strongly holding them. Please tell me if I've missed any marks.

vwallen:
Converting this to `draft` to retain the conversation. Will take up these changes and split up into appropriate new PRs

jamesnw:
The split out PRs have been merged 🎉

PR #2120, opened 2024-10-30
Make custom ellipses identifier match name

ddbeck:
This is not semver major because we haven't released it yet.


jamesnw (approved):
Sorry about that! You'll need to re-run dist to get the tests to pass. Thanks for catching.

PR #2129, opened 2024-10-31
Add CSS.supports() and missing CSSOM keys

jamesnw:


features/css-supports.yml
```diff
@@ -0,0 +1,7 @@
+name: CSS.supports()
+description: The `CSS.supports()` static method returns whether the browser supports a given CSS feature.
```
ddbeck, line 2: Maybe we can be more specific here? I don't know if this is too much:

```suggestion
description: The `CSS.supports()` static method returns whether the browser supports a CSS declaration (given by two arguments, a property and value) or an `@supports` at-rule condition string.
```

ddbeck (approved):
Thank you! One optional description suggestion for you

PR #2135, opened 2024-11-01
Add rhythmic sizing

jamesnw:
This could also be called `line-height-step` since that's the only implemented (behind a flag) key, but I think the feature developers will look for is the full feature. As more is implemented, the description will need to be updated, or potentially split at that time.


features/rhythmic-sizing.yml
```diff
@@ -0,0 +1,6 @@
+name: Rhythmic sizing
+description: The `line-height-step` CSS property sets the spacing between body text baselines and rounds the spacing for taller lines like headings up to the next multiple of the body text spacing to create vertical rhythm. Also called rhythmic sizing.
```
ddbeck, line 2: I think "creat[ing] vertical rhythm" is the central aim of this, so I think it makes sense to bring it forward in the description. I don't think we need an also-known-as when it's already the name of the feature.

```suggestion
description: The `line-height-step` CSS property creates consistent vertical rhythm by setting the spacing between body text baselines. It rounds taller lines like headings up to the next multiple of the body text spacing.
```

ddbeck (approved):
I like this. Optional description suggestion for you.

Issue #2133, opened 2024-11-01
Define how the core browser set was determined


tomayac:
Based on an [interesting question asked on Mastodon](https://techhub.social/@develwithoutacause/113398726006901259) by @dgp1130. I was intrigued, but couldn't find based on what criteria the core browser set was determined.

ddbeck:
Hi @tomayac, I'll do my best to answer your as succinctly as I can. I'll respond to the more specific "What if…?" question separately.

In the strictest terms, the core browser set was selected by the web-features [owners group](https://github.com/web-platform-dx/web-features/blob/main/GOVERNANCE.md#list-of-current-owners) in a consensus-seeking decision-making process. We worked on the overall definition, including the browser set and releases timeline, until we reached an actual consensus. This process took about 6 months. Most discussions happened in WebDX meetings, the WebDX matrix chat, and [in issues labeled here](https://github.com/web-platform-dx/web-features/issues?q=is%3Aissue+label%3A%22baseline+definition%22+is%3Aclosed).

More broadly, we picked the browser set to accommodate a number of requirements and constraints:

- Developer expectations (e.g., through surveys, direct feedback on an earlier definition, and representation in the owners group)
- Developer behavior (e.g., looking at, for instance, open source browserslist configs)
- Baseline consumer expectations (i.e., what would work for caniuse and MDN)
- Upstream data availability and quality (i.e., what data sources were actually available to us, for support and usage)

A lot of work went into trying to find some statistical rule (e.g., using StatCounter usage thresholds) that would encompass everything. Unfortunately, data quality and other issues prevented that from happening. So the governance group reviewed all the information and made a judgment about a browser set that we hoped would be satisfactory to developers.

The formal [Baseline definition document](https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md#baseline) contains a lot of information about our shared motivations, at the time it was adopted.

ddbeck:
To @dgp1130's question, which I'll excerpt here:

> Presumably the browser set would evolve if a new browser gets enough market share to justify its inclusion.
>
> What happens if that browser is missing a particular feature supported by the other browsers?
>
> Does that imply that Baseline 2025 could hypothetically include *less* features than Baseline 2024 if it included a new browser? I feel like *losing* features is something you wouldn't expect in Baseline, much like you don't expect a version to go backwards.
>
> Is there any expected outcome for this particular scenario?

[...]

PR #2142, opened 2024-11-04
Baseline defintion: flag undefined status situations

ddbeck:
https://github.com/web-platform-dx/web-features/issues/2133 prompted me to add two non-normative notes to the Baseline definition, explicitly highlighting some (IMO, intentionally) undefined qualities of the definition.

### DIFF
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -168,6 +168,15 @@ The _core browser set_ shall be defined as:
 * Mozilla Firefox (Android)
 * Mozilla Firefox (desktop)
 
+> [!NOTE]
+> How Baseline statuses change based on the introduction of a new browser to the existing core browser set is not yet defined.
+> If you need this to adopt Baseline in your tool or process, file an issue describing your use case.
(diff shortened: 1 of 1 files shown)


ddbeck:
If we wanted to, we could make some statement about what we expect to happen if we were to add a browser and that browser did not implement the full Baseline feature set at the time the browser was added. We could do that as part of this PR or in a later one.

Some ideas for that, from strongest to weakest:

- Freeze the browser set for those features (i.e., only judge existing Baseline low/high features on the "old" browser set)
- Declare some minimum transition period (e.g., additions to the core browser set take 30 months to negatively effect existing Baseline features)
- Offer a warning period (e.g., we won't change the core browser set in a way that would remove features from the Baseline feature set without giving at least 3 months notice)

captainbrosset (approved):
Thanks for adding this, Daniel. At least adding some notes about how the situation is not yet defined is useful.

ddbeck:
I don't think there's any need or interest in expanding upon these notes, so I'm going to merge this now. Thank you!

Tonyacr7 (approved):

Tonyacr7 (approved):

PR #2148, opened 2024-11-05
Adds calc-size feature

vwallen:
Split out from original PR here: https://github.com/web-platform-dx/web-features/pull/2111


features/calc-size.yml
```diff
@@ -0,0 +1,6 @@
+name: calc-size()
+description: The `calc-size()` CSS function computes mathematical expressions that include `height` and `weight` keyword values, for example `calc-size(fit-content, size / 2)`. Not to be confused with the `calc()` CSS function, which cannot use keyword values.
```
ddbeck, line 2:
```suggestion
description: The `calc-size()` CSS function computes mathematical expressions that include `height` and `width` keyword values, for example `calc-size(fit-content, size / 2)`. Not to be confused with the `calc()` CSS function, which cannot use keyword values.
```

ddbeck (approved):
One little fix, but otherwise ready to merge. Thank you!

PR #2149, opened 2024-11-05
Adds CSS abstract types to existing property features

vwallen:
Split off from existing PR: https://github.com/web-platform-dx/web-features/pull/2111

Keys related to abstract types are distributed to existing property features most associated with that type, where that would not change the support status of the feature

A few orphan keys remain with no associated properties:
```
  - css.types.frequency
  - css.types.frequency-percentage
  - css.types.frequency.hz
  - css.types.frequency.khz
  - css.types.time-percentage
```


jamesnw (approved):
This distribution makes sense to me. 

For the remaining keys-

- Frequency, could they be their own feature?
-  time-percentage- could that go in to `animations-css` with `time`?

Alternatively, should they get added to some sort of CSS catchall? Or removed from BCD, since they aren't used at all?

@ddbeck What should we do about the keys that are in BCD, and in the spec, but are not actually usable anywhere?

ddbeck:
> @ddbeck What should we do about the keys that are in BCD, and in the spec, but are not actually usable anywhere?

I've opened https://github.com/mdn/browser-compat-data/pull/24969 to remove the `<frequency>` and `<frequency-percentage>` types. I'm very confident they do not belong in BCD, at all.

That leaves `<time-percentage>`. Is that also not used in any actual properties? If so, I'd be happy to open a PR against BCD to remove that too.

ddbeck (approved):
This all looks right to me. If `<time-percentage>` is definitely unused, then this is ready to merge. Thank you!

PR #2151, opened 2024-11-05
Adds additional font related keys

vwallen:
Most keys fit into existing features. `colorv0` is given its own feature to match `colrv1` and because it has its own `caniuse` key


features/font-synthesis.yml
```diff
@@ -2,3 +2,16 @@ name: font-synthesis
 description: The `font-synthesis` CSS shorthand property disables all font synthesis except the given kinds. To disable a specific kind of font synthesis, instead use the longhand properties such as `font-synthesis-style` and `font-synthesis-weight`.
 spec: https://drafts.csswg.org/css-fonts-4/#font-synthesis
 group: font-synthesis
+status:
+  compute_from: css.properties.font-synthesis
+compat_features:
+  - css.properties.font-synthesis
+  # The font-synthesis shorthand and longhand properties work inversely (e.g.
+  # `font-synthesis-wieght: none` affects weight, while `font-synthesis: weight`
```
jamesnw, line 10:
```suggestion
  # `font-synthesis-weight: none` affects weight, while `font-synthesis: weight`
```

jamesnw (approved):
This looks good to me, but I'll leave open for a bit in case a non-Oddbirder wants to review.

features/colrv0.yml
```diff
@@ -0,0 +1,7 @@
+name: COLRv0
+description: COLRv0 is a font format that supports multi-color glyphs.
```
ddbeck, line 2: Can we say something, either here or in v1, that distinguishes the two formats?

vwallen: Updated

ddbeck (approved):
Perfect. Thank you!

PR #2167, opened 2024-11-06
Add text-box feature

jamesnw:
Builds on https://github.com/web-platform-dx/web-features/pull/1006, but I made the feature about the shorthand and both longhand properties.


ddbeck (approved):
This is nice. Thank you!

PR #2168, opened 2024-11-06
Selectors - and indexed pseudo classes

jamesnw:


features/nth-child.yml
```diff
@@ -1,6 +1,19 @@
 name: ":nth-child() (initial support)"
-description: "The `:nth-child()` and `:nth-last-child()` CSS functional pseudo-classes match elements by the relative position of elements (first, second, third, fourth, and so on), counted from the first or last sibling."
+description: "The `:nth-child()` and `:nth-last-child()` CSS functional pseudo-classes match elements by their relative position, counted from the first or last sibling. The `:first-child`, `:last-child`, and `:only-child` pseudo-classes represent common use cases."
```
captainbrosset, line 2: Position within what?
Possible suggestion:

```suggestion
description: "The `:nth-child()` and `:nth-last-child()` CSS functional pseudo-classes match elements based on their relative position within a list of elements, counted from the first or last sibling. The `:first-child`, `:last-child`, and `:only-child` pseudo-classes represent common use cases."
```

Also, "represent common use cases" feels improvable. This doesn't describe what these pseudos do.

features/nth-child.yml.dist
```diff
@@ -40,3 +79,16 @@ compat_features:
...
+  #   chrome: "57"
+  #   chrome_android: "57"
+  #   edge: "79"
+  #   firefox: "52"
+  #   firefox_android: "52"
+  - css.selectors.first-child.no_parent_required
+  - css.selectors.last-child.no_parent_required
+  - css.selectors.nth-child.no_parent_required
+  - css.selectors.nth-last-child.no_parent_required
+  - css.selectors.only-child.no_parent_required
```
captainbrosset, line 94: I'm unclear as to what these subfeatures cover. Can you help me understand?
Would they make sense as a separate feature?

captainbrosset (changes requested):

features/nth-child.yml.dist
jamesnw: Selectors level 3 had text referring to the children's parent, but selectors level 4 adds the note-

```
As there was no reason to exclude them from matching elements without parents, or with non-element parents, they have been rephrased to refer to an element’s relative index amongst its siblings.
```

For instance, the document root matches `:first-child` despite having no parent. 

This doesn't seem like a feature to me. It also looks like Safari data might be incorrect, as `document.documentElement.matches(":first-child")` returns true in Safari.

features/nth-child.yml
```diff
@@ -1,6 +1,19 @@
 name: ":nth-child() (initial support)"
```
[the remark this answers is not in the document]
jamesnw: It looks like this was added in the original feature definition in https://github.com/web-platform-dx/web-features/pull/1130, and used to differentiate with the `nth-child-of` feature.

features/nth-of-type.yml
```diff
@@ -0,0 +1,10 @@
+name: ":nth-of-type() pseudo-classes"
+description: "The `:nth-of-type()` CSS functional pseudo-classes match elements based on their position among siblings of the same type, counted from the first or last sibling. The `:only-of-type` pseudo-class matches any element that is the only child of its type, of its parent."
```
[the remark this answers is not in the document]
jamesnw: I updated this- it's a lot to cram into 300 characters, and I'd welcome other verbage.

features/nth-child.yml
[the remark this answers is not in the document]
captainbrosset: Gotcha. I would still remove this suffix here. nth-child is, in my mind, different from the nth-child-of-type thing, and the latter is not merely an evolution of the former.

features/nth-child.yml.dist
captainbrosset: Thanks! Keeping them here, ignored, seems right then.

PR #2169, opened 2024-11-06
Selectors- add baseline false features

jamesnw:
See #2170 for more info


features/target-within.yml
```diff
@@ -0,0 +1,6 @@
+name: :target-within
+description: The `:target-within` CSS pseudo-class selects the element with an id matching the URL fragment and its ancestors.
```
ddbeck, line 2:
```suggestion
description: The `:target-within` CSS pseudo-class matches the element with an ID matching the URL fragment and its ancestors.
```

features/time-relative-selectors.yml
```diff
@@ -0,0 +1,8 @@
+name: Time-dimensional pseudo-selectors
+description: The `:future` and `:past` CSS pseudo-selectors style upcoming or prior elements during media playback.
```
ddbeck, line 2:
```suggestion
description: The `:future` and `:past` CSS pseudo-classes match upcoming or prior elements during media playback.
```

features/time-relative-selectors.yml
```diff
@@ -0,0 +1,8 @@
+name: Time-dimensional pseudo-selectors
```
ddbeck, line 1: Neither the name nor the description has terminology in common with the file name. Can something change here to make them line up with each other? I kinda like "time-relative" as being more approachable, but either is fine.

ddbeck (approved):
I've noted some tiny issues with an ID and some descriptions. Feel free to merge when you've resolved those. Thank you!

features/time-relative-selectors.yml
jamesnw: Good point- I agree that time-relative is more approachable and went with that.

PR #2171, opened 2024-11-06
Selectors- Add baseline high

jamesnw:
See #2170 for more info


features/link-selectors.yml
```diff
@@ -0,0 +1,15 @@
+name: Link selectors
+description: The `:link` CSS pseudo-class matches unvisited links, `:visited` matches visited links, and `:any-link` matches both.
+spec: https://drafts.csswg.org/selectors-4/#location
+caniuse: css-any-link
+group: selectors
+status:
+  compute_from: css.selectors.any-link
+compat_features:
+  - css.selectors.any-link
```
ddbeck, line 9: I had no idea this one existed—every day I'm learning about some new-to-me thing on this project. Cool, but slightly concerning.

features/scope-pseudo.yml
```diff
@@ -0,0 +1,7 @@
+name: :scope pseudo-class
```
ddbeck, line 1: I'm assuming we're only saying pseudo class because of `@scope`:

```suggestion
name: :scope (pseudo-class)
```

features/scope-pseudo.yml
```diff
@@ -0,0 +1,7 @@
...
+description: The `:scope` CSS pseudo-class represents the scoping root, for instance the element that `querySelector()` is called on.
```
ddbeck, line 2:
```suggestion
description: The `:scope` CSS pseudo-class matches the scoping root, for instance the element that `querySelector()` is called on.
```

ddbeck (approved):
Some optional description suggestions here, but I'm happy with this merging whenever you think it's ready.

features/scope-pseudo.yml
captainbrosset: This description should also mention that the pseudo-class is useful when used inside `@scope` blocks.

MDN:
> When used inside a [@scope](https://developer.mozilla.org/en-US/docs/Web/CSS/@scope) block, :scope matches the block's defined scope root.

https://developer.mozilla.org/en-US/docs/Web/CSS/:scope

captainbrosset (changes requested):

features/scope-pseudo.yml
jamesnw: Good catch. Added.

PR #2186, opened 2024-11-07
Add more keys to flexbox

jamesnw:
One thing to note- there are some property keys here like `css.properties.align-content` or `css.properties.align-items` that also work in `grid` (and elsewhere). I chose to put them with flexbox, since that was where they were introduced.


captainbrosset (approved):

captainbrosset:
I agree that's probably the right thing to do now, but it's worth asking ourselves whether we should re-think about this.
There are multiple cases where BCD keys fit into multiple different features.

The place/align/justify CSS properties can/will be used in grid, flex, multi-col, block.
The launchHandler API can be used in PWA file handling, but soon in other types of PWA opening scenarios too.

jamesnw:
> There are multiple cases where BCD keys fit into multiple different features.

Agreed. Linking this to #1173 as another use case.

PR #2189, opened 2024-11-07
Add align-content in block layout

jamesnw:
This is handled a bit different than the other `align-content` keys. The others are with their layout context (grid/flex), but we don't have a block layout feature (and don't need one, I don't think).

In addition, I think this is a bit of a standalone feature, for instance, given this post- https://web.dev/blog/align-content-block. There are also a few SO posts asking about `align-content` in block layouts.


features/align-content-block.yml
```diff
@@ -0,0 +1,6 @@
+name: align-content in block layouts
+description: The `align-content` property vertically aligns content in block layouts, and no longer requires a flex or grid layout.
```
ddbeck, line 2: Optional description suggestion: "no longer" feels like something that will age poorly. It's a good distinction to make here, but maybe we can do something like this that'll feel less like a release notes entry:

```suggestion
description: The `align-content` property vertically aligns content in block layouts, like it does in flex or grid layouts.
```

ddbeck (approved):

features/align-content-block.yml
jamesnw: Thanks! I was struggling with that exactly, and your update works great!

PR #2191, opened 2024-11-07
Add a semi-automated release workflow

ddbeck:
This PR adds a workflow to run on demand for new releases. It's more of a proof of concept at this point—there's still some work to do to automatically clean up the release notes, prepopulate with some stats, and so on. But at very least, this means that one owner can cut a release, without the need for the (trivial) review of the `package.json` file.

Example of the proposed GitHub workflow dispatch UI


Elchi3 (approved):
I'm very excited about this! :)

PR #2197, opened 2024-11-08
Adds CSS filter effects

vwallen:


features/filter.yml
```diff
@@ -0,0 +1,18 @@
+name: filter
+description: The `filter` CSS property applies one or more graphic effects to an element. Filter functions, such as `blur()` and `drop-shadow()`, can be used or combined to produce different effects.
```
vwallen, line 2: This use of "applies" rather than "sets" is off guidelines, but I think is warranted vs less clear alternatives such as "The `filter` CSS property sets a list of graphic effects to apply to an element."

features/svg-filters.yml
vwallen, line 1: Because keys were added to such a long list, these have been sorted for easier reference in the future

features/filter.yml
ddbeck: Yes, this is a good deviation from the rules. "Applying an effect" seems more idiomatic than "setting an effect" but I don't think that would be true for other uses. 👍 

One optional suggestion for the "can be used" text:

```suggestion
description: The `filter` CSS property applies one or more graphic effects to an element. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects.
```

ddbeck (approved):

PR #2206, opened 2024-11-11
Adds background features and updates keys on existing background features

vwallen:


features/background-origin.yml
```diff
@@ -0,0 +1,10 @@
+name: background-origin
+description: The `background-origin` CSS property sets the background origin position relative to the border and padding of an element.
```
ddbeck, line 2: Suggestion to avoid circularity:

```suggestion
description: The `background-origin` CSS property sets the background starting position relative to the border and padding of an element.
```

features/background-position.yml
```diff
@@ -0,0 +1,22 @@
+name: background-position
+description: The `background-position` CSS property sets the initial position of background images relative to the background origin.
```
ddbeck, line 2: I'm not 100% certain about this one, but I kinda think of this as the property that offsets the background.

```suggestion
description: The `background-position` CSS property offsets the initial position of background images relative to the background origin.
```

ddbeck (approved):
I'm happy with this. I have some optional description suggestions, but this is ready to merge.

Issue #1894, opened 2024-10-02
compute_from suppresses deprecation warnings


jamesnw:
Minimal nonsense reproduction, `api.SVGAnimateColorElement` is deprecated, `html.elements.del` is not.

```yml
name: deprecated
description: deprecated
compat_features:
  - api.SVGAnimateColorElement
  - html.elements.del
```

`npm run dist -- path/to/file` prints the warning:

```
warn: [feature id]: contains at least one deprecated compat feature and can never be Baseline. Was this intentional?
```

Adding a status.compute_from suppresses the deprecation warning. 

```yml
name: deprecated
description: deprecated
status:
  compute_from: html.elements.del
compat_features:
  - api.SVGAnimateColorElement
  - html.elements.del
```

`npm run dist -- path/to/file` does not print a warning.

Ideally, it would also show the deprecated keys, but it looks like that is a bit harder to surface.

PR #2210, opened 2024-11-12
Error on deprecated compat keys

ddbeck:
Until we have a representation for discouraged features, we should treat using deprecated BCD keys as an error.

This also silences the compat key tag warning, which isn't very useful right now.


jamesnw (approved):

ddbeck:
OK, I've changed this slightly to warn for drafts, but error for published features. This means that changing `ink` is required to merge this.

ddbeck:
Also, note that this is not a complete resolution. We might still have other problematic keys. See https://github.com/web-platform-dx/web-features/issues/1894

ddbeck:
And we've now got https://github.com/web-platform-dx/web-features/pull/2227 to fix ink itself

Elchi3 (approved):

PR #2212, opened 2024-11-12
Removes orphaned drafts

vwallen:
These drafts are not updated or created by `update-drafts`. In the case of `css-grid-2.yml` all of the keys are already present in existing features.


jamesnw (approved):
`http.headers.Expect-CT` in rfc9163 was marked deprecated 2 months ago. I'm unsure why these 2 specs weren't deleted.

ddbeck:
> I'm unsure why these 2 specs weren't deleted.

It's because the update-drafts script skips over specs that have zero standard, non-deprecated keys left. I opened https://github.com/web-platform-dx/web-features/pull/2212 to fix this.

ddbeck:
Actually, #2218 only fixed the RFC spec. I'm not sure what's going on with the css-grid-2 spec.

ddbeck:
OK, I figured it out. The update-drafts script normalizes CSS URLs to the latest spec level. So when the latest advances, the previous level gets left behind. That's a more dramatic change; I'll open another PR to show what would happen if we changed that.

https://github.com/web-platform-dx/web-features/blob/edcd078d434cb2c4c4d476536f2cd24b1fc3cf69/scripts/update-drafts.ts#L64-L67

ddbeck:
https://github.com/web-platform-dx/web-features/pull/2225 would fix (and close) this.

PR #2213, opened 2024-11-12
Advance review date on the Baseline definition

ddbeck:
Today, the web-features owners group met for our quarterly governance call. We reviewed the definition and decided to make no changes today and bring the definition up for review again next year.

That said, we acknowledge that work at the end of this year and the beginning of next (such as dealing with discouraged features) might lead to changes to the Baseline definition sooner. In any case, we'll revisit the definition in about six months.

(We agreed on the call to this date change, so I wasn't planning to request reviews from the entire owners group.)


atopal (approved):
LGTM!

Elchi3 (approved):

PR #2226, opened 2024-11-13
Uncomment keys excluded for non-real values, now fixed

ddbeck:
A few such keys remain in drafts, that I expect to be taken care of in https://github.com/web-platform-dx/web-features/pull/2044.


Elchi3 (approved):
Very cool! 👍

PR #2228, opened 2024-11-13
Check every key (regardless of compute_from) for discouraged condition

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/1894. If `compute_from` is set, then explicitly check every key's `deprecated` value.

~~Right now this is not _great_—it's quite slow. I need to do some sort of refactoring here to not run computeBaseline on every key twice. It's too late in the day for me to think that through.~~


Elchi3 (approved):

jamesnw (approved):
🚀

ddbeck:
OK, then if we're going down this route we have a little bit of work to do. These are the keys currently failing the test:

- [ ] api.FontFaceSet.FontFaceSet 
- [ ] api.CanMakePaymentEvent.methodData 
- [ ] api.CanMakePaymentEvent.modifiers 
- [ ] api.CanMakePaymentEvent.paymentRequestOrigin 
- [ ] api.CanMakePaymentEvent.topOrigin 
- [ ] api.PaymentManager.instruments 
- [ ] api.PaymentRequestEvent.instrumentKey 
- [ ] api.PaymentRequest.merchantvalidation_event 
- [ ] api.PaymentRequest.shippingAddress 
- [ ] api.PaymentRequest.shippingOption 
- [ ] api.PaymentRequest.shippingType 
- [ ] api.PaymentRequest.shippingaddresschange_event 
- [ ] api.PaymentRequest.shippingoptionchange_event 
- [ ] css.at-rules.media.prefers-color-scheme.no-preference 
- [ ] api.SVGAnimateColorElement

scripts/dist.ts
```diff
@@ -238,12 +238,12 @@ function toDist(sourcePath: string): YAML.Document {
 
     if (!source.draft_date) {
       logger.error(
-        `${id}: contains at least one deprecated compat feature and can never be Baseline. This is forbidden for published features.`,
+        `${id}: contains at least one deprecated compat feature. This is forbidden for published features.`,
```
ddbeck, line 241: This is kinda annoying in some cases, because you never get the one key that caused it. I should refactor this a bit, to conditionally check deprecated for each key up here.

jamesnw:
@vwallen Can you open PRs for the remaining keys in [this comment](https://github.com/web-platform-dx/web-features/pull/2228#issuecomment-2476726277)?

jamesnw:
@ddbeck The blocking keys have been removed, and it looks like none have been added in the interim.

ddbeck:
Great. Let's merge this now.

PR #2229, opened 2024-11-13
Add initial CSS pseudoclasses

jamesnw:
This splits off the pseudoclasses from #2170.


features/root.yml
```diff
@@ -0,0 +1,6 @@
+name: :root
+description: The `:root` pseudo-class matches the root element of the document.
```
ddbeck, line 2: To make this a little more descriptive:

```suggestion
description: The `:root` pseudo-class matches the root element of the document, usually the `<html>` element.
```

features/target.yml
```diff
@@ -0,0 +1,6 @@
+name: :target
+description: The `:target` CSS pseudo-class matches the element with an id matching the URL fragment.
```
ddbeck, line 2: nit, casing

```suggestion
description: The `:target` CSS pseudo-class matches the element with an ID matching the URL fragment.
```

ddbeck (approved):
I like these. Some tiny suggestions for the descriptions, but you're welcome to merge this. Thank you!

PR #2232, opened 2024-11-14
Add WebRTC

autonome:


autonome:
@ddbeck ready for a first pass on the approach.

autonome:
There are also a few candidates for sub-features if we want to cleave those off into their own thing, eg the error stuff.

ddbeck (approved):
Thank you!

PR #2233, opened 2024-11-14
Add Browserlist config

pepelsbey:
Adds the “Using Baseline to target browser support” section to the “[Baseline in your project](https://web-platform-dx.github.io/web-features/use-baseline/)” page.

image

### DIFF
--- a/gh-pages/src/use-baseline.md
+++ b/gh-pages/src/use-baseline.md
@@ -13,6 +13,25 @@ You can use Baseline to:
 
 When visiting sites like [Can I Use](https://caniuse.com), [MDN Web Docs](https://developer.mozilla.org/), and [RUM Archive Insights](https://rumarchive.com/insights/#baseline) check the status of a feature to learn if it available across browsers.
 
+## Using Baseline to target browser support
+
+If you want to use Baseline status to target browsers you support, you can use the following [Browserlist](https://browsersl.ist/) configuration in your `package.json` file:
+
+```js
+"browserslist": [
+  "Chrome > 0 and last 2.5 years",
+  "ChromeAndroid > 0 and last 2.5 years",
+  "Edge > 0 and last 2.5 years",
+  "Firefox > 0 and last 2.5 years",
+  "FirefoxAndroid > 0 and last 2.5 years",
+  "Safari > 0 and last 2.5 years",
+  "iOS > 0 and last 2.5 years",
+  "not dead"
+]
+```
+
+This configuration matches the Baseline widely available stage. Tools like [Babel](https://babeljs.io/), [PostCSS Preset Env](https://preset-env.cssdb.org/), [Lightning CSS](https://lightningcss.dev/), and others will use it to determine the level of browser support and transpile your code or include needed polyfills.
+
 ## Let others know when a feature works across browsers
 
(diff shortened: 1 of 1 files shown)


tidoust:
I note the plan to transition @tonypconway's work on [bl2bl](https://github.com/tonypconway/bl2bl/) to under web-platform-dx. My understanding is that the Browserlist configuration is slighty more complex than a `last 2.5 years` so as not to miss the last version released right before that. In any case, it will be useful to link to the code from the "Baseline in your project" page.

pepelsbey:
The main idea behind this PR is to offer a simple “set it and forget it” config. This is the best chance for Baseline to get traction and usage in existing toolchains. If the config doesn’t give the true “widely available” set and could be improved, that would be preferable. If not, then contributing to Browserlist to extend its capabilities is an option, too.

tonypconway:
Thanks for looping me in @tidoust. 

> a simple “set it and forget it” config. This is the best chance for Baseline to get traction and usage in existing toolchains

Totally agree @pepelsbey . As François mentioned, and I explained [here](https://github.com/browserslist/browserslist/issues/771#issuecomment-2188558343) and @ddbeck confirmed [here](https://github.com/browserslist/browserslist/issues/771#issuecomment-2202729885), just using `[Browser] last 2.5 years` doesn't actually give you the right feature set for Widely Available because you need to include the last version of each core browser _before_ that cut off date.  If you don't, you'll end up unnecessarily polyfilling, or accidentally excluding users if you're using browserslist to power a "your browser is not supported" message.  Same goes for using a particular year, you can't just have `[Browser] since 2020-12-31` for the 2020 feature set - you'd only include browsers _after_ that date, when you need to include the last one before.  I don't think there's a way to get those browser versions without either changing browserslist, or calculating your min versions and updating your `browserslist` config on the fly.

There was already [a PR created for browserlist to support these queries](https://github.com/browserslist/browserslist/pull/833) but the maintainers are currently not willing to move forward with it because they have concerns around how users of browsers like UC Browser and QQ are represented.  I'm working on a few things to address this:

1. the [bl2bl](https://github.com/tonypconway/bl2bl/) module that Francois alluded to.  As a stop gap, this should address the "set it and forget it" piece you mentioned.  Simply add the module to your project, add a config to your package.json, and add some commands to your build process.  Please do take a look at it and feel free to add issues or make PRs.  It's a bit muddled at the moment and I'd love to get it cleaned up a bit before we move it over to this org.  This already has support for including versions of browsers that implement Chromium and appear in [MDN browser-compat-data's browsers data](https://github.com/mdn/browser-compat-data/tree/main/browsers).
2. I'm working with the very generous @logocomune who owns useragents.io on building a best effort mapping of Chromium downstream browsers like UC, QQ and Coc Coc which aren't in MDN's BCD.  Even Yandex browser, which doesn't feature in browserslist!
3. We're talking to different analytics providers about making it easy to understand the coverage you get from different Baseline versions.  At the same time, I'm also trying to create a Looker Studio template that takes a browser version report from the Explore UI in Google Analytics and gives you a breakdown of which Baseline Years are represented and a percentage of users that you'll support well if you use Widely Available.

Once these things are in place - especially 1 - I would be very happy to update this to reflect bl2bl usage.  And if/when browserslist does include Baseline queries, we can revisit the guidance and deprecate bl2bl!

pepelsbey:
Thank you everyone for the discussion! Closing the PR since the suggested query doesn't accurately represent the widely available Baseline.

PR #2247, opened 2024-11-14
Revise `<input>` type descriptions

ddbeck:
I think these are complete now. I went and read through all of them, then revised them as a group.

### DIFF
--- a/features/button.yml
+++ b/features/button.yml
@@ -1,5 +1,5 @@
 name: <button>
-description: The `<button>` element is an interactive element activated by a user gesture that performs an action such as submitting a form, or opening a dialog.
+description: The `<button>` HTML element represents a pushbutton that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default.
 spec: https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element
 group:
   - forms
--- a/features/button.yml
+++ b/features/button.yml
@@ -1,5 +1,5 @@
 name: <button>
-description: The `<button>` HTML element represents a pushbutton that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default.
+description: The `<button>` HTML element represents a button that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default.
 spec: https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element
 group:
   - forms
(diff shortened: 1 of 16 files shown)


features/input-submit.yml
```diff
@@ -1,5 +1,5 @@
 name: '<input type="submit">'
-description: The `<input type="submit">` HTML element submits its associated `<form>`.
+description: The `<input type="submit">` HTML element sends the data from its associated `<form>` to a server.
```
autonome, line 2: (non blocking comment) Intercepting a form `submit` is pretty common for various reasons, might never go to a server.

features/button.yml
```diff
@@ -1,5 +1,5 @@
 name: <button>
-description: The `<button>` element is an interactive element activated by a user gesture that performs an action such as submitting a form, or opening a dialog.
+description: The `<button>` HTML element represents a pushbutton that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default.
```
autonome, line 2: `pushbutton` doesn't seem to be used in MDN or the spec. A quick search shows `push button` (with a space) to be used sometimes, eg shows up in W3Schools and various Q&A sites. Not a blocker, but seems overkill to bring a new term into use here. Would prefer to follow whatever the spec/MDN language.

autonome (approved):

features/button.yml
ddbeck: Ah, I saw it in this table here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types. Yeah, we can just not do this.

```suggestion
description: The `<button>` HTML element represents a button that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default.
```

features/input-button.yml
```diff
@@ -1,5 +1,5 @@
 name: '<input type="button">'
-description: The `<input type="button">` HTML element represents a button with programmable behavior, and which is inert by default.
+description: The `<input type="button">` HTML element represents a pushbutton that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default. Not to be confused with the `<button>` elment, which contains HTML content.
```
ddbeck, line 2:
```suggestion
description: The `<input type="button">` HTML element represents a button that triggers some action, such as submitting a form or opening a dialog, styled as a labeled rectangular box by default. Not to be confused with the `<button>` elment, which contains HTML content.
```

features/input-submit.yml
ddbeck: So maybe something like this?

```suggestion
description: The `<input type="submit">` HTML element represents a button that triggers an action on its associated `<form>`, such as sending the form data to a server.
```

features/input-reset.yml
```diff
@@ -1,5 +1,5 @@
 name: '<input type="reset">'
-description: The `<input type="reset">` HTML element is rendered as a button, and resets all inputs in the form to their initial values.
+description: The `<input type="reset">` HTML element represents a pushbutton that sets all form fields to their initial values.
```
ddbeck, line 2:
```suggestion
description: The `<input type="reset">` HTML element represents a button that sets all form fields to their initial values.
```

PR #2251, opened 2024-11-15
Update the web-animations description

### LINKED ISSUE #2244: Revise web-animations feature
### Specification

https://www.w3.org/TR/web-animations-1/

### Description

The feature description for web-animations describes it as "The `animate()` method of `Element` objects", however animations can also be constructed with the Animation constructor and the web animations API can be used on CSS animations and transitions after getting the Animation objects using the `getAnimations()` method on a document or `Element`.

I propose we revise the description to clearly include these other means of using the web animations APIs.

### Documentation

N/A This is a proposed revision to the existing feature

### Browser support

N/A This is a proposed revision to the existing feature

flackr:
Update the web-animations description to include
* Construction of an `Animation` directly
* Querying and modifying CSS animations and transitions retrieved via `getAnimations()`

This fixes #2244.

### DIFF
--- a/features/web-animations.yml
+++ b/features/web-animations.yml
@@ -1,5 +1,5 @@
 name: Web animations
-description: The `animate()` method of `Element` objects programmatically animates elements over time and can synchronize the animations of multiple elements.
+description: The programmatic Animation interface for animating elements over time and synchronizing the animations of multiple elements. This can be used to construct new animations through the `animate()` method or `Animation` constructor as well as query and modify CSS animations and transitions retrieved via the `getAnimations()` method.
 spec: https://drafts.csswg.org/web-animations-1/
 caniuse: web-animation
 group: animation


captainbrosset (changes requested):
Suggesting some changes to the description. 
Also based on my comment here https://github.com/web-platform-dx/web-features/issues/2244#issuecomment-2624565234 I would add the following BCD keys to the feature:

* api.Document.getAnimations
* api.Element.getAnimations
* api.ShadowRoot.getAnimations

features/web-animations.yml
```diff
@@ -1,5 +1,5 @@
 name: Web animations
-description: The `animate()` method of `Element` objects programmatically animates elements over time and can synchronize the animations of multiple elements.
+description: The programmatic Animation interface for animating elements over time and synchronizing the animations of multiple elements. This can be used to construct new animations through the `animate()` method or `Animation` constructor as well as query and modify CSS animations and transitions retrieved via the `getAnimations()` method.
```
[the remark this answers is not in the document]
flackr: This looks good. I suppose we could omit the Animation interface altogether and just call out animate() and getAnimations - though for use counting purposes I feel like it's potentially missing a lot of uses but if it's reasonable that the animation constructor implicitly counts as use of the feature then sounds good.

features/web-animations.yml
```diff
@@ -1,5 +1,5 @@
 name: Web animations
-description: The `animate()` method of `Element` objects programmatically animates elements over time and can synchronize the animations of multiple elements.
+description: The `animate()` method, together with the `Animation` interface, animates an element over time, possibly synchronizing the animations of multiple elements. The `getAnimations()` method queries and modifies current CSS animations, transitions, and web animations.
```
ddbeck, line 2: Hmm, after reading this it's almost as if we're trying to do too much to tease out the individual pieces of the API. Instead, I'm wondering if we can't give an overview and hint at what things to go look up in the docs. What do you think of something like this?

```suggestion
description: The `animate()` method of `Element` objects, the `document.getAnimations()` method, and the `Animation` interface animate elements over time and coordinate elements' scripted animations, CSS animations, and transitions.
```

features/web-animations.yml
```diff
@@ -1,5 +1,5 @@
...
 spec: https://drafts.csswg.org/web-animations-1/
 caniuse: web-animation
 group: animation
```
ddbeck, line 5: It's out of range of the diff, so I can't directly suggest it, but to match the description and fully resolve https://github.com/web-platform-dx/web-features/issues/2244 we also need to change which keys drive the headline status. This will also line us more closely with caniuse.

You'll need this in the current YAML file:

```
status:
  compute_from:
    - api.Element.animate
    - api.Animation
    - api.Document.getAnimations
```

and you'll need to run `npm install` and `npm run dist` and commit the results.

ddbeck (changes requested):
OK, I've caught up on the discussion here and in the original issue. I've got some ideas for the description (to avoid losing the context of `Element` and such) and an idea to fix the computed status too. Thank you!

PR #2252, opened 2024-11-15
Adds additional contain features and keys

vwallen:
Adds a feature for the `contain` property itself, to include the higher level keywords values `none`, `strict`, and `content` which are either the default (no containment) or aliases for already included keyword values. 

The main key `css.properties.contain` is already included in a keyword feature. I'm not sure of the protocol for moving a key from one feature to another, but if the `contain` feature is accepted as-is, it should reclaim that key.

Also adding keys to the `content-visibility` feature which required an update `compute_from` to maintain consistent support reporting


features/contain.yml
```diff
@@ -0,0 +1,11 @@
+name: contain
+description: "The `contain` CSS property sets limits to the scope of styles, layout, and paint rendering for speed and efficiency. The `none` keyword value disables containment. The `strict` and `content` keyword values are equivalent to `contain: size layout style paint` and `contain: layout style paint` respectively."
```
jamesnw, line 2:
```suggestion
description: "The `contain` CSS property sets limits to the scope of styles, layout, and paint rendering for speed and efficiency. The `none` keyword value disables containment, `strict` is equivalent to `contain: size layout style paint`, and `content` is equivalent to `contain: layout style paint`."
```
Optional nit.

features/contain.yml
```diff
@@ -0,0 +1,11 @@
...
+spec: https://drafts.csswg.org/css-contain-2/#contain-property
+group: containment
+caniuse: css-containment
+compat_features:
+  # Already included in `contain: size`
+  #- css.properties.contain
```
jamesnw, line 8: It's fine to move keys around as needed, and I think it makes sense to move this here.

features/content-visibility.yml.dist
```diff
@@ -39,3 +42,12 @@ compat_features:
...
   - api.ContentVisibilityAutoStateChangeEvent.skipped
+
+  # baseline: false
+  # support:
+  #   chrome: "108"
+  #   chrome_android: "108"
+  #   edge: "108"
+  #   firefox: "130"
+  #   firefox_android: "130"
+  - api.Element.contentvisibilityautostatechange_event
```
jamesnw, line 53: Can you confirm this is actually not supported in Safari? I'm wondering if this just got missed in BCD. We could potentially avoid a compute_from if it is actually supported.

jamesnw (changes requested):
A few minor changes, but looks good!

features/content-visibility.yml.dist
jamesnw: Per our testing, this is supported, so I opened https://github.com/mdn/browser-compat-data/pull/25186. 

This will mean we won't need a compute_from, but we'll be blocked on that issue. @ddbeck  Can you help with the BCD issue when you have a chance?

jamesnw: This is merged on BCD, and should be included in BCD 5.6.19.

features/content-visibility.yml.dist
```diff
@@ -2,16 +2,13 @@
 # Do not edit this file by hand. Edit the source file instead!
 
 status:
-  baseline: low
-  baseline_low_date: 2024-09-16
+  baseline: false
```
ddbeck, line 5: Since we're regressing the status, would you add a comment to the authored yaml file with a note as in https://github.com/web-platform-dx/web-features/pull/1973?

jamesnw: This should no longer be a regression with https://github.com/mdn/browser-compat-data/pull/25186 in 5.6.19.

jamesnw: There are now no status changes here.

jamesnw (approved):

PR #2274, opened 2024-11-19
Add requestIdleCallback feature

autonome:


features/requestidlecallback.yml
```diff
@@ -1,7 +1,7 @@
-draft_date: 2024-09-09
-name: requestIdleCallback() Cooperative Scheduling of Background Tasks
-description: TODO
+name: requestIdleCallback()
+description: "The `requestIdleCallback()` API queues JavaScript to be executed in idle browser time, either at the end of a frame or when the user is inactive."
```
ddbeck, line 2: Suggesting rewording to have some symmetry with `request-animation-frame`. Also adding the "also known as" for how MDN refers to this API.

```suggestion
description: "The `requestIdleCallback()` API queues a function that runs in idle browser time, either at the end of a frame or when the user is inactive. Also known as the background tasks API."
```

ddbeck (approved):
I like this! Minor suggestion on the description, for your consideration. Please merge if you think it's ready. Thank you!

PR #2279, opened 2024-11-19
Add mixed type parameters to lab/lch, oklab/oklch

jamesnw:
These were excluded because compute_from didn't yet exist. There's an argument to be made that these should be considered baseline in August instead of May, when mixed type parameters were added, and the compute_from should be excluded, but I could be convinced either way.


ddbeck (approved):
I think we still need the compute_from to align with caniuse. This is good enough for today. Thank you!

PR #2290, opened 2024-11-21
fix grammar conflict in web-features.md

brandondrew:
"an entire JavaScript APIs" doesn't work because it simultaneously requires singularity and plurality.  

Since the sentence begins with a very small feature consisting of a singular property and a singular technology and works its way up to a vast feature consisting of some of everything, I'm guessing that the API is meant to be singular—giving two more simple examples before jumping to the complex example.


ddbeck (approved):
Thank you and welcome to the project, @brandondrew!

brandondrew:
@ddbeck Thank you!

PR #2296, opened 2024-11-21
Add more DOMParser keys

autonome:


autonome:
Hrm, maybe should compute this? Baseline high is 6+ yrs ago so maybe fine?

ddbeck (approved):

PR #2298, opened 2024-11-21
Add FormDataEvent

autonome:


autonome:
Oh, and the `FormData` keys are split across XHR and Fetch, but `FormData` is not limited to those. Maybe that's worth a re-look at some point. The spec/docs for `FormDataEvent` are even clearer that this is not limited to XHR/Fetch, so I put it with forms.

autonome:
Erp, wrong base.

PR #2301, opened 2024-11-21
Early customizable select feature

captainbrosset:
Customizable select has been in discussions and in various states of prototypes for a few years already. It's feeling like the proposal is stabilizing now, and the feature has reached WHATWG stage 2 ([soon stage 3](https://github.com/whatwg/html/issues/10750#issuecomment-2477680683)). See [this developer.chrome.com blog about it](https://developer.chrome.com/blog/rfc-customizable-select?hl=en).

We've talked about how some (most?) features might need to get created early in the making, and I wanted to see what it felt like for this one.

@sachag had asked for it in the past, in order to be able to link to it from State of HTML, where it's mentioned.


features/customizable-select.yml
```diff
@@ -0,0 +1,11 @@
+name: Customizable <select>
+description: The `<select>` element's appearance, including the button, selected option, picker dropdown, and options, can be customized using CSS.
+spec: https://open-ui.org/components/customizableselect/
```
josepharhar, line 3: This link should be updated to https://open-ui.org/components/customizableselect/

captainbrosset:
Chromium I2S is out: https://groups.google.com/a/chromium.org/g/blink-dev/c/kN5LTzuTLVs/m/6HqTsmk3EQAJ

features/customizable-select.yml
captainbrosset: Isn't this the same link that's already there in the file?

captainbrosset:
> One obvious problem with creating this early feature is the lack of official spec. I think the most official artefact we have for this feature is the explainer on openui.org, which is what I've used for now, but which breaks the build.

Based on the [html spec PR](https://github.com/whatwg/html/issues/9799), there are actually multiple spec additions that cover what's needed for customizable select:

Spec PRs:

- https://github.com/whatwg/html/pull/10557
- https://github.com/whatwg/html/pull/10586
- https://github.com/whatwg/html/pull/10629
- https://github.com/whatwg/html/pull/10633
- https://github.com/whatwg/html/pull/10670


CSS spec PRs:

- https://github.com/w3c/csswg-drafts/pull/10691
- https://github.com/w3c/csswg-drafts/pull/10865
- https://github.com/w3c/csswg-drafts/pull/10936
- https://github.com/w3c/csswg-drafts/pull/10986

Accessibility PRs:

- https://github.com/w3c/aria/pull/2369
- https://github.com/w3c/aria/pull/2344
- https://github.com/w3c/html-aria/pull/528
- https://github.com/w3c/aria/pull/2360

It might make sense to just allow the OpenUI explainer to our list of spec instead of waiting for all of these PRs to get merged and then to reference them all in the feature.

captainbrosset:
@ddbeck I think we should merge this pretty much as is, allowing the OpenUI explainer URL. Customizable select has been in the works for many years, it needs to exist on our repo. We can use this as a guinea pig for early features if we want.

scripts/specs.ts
```diff
@@ -100,6 +100,10 @@ const defaultAllowlist: allowlistItem[] = [
     [
         "https://github.com/tc39/proposal-import-attributes/tree/abca60286360b47f9a6be25a28f489c2cb157beb",
         "Allowed because import assertions were replaced in-place by import attributes. Remove this exception when javascript.statements.import.import_assertions is dropped from BCD in June 2026."
+    ],
+    [
+        "https://open-ui.org/components/customizableselect/",
+        "Allowed because customizable select doesn't yet have a spec URL. And even when it does, it will be defined in multiple HTML, CSS, and ARIA specs. The OpenUI explainer is the only central place that defines the component at the moment."
```
ddbeck, line 106: Is there a good issue or PR to link to that would prompt us to remove this exception? Basically, I want to know where to look when we have "real" spec links to replace this with.

Something like this?

```suggestion
        "Allowed because customizable select doesn't yet have a spec URL. And even when it does, it will be defined in multiple HTML, CSS, and ARIA specs. The OpenUI explainer is the only central place that defines the component at the moment. Remove when SOMEURL merges."
```

ddbeck (approved):
I'm happy with this, but before merging, I'd like the spec link exception to point at something that'll tell us when the specification story is resolved.

scripts/specs.ts
captainbrosset: There are multiple PRs, not just one. See my previous comment: https://github.com/web-platform-dx/web-features/pull/2301#issuecomment-2624608170
Out of those, I'm guessing it would be enough to link to the following ones:

https://github.com/whatwg/html/pull/10586
https://github.com/whatwg/html/pull/10629
https://github.com/whatwg/html/pull/10633

@josepharhar are you able to help select one (or more, but ideally a small number) spec PR that would basically signify "that's it, customizable select is now in the spec"?

josepharhar: Sorry for being slow - if I had to choose one of the PRs, I'd choose the base appearance rendering PR as the best indicator for the spec of customizable select: https://github.com/whatwg/html/pull/10629

captainbrosset: I've added them all in the meantime 😆
Thank you.

PR #2304, opened 2024-11-21
Add new contrast-color feature

captainbrosset:
Don't look at the MDN page for this feature, it's outdated. See https://github.com/mdn/content/issues/36904.


features/contrast-color.yml
```diff
@@ -0,0 +1,8 @@
+name: contrast-color()
+description: "The `contrast-color()` CSS function picks a color that has guaranteed contrast with the color passed as an argument."
+spec: https://drafts.csswg.org/css-color-5/#contrast-color
+group: color-types
+compat_features:
+  # BCD is out of date with the spec.
+  # See https://github.com/mdn/content/issues/36904
+  - css.types.color.color-contrast
```
jamesnw, line 8: As I look at this more, it seems like both the name and the functionality are shifting (and the level 6 specification is very different). What if we mark `css.types.color.color-contrast` as either non-standard_track or deprecated, and then treat this feature similarly to #2301, as a placeholder for future expected functionality?

jamesnw: I opened up https://github.com/mdn/browser-compat-data/pull/25305 to specify `contrast-color` as non-standard track. This solves the immediate problem of how to represent it as a feature, but doesn't address the question of reserving keys for future functionality.

jamesnw: @captainbrosset FYI, `css.types.color.color-contrast` has been removed from BCD.

captainbrosset:
I spent some time looking into CSSWG conversations and specs, but couldn't form a definitive point of view on what's needed here.
I think it makes sense to keep this feature, but remove the BCD key (since it got deleted anyway), and ship this as a way to hold the ID for a future feature.
Either that, or close the PR without merging, since I don't know how to describe this yet.

features/contrast-color.yml
captainbrosset: Thanks

ddbeck:
@captainbrosset I'm trying to work this out in terms of whether or not we can be helpful developers. I think there are two plausible routes here: disambiguation or pretending it doesn't exist yet.

If we think this idea has enough attention that a lot of developers might have heard about it (from places that are not CSSWG issues), then main thing we can do at this point seems to be to provide disambiguation:

Looking around, it does appear that there have been articles on both `color-contrast()` and `contrast-color()` ([example](https://www.smashingmagazine.com/2023/01/creating-high-contrast-design-system-css-custom-properties/) and [example](https://lea.verou.me/blog/2024/contrast-color/) and [example](https://www.smashingmagazine.com/2022/05/accessible-design-system-themes-css-color-contrast/)). The names here are _super_ confusing—we can help with that.

In that scenario, it might make sense to create a discouraged `color-contrast()` feature (that looks a lot like `<portal>` from https://github.com/web-platform-dx/web-features/pull/2593) pointing to `contrast-color()`. They'd both be unimplemented, but at least it'd be sort of honest about the messy history and confusing names.

On the other hand, if we think that this is still obscure, flying under the radar, then let's close this PR and put it off until there's more developer interest.

jamesnw:
I could go either way, but given that `color-contrast()` appeared in the 2022 State of CSS Survey, I'm leaning towards recommending a discouraged `color-contrast()` feature that points to `contrast-color()`. There isn't a ton of chatter about the feature, but enough that disambiguation would be useful.

Although, to make the other case, early specs often go through bike shedding, and I'm not sure where to set the bar for disambiguating renamed features. Another instance would be `inset-area` getting renamed to `position-area`, after being present in several Chromium versions. We were able to avoid that question because it is part of the larger Anchor Positioning feature. 

I think as long as we are okay with setting a precedent that may lead to more than just a handful of discouraged disambiguation keys, this one makes sense to disambiguate.

captainbrosset:
I could also go both ways.

The 2 smashing articles you found both talk about the old `color-contrast` function (rather than the soon-to-come `contrast-color` function). And out of them, only one really talks about the function itself. The other one just mentions it in passing.
That one article was likely written when safari was implementing behind a flag, as a way to get people excited about the bleeding edge.
So, yeah, not a lot of chatter apart from the usual bleeding edge blogs.

Which makes me think that there's value in publishing a discouraged feature for `color-contrast`, maybe. People might have learnt about it via these articles, and might be under the impression that it will soon be available.
But then again, we could also do nothing. The lack of a feature altogether, in our catalog, could be enough of a sign that `color-contrast` is not a feature worth learning about.

Is that last point true though? It makes a lot of sense to publish discouraged features for things that did exist for a while, and maybe don't anymore, or do but have better alternatives. At first sight, it seems to make less sense to do so for features that never saw the light of day. But, when a feature is born, we don't yet know whether it will see day light. As Chromium starts to create web-features for all new things as part of its shipping process, we will have a lot of features that are Chromium-only-behind-flag things. And these can either continue on, or get merged or renamed, or just plain stop.

So, I guess my conclusion is that I'm leaning towards creating a discouraged feature, just to put ourselves in the position of having done so while the work was happening in Safari, and later having to live with that ID being reserved, and that feature now being discouraged.

captainbrosset:
But, if we do create discourage color-contrast feature, then we also need to create a contrast-color feature to point to (via the `alternatives` field). But that feature doesn't exist yet anywhere, so we'd essentially be reserving the ID.

jamesnw (approved):
I think this looks good!

It is missing a `dist` file for `contrast-color`.

PR #2308, opened 2024-11-21
Adds pointer-events feature

vwallen:


features/pointer-events.yml
```diff
@@ -0,0 +1,6 @@
+name: pointer-events
+description: The `pointer-events` CSS property sets whether an element triggers pointer events.
```
jamesnw, line 2: This isn't quite right- it's whether the element can be the target of pointer events. It would be nice to explain what a pointer is, and also what types of events. Maybe something like this?

```suggestion
description: The `pointer-events` CSS property sets whether a user can interact with an element using a mouse, touch, or other pointing input device.
```

jamesnw (changes requested):

jamesnw:
Could we link to this? https://caniuse.com/pointer-events

autonome:
bah sorry, i had this in #2285. i'll remove it and let you take it from here. but check and see if the keys are same. iirc i got there b/c i found some other keys outside the draft doc.

jamesnw:
> bah sorry, i had this in #2285. i'll remove it and let you take it from here. but check and see if the keys are same. iirc i got there b/c i found some other keys outside the draft doc

@autonome To clarify, you're talking about the `pointer-events` feature id, right? But this is a good question. Are `api.PointerEvent` and `css.properties.pointer-events` sufficiently separate features? I definitely can see people disabling interaction on an element without wanting to handle events in JS, so I think there's a case to be made for them being separate. 

If we go that way, then we'll need to differentiate the names. I think `pointer-events-api` might make the most sense for the api portion? I'm not thinking of a good differentiator for the CSS portion.

autonome:
ack, renaming on my end for the JS API, these are separate features!

Elchi3 (approved):

PR #2311, opened 2024-11-22
Update relative-color feature keys

jamesnw:
Essentially, this makes the decision that relative colors were not baseline low until all of the channel calculation oddities were resolved. Because this happened after `currentColor` issues were resolved, we don't need to make a decision on whether that impacts baseline.


jamesnw:
The change should be included in BCD 5.6.19.

autonome (approved):
LGTM

PR #2313, opened 2024-11-22
Add color feature

jamesnw:


autonome (approved):
I have a vague concern over name disambiguation given how many color related features there are. But this property certainly wins on historical precedence, and references seem to all just say "Color" so 👍🏼

Issue #204, opened 2023-05-24
Consider providing an API for Feature Set


atopal:
In addition to providing a package, we should also look into providing an API that authors can use to build components for easy integration of feature set statuses on their own sites.

Reference: https://twitter.com/EisenbergEffect/status/1661023585002217474

ddbeck:
Tweets are no longer available without being signed into Twitter. The text of the linked tweet:

> On a related note: It would be really cool if there were a Baseline Web API I could call and even cooler if there were a Web Component and/or embed I could drop into my content like the one you have in this post. I would start using those immediately.

ddbeck:
> It would be really cool … if there were a Web Component and/or embed I could drop into my content like the one you have in this post. I would start using those immediately.

This part is now satisfied by https://github.com/web-platform-dx/baseline-status

We don't yet have an API to call, only the JSON release assets.

atopal:
There is an API, that we should make official. It can be used like this: https://api.webstatus.dev/v1/features/growable-sharedarraybuffer and it's the one used in baseline-status.

tidoust:
If the API merely exposes the contents of the package, it could make sense to provide it along with the package, both to be in control of the update workflow (I do not know when the API endpoint receives updates) and to make it possible for people to create and use their own API endpoint if they so wish (e.g., because they do not want to rely on any third party).

atopal:
There are now docs for how to use the webstatus.dev API: https://web.dev/articles/web-platform-dashboard-baseline

jcscottiii:
Feel free to explore using this as well: https://editor.swagger.io/?url=https://raw.githubusercontent.com/GoogleChrome/webstatus.dev/refs/heads/main/openapi/backend/openapi.yaml

PR #2315, opened 2024-11-25
Add API for features

jamesnw:
This adds a simple static api to the GitHub Pages site. 

- `/api/v1/features.json` returns a list of all keys.
- `/api/v1/features/[id].json` returns a single feature (with the addition of the id).

We could easily add additional endpoints down the road.

While there is an API available at `https://api.webstatus.dev/v1/features/[id]`, I think this has some benefits-

- Anyone can (very) simply host their own version
- The returned data is in the same format as the web-features data (the webstatus.dev API renames fields and changes the structure slightly)
- We have better control over the update cadence

@tidoust says it better here- https://github.com/web-platform-dx/web-features/issues/204#issuecomment-2360935818.

This does add a dependency from the `gh-pages` submodule on `packages/web-features`, and I think currently the API would get updated whenever the gh-pages repo is updated. We need to consider if the API should be updated with every update to main, or only on releases.

- [ ] Document API
- [ ] Configure API update cadence

### DIFF
--- a/.github/workflows/gh-pages-build-deploy.yml
+++ b/.github/workflows/gh-pages-build-deploy.yml
@@ -41,7 +41,11 @@ jobs:
         with:
           cache: npm
           node-version-file: .node-version
-      - name: NPM Install
+      - name: NPM Install web-features
+        run: npm ci
+      - name: NPM Build web-features
+        run: npm run build
+      - name: NPM Install gh-pages
         run: npm ci
         working-directory: gh-pages/
       - name: Build pages
--- a/gh-pages/src/api/api.11tydata.js
+++ b/gh-pages/src/api/api.11tydata.js
@@ -1,0 +1,7 @@
+module.exports = async function () {
+    const { features } = await import('../../../packages/web-features/index.js');
+    return {
+        features: Object.keys(features).map(id => ({ ...features[id], id })),
+        featureList: { features: Object.keys(features) },
+    };
+}
--- a/gh-pages/src/api/feature.liquid
+++ b/gh-pages/src/api/feature.liquid
@@ -1,0 +1,9 @@
+---
+pagination:
+  data: features
+  size: 1
+  alias: feature
+permalink: "api/v1/features/{{ feature.id | slugify }}.json"
+---
+{{ feature | json }}
+
--- a/gh-pages/src/api/features.liquid
+++ b/gh-pages/src/api/features.liquid
@@ -1,0 +1,5 @@
+---
+permalink: "api/v1/features.json"
+---
+{{ featureList | json }}
+


captainbrosset (approved):
Love it. 🛳️ it

captainbrosset:
Discussed on the WebDX CG call now. This would be directly from the web-features repo, in the same format as the web-features data. So a nice alternative to the existing Google API.
Still needs to be figured out: how do we version, what cadence do we release, how do we incorporate it into the flow?

There's a use case for this at OddBird for their Baseline web component. It's nice because it allows to fetch just one feature too, not everything.

jcscottiii:
Semi-related question that should not block this: Is there a particular thing that api.webstatus.dev is lacking? This would be helpful to let leadership know that the API is missing something to the point the community needs alternatives.

jamesnw:
> Semi-related question that should not block this: Is there a particular thing that api.webstatus.dev is lacking? This would be helpful to let leadership know that the API is missing something to the point the community needs alternatives.

I don't think this is a case of api.webstatus.dev missing something, but some of the motivations for a separate API were:

1.  Potential for delay until api.webstatus.dev is updated, this could be integrated into the web-features release.
2. The api.webstatus.dev API (at least at the time) had some differences from web-features, with different names for some of the keys and some info missing. It's been a bit since I looked at this, so I don't remember the specifics, just noting that api.webstatus.dev does more than just dump the contents.
3. Another option for developers who want or need to avoid calling a Google API (or other trackable endpoint). Github Pages alleviates this somewhat, and this could even be self hosted with minimal effort. 

> There's a use case for this at OddBird for their Baseline web component.

Link, if helpful- https://github.com/oddbird/browser-support?tab=readme-ov-file

captainbrosset:
@jamesnw worth noting too that the explorer site also exposes the data as JSON. Go to any feature page and add `.json` at the end. E.g.: https://web-platform-dx.github.io/web-features-explorer/features/referencetarget.json

This returns the web-features data (unchanged) + the additional data coming from the https://github.com/web-platform-dx/web-features-mappings repo.

jamesnw:
@captainbrosset Excellent! That checks a lot of boxes for me, and has an existing update cadence. The additional data is also nice. 

I would be fine closing this PR and using that, unless someone has a reason why it should exist on the marketing site rather than the explorer.

jcscottiii:
> 1. Potential for delay until api.webstatus.dev is updated, this could be integrated into the web-features release.

We currently update the data twice a day automatically. Previously, we were tracking what was on main (bleeding edge) but we actually slowed down to track the releases now. If that is something you would be interested in, please let us know!

> 2. The api.webstatus.dev API (at least at the time) had some differences from web-features, with different names for some of the keys and some info missing. It's been a bit since I looked at this, so I don't remember the specifics, just noting that api.webstatus.dev does more than just dump the contents.

This was definitely a case early on for one reason: 1. There were some bugs  😅 (This may have caused some of the latency issues observed in point 1). But now with supporting feature evolution in V3, we ironed out a lot of those bugs.

> 3. Another option for developers who want or need to avoid calling a Google API (or other trackable endpoint). Github Pages alleviates this somewhat, and this could even be self hosted with minimal effort.

If that is a concern, I think @captainbrosset's suggestion of using what the explorer exposes is a great alternative!

captainbrosset:
@jamesnw feel free to close this if you're happy with using the JSON endpoint I mentioned earlier.
The update cadence is once a day.

PR #2328, opened 2024-11-25
Add :has-slotted feature

jamesnw:
Because `:has-slotted` is tagged as `web-features:slot` in BCD, I needed to move the keys in `slot` to the yml to override.


features/has-slotted.yml
```diff
@@ -0,0 +1,8 @@
+name: :has-slotted
+description: The `:has-slotted` CSS pseudo-class matches `<slot>` elements with any slotted content, including whitespace, text nodes, or elements. It can be used to determine that a `<slot>`'s fallback content is not displayed.
```
ddbeck, line 2: I had the "oh neat" moment on the second sentence. I suggest making that more prominent:

```suggestion
description: The `:has-slotted` CSS pseudo-class matches `<slot>` elements where the fallback content is not shown. The pseudo-class matches any slotted content, including white space, text nodes, or elements.
```

ddbeck (approved):
Looks good. Feel free to merge after you consider my description suggestion.

PR #2330, opened 2024-11-25
Adds text stroke and fill compat feature

vwallen:
These compat features do not have equivalents in the CSS spec. Is it worth including that in some way in the description?

For example: "These features are part of the compatibility specification" or "These are standardized compatibility properties for features not included in the CSS specification."


features/text-stroke-fill.yml
```diff
@@ -0,0 +1,10 @@
+name: Text stroke and fill
```
jamesnw, line 1:
```suggestion
name: Text stroke and fill  (compatibility prefixes)
```
This mirrors the `resolution-compat` and `supports-compat` features. Those are a bit different since there are recommended alternatives, but I think the qualifier still applies.

features/text-stroke-fill.yml
```diff
@@ -0,0 +1,10 @@
...
+description: The `-webkit-text-stroke-width` and `-webkit-text-stroke-color` CSS properties set the thickness and color of text outlines. The `-webkit-text-fill-color` sets the color within text character outlines. Both default to the text `color`.
```
jamesnw, line 2:
```suggestion
description: The `-webkit-text-stroke-width` and `-webkit-text-stroke-color` CSS properties set the thickness and color of text outlines. The `-webkit-text-fill-color` sets the color within text character outlines. Both default to the text color.
```
Since it's technically `currentColor`, I think without the ticks makes sense here.

jamesnw (approved):
Some minor suggestions, otherwise looks good.

Elchi3 (approved):

PR #2332, opened 2024-11-25
Adds missing keys to svg feature

vwallen:


Elchi3 (approved):
Looks good to me. Minor: move it to line 607?

vwallen:
> Looks good to me. Minor: move it to line 607?

So I better understand, why is that better than grouping these with the property?

vwallen:
Ugh, you're right and that was my intent, but I missed it

Elchi3 (approved):
No worries and thank you! :)

PR #2345, opened 2024-11-26
Remove redundant compat_features keys from features with comments

queengooborg:
This PR removes the redundant `compat_features` keys from web features that have comments around them.  This is an important manual step required to pave the way towards automated removal of `compat_features` keys.

Note: some wording in the comments was updated to maintain context without the presence of the `compat_features` value.


features/source.yml
```diff
@@ -6,14 +6,7 @@ group:
...
-  - html.elements.source.media
-  # Included in srcset (see #1448)
-  # - html.elements.source.sizes
-  - html.elements.source.src
-  # Included in srcset (see #1448)
-  # - html.elements.source.srcset
-  - html.elements.source.type
-  - html.elements.source.width
+# Included in srcset (see #1448)
+# compat_features:
```
ddbeck, line 10:
```suggestion
```

Elchi3 (approved):
I don't see a build failure, so I'm assuming the dists are fresh. Merging.

PR #2351, opened 2024-11-27
Add more media keys and features

autonome:
Pulled in all the media-related spec-drafts. Some new features, some merges of keys.

Feedback on these would be good:

- I went for a "media-*" pattern for keys/filename. CSS and media APIs both falling in that pattern isn't ideal, but doesn't seem like a huge deal.

- There's a `media-elements` group, but maybe we need a `media-apis` or `media-handling` group also?


features/media-capture-region.yml
ddbeck, line 1: → `region-capture.yml`

(to match the name)

features/media-playback-quality.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-09-09
-name: Media Playback Quality
-description: TODO
+name: Media playback quality
+description: The Media Playback Quality API contains metrics that can be used to determine the playback quality of a video, such as the creation time, and how many frames were dropped.
```
ddbeck, line 2:
```suggestion
description: The `getVideoPlaybackQuality()` method of `HTMLVideoElement` returns metrics that can be used to determine the playback quality of a video, such as how many frames were dropped.
```

features/media-recording.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-10-21
-name: MediaStream Recording
-description: TODO
+name: MediaStream recording
+description: The MediaStream Recording API captures the data generated by a `MediaStream` or `HTMLMediaElement` object for analysis, processing, or saving to disk. It's also known as the Media Recording API or the MediaRecorder API, and is closely affiliated with the Media Capture and Streams API and the WebRTC API.
```
ddbeck, line 2:
```suggestion
description: The `MediaRecorder` API captures data generated by `MediaStream` or `HTMLMediaElement` objects for analysis, processing, or saving to disk. Also known as the Media Recording API or the MediaRecorder API.
```

features/media-recording.yml
ddbeck, line 1: → `mediastream-recording.yml`

(to match up with the name)

features/media-transform.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-09-09
-name: MediaStreamTrack Insertable Media Processing using Streams
-description: TODO
+name: MediaStreamTrack insertable media processing using streams
```
ddbeck, line 1: This seems to be more common (e.g., on MDN and developer.chrome.com)

```suggestion
name: Insertable streams for MediaStreamTrack
```

features/media-transform.yml
ddbeck, line 1: → `insertable-streams.yml`

(to align with the name below)

features/media-transform.yml
```diff
@@ -1,6 +1,5 @@
...
+description: The MediaStreamTrack Insertable Media Processing using Streams API, also known as MIMPuS, gives direct access to media streams for manipulations such as merging streams, real time effects, and content overlays.
```
ddbeck, line 2: If the spec is going to lead with funny hats, then I think we should too:

```suggestion
description: The `MediaStreamTrackProcessor` and `MediaStreamTrackGenerator` APIs modify the contents of a media track for merging streams, applying real-time effects like funny hats, and content overlays.
```

ddbeck (changes requested):
Several suggestions here, but nothing earth shattering. Thank you!

features/media-capabilities.yml
```diff
@@ -0,0 +1,17 @@
+name: Media capabilities
+description: The `MediaCapabilities` API lets you query the decoding abilities of the device, system and browser about supported codecs, resolutions, and bitrates.
+spec: https://w3c.github.io/media-capabilities/
+# Computing status because core feature is baseline but this decoding
+# parameter is not, but doesn't prevent use.
+#  - api.MediaCapabilities.decodingInfo.configuration_keySystemConfiguration_parameter
+status:
+  compute_from: api.MediaCapabilities
```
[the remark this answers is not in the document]
autonome: Just this key, all its subkeys, or merge the whole feature?

Media properties encoding/decoding/bitrate/etc seems orthogonal to whether it's encrypted or not.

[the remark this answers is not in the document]
autonome: Got it! Moved, but still had to keep the computed status b/c of a different key in there.

ddbeck (approved):

PR #2352, opened 2024-11-27
Include test:types in npm run test

jamesnw:
This will ensure the type tests are run on PRs as well.


ddbeck (approved):

ddbeck:
Thank you! Good catch.

PR #2358, opened 2024-11-29
Add more streams keys/features

autonome:
Caniuse considers BYOB as a part of core feature, so keeping that as part of the core feature and computing status to match.

Hm, but there's an issue here... the BCD keys show Safari is false while Caniuse shows partial, but Baseline low for the feature as a whole.

Reviewers, what say you?


features/streams.yml
ddbeck, line 1: This regresses the Baseline status (correctly, to match caniuse), so we need a comment documenting that. See https://github.com/web-platform-dx/web-features/pull/1973/files for examples to copy from.

I think the "reason" here is that we're aligning to caniuse more closely, not that anything actually changed on the platform.

features/streams.yml
```diff
@@ -3,40 +3,92 @@ description: The streams API creates, composes, and consumes continuously genera
...
+  - api.TransformStream.transferable
+  - api.TransformStream.writable
+  - api.TransformStreamDefaultController
+  - api.TransformStreamDefaultController.desiredSize
+  - api.TransformStreamDefaultController.enqueue
+  - api.TransformStreamDefaultController.error
+  - api.TransformStreamDefaultController.terminate
+  - api.WritableStream.close
+  - api.WritableStream.transferable
+  # TODO: consider abortable stream feature
```
ddbeck, line 61:
```suggestion
```

ddbeck (commented):
This needs a comment to explain the status regression (see inline for details), but is the right thing to do. Close, on this one.

ddbeck (approved):

PR #2363, opened 2024-11-29
Add features and keys from CSSOM View Module spec

autonome:
Epic bucket of stuff. This PR parcels out almost everything in the CSSOM View spec across a number of features.

While it looks big, there's ultimately not a lot of controversial stuff in here. Many of the keys seemed to have a straightforward home already in existing features.

New features split out so far are Caniuse features:

- Scrolling APIs on elements
- Window.matchMedia + MediaQueryList
- Window.devicePixelRatio

New features split out, but not based on Caniuse features:
- Window.screen

The only keys remaining from that spec now are:
- feature policy, will go into that feature at some point
- the `getBoxQuads` keys, which are behind a flag on Firefox since ~2014

Something that feels ok but not ideal is putting all the size/positional bits of `Element` and `Window` into `DOM` and `Window` features, respectively. They are certainly properties of those, and it didn't move any baselines, but they are a different set of use-cases. But this option is likely better than a new feature for this, as I didn't find any great evidence showing developer POV around CSSOM View as a whole thing on its own.


features/scroll-elements.yml
```diff
@@ -0,0 +1,43 @@
+name: Scroll methods on elements
+description: Scrolling methods such as `scroll`, `scrollBy` and `scrollTo` change the scroll position of an element. Similar to setting th `scrollTop` and `scrollLeft` properties, but also allows options to be passed to define the scroll behavior.
```
jamesnw, line 2:
```suggestion
description: Scrolling methods such as `scroll()`, `scrollBy()`, and `scrollTo()` change the scroll position of overflow content within an element. Similar to setting `scrollTop` and `scrollLeft` properties, but also allows options defining whether the scroll should animate smoothly or jump.
```

I don't love the "overflow content" portion, but think it could be useful to specify you're calling it on the container that has overflow, not on an element within the scroll container.

Since `scrollTo` is just an alias for `scroll` (I think? Is there a difference?) do we need to mention both?

features/devicepixelratio.yml
```diff
@@ -0,0 +1,6 @@
+name: devicepixelratio
```
jamesnw, line 1:
```suggestion
name: devicePixelRatio
```
I think? we should camelCase it like the property?

features/devicepixelratio.yml
```diff
@@ -0,0 +1,6 @@
...
+description: The `Window.devicePixelRatio` read-only property returns the ratio of the vertical size of one physical pixel on the current display device to the size of one CSS pixel.
```
jamesnw, line 2:
```suggestion
description: The `window.devicePixelRatio` read-only property returns the ratio of the size of one CSS pixel to the vertical size of one physical pixel on the current display device.
```

I _think_ the ratio goes the other way, but double check me.

jamesnw (approved):
A few suggestions, but nothing blocking. Looks good to me!

features/scroll-elements.yml
autonome: I cannot find a difference, looks like an alias, will update to note that.

features/devicepixelratio.yml
autonome: Yeah it's funky to describe. I basically copied the Caniuse description because MDN's made things more muddy.

Yes, it's "physical" pixel / CSS pixel so i'll update to your suggestion which makes that clearer.

For reference, here's the spec:

> The devicePixelRatio attribute must return the result of the following determine the device pixel ratio algorithm:
> 
> - If there is no output device, return 1 and abort these steps.
> - Let CSS pixel size be the size of a [CSS pixel](https://drafts.csswg.org/css-values-4/#px) at the current [page zoom](https://drafts.csswg.org/cssom-view/#page-zoom) and using a [scale factor](https://drafts.csswg.org/cssom-view/#scale-factor) of 1.0.
> - Let device pixel size be the vertical size of a device pixel of the output device.
> - Return the result of dividing CSS pixel size by device pixel size.

Boris and PPK's write-ups are good too:

- https://web.dev/articles/high-dpi
- https://www.quirksmode.org/blog/archives/2012/06/devicepixelrati.html

ddbeck (commented):
A couple minor things here, but this is quite close.

features/visual-viewport.yml
```diff
@@ -1,17 +1,18 @@
 name: Visual viewport API
 description: The `visualViewport` API provides a way to query and modify the user-visible viewport of a web page.
 spec: https://drafts.csswg.org/cssom-view-1/#visualViewport
-# The resize and scroll events are excluded as the onresize and onscroll
-# properties were missing in Chrome 61, but were added in Chrome 62.
```
captainbrosset, line 5: I don't really understand this old comment. What's making us change our mind and now include these 2 keys?

captainbrosset (approved):

ddbeck (approved):
OK, one last nit in a line comment, then we're ready to merge. Thank you!

PR #2378, opened 2024-12-02
Add File API

autonome:


autonome:
A couple of notes:

- The name seems a bit bare, and this is often referred to as "the File API" so we could put API in there if someone feels strongly about it
- The name also doesn't center the entry point, which is `FileList`... seems odd but I've never seen anyone really talk about this like "the FileList API, so I dug around the internet and didn't see any evidence of anyone doing so
- Re entry point - I put `File` key in the computed status as a node to the relation, but maybe that isn't needed

features/file.yml
```diff
@@ -1,7 +1,12 @@
-draft_date: 2024-10-21
-name: File API
-description: TODO
+name: File
+description: The `File` API provides information about files and allows JavaScript in a web page to access their content.
```
jamesnw, line 2: Should this also mention the ability to create files as well? I've used this for creating CSVs to be downloaded or images for display/download, for example.

autonome: yes, thanks! updated

jamesnw (approved):
I'm on the line about the name, with a slight preference for "File API", but will defer to others.

features/file.yml
```diff
@@ -1,7 +1,12 @@
-draft_date: 2024-10-21
 name: File API
-description: TODO
+description: The `File` API provides information about files and allows JavaScript in a web page to create new files, and access existing files and their contents.
```
ddbeck, line 2: We're it not for caniuse and a lot of references to "File API" on Stack Overflow and elsewhere, I'd say this is one of those things that exists spread around the platform (like `ImageData`). So I suggest we try to show that this is an intermediary type:

```suggestion
description: The File API represents and reads from file data and metadata. Typically, you get files through other APIs, such as the `<input type="file">` element or drag-and-drop events. You may also send file data to other APIs, such as creating images or drawing on canvases.
```

ddbeck (approved):
OK, I think this is ready to go, but I'd like to suggest a pretty significant rewrite of the description. Take a look and merge (or ask for re-review) at your convenience.

PR #2380, opened 2024-12-03
Rename `:not` to `:not()` for consistency

nt1m:
Other functional pseudo-classes show the parentheses.


jamesnw (approved):
Thanks for catching that!

PR #2387, opened 2024-12-03
Add Text Encoding API

autonome:


autonome:
Kind of funky name... Caniuse goes the long route, so stuck with that.

features/text-encoding.yml
```diff
@@ -1,7 +1,12 @@
-draft_date: 2024-10-21
-name: Encoding Standard
-description: TODO
+name: Text encoding and decoding
+description: " The `TextEncoder` API generates a byte stream with UTF-8 encoding from a stream of code points, and `TextDecoder` does the reverse."
```
jamesnw, line 2:
```suggestion
description: "The `TextEncoder` API transforms a stream of code points into a byte stream with UTF-8 encoding, and `TextDecoder` does the reverse."
```

Optional- if you're describing this process directionally, I find it a bit clearer to go input to output with the TextEncoder portion.

jamesnw (approved):

PR #2388, opened 2024-12-03
Add `discouraged` to schema (and `with` statement demo feature)

ddbeck:
This adds `discouraged` to the schema. The idea here is that consumers would be able to show something like an MDN deprecation banner, saying that a feature is discouraged from use (citing a source) and suggesting an alternative (if one exists).

There are a couple of things missing here:

- There is no documentation. I'd like to get some feedback on the overall approach here before writing up how consumers are meant to use this or how feature authors are to decide whether to set `discouraged` and set useful `according_to` and `alternative` values.
- This depends on https://github.com/web-platform-dx/web-features/pull/2386 as a valid alternative feature.

Someday, I'd like to have a proper status (with iconography) for this condition that would take the place of Baseline for such features. This is a prerequisite for doing this, but doesn't actually affect the Baseline definition today.


features/symbol.yml
```diff
@@ -23,6 +23,5 @@ compat_features:
   - javascript.builtins.Symbol.toString
   - javascript.builtins.Symbol.toStringTag
   - javascript.builtins.Symbol.toStringTag.dom_objects
-  - javascript.builtins.Symbol.unscopables
```
ddbeck, line 26: This key isn't _itself_ deprecated, but it exists to use with a deprecated feature. I've moved it into the deprecated feature.

features/with.yml.dist
```diff
@@ -0,0 +1,50 @@
...
+status:
+  baseline: false
+  support:
+    chrome: "38"
+    chrome_android: "38"
+    edge: "12"
+    firefox: "48"
+    firefox_android: "48"
+    safari: "10"
+    safari_ios: "10"
```
ddbeck, line 13: Given the current Baseline definition, this is correct. That said, I expect consumers (like MDN and caniuse) to choose to suppress the Baseline information when there's a discouragement.

schemas/data.schema.json
```diff
@@ -57,6 +57,30 @@
...
+              "description": "IDs for features that substitute some or all of this feature's utility",
+              "items": {},
+              "type": "array"
+            }
+          },
+          "required": [
+            "according_to"
+          ],
+          "type": "object"
+        },
```
ddbeck, line 83: This is a semver minor event.

jamesnw:
To stress test this a bit, I went through the CSS keys that are deprecated and standard track. I'm thinking I'll create the following features:

## Clip
alternative: `clip-path`

- "css.properties.clip"
- "css.properties.clip.auto"
- "css.types.shape"
- "css.types.shape.rect"

## Media Queries (deprecated)
 alternative: media-queries

- "css.at-rules.media.device-aspect-ratio"
- "css.at-rules.media.device-height"
- "css.at-rules.media.device-width"

## Writing mode (deprecated)
alternative: writing-mode

- "css.properties.writing-mode.lr"
- "css.properties.writing-mode.lr-tb"
- "css.properties.writing-mode.rl"
- "css.properties.writing-mode.rl-tb"
- "css.properties.writing-mode.tb"
- "css.properties.writing-mode.tb-rl"

## color-adjust 
alternative: print-color-adjust

- "css.properties.color-adjust"

## SVG (deprecated)

- Existing keys in draft, plus `css.properties.glyph-orientation-vertical` which was missed.

## IME Mode
(no alternative)

- "css.properties.ime-mode"

## text-decoration-line :blink
alternative: animations-css

- "css.properties.text-decoration-line.blink"

## word-break: break-word
alternative: Use overflow-wrap: anywhere instead (note we only can link to overflow-wrap)

- "css.properties.word-break.break-word"

## overflow: overlay
alternative: Use overflow: auto  instead (note we can only link to overflow-shorthand)

- "css.types.overflow.overlay"

# Overview

Note that the list isn't long (and makes me want to double check my query), but it does surface a question around alternatives-

What should we do when the alternative is just a single key within a larger feature? Can we mention the alternative in the description when possible?

schemas/data.schema.json
```diff
@@ -57,6 +57,30 @@
...
+          "additionalProperties": false,
+          "description": "Whether developers are formally discouraged from using this feature",
+          "properties": {
+            "according_to": {
+              "description": "Links to a formal discouragement notice, such as specification text, intent-to-unship, etc.",
+              "items": {
+                "description": "URI",
+                "format": "uri",
+                "type": "string"
+              },
```
captainbrosset, line 70: What do you think about making this an object with both URI and link text?

Having the URI only forces consumers to come up with some link text on their own (or display the full URI, which isn't great).
This would avoid patterns like: "Discouraged, click _here_ to learn why".

captainbrosset (commented):
I like this.
Just left one comment to help consumers.

My more important general comment is how do we ensure that this isn't used for things we don't agree with. Because this is not data we can really check, there isn't much, beside thorough reviewing, to ensure this isn't used incorrectly.

Not sure what incorrect usage there could be, but I want to avoid, for example, cases where someone marks a feature as discouraged because it's just not implemented in browser X, and browser X vendor opposes that feature.

atopal:
> I like this. Just left one comment to help consumers.
> 
> My more important general comment is how do we ensure that this isn't used for things we don't agree with. Because this is not data we can really check, there isn't much, beside thorough reviewing, to ensure this isn't used incorrectly.
> 
> Not sure what incorrect usage there could be, but I want to avoid, for example, cases where someone marks a feature as discouraged because it's just not implemented in browser X, and browser X vendor opposes that feature.

That is a concern for sure, and it's independent of the wording. Even going with "deprecated", unless that comes out of the standardization process, it's not clear what should happen next. If one vendor considers an API as deprecated, but another one doesn't. Is that feature deprecated? Not hypothetical either, see [Battery Status](https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API).

atopal:
Maybe we limit this to cases where it's unambiguous (either from the spec or alignment of vendors) and explicitly exclude cases where there is disagreement for now?

Elchi3 (approved):
I've applied this to https://github.com/web-platform-dx/web-features/pull/2076 and it works for me. I think this could land and "reason" could be added later among other things.

Suppressing the baseline info should probably be strongly encouraged somehow, but I guess it is fine to expose `false` for now. Adding "discouraged" as a new baseline enum would certainly be more bold and maybe that would be the correct thing but maybe not now to avoid being too disruptive. I guess it would still be a possibility to change it later in a breaking release if desired.

captainbrosset:
This PR was discussed during today's WebDX CG call ([minutes](https://docs.google.com/document/d/1ree75ImLZjf60lTZ3BhCaLHygxgywr7SBXp-q0xPs8A/edit?tab=t.51kzxrqbshbo#heading=h.fnbp3elwn1ic)).

ddbeck:
I've pushed some guidelines and changes to the README for consumers. I hope it answers most of the remaining questions.

> What should we do when the alternative is just a single key within a larger feature? Can we mention the alternative in the description when possible?

Eventually, I think we should have a way to refer to inner/sub-features, but I don't think we have to have it here. I suspect that after we've enumerated the discouraged features we can revisit this more easily.

For now, I think we can satisfy the basic case of "see also: [something else to use instead]" without providing a detailed explanation for what we're referring to, specifically.

> Not sure what incorrect usage there could be, but I want to avoid, for example, cases where someone marks a feature as discouraged because it's just not implemented in browser X, and browser X vendor opposes that feature.

> Maybe we limit this to cases where it's unambiguous (either from the spec or alignment of vendors) and explicitly exclude cases where there is disagreement for now?

I included some text covering this in the new guidelines. The main thrust of it is: only set `discouraged` when a feature has consensus for discouragement. In most cases, this will come affirmatively from the specification (text saying don't use it or a transparent decision to drop something from a specification). Less frequently, this could come from the implementers, if all of them did so (e.g., you have a single-browser feature and that vendor starts telling developers to stop using that feature).

packages/web-features/README.md
```diff
@@ -27,13 +25,18 @@ import schema from "web-features/data.schema.json" with { type: "json" };
...
-If you're using `web-features` to render Baseline iconography or browser logos with support markers, then you must follow these procedures to ensure consistent usage.
+If you're using `web-features` to render Baseline iconography or browser logos with support markers, then you must follow the [name and logo usage guidelines](https://web-platform-dx.github.io/web-features/name-and-logo-usage-guidelines/).
 
 For Baseline iconography, follow this procedure for each feature:
 
 1. If `status.baseline` is `"high"`, then show an affirmative "widely available" icon.
 1. If `status.baseline` is `"low"`, then show an affirmative "newly available" icon.
 1. If `status.baseline` is `false`, then show a "limited availability" non-Baseline icon.
+
+   **Note**: All features that have the `discouraged` property are, by definition, non-Baseline.
```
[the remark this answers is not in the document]
ddbeck: Yes, good idea. 👍

index.ts
```diff
@@ -182,6 +182,16 @@ for (const [key, data] of yamlEntries('features')) {
...
+// Assert that discouraged feature's alternatives are valid
+for (const [id, feature] of Object.entries(features)) {
+    for (const alternative of feature.discouraged?.alternatives ?? []) {
+        console.log(`Confirming ${alternative} in feature set`);
+        if (!(alternative in features)) {
+            throw new Error(`${id}'s alternative "${alternative}" is not a valid feature ID`);
+        }
+    }
+}
+
```
ddbeck, line 194: This is succeeding the test workflow on GitHub Actions, but failing locally (it **should** fail). I think this is somewhat worrying. Maybe someone else can test this locally to reproduce?

jamesnw: @ddbeck If I'm not mistaken, `actions/checkout` checks out a commit that is the result of merging the PR branch into the target branch, and not on the PR branch itself. Since `destructuring` is now present on main, it doesn't fail. To test, you could merge main into your local branch and verify it passes, and then push a commit with a fake alternative id and verify it fails successfully.

docs/guidelines.md
```diff
@@ -309,3 +309,42 @@ For example, don't assign a feature to both `css` and `fonts`, since `css` is th
...
 Do assign features to groups when there's an opportunity for future feature composition (see [#971](https://github.com/web-platform-dx/web-features/issues/971)).
 For example, several features for the JavaScript `Array` interface are members of the `array` group.
+
+## Discouraged
+
+Rarely, the developers should not use a platform feature because it's the consensus of relevant stakeholders (i.e., a standards body and implementers), even if that feature is still implemented in browsers.
+Mark a feature as discouraged when:
+
+- The specification adopts clear language directing developers to stop using that feature or to prefer to an alternative.
+  This is often through terms like "deprecated", "obsolete", "legacy", or "non-standard."
```
captainbrosset, line 319: I would omit the "non-standard" part here. There are many non-standard (yet) features that are not discouraged, per se.

captainbrosset (approved):

docs/guidelines.md
ddbeck: OK, I've removed this with fc94af6beac7a6f53ed19977961ddf452f14a9bc. I think it's _true_—particularly when a spec absorbs some but not all functionality from a non-standard predecessor—but it probably doesn't add clarity here (without more explanation).

index.ts
ddbeck: Ah, this makes sense now. Thanks for the explanation, @jamesnw!

schemas/data.schema.json
ddbeck: I think this is a good idea, but I'm slightly concerned about us making guesses about what kind of text that consumers would like. I could imagine several different plausible ways to turn discouragement citations into text (I had a try below). I think it might be better to see how MDN and others actually end up using this and what works with developers, before committing to a single way of doing link text.

I do think we should revisit this, like the reason enums.

---

Possible ways to show this stuff:

- Spec owner:

   > Most developers shouldn't use `with` in new websites (according to [TC39](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-with-statement)). Consider using [Destructuring](#) instead.
   
- Wikipedia style:

  > Most developers shouldn't use `with` in new websites<sup><a href="https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-with-statement">[1]</a></sup>. Consider using [Destructuring](#) instead.
  

- Fetching the page title:

  > Most developers shouldn't use `with` in new websites (see [ECMAScript® 2025 Language Specification](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-with-statement)). Consider using [Destructuring](#) instead.

- Domain name:

  > Most developers shouldn't use `with` in new websites ([tc39.es](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-with-statement)). Consider using [Destructuring](#) instead.

All are possible without a handwritten text field.

ddbeck:
FYI: I'm planning to merge this PR after the last BCD and web-features releases of the year, probably the afternoon of Friday, December 20.

PR #2394, opened 2024-12-04
Add Content Index API

autonome:
This feels like a standalone feature. An alternative approach could be to roll it into ServiceWorker.

I checked and apparently it's in Android but not desktop.


jamesnw (approved):
I agree this makes sense as a separate feature.

PR #2395, opened 2024-12-04
Add LaunchHandler API

autonome:


autonome:
Need to figure out why there's no manifest key...

features/app-launch.yml
```diff
@@ -0,0 +1,10 @@
+name: Launch handler
+description: The Launch Handler API allows developers to control how a progressive web app (PWA) is launched, for example if it uses an existing window or creates a new one, and how the app's target launch URL is handled.
+spec: https://wicg.github.io/web-app-launch/
+group: progressive-web-app
+compat_features:
```
captainbrosset, line 5: I think we need to add `html.manifest.launch_handler` and `html.manifest.launch_handler.client_mode`.

features/app-launch.yml
```diff
@@ -0,0 +1,10 @@
...
+  - api.LaunchParams
+  - api.LaunchParams.targetURL
+  - api.LaunchQueue
+  - api.LaunchQueue.setConsumer
+  - api.Window.launchQueue
```
captainbrosset, line 10: I _think_ there might be two separate features here.

Developers can use the `launch_handler` manifest member to decide how the PWA launches (new or existing window). That part is captured by the 2 BCD keys I suggested before.

In addition, they can write code to handle whatever _thing_ the PWA was opened with. A PWA can be opened as a result of multiple things:

- In the file explorer/Finder, the user opens an file that's associated with the PWA.
- From another app, the user shares a file/url/text/image by using the OS built-in share dialog, and chooses the PWA as a target.
- In another app, the user activates A URL that's associated with this app.

All of the above have the ability to trigger the launch of the PWA, and the PWA needs to be able to handle the thing that was used to open it (file, shared content, link). This is done with `launchQueue` and `launchParams`.

As a developer, I see the two above scenarios as separate. I can write code to handle shared or associated content when my app launches. **And** I can decide if a new instance of my app is spawned when the PWA is launched.

captainbrosset (changes requested):

features/app-launch.yml
captainbrosset: How about naming the manifest-based feature **app-launch* (like now), and the launchQueue/Params feature **app-launch-consumer**?

@tomayac: can you take a look and let us know if you think developers think of the 2 things I described in my previous comment as, indeed, separate?

tomayac: I think they need to be thought of holistically, especially in the [`focus-existing`](https://developer.chrome.com/docs/web-platform/launch-handler#:~:text=focus%2Dexisting%3A%20The%20most%20recently%20interacted%20with%20browsing%20context%20in%20a%20web%20app%20window%20is%20chosen%20to%20handle%20the%20launch.%20A%20new%20LaunchParams%20object%20with%20its%20targetURL%20set%20to%20the%20launch%20URL%20will%20be%20enqueued%20in%20the%20document%27s%20window.launchQueue.) case. 

Note that the launch queue stuff is also shared by [`file_handlers`](https://developer.mozilla.org/en-US/docs/Web/Manifest/file_handlers):

- `api.LaunchParams`
- `api.LaunchParams.files` (new, above it was `targetURL`)
- `api.LaunchQueue`
- `api.LaunchQueue.setConsumer`
- `api.Window.launchQueue`

captainbrosset: Thanks Thomas.

Regarding file handing:

We already have a [file handler feature](https://github.com/web-platform-dx/web-features/blob/main/features/app-file-handlers.yml) on the repo and we decided to link ot to `html.manifest.file_handlers` and `api.LaunchParams.files`. That's because the parent BCD key `api.LaunchParams` can be used for more things than just file handling. And because `api.LaunchQueue*` can also be used for more things.

tomayac: Coincidentally, I had to dig this up today as well. Here are the launch triggers: https://github.com/WICG/web-app-launch/blob/main/launch_handler.md#background.

PR #2397, opened 2024-12-04
Add Reporting API

autonome:


features/reporting.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-10-21
 name: Reporting API
-description: TODO
+description: The Reporting API is a generic way to make reports available based on various platform features in a consistent manner, such as information about Content Security Policy, Permissions-Policy, or feature deprecations.
```
captainbrosset, line 2: I'm not very familiar with this API, but it seems helpful to me to mention the server endpoint aspect here.

```suggestion
description: The Reporting API makes consistent reports about Content Security Policy violations, Permissions-Policy violations, deprecated feature usage, crashes, and other web platform features available to your server endpoints.
```

features/reporting.yml
```diff
@@ -1,6 +1,5 @@
...
 spec: https://w3c.github.io/reporting/
 compat_features:
```
captainbrosset, line 4: Should this BCD key (and sub-keys) be listed here too: api.SecurityPolicyViolationEvent?

captainbrosset (approved):

features/reporting.yml
autonome: Those keys are in the CSP feature. The Reporting API is topic-agnostic, so topic-specific things probably have a better home w/ their topic area.

PR #2405, opened 2024-12-06
Add more HTML spec keys

autonome:


Alexizx00789999 (approved):

features/channel-messaging.yml.dist
```diff
@@ -12,7 +12,7 @@ status:
     firefox: "41"
     firefox_android: "41"
     safari: "5"
-    safari_ios: "4.2"
+    safari_ios: "5"
```
autonome, line 15: this is interesting!

features/view-transitions.yml
```diff
@@ -5,6 +5,13 @@ caniuse: view-transitions
 group: view-transitions
 compat_features:
   - api.Document.startViewTransition
+  - api.PageRevealEvent
```
jamesnw, line 8: The [caniuse](https://caniuse.com/view-transitions) mentioned lists Chrome 111 and Safari 18. 

I think this needs a compute_from `api.ViewTransition` to match, or at the very least, the new keys added in Chrome 128 don't need to be moving back the date.

features/channel-messaging.yml.dist
jamesnw: I think this is an error in BCD. 

`api.MessagePort` has safari_ios set to `mirror`, and `safari 5` mirrors `safari_ios 4.2`.

`api.MessagePort.worker_support` has `safari_ios` explictly set to `5`, and probably should be `mirror` instead.

This is based on a guess of what seems most likely, not on actually testing functionality or finding release notes.

autonome: Can't find anything in release notes yet, but filed mdn/browser-compat-data/pull/25407

autonome: Going to leave this as is and resolve, since it'll be automatically fixed here on rebuild if/when the BCD fix ships.

features/view-transitions.yml
autonome: Thanks, fixed the first bit.

Re #2435, I'll leave this as-is for now, which'll make it easier to find these related keys if that issue is fixed. Otherwise they'd be hidden away in drafts or BCD.

features/beforeunload.yml
```diff
@@ -1,9 +1,12 @@
 name: Beforeunload
 description: "The `beforeunload` event is fired when the current window is about to be unloaded. Typically this is used to display a dialog to confirm if users really want to leave the page when there is unsaved data that would be lost."
 spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-beforeunloadevent-interface
+status:
+  compute_from: api.Window.beforeunload_event
```
autonome, line 5: This is the entry point, and aligns w/ the actual core feature availability better.

features/offscreen-canvas.yml
```diff
@@ -41,15 +43,21 @@ compat_features:
   - api.OffscreenCanvasRenderingContext2D.fillRect
   - api.OffscreenCanvasRenderingContext2D.fillStyle
   - api.OffscreenCanvasRenderingContext2D.fillText
+  - api.OffscreenCanvasRenderingContext2D.filter
```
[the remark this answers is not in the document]
autonome: Thanks, I had same thought when adding them. But the group guidance has always been "start big, and split off smaller features over time as needed" vs "make too many features that are later harder to correct", hence this inclusion of non-baseline keys... not just here, but in many many places across the Web Features set.

As I've said more than a few times now: I'd love a consistently applied criteria and process for determining developer POV. We do not have that. I've suggested things like personas, "area peers", a dedicated standards contact... there are so many ways to mitigate the churn and better reflect developer POV which I hope are adopted over time.

features/dom.yml
```diff
@@ -123,6 +124,8 @@ compat_features:
   - api.Document.replaceChildren
   - api.Document.scripts
   - api.Document.title
+  - api.Document.write
```
ddbeck, line 127: Drop this key please. It should've been marked as deprecated. I've sent https://github.com/mdn/browser-compat-data/pull/25468

features/dom.yml
```diff
@@ -92,6 +92,7 @@ compat_features:
   - api.Document.createTreeWalker.whatToShow_filter_parameters_optional
   - api.Document.currentScript
   - api.Document.defaultView
+  - api.Document.designMode
```
ddbeck, line 95: Googling "designMode" had me thinking this might almost be a feature by itself, albeit a novelty. 😄

https://www.reddit.com/r/webdev/comments/npvmo0/that_feeling_when_you_first_discovered/

(It's fine to leave this here.)

features/gethtml.yml
```diff
@@ -0,0 +1,5 @@
+name: getHTML
+description: The `getHTML()` method serializes an element's DOM to an HTML string.
+spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#html-serialization-methods
+compat_features:
+  - api.Element.getHTML
```
ddbeck, line 5:
```suggestion
description: The `getHTML()` method of `Element` and `ShadowRoot` objects serializes an element's DOM to an HTML string.
spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#html-serialization-methods
compat_features:
  - api.Element.getHTML
  - api.ShadowRoot.getHTML
```

ddbeck (changes requested):
Some small changes in line comments.

features/dom.yml
autonome: yeah there's an alternate timeline where this got big in authoring tools...

features/dom.yml
```diff
@@ -148,6 +150,7 @@ compat_features:
   - api.Element.append
   - api.Element.attributes
   - api.Element.before
+  - api.Element.beforematch_event
```
captainbrosset, line 153: Reading the MDN docs for this event, it seems like it belongs to hidden-until-found instead.

captainbrosset (changes requested):
I found one key that might need to be moved to another feature. But otherwise looks great.

captainbrosset (approved):

PR #2418, opened 2024-12-10
Add Prefetch, Resource Hints and more Fetch keys

autonome:
~~Draft, some of these need better home~~

Ok, a tangled web of features here. Easier to discuss in one place since they're interrelated.

## Prefetch vs Speculation Rules

Speculation rules is broader. Prefetching is one `purpose`, but there can and will be more (eg prerender).

Developer POV seems to more be about "prefetch" than "speculation rules" as a singular thing, so I kept it split out - and that aligns w/ spec as well where it's separate.

Speculation Rules is more modern flexible approach that perhaps should be centered, but unless we're trying to actively shape how developers think about this, we should probably keep Prefetch as a standalone feature.

## Resources hints - ~~together or~~ separate

A bunch of `rel` attribute values on the `<link>` tag related to performance, and collectively called `resource hints` in more than a few places, eg https://web.dev/learn/performance/resource-hints.

- Spec doesn't treat them as a group
- Caniuse has separate features for four of them

Bunched for now, feedback wanted on whether to group or not.

Update: `modulepreload` is separate feature already and a Caniuse feature so splitting this into separate features.

## Authorization header removed in cross-origin redirects

I wiggled a bit on whether to put the `authorization_removed_cross_origin` keys all in their features or in `Fetch`, where that "feature" is specified.

1) Calling it a "feature" is a bit of a stretch

2) If you're a developer, you want to know whether network access method X supports feature Y

So I left the keys in the implementing network access method.


autonome:
Ready for review, with a few specific bits to note:

- Caniuse for `dns-prefetch` starts at Chrome 4 but BCD has it starting at 46?!
- `api.Request.duplex`... maybe should be a feature
- two prefetch features, but are quite distinct so made them separate
- no biggie but `modulepreload` was done earlier and now stands out with a different feature key format from the other `link-rel-*` features

features/link-rel-prefetch.yml
```diff
@@ -0,0 +1,6 @@
+name: '<link rel="prefetch">'
+description: The `rel="prefetch"` attribute for the `<link>` HTML element hints to browsers that the user is likely to need the target resource for future navigations, so the browser should preemptively fetch and cache the resource.
```
ddbeck, line 2:
```suggestion
description: The `rel="prefetch"` attribute for the `<link>` HTML element is a hint to the browser that the user is likely to navigate to a resource, so the browser should preemptively fetch and cache the resource.
```

features/link-rel-preload.yml
```diff
@@ -0,0 +1,14 @@
+name: '<link rel="preload">'
+description: The `rel="preload"` attribute for the `<link>` HTML element lets you declare fetch requests in the HTML's `<head>, specifying resources such as images or style sheets that your page will need very soon, which you want to the browser to start loading before main rendering begins.
```
ddbeck, line 2: This one is described in rather different terms. Is it also a hint? Or does this feature have stronger guarantees about what's to happen when it's used?

features/prefetch.yml
```diff
@@ -0,0 +1,7 @@
+name: Prefetch
+description: The `Sec-Purpose` HTTP header with a value of `prefetch` signals that a resource can be requested in anticipation of a near future navigation, such as a page linked in search results or a link that a user has hovered over.
+spec: https://wicg.github.io/nav-speculation/prefetch.html
+compat_features:
+  - api.PerformanceResourceTiming.deliveryType.navigational-prefetch
```
ddbeck, line 5: I'm not sure what to do with this key. It could belong with `resource-timing` or `link-rel-prefetch`. @Elchi3 do you have an opinion about it?

features/prefetch.yml
```diff
@@ -0,0 +1,7 @@
...
+  - http.headers.Sec-Purpose
+  - http.headers.Sec-Purpose.prefetch
```
ddbeck, line 7: AFAICT, these are request headers sent when making a request from a `<link rel="prefetch">`. I think these keys should go in the `link-rel-prefetch` feature instead.

There won't be anything left in this feature after this, so it can be deleted.

ddbeck (changes requested):

features/link-rel-dns-prefetch.yml
```diff
@@ -0,0 +1,6 @@
+name: '<link rel="dns-prefetch">'
```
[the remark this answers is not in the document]
autonome: I followed the example set in `features/modulepreload.yml` throughout these features.

I like that these are consistent, as naming features that are attribute-value-dependent is challenging enough already.

Consistency gives developers a familiar pattern to recognize in a sea of features.

But fine changing to whatever you think makes sense, seeing as we haven't taken this approach across all of web features.

features/prefetch.yml
autonome: I wrote rationale for prefetch as a separate feature in comment #0, but don't feel strongly about it. There was no easy out in this stack of vaguely-intersecting keys.

autonome: Yeah after removing the header keys, this doesn't make sense as a single key feature.

In previous situations like this, we've moved the key into the feature the key depends on, which would be `link-rel-prefetch` in this case.

autonome: Moved all these keys into `link-rel-prefetch` for now.

features/fetch.yml
```diff
@@ -36,6 +38,7 @@ compat_features:
   - api.Request.credentials
   - api.Request.credentials.default_same-origin
   - api.Request.destination
+  - api.Request.duplex
```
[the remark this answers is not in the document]
autonome: Yeah whether or not full duplex happens (certainly that will not be soon), this property does live here to describe the stream. In the comment above I was thinking of the duplex support in WebTransport, but this doesn't have anything to do with it.

features/link-rel-dns-prefetch.yml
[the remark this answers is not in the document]
ddbeck: OK, I'm happy with that. While I _like_ consistency, I think we're obliged to describe features as they already exist in the world, even if that leads to inconsistencies. In this case, there's not a strong consistency in any case, so I thought it'd be good to raise the possibility of being (correctly) inconsistent before we moved on. I'm satisfied with this.

ddbeck (commented):
I've re-reviewed and the only thing outstanding now is the `link-rel-preload` description. This is very close!

features/link-rel-preload.yml
autonome: Interesting, where the other hints use RFC 2119 language in the spec, `preload` uses "will":

> The [preload](https://html.spec.whatwg.org/multipage/links.html#link-type-preload) keyword indicates that the user agent will preemptively [fetch](https://fetch.spec.whatwg.org/#concept-fetch) and cache the specified resource

After reading through all of that and a few guides, I think it's likely because it MUST not block `load` so it only guarantees that the fetches are queued with a priority flag... but can't guarantee other than that. It specifically says these are not actually loads, just requests-to-load.

I'll tweak to make that part clearer.

features/cors.yml
```diff
@@ -17,3 +17,4 @@ compat_features:
   - http.headers.Access-Control-Max-Age
   - http.headers.Access-Control-Request-Headers
   - http.headers.Access-Control-Request-Method
+  - http.headers.Cross-Origin-Resource-Policy
```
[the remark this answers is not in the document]
autonome: Going to unbundle this and the CORB key since there's enough hot stuff in this PR already and getting that input might take a while.

features/fetch.yml
```diff
@@ -80,3 +83,4 @@ compat_features:
   - api.fetch.blob_data_support
   - api.fetch.init_keepalive_parameter
   - api.fetch.init_referrerPolicy_parameter
+  - http.headers.X-Content-Type-Options
```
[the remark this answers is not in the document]
autonome: Going to unbundle this and the CORP key since there's enough hot stuff in this PR already and getting that input might take a while.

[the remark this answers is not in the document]
autonome: #2455

autonome:
- Pushed the CORS/CORP/CORB keys over to #2455 so we don't lose that context.
- Updated the `preload` description
- Bundled all of `prefetch` into the `link-rel-prefetch`, which aligns w/ how we've handled key dependency in the past

features/link-rel-preload.yml
```diff
@@ -0,0 +1,14 @@
+name: '<link rel="preload">'
+description: The `rel="preload"` attribute for the `<link>` HTML element lets you declare fetch requests in the HTML's `<head> for resources such as images or style sheets that your page will need very soon, so the browser can queue them at a high priority for loading before main rendering begins.
```
ddbeck, line 2: Continues https://github.com/web-platform-dx/web-features/pull/2418#discussion_r1888567948

OK, cool. Some editing here to make this a bit tighter and consistent (e.g., we haven't previously mentioned `<head>` in these features):

```suggestion
description: The `rel="preload"` attribute for the `<link>` HTML element requests resources, such as images or style sheets, that the page needs soon, so the browser may prioritize them for loading before rendering begins.
```

ddbeck (approved):
OK, I have one line comment with suggestions for the new preload description, but otherwise this is good. Please merge after you've considered my suggestion.

Issue #2413, opened 2024-12-09
"events" feature seems overly generic


vmpstr:
https://webstatus.dev/features/events feature seems to represent all events. I want to double check that this is an important aspect to capture, since all browsers (that I know of) have some sort of events that happen.

The availability dates are also suspicious (e.g. available in Chrome since 2021-04-13 and baseline since 2015-07-29)

ddbeck:
Thanks for opening this issue, @vmpstr.

The feature description could use some work. The feature represents the machinery of representing, dispatching, and receiving events (i.e., the `Event` and `EventTarget` APIs), while specific events live with more specific features (e.g., the `toggle` event lives with the `<details>` element feature). The description doesn't communicate this at all. I'll open a PR to revise it.

I'm not sure where the "Available since" dates on webstatus.dev are coming from, but they don't seem to be from web-features (web-features reports the first release of each included browser as supporting this feature). I suspect it's something to do with the mapping with web platform tests, but I don't know for certain. I'd recommend that you open an issue with https://github.com/GoogleChrome/webstatus.dev/ about those dates.

ddbeck:
I've opened https://github.com/web-platform-dx/web-features/pull/2421 to revise the description. I'd welcome your review on it. Thanks again!

PR #2421, opened 2024-12-10
Revise `event` description to make it clear that it's generic

ddbeck:
This is one of those foundational features that's both difficult to avoid and hard to make explicit inside our length constraints.

Fixes https://github.com/web-platform-dx/web-features/issues/2413, I hope.


jamesnw (approved):
LGTM!

PR #2436, opened 2024-12-12
Update http3 for Safari support

### LINKED ISSUE #2369: Safari supports HTTP/3
caniuse was recently updated to reflect that Safari on both iOS and macOS fully supports HTTP/3. 

https://caniuse.com/http3

https://web-platform-dx.github.io/web-features-explorer/features/http3/

jamesnw:
Since HTTP/3 has no representation in BCD, the following version numbers are from https://caniuse.com/http3.

Closes #2369

### DIFF
--- a/features/http3.yml
+++ b/features/http3.yml
@@ -6,10 +6,13 @@ caniuse: http3
 # from https://caniuse.com/http3. TODO: capture partial support in Safari with
 # notes. caniuse says "Enabled by default for a portion of users."
 status:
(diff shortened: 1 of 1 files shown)


ddbeck (approved):

ddbeck:
It seems like you shouldn't have to care about the Baseline status on a feature like this—just the versions where the feature shipped. I filed https://github.com/web-platform-dx/web-features/issues/2438.

PR #2428, opened 2024-12-11
Auto-advance date of the `next` tag

ddbeck:
Removes softprops/action-gh-release in favor of the `gh` CLI.

Fixes https://github.com/web-platform-dx/web-features/issues/2068

Closes https://github.com/web-platform-dx/web-features/pull/2425

Issue #2068, opened 2024-10-23
@next GitHub release date does not advance


ddbeck:
The https://github.com/web-platform-dx/web-features/releases/tag/next release is continuously updated with the latest (test-passing) commit. However, because it updates an existing tag instead of creating a new one, the release date never advances on GitHub, so it sorts oddly, to the fixed time we created the workflow, well before v2.0.0.

We ought to use some steps as described in https://github.com/softprops/action-gh-release/issues/171#issuecomment-1229559873 to automatically advance the tag date on each publish for the @next release.

ddbeck:
While I'm at it, I think it'd be a good idea to stop using `softprops/action-gh-release` itself. The latest update https://github.com/web-platform-dx/web-features/pull/2131 had me dig into what was being released and why (the release notes were vague). That led me to find that the maintainership of the action has been handed over to someone else, but there's no documentation about it. Seems… iffy.

I think we can do the same with the `gh` CLI, something like:

```yaml
run: >
  gh release create
  --verify-tag
  --title 'web-features@next'
  --notes 'This is a continuously-updated prerelease generated from `main` (currently at ${{ steps.timestamp_and_hash.outputs.SHORT_HASH }}).'
  --prerelease
  --latest=false
  next
  <artifact paths go here>
```

(Of course, we'll have to delete and recreate the `next` tag in the step preceding this one.)

PR #2453, opened 2024-12-17
Revert "Auto-advance date of the `next` tag (#2428)"

ddbeck:
This produces a bunch of needless notifications for release watchers on GitHub. I think there's a way to work around this, but it needs some testing first.

This partially reverts commit 766b40e8fcdd82d59de4979d4a37714e4c8f9b2b and #2428.

### DIFF
--- a/.github/workflows/publish_next_web-features.yml
+++ b/.github/workflows/publish_next_web-features.yml
@@ -74,27 +74,15 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - name: Remove the existing pre-release
-        run: gh release delete "$TAG" --cleanup-tag --yes || true
-        env:
-          GH_TOKEN: ${{ github.token }}
-          TAG: ${{ env.dist_tag }}
-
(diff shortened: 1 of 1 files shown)


szepeviktor:
Were you thinking about a daily pre-release?

ddbeck:
What I _want_ is a continuously-updated release, so that it's easy to get at artifacts for the head of `main`. That is, I want to be able to `curl https://github.com/web-platform-dx/web-features/releases/download/next/data.extended.json` to always have the latest release data.

The existing workflow does that. However, the date never changes (previously, it appeared as if the `next` release was stuck in September or so, even though it changed several times a day). Deleting and re-creating the release fixes the date, but (misleadingly) makes it appear that there's a "new" release on every commit.

A little more detail here: https://github.com/web-platform-dx/web-features/issues/2068

szepeviktor:
> so that it's easy to get at artifacts

I see!
You could generate [GitHub Artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) with a workflow for `main`.

PR #2456, opened 2024-12-17
Add protocols supported by registerProtocolHandler

autonome:
I can't remember if we talked about how to handle this already or not.

If the "feature" is that it is possible for a web page to register itself as the handler for the scheme, then it seems like these are not separate independent features. So I've added them to `registerProtocolHandler` feature for now.


ddbeck (approved):
Makes sense. If folks want these broken out, we should do that when there's interest.

PR #2457, opened 2024-12-17
Add missing webvtt keys

jamesnw:
Adds `api.VTTCue.region` to WebVTT Cue Settings, and `api.VTTRegion.lines` to WebVTT regions. Neither fit 100% with the existing support story, but Cue hides it with compute_from, and regions is still baseline:false.


features/webvtt-cue-settings.yml
```diff
@@ -13,6 +13,7 @@ compat_features:
   - api.VTTCue.align
   - api.VTTCue.line
   - api.VTTCue.position
+  - api.VTTCue.region
```
ddbeck, line 16: I think this belongs with `webtt-regions`. The type of `region` is `VTTRegion`, so it seems kinda weird that it's not with the rest of that API.

ddbeck (changes requested):

ddbeck (approved):

PR #2459, opened 2024-12-17
Update scroll to text fragment

jamesnw:
See #1007 and #732 for context.

Essentially, the feature itself is not in BCD, so the status in manually set. This adds Firefox support, which makes it baseline low (and matches [Caniuse](https://caniuse.com/url-scroll-to-text-fragment)).

Separately, feature detection was not added in Safari, and has been sitting behind a feature flag. It will be enabled by default in Safari at some later date, per https://github.com/WebKit/WebKit/pull/36413.

This adds the keys for feature detection. We'll need to do more cleanup once Safari supports feature detection.


ddbeck:
> Essentially, the feature itself is not in BCD, so the status in manually set. This adds Firefox support, which makes it baseline low (and matches [Caniuse](https://caniuse.com/url-scroll-to-text-fragment)).

I think this might actually exist as `html.elements.a.text_fragments`. It's slightly vague, but the version numbers line up.

features/scroll-to-text-fragment.yml
```diff
@@ -7,16 +7,19 @@ group:
...
     safari_ios: "16.1"
 # document.fragmentDirective cannot be considered part of this feature due to
 # https://webkit.org/b/273466. TODO: Consider partial support for this feature
 # depending on https://github.com/WICG/scroll-to-text-fragment/issues/257.
-# compat_features:
-#   - api.Document.fragmentDirective
-#   - api.FragmentDirective
+compat_features:
+  - api.Document.fragmentDirective
+  - api.FragmentDirective
```
ddbeck, line 25: I _think_ the right thing to do here now is:

- Add `html.elements.a.text_fragments` to `compat_features` 
- Use `compute_from: html.elements.a.text_fragments` to do the override (instead of the manual one here)

ddbeck (changes requested):
Oh wait I'm undercaffeinated. I think that key solves some problems. Here's an actual review.

features/scroll-to-text-fragment.yml
jamesnw: Oh excellent! That makes things much better, thanks for tracking that down.

ddbeck (approved):

PR #2464, opened 2024-12-18
Add a fenced frame and a permissions key

autonome:
`api.Permissions.request` maybe could be a separate feature, but uses the existing permissions machinery so rolling into there for now.


ddbeck:
> `api.Permissions.request` maybe could be a separate feature, but uses the existing permissions machinery so rolling into there for now.

I genuinely couldn't come with a strong argument for or against, so I went in favor of parsimony. We'll break it out if there's interest. Thank you!

PR #2466, opened 2024-12-18
Add private network access

autonome:
Seems clearly referred to as a standalone feature in developer conversations, but also there's not much there there in BCD yet around it.

NOTE: see https://github.com/mdn/browser-compat-data/issues/25467

### DIFF
--- a/features/private-network-access.yml
+++ b/features/private-network-access.yml
@@ -1,5 +1,5 @@
-name: Private Network Access
-description: The `Access-Control-Request-Private-Network: true` and `Access-Control-Allow-Private-Network: true` HTTP request and response headers establish permission for websites to make requests to internal networks, such as to routers.
+name: Private network access
+description: "The `Access-Control-Request-Private-Network: true` and `Access-Control-Allow-Private-Network: true` HTTP request and response headers establish permission for websites to make requests to internal networks, such as to routers."
 spec: https://wicg.github.io/private-network-access/
 compat_features:
   - http.mixed-content.private_network_access


features/private-network-access.yml
```diff
@@ -0,0 +1,7 @@
+name: Private Network Access
+description: The Private Network Access feature is a set of modifications to Fetch and HTML which are intended to mitigate the risks associated with unintentional exposure of devices and servers on a client’s internal network to the web at large.
+spec: https://wicg.github.io/private-network-access/
+compat_features:
+  - http.mixed-content.private_network_access
+  # This key is in the Fetch feature, maybe should move here.
```
jamesnw, line 6:
```suggestion
  # TODO: This key is in the Fetch feature, maybe should move here.
```
Or if this is a candidate for the key being in both, what was the keyword we were going to use?

features/private-network-access.yml
```diff
@@ -0,0 +1,7 @@
+name: Private Network Access
+description: The Private Network Access feature is a set of modifications to Fetch and HTML which are intended to mitigate the risks associated with unintentional exposure of devices and servers on a client’s internal network to the web at large.
```
jamesnw, line 2:
```suggestion
description: Private Network Access modifies Fetch and HTML to mitigate the risks associated with unintentional exposure of devices and servers on a client’s internal network to the web at large.
```

jamesnw (approved):
Some minor wording suggestions, but LGTM!

features/private-network-access.yml
ddbeck: The description talks about modifying other features. It's not clear how this is a feature to developers. I had a skim through the spec, so I don't really understand this thing, but I suspect we want to say something more in this direction:

```suggestion
description: The `Access-Control-Request-Private-Network: true` and `Access-Control-Allow-Private-Network: true` HTTP request and response headers establish permission for websites to make requests to internal networks, such as to routers.
```

Curiously, BCD doesn't appear to represent the request header. If we went this route, we probably ought to file a bug on BCD reporting it as missing.

features/private-network-access.yml
ddbeck: I'd say this key certainly belongs in this feature.

features/private-network-access.yml
autonome: Filed mdn/browser-compat-data/issues/25467

autonome: Committed this change, which maybe is enough until BCD is sorted.

autonome: Or we could leave this open and blocked until then.

ddbeck: As long as the `Request.targetAddressSpace` property is representative of browser support for this (which it appears to be, at least for now) we don't have to block on the missing keys. Let's go with this.

ddbeck (approved):

PR #2467, opened 2024-12-18
Add private click measurement

autonome:
Welp, this seems both integrated and different from attribution reporting. Competing approaches maybe? Need to do some research here.

Ok looks like these might be competing approaches, would need to be separate features at this point.

2022 seems like so long ago. I'd forgotten about PCM.


features/private-click-measurement.yml
```diff
@@ -0,0 +1,6 @@
+name: Private click measurement
+description: Private Click Measurement, or PCM, is an API for measuring ad clicks across websites.
+spec: https://privacycg.github.io/private-click-measurement/
+compat_features:
+  - api.HTMLAnchorElement.attributionSourceId
+  - html.elements.a.attributionsourceid
```
ddbeck, line 6: Weird how BCD doesn't have data for the `attributiondestination` bit. I've filed https://github.com/mdn/browser-compat-data/issues/25595

ddbeck (approved):
This is ready to go, if you like my description suggestion. Otherwise we can go another round here.

(I'm not sure how much you'll be on this in this calendar year. We have a chat schedule later today, so I figure I'll figure out how to move this forward then.)

PR #2469, opened 2024-12-18
Add secure context and iframe to Gamepad features

jamesnw:
I suspect these were surfaced after initial work on #2434 was started.

### DIFF
--- a/features/gamepad.yml
+++ b/features/gamepad.yml
@@ -4,7 +4,7 @@ spec: https://w3c.github.io/gamepad/
 caniuse: gamepad
 group: gamepad
 status:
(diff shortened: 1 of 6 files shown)


features/gamepad.yml
ddbeck, line 1: This is not your fault, but I noticed it while we were here: the `compute_from` here ought to be `api.Navigator.getGamepads` not `api.Gamepad`. Would you mind fixing it?

ddbeck (approved):
One minor suggestion, but you're free to merge with or without it. Thank you!

features/gamepad.yml
jamesnw: Sure! Done.

PR #2474, opened 2024-12-19
Add undist script

jamesnw:
As we migrate more keys to BCD, adding or changing keys will be more challenging. We essentially will need to "undist" the .dist file, and move the keys without comments over to the .yml file.

Some examples-

```bash
# Move compat_features keys to webgpu.yml
npm run undist features/webgpu.yml.dist

# It works for paths ending in ".yml", ".yml.dist" or without either
npm run undist features/lh

# It sorts the keys by default, but this can be disabled
npm run undist features/is -- --no-s

# You can add keys, either as multiple options or comma separated
npm run undist features/modal -- -a new.key -a another.new.key -a multiple,keys

# I'll probably run dist immediately after
npm run undist features/inert -- -a new.key; npm run dist

```
I'm wondering if the positional argument should be a feature id rather than a path, as this is only useful for features, and not drafts.

I opted to make this work on a single file rather than multiple to enable the add functionality, as that use case seems the most common.


scripts/undist.ts
```diff
@@ -0,0 +1,80 @@
...
+  return {
+    yml: `${base}.yml`,
+    dist: `${base}.yml.dist`,
+  };
+}
+
+async function main() {
+  const { path, sort, add = [] } = argv;
+  // Accept `-a key1,key2,key3` as well as `-a key1 -a key2 -a key3`
+  const additions = add.flatMap((addition) => addition.split(","));
```
ddbeck, line 44: Both `-a key1 key2 key3` and `-a key1 -a key2 -a key3` already work. Does a comma-separated version allow for something that the other two forms don't already cover? My inclination is to not support an unusual shell notation, unless there's an explicit use case for it.

scripts/undist.ts
```diff
@@ -0,0 +1,80 @@
...
+  let base = path;
+  if (path.endsWith(".yml.dist")) {
+    base = path.slice(0, -9);
+  } else if (path.endsWith(".yml")) {
+    base = path.slice(0, -4);
+  }
+  return {
+    yml: `${base}.yml`,
+    dist: `${base}.yml.dist`,
+  };
```
ddbeck, line 38: Importing [`node:path`](https://nodejs.org/api/path.html) could avoid some magic numbers. Something like this:

```suggestion
  const { name, ext } = path.parse(path);
  const id = ext === ".dist" ? path.basename(name) : name;

  return {
    yml: `${id}.yml`,
    dist: `${id}.yml.dist`,
  };
```

scripts/undist.ts
```diff
@@ -0,0 +1,80 @@
...
+  const distData = YAML.parse(distContents);
+
+  const compatFeatures = distData.compat_features;
+  if (!compatFeatures) {
+    throw new Error(`No 'compat_features' found in ${dist}`);
+  }
+  if (additions.length > 0) {
+    compatFeatures.push(...additions);
+  }
+  if (sort) compatFeatures.sort();
```
ddbeck, line 70: optional style nit:

```suggestion
  if (sort) {
    compatFeatures.sort();
  }
```

ddbeck (approved):
This is good. I have a few optional suggestions for you, @jamesnw. But you're welcome to merge at your discretion.

scripts/undist.ts
jamesnw: Good call- removed!

PR #2475, opened 2024-12-19
Add style attributes

jamesnw:


features/style-attributes.yml
```diff
@@ -0,0 +1,5 @@
+name: Style attributes
```
captainbrosset, line 1:
```suggestion
name: Style attribute
```

features/style-attributes.yml
captainbrosset, line 1: Is there a reason why the feature ID is plural? I would suggest `style-attribute.yml`.

captainbrosset (changes requested):

features/style-attributes.yml
ddbeck: Since we only need a suffix to disambiguate, I'd suggest using the more common `-attr`: `sytle-attr.yml`.

features/style-attributes.yml
```diff
@@ -0,0 +1,5 @@
...
+description: The `style` global HTML attribute applies inline CSS styling declarations to individual elements.
+spec: https://html.spec.whatwg.org/multipage/dom.html#the-style-attribute
```
ddbeck, line 3: I guess this should be in a CSS or HTML group, or both?

features/style-attributes.yml
ddbeck: I'd like to go through all the attr features and harmonize the way we're doing this (some are the attr name alone, some are like "translate attribute", some are like this). I suspect most of them should be in the style of `attrname (attribute)` when they disambiguate themselves from something else (like this does, for `<style>`):

```suggestion
name: style (attribute)
```

ddbeck (changes requested):

features/style-attributes.yml
jamesnw: Good catch. If this isn't a candidate to be in both, I'm not sure what would be 😄

captainbrosset (approved):

ddbeck (approved):

PR #2485, opened 2024-12-20
Add SVG deprecated as discouraged feature

jamesnw:


features/ping.yml
```diff
@@ -2,12 +2,8 @@ name: ping
 description: The `ping` attribute for `<a>` elements sets a URL to send a POST request to when the `<a>` is clicked. This is typically used for outbound link reporting.
 spec: https://html.spec.whatwg.org/multipage/links.html#ping
 caniuse: ping
-status:
-  compute_from: html.elements.a.ping
```
jamesnw, line 6: This was added to hide the deprecated SVG elements, and so it can be removed.

features/svg-deprecated.yml
```diff
@@ -1,7 +1,11 @@
-draft_date: 2024-10-085
-name: svg-deprecated
-description: Placeholder for deprecated or non-real features
+name: SVG (deprecated)
+description: SVG 2 deprecated or removed a significant number of features defined in SVG 1.1, including features related to fonts, alternate glyphs, the `xlink` namespace, and the `<cursor>` and `<tref>` elements.
```
jamesnw, line 2: This was a bit difficult to write, as this is a junk drawer feature, and its entire purpose for existing is to describe that deprecations occurred, and not the features themselves.

features/svg-deprecated.yml
```diff
@@ -1,7 +1,11 @@
...
 spec: https://svgwg.org/svg2-draft/
+discouraged:
+  according_to:
+    - https://svgwg.org/svg2-draft/changes.html#substantial
```
jamesnw, line 6: Rather than link to the individual parts that deprecate each feature, this link enumerates them for us.

features/svg.yml
```diff
@@ -428,7 +428,6 @@ compat_features:
   - svg.elements.script
   - svg.elements.script.href
   - svg.elements.script.type
-  - svg.elements.script.xlink_href
```
jamesnw, line 431: Will be marked as deprecated in https://github.com/mdn/browser-compat-data/pull/25477, but that doesn't block this change.

jamesnw (commented):
Additional context- https://github.com/w3c/svgwg/wiki/SVG-2-new-features#removed

features/svg.yml
jamesnw: This will be resolved in #2500

features/svg-deprecated.yml
```diff
@@ -1,7 +1,13 @@
-draft_date: 2024-10-085
-name: svg-deprecated
-description: Placeholder for deprecated or non-real features
-spec: https://svgwg.org/svg2-draft/
+name: SVG (deprecated)
+description: SVG 2 deprecated or removed a significant number of features defined in SVG 1.1, including features related to fonts, alternate glyphs, the `xlink` namespace, and the `<cursor>` and `<tref>` elements.
```
ddbeck, line 2: OK, this took a bit of digging, but I think I now have a better idea of how to represent this.

In general, I think we should talk about SVG 1.1 as having some now-discouraged facets. This is a bit awkward, but I think it's a lit bit more in the spirit of describing a "feature" rather than a "change."

(Another approach here would be to break this into pieces, describing each thing as a feature—but that's a lot of work for an audience that probably does not exist. I think we should go ahead and make this a junk drawer feature, holding open the possibility of splitting this up in the future.)

While we're here, I think we should ignore two pieces, both of which will be eligible for removal from BCD in less than three weeks:

- `<tref>`
- `<cursor>`

```suggestion
name: SVG 1.1 (discouraged)
description: The SVG 1.1 image format has several components that were excluded from SVG 2, such as fonts, alternate glyphs, and the `xlink` namespace.
```

features/svg-deprecated.yml
```diff
@@ -1,7 +1,13 @@
...
+spec: https://www.w3.org/TR/SVG11/
+discouraged:
+  according_to:
+    - https://svgwg.org/svg2-draft/changes.html#substantial
+  alternatives:
+    - svg
+    - cursor
+    - font-face
 compat_features:
```
ddbeck, line 11: I can't suggest on the stuff outside the diff, but I propose dropping the keys relating to `<tref>` and `<cursor>`.

ddbeck (changes requested):
Thanks for this @jamesnw. I'm not sure how much (if any) time you'll be spending on this stuff in this calendar year, so I've made some suggestions. It looks like I'll catch up with you tomorrow and we can make a plan for handling the follow up work here.

features/ping.yml
```diff
@@ -2,12 +2,8 @@ name: ping
...
 compat_features:
   - api.HTMLAnchorElement.ping
   - api.HTMLAreaElement.ping
-  - api.SVGAElement.ping
   - html.elements.a.ping
   - html.elements.area.ping
```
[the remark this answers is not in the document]
jamesnw: Hmm- you're right. I'm not remembering how they got sorted into the deprecated ones. I'll revert that portion.

features/svg-deprecated.yml
jamesnw: This makes sense to me, and I agree the audience is likely not there for splitting them up.

ddbeck (approved):
Thank you, @jamesnw!

PR #2486, opened 2024-12-20
Discouraged CSS features

jamesnw:
Credit to @vwallen for the content.


captainbrosset (approved):
Approved with a suggestion for the missing description.

features/overflow-overlay.yml
```diff
@@ -0,0 +1,12 @@
+name: "`overflow: overlay`"
+description: "The `overflow: overlay` CSS declaration sets the behavior for when content doesn't fit in an element. It is superceded by `overflow: auto`."
```
ddbeck, line 2: Yikes, this one is messy. According to https://caniuse.com/css-overflow-overlay, this has a rather different between versions that BCD doesn't capture at all. I suspect what this needs is new support statements in BCD, showing `partial_implementation` when browsers had the non-standard behavior (i.e., before they aliased to `auto`).

Maybe we thread this needle like this:

```suggestion
description: "The `overflow: overlay` CSS declaration is an alias to `overflow: auto`. However, non-standard historic implementations change the appearance of scrollbars when content doesn't fit in an element."
```

And then keep this whole file in `drafts` until we (I?) can do the work upstream.

(cc: @atopal this is another good example of a "feature" that is interoperable—`overflow: overlay` aliasing to `overflow: auto`—but a bad thing to describe as Baseline)

features/text-decoration-line-blink.yml
```diff
@@ -0,0 +1,13 @@
+name: "`text-decoration-line: blink`"
+description: "The deprecated `text-decoration: blink` CSS declaration used to make text alternate between being visible and invisible."
```
ddbeck, line 2: Let's write these in the usual way (present-tense, not incorporating the "deprecation"):

```suggestion
description: "The `text-decoration: blink` CSS declaration flashes text between visible and invisible."
```

features/word-break-break-word.yml
```diff
@@ -0,0 +1,11 @@
+name: "`word-break: break-word`"
+description: "The `word-break: break-word` CSS declaration sets word breaks to occur according to their customary rules. It is superceded by `overflow-wrap: anywhere`"
```
ddbeck, line 2: I think this is more inline with our "not to be confused with" and "also known as". I suspect we'll soon need a style guide entry for these.

```suggestion
description: "The `word-break: break-word` CSS declaration sets word breaks to occur according to their customary rules. Superseded by `overflow-wrap: anywhere`"
```

features/writing-mode-deprecated.yml
```diff
@@ -0,0 +1,16 @@
+name: Writing mode (deprecated)
+description: The `lr`, `lr-tb`, `rl`, `rl-tb`, `tb`, and `tb-rl` keyword values represent writing modes for elements in SVG 1.1 documents.
```
ddbeck, line 2: Let's line this up closer to the other writing-mode features

```suggestion
name: writing-mode SVG 1.1 values
description: The `lr`, `lr-tb`, `rl`, `rl-tb`, `tb`, and `tb-rl` keyword values for the `writing-mode` CSS property set the text direction for elements in SVG 1.1 documents.
```

features/writing-mode-deprecated.yml
ddbeck, line 1: `features/writing-mode-deprecated.yml` → `features/writing-mode-svg-values.yml`

ddbeck (changes requested):
Some ideas here, mostly to avoid having to stuff discouraged information into the IDs and descriptions.

features/word-break-break-word.yml
```diff
@@ -0,0 +1,11 @@
+name: "`word-break: break-word`"
+description: "The `word-break: break-word` CSS declaration sets word breaks to occur according to their customary rules. Superseded by `overflow-wrap: anywhere`"
```
ddbeck, line 2:
```suggestion
description: "The `word-break: break-word` CSS declaration sets word breaks to occur according to their customary rules. Superseded by `overflow-wrap: anywhere`."
```

ddbeck (approved):
One last nit, but this is ready to merge at your convenience.

features/overflow-overlay.yml
ddbeck: After doing the work on the compat data, I'm now suggesting:

```suggestion
description: "The `overflow: overlay` CSS declaration is an alias to `overflow: auto`. Historically, it caused non-standard behavior, allowing scrollbars to overlay content without taking up layout space."
```

ddbeck (changes requested):

PR #2491, opened 2024-12-28
Split streams into more features

MattiasBuelens:
Prompted by the discussion in https://github.com/web-platform-tests/wpt/pull/49829, I had a look at the current list of features in the [streams](https://webstatus.dev/?q=group%3Astreams) group. Currently, there's only the "base" `streams` feature and a single `async-iterable-streams` extension for async iteration. However, there have been several additions over the years since we shipped BYOB support:

  * Transferable streams, [added to the standard in August 2020](https://github.com/whatwg/streams/pull/1053) and [shipped in Chrome 87 and Firefox 103](https://caniuse.com/mdn-api_readablestream_transferable)
  * `ReadableStream.from(asyncIterable)`, [added to the standard in June 2023](https://github.com/whatwg/streams/pull/1083) and [shipped in Firefox 117](https://caniuse.com/mdn-api_readablestream_from_static)
  * `Transformer.cancel()`, [added to the standard in September 2023](https://github.com/whatwg/streams/pull/1283) but not yet shipping in any browser
  * `ReadableStreamBYOBReader.read(view, {min})`, [added to the standard in November 2023](https://github.com/whatwg/streams/pull/1145) and [shipped in Firefox 134](https://caniuse.com/mdn-api_readablestreambyobreader_read_options_min_parameter)

I suggest we add the following new features, and move them out of the base `streams` feature:
* `transferable-streams`
* `readablestream-from`
* `transformstream-transformer-cancel`
  * There isn't a BCD feature for this yet, but I'm working on adding one: https://github.com/mdn/browser-compat-data/pull/25519
* `readablestreambyobreader-read-min`


MattiasBuelens:
@jcscottiii FYI

features/readablestream-from.yml
```diff
@@ -0,0 +1,6 @@
+name: ReadableStream.from()
+description: The `ReadableStream.from()` static method allows you to wrap iterable and async iterable objects (for example, arrays or async generators) as a readable stream.
```
captainbrosset, line 2:
```suggestion
description: The `ReadableStream.from()` static method wraps iterable and async iterable objects, such as arrays or async generator functions, as a readable stream.
```

features/readablestreambyobreader-read-min.yml
```diff
@@ -0,0 +1,6 @@
+name: "ReadableStreamBYOBReader.read(): min option"
+description: The `min` option for `ReadableStreamBYOBReader.read(view, options)` allows you to specify the minimum number of elements that must be available before the read request will fulfill.
```
captainbrosset, line 2:
```suggestion
description: The `min` option for `ReadableStreamBYOBReader.read(view, options)` sets the minimum number of elements that must be available before the read request will fulfill.
```

features/transferable-streams.yml
```diff
@@ -0,0 +1,8 @@
+name: Transferable streams
+description: Streams are transferable objects, allowing you to move them from one context to another (for example, to a worker).
```
captainbrosset, line 2:
```suggestion
description: Streams are transferable objects, which can be moved between contexts such as workers.
```

features/transformstream-transformer-cancel.yml
```diff
@@ -0,0 +1,6 @@
+name: "TransformStream: transformer.cancel method"
+description: The `cancel` method of a `TransformStream` transformer allows you to clean up resources when the readable side is cancelled or the writable side is aborted.
```
captainbrosset, line 2:
```suggestion
description: The `cancel` method of a `TransformStream` transformer cleans up resources when the readable side is cancelled or the writable side is aborted.
```

captainbrosset (changes requested):
Given the differences in browser support and ship dates, I agree that it makes sense to split Streams into multiple sub-features. These seem separate enough to me.

That said, I don't have enough experience with streams to know whether that's also how developers think of them.
The key aspect we use to define features on this repo is developer perception.
Some of the questions we try to answer to split things up might be:
- Do web developers think of these features independently from each other?
- Can they use streams already without these later additions?
- Are these later additions improvements over the base feature?
- Is there evidence of these additions being talked about on the web (StackOverflow, blogs, etc.)?

Let us know your thoughts. I'm not saying this isn't the right way to split, merely just leaving thought to provoke conversation.

Beyond that, I made a few tiny description proposals. Also, this should probably be merged again on the latest main branch and dist files should be [re-generated](https://github.com/web-platform-dx/web-features/blob/main/docs/CONTRIBUTING.md#feature-dist-files).

captainbrosset (approved):
This looks good to me. Thank you for providing the detailed rationale about these sub-features.

I don't think this needs to block this PR, but I do have one follow-up question: should the parent streams feature be Baseline Widely Available then?
Looking at that feature, it uses `compute_from` with the `api.ReadableStreamBYOBReader` BCD key, which makes it not available on Safari. But, reading the [Streams guide](https://web.dev/articles/streams) you provided, it feels like the streams feature should be baseline if the base use cases have been available for a while already.

MattiasBuelens:
> I don't think this needs to block this PR, but I do have one follow-up question: should the parent streams feature be Baseline Widely Available then?
> Looking at that feature, it uses `compute_from` with the `api.ReadableStreamBYOBReader` BCD key, which makes it not available on Safari. But, reading the [Streams guide](https://web.dev/articles/streams) you provided, it feels like the streams feature should be baseline if the base use cases have been available for a while already.

Personally, I would like to see BYOB become a separate feature. It's a bit weird how e.g. [Compression Streams](https://webstatus.dev/features/compression-streams) (which builds upon Streams) is Baseline Newly Available, while [Streams](https://webstatus.dev/features/streams) is only Baseline Limited Available.

There was a discussion regarding BYOB as part of #2358. [The decision there](https://github.com/web-platform-dx/web-features/pull/2358#discussion_r1891357379) was to keep BYOB as part of the base feature, in order to match the status quo on https://caniuse.com/streams. We could consider splitting up the feature in *both* Baseline and caniuse, but I don't know what the impact of such a split would be on developers using this feature data today. 🤷‍♂️

ddbeck:
I have a call with Alexis from caniuse scheduled for today. I'm putting this PR on our agenda, to learn more about developer impact of feature splitting on caniuse. I'll report back when I learn something.

ddbeck:
After some discussion with Alexis, this is what I learned:

- Historically, caniuse has not split features very often, but not never. As he put it, features on caniuse aren't always the last word on how we should perceive those features, so there's nothing in principle that would stop changing an existing feature on caniuse.
- Alexis said he'd be receptive to a PR that splits BYOB into its own feature on caniuse, modifying the existing streams feature to cover the API without BYOB.
- We also discussed the possibility of a transitional phase, where we create a BYOB feature and set the `caniuse: streams` value on it, until caniuse has been updated to reflect both general Streams and BYOB.

For the purposes of this PR, I'd be open to seeing a proposed BYOB feature that takes the `streams` mapping.

I'd also appreciate some assistance on contributing the splitting work to caniuse. I think the hardest part would be sifting through caniuse's notes, to determine which notes would belong with which feature (though I would, of course, welcome a complete PR to [Fyrd/caniuse](https://github.com/Fyrd/caniuse) to make the split).

MattiasBuelens:
Thanks @ddbeck! I've implemented your proposal with BYOB as a separate `readable-byte-streams` feature that maps to caniuse's `streams`.

* The main `streams` feature is now computed from `ReadableStream` + `WritableStream` + `TransformStream`, representing the three main classes for streams. This moves the feature to Baseline Widely Available.
* The `readable-byte-streams` feature has the same status as the original `streams` feature, remaining at Baseline Limited Availability (due to missing support in Safari).

I'll see if I can find my way around the Caniuse codebase and make a similar split there. 🙂

features/readable-byte-streams.yml
```diff
@@ -0,0 +1,26 @@
+name: Readable byte streams
+description: Readable byte streams allow you to efficiently read bytes from a stream into a buffer without incurring extra copies.
```
ddbeck, line 2: For most features, we try to mention the most likely "entry point" to that feature in the description and then frame the feature in that context—in other words, what would a developer literally _do_ with this feature? That said, I don't know a lot about this feature, other than what I've skimmed from the MDN and web.dev articles—so I'm really leaning on your input here. What do you think of something along these lines?

Also, given our discussion so far, I figure we should mention BYOB here. We have an "also known as" convention for feature names that I've suggested here.

```suggestion
description: A `ReadableStream` constructed with `{ type: "bytes" }` reads bytes from a stream without making extra copies, improving efficiency for streams of large chunks. Also known as BYOB or bring your own buffer.
```

features/readablestream-from.yml
```diff
@@ -0,0 +1,6 @@
+name: ReadableStream.from()
+description: The `ReadableStream.from()` static method wraps iterable and async iterable objects, such as arrays or async generator functions, as a readable stream.
```
ddbeck, line 2: "wraps … as" reads a little funny to me (though I see that MDN uses the same language). What do you think of this instead, which borrows a construction from the `iterator-methods` feature description?

```suggestion
description: The `ReadableStream.from()` static method converts an iterable or async iterable object, such as an array or async generator function, into a readable stream.
```

features/transferable-streams.yml
```diff
@@ -0,0 +1,8 @@
+name: Transferable streams
+description: Streams are transferable objects, which can be moved between contexts such as workers.
```
ddbeck, line 2: "Between" implies two things and I think making it explicit is nice. Does this wording work for you?

```suggestion
description: Streams are transferable objects, which can be moved between contexts such as windows and workers.
```

features/transformstream-transformer-cancel.yml
```diff
@@ -0,0 +1,6 @@
+name: "TransformStream: transformer.cancel method"
```
ddbeck, line 1: To align with our naming conventions here:

```suggestion
name: TransformStream transformer cancel() method
```

features/transformstream-transformer-cancel.yml
```diff
@@ -0,0 +1,6 @@
...
+description: The `cancel` method of a `TransformStream` transformer cleans up resources when the readable side is cancelled or the writable side is aborted.
```
ddbeck, line 2: Style nits:

```suggestion
description: The `cancel()` method of a `TransformStream` transformer cleans up resources when the readable side cancels or the writable side aborts.
```

features/readablestreambyobreader-read-min.yml
```diff
@@ -0,0 +1,6 @@
+name: "ReadableStreamBYOBReader.read(): min option"
```
ddbeck, line 1: I'm still not sure whether developers think of this as a distinct feature or an embellishment on the BYOB feature above. In other features, we've taken narrow options like this to be "later additions" to an existing feature. Searching Stack Overflow doesn't turn up much for me (though I'm not sure I'm using the right terminology). On the other hand, this is _very_ new, so maybe it's an unsettled question in general.

Basically, how sad would you be if we stuffed this compat_features key back into `readable-byte-streams`?

(We're working on some ways to improve this in the data model—to make it easier to split and merge things—so this isn't to say it must _always_ be this way, but this the ongoing negotiation we've had to do in some features.)

ddbeck (commented):
I think this is getting very close to landing. I have some suggestions around names and descriptions in line comments. The major thing I'm wondering about here is the `min` option and whether that justifies its own feature.

Thanks for your ongoing effort into this!

captainbrosset (approved):

PR #2510, opened 2025-01-09
BeforeInstallPrompt feature

captainbrosset:
Learn more:

* https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event
* https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Trigger_install_prompt#adding_an_in-app_install_ui
* https://web.dev/learn/pwa/installation-prompt#the_beforeinstallprompt_event
* https://web.dev/articles/customize-install
* https://developer.chrome.com/blog/app-install-banners-native#show_the_install_prompt


features/in-app-install.yml
```diff
@@ -0,0 +1,12 @@
+name: In-app install experience
+description: On browsers that support installing Progressive Web Apps (PWAs), the `beforeinstallprompt` event can be used to defer the browser installation flow until when the user interacts with an in-app install button.
+spec: https://wicg.github.io/manifest-incubations/
```
ddbeck, line 3: So this is pretty interesting. In the course of trying to figure out whether there was a more specific link to use here, I got to wondering why MDN doesn't link to this spec. That turned up these PRs:

- https://github.com/mdn/content/pull/26828
- https://github.com/mdn/browser-compat-data/pull/19910

This puts us in a slightly awkward spot here. BCD shows unqualified support for the API, even where it departs from the spec (because it effectively pretends the spec doesn't exist). We have some options here:

- If we were to link to the spec, we're sort of implying that the Chrome and Edge support for this feature reflects support for this specification.
- Not linking to a specification at all is a difficult alternative (not least of which being that the schema requires it)—what even is the feature, at that point, if we can't point to a definition?
- We could go upstream and unroll some of Will's changes to the compat data, breaking correspondence between MDN content and BCD.
- Override BCD's status.

Or maybe something I haven't thought of yet.

features/in-app-install.yml
```diff
@@ -0,0 +1,12 @@
+name: In-app install experience
```
ddbeck, line 1: I'm not very excited about this name. It's not very specific and it doesn't seem to be widely used (or perhaps used at all) in the existing blog posts and docs about this. I think good alternatives might be (in my order of preference):

- beforeinstallprompt
- App install prompt
- App install prompt event

In any case, the ID would need to change too.

features/in-app-install.yml
```diff
@@ -0,0 +1,12 @@
...
+description: On browsers that support installing Progressive Web Apps (PWAs), the `beforeinstallprompt` event can be used to defer the browser installation flow until when the user interacts with an in-app install button.
```
ddbeck, line 2:
```suggestion
description: The `beforeinstallprompt` event fires when a Progressive Web App (PWA) is installable, typically on page load in browsers that support installing PWAs. You can use this event to ask users to install within your site, instead of using a generic prompt provided by the browser.
```

ddbeck (changes requested):
Some minor issues around names and descriptions, but BCD is giving me a little headache here.

features/in-app-install.yml
captainbrosset: Thanks, I'll use this but tweak it a little bit, mostly because the in-app install option doesn't "replace" the browser prompt altogether. It delays it until the user interacts with whatever in-app UI you're displaying.

features/in-app-install.yml
captainbrosset: I would personally link to the spec like what's done here. There's no other spec to link to, but leave a comment saying that the only implementation that currently exists is in Chromium, and that it differs from the spec. And then, work with WICG to update the spec, which I think https://github.com/WICG/manifest-incubations/issues/36 is about.

ddbeck (approved):
Thank you, @captainbrosset. I like your description and I'm happy with this compromise around the spec.

PR #2511, opened 2025-01-09
Obsolete document color properties

captainbrosset:
This PR adds a discouraged feature for the following obsolete document properties:

  - bgColor
  - fgColor
  - alinkColor
  - linkColor
  - vlinkColor


features/document-colors.yml
```diff
@@ -0,0 +1,15 @@
+name: Document colors
+description: A document's `bgColor`, `fgColor`, `alinkColor`, `linkColor`, and `vlinkColor` properties get and set the background, foreground, active link, link, and visited link colors of that document.
+spec: https://html.spec.whatwg.org/multipage/obsolete.html
+discouraged:
+  according_to:
+    - https://html.spec.whatwg.org/multipage/obsolete.html
+  alternatives:
+    - background-color
+    - color
```
jamesnw, line 9:
```suggestion
    - color
    - css-object-model
```
I think this would be a good alternative as well, as it is how you would manipulate the document's styles. Potentially also `get-computed-style` as an entry point to that?

jamesnw (approved):

features/document-colors.yml
captainbrosset:
```suggestion
    - color
    - css-object-model
    - get-computed-style
```

Agreed! Thanks.

PR #2517, opened 2025-01-10
Separate fetch request streaming from fetch

captainbrosset:
Why is this a separate feature? Because it's a later addition to fetch that's not yet implemented everywhere and which is backed by existing developer needs. See:

- Interop 25 request: https://github.com/web-platform-tests/interop/issues/820
- 72 +1s on this crbug: https://issues.chromium.org/issues/40505032


features/fetch-request-streaming.yml
```diff
@@ -0,0 +1,5 @@
+name: Fetch request streaming
```
ddbeck, line 1: I don't love gerund feature names but avoiding ambiguity here is hard. After reading a bunch of StackOverflow questions and answers, I think this might be the least-worst option. What do you think of this alternative?

```suggestion
name: Fetch request streams
```

(This would require updating the ID too.)

features/fetch-request-streaming.yml
```diff
@@ -0,0 +1,5 @@
...
+description: Setting the `body` property of a request to a `ReadableStream`, when using the `fetch()` method, allows you to stream data to a server.
```
ddbeck, line 2: I think it'd be nice bring the "stream data to the server" language — which is the main story here — to the beginning of the description, though it comes at the cost of delaying the `ReadableStream` mention. What do you think of this instead?

```suggestion
description: A `fetch()` request streams data to the server when a request's `body` property is a `ReadableStream` object.
```

ddbeck:
@captainbrosset do you want to come back to this? I think it would be nice to wrap this one up, especially given the Interop angle.

captainbrosset:
> @captainbrosset do you want to come back to this? I think it would be nice to wrap this one up, especially given the Interop angle.

Ah yes, thanks for the review and ping. This had fallen off my radar. I've addressed your comments.

ddbeck (approved):
Thank you, Patrick!

captainbrosset:
Thanks for changing the file name. I was on mobile when I sent my last comment, and although I could just commit your suggestions, I didn't think of renaming the file (nor could). Thanks for taking care of that.

PR #2529, opened 2025-01-13
Add parens to getHTML() name

foolip:


ddbeck (approved):
Welcome back, Philip!

PR #2545, opened 2025-01-17
Gh pages adding rumvision

ErwinHofmanRV:
I saw a missing word when reading the gh-pages and added RUMvision to **in the wild** section


ddbeck (approved):
Thank you, @ErwinHofmanRV!

PR #2549, opened 2025-01-18
`wasm-string-builtins`: add keys tagged in BCD

ddbeck:
These keys are tagged as part of the feature in BCD, but not here.

### DIFF
--- a/features/wasm-string-builtins.yml.dist
+++ b/features/wasm-string-builtins.yml.dist
@@ -6,7 +6,23 @@ status:
   support:
     chrome: "130"
     edge: "130"
-    firefox: "134"
-    firefox_android: "134"
 compat_features:
+  # baseline: false
+  # support:
+  #   chrome: "130"
+  #   edge: "130"
+  #   firefox: "134"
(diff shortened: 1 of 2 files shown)


features/wasm-string-builtins.yml.dist
```diff
@@ -6,7 +6,23 @@ status:
...
+  # support:
+  #   chrome: "130"
+  #   chrome_android: "130"
+  #   edge: "130"
+  - webassembly.api.Module.Module.compile_options
+  - webassembly.api.compileStreaming_static.compile_options
+  - webassembly.api.compile_static.compile_options
+  - webassembly.api.instantiateStreaming_static.compile_options
+  - webassembly.api.instantiate_static.compile_options
+  - webassembly.api.validate_static.compile_options
```
ddbeck, line 28: I believe the Firefox data for these key are wrong. See https://github.com/mdn/browser-compat-data/pull/25706. This is pre-Baseline, so we probably don't need to block on that.

Elchi3 (approved):
Very nice! 👍

PR #2561, opened 2025-01-21
Disallow unused variables

ddbeck:
This enables an eslint rule to prevent unused variables everywhere, cementing a suggestion from @Elchi3 in https://github.com/web-platform-dx/web-features/pull/2544#discussion_r1923807759.


Elchi3 (approved):
Great!! 👍

PR #2562, opened 2025-01-21
Add `element-capture` feature

ddbeck:
Description based on reading https://developer.chrome.com/docs/web-platform/element-capture.


features/element-capture.yml
```diff
@@ -0,0 +1,7 @@
+name: Element capture
+description: The `RestrictionTarget` API captures video from a specific element, excluding content which might occlude the element itself, such as video conferencing controls.
```
foolip, line 2: Although you need to call `RestrictionTarget.fromElement()` first, I think it makes more sense to call out the `restrictTo()` method as the entrypoint for this feature. Something like

> The `restrictTo()` method for screen capture media track ...

Spelling out exactly what a "screen capture media track" is would be a bit of work, it's a `BrowserCaptureMediaStreamTrack` which can be returned from `getDisplayMedia()` (but isn't always IIUC).

foolip (approved):
Description could be polished perhaps, but no further review needed.

features/element-capture.yml
ddbeck: I don't think we have to explain that much. Revised with cf5715e7cf5374751f013c350d57bbe9b9d4f70d. Thank you!

PR #2565, opened 2025-01-22
Add feature for discouraged import assertions

ddbeck:
Prompted by https://github.com/web-platform-dx/web-features/pull/2564.

This feature ought to be tombstoned (see https://github.com/web-platform-dx/web-features/issues/91#issuecomment-2596266595 for discussion on that) but for now we have no-support compat data, so it's pretty straight forward to include it now, without doing anything special (excepting the spec URL).


features/import-assertions.yml
```diff
@@ -0,0 +1,13 @@
+name: Import assertions
+description: 'Module `import … assert { type: "json" }` and `import … assert { type: "css" }` JavaScript statements load JSON data.'
```
captainbrosset, line 2:
```suggestion
description: 'Module `import … assert { type: "json" }` and `import … assert { type: "css" }` JavaScript statements load JSON and CSS data.'
```

scripts/specs.ts
```diff
@@ -96,6 +96,10 @@ const defaultAllowlist: allowlistItem[] = [
     [
         "https://immersive-web.github.io/webvr/spec/1.1/",
         "Allowed because this is the legacy spec that defines WebVR."
+    ],
+    [
+        "https://github.com/tc39/proposal-import-attributes/tree/abca60286360b47f9a6be25a28f489c2cb157beb",
+        "Allowed because import assertions were replaced in-place by import attributes. Remove this exception when javascript.statements.import.import_assertions is dropped from BCD in June 2026."
```
captainbrosset, line 102: Nice-to-have: a script that goes over this array and checks if the specs listed here are actually used in features. This way we don't need to remember to remove items.

captainbrosset (approved):

PR #2573, opened 2025-01-24
Add `inputmode` global HTML attribute

ddbeck:
One thing that I know is an issue:

Safari's reported support differs between caniuse and BCD. Honestly, I don't know who's "wrong" in this scenario—it's not clear to me how to test whether an attribute is invalid, if there's no discernible behavior anyway.


captainbrosset:
I think BCD is correct here. `inputmode` is implemented in Safari across macOS and iOS. It's just a lot more visible on iOS because of the virtual keyboard. But, even on macOS the attribute does have an effect. I just tested and I get a different suggestion popups on input fields with `inputmode=tel` and `inputmode=email` than I do on other types of inputs.

The point here is that the spec doesn't say exactly _what_ should happen. This is for the user agent to decide, depending on what works best for the user, on the particular device they're using. It makes no sense for Safari on macOS to change the keyboard, because it's a physical keyboard, so instead they use the `inputmode` attribute to offer some contextual auto-completion help.

Either way, the attribute, and reflected DOM property, does work on Safari macOS too.

features/inputmode.yml
```diff
@@ -0,0 +1,7 @@
+name: inputmode
+description: The `inputmode` global HTML attribute gives a hint about what type of data is to be entered into an `<input>` element, such as `text`, `numeric`, or `email`, so the browser may choose a virtual keyboard.
```
captainbrosset, line 2:
```suggestion
description: The `inputmode` global HTML attribute gives a hint about what type of data is to be entered into an `<input>` element, such as `text`, `numeric`, or `email`, so the browser may help the user enter data using an adapted input mechanism, such as an on-screen virtual keyboard.
```

captainbrosset (changes requested):

features/inputmode.yml
ddbeck: I like this addition a lot!

ddbeck:
Thanks, @captainbrosset. I sent a PR to caniuse to update the data there (https://github.com/Fyrd/caniuse/pull/7256), so I think this is ready for re-review.

captainbrosset (approved):

PR #2589, opened 2025-01-28
Use GitHub Discussions for periodic announcements

ddbeck:
A couple of weeks ago, it came up on the WebDX call that we should have a dedicated announcements channel for web-features consumers. This is my proposal for doing that.

This PR adds documentation describing a process for publishing project announcements to GitHub Discussions.

If we choose to merge this PR, then we'll need to do the following just before merging:

- [x] Turn on Discussions in the repository settings.
- [x] Create a thread to post announcements to that is open to subscribing for notifications, but not open generally to replies (i.e., is locked)
- [x] Turn off all other discussions (we don't really have a use for non-announcements yet).
- [x] Update the documentation to link to the thread.

After merging, we should:

- [ ] Link to discussion thread and this PR's documentation in the next release's release notes.

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -4,12 +4,12 @@
 
 web-features uses a GitHub Discussions thread to give notice of breaking changes and other major changes to interested web-features consumers.
 
-If you want to receive these announcements, subscribe to the [Upcoming changes](https://github.com/web-platform-dx/web-features/discussions/123456789) <!-- TODO: link to the actual discussion thread, after it's created --> announcements thread or to all discussions through the repository's _Watch_ menu.
+If you want to receive these announcements, subscribe to the [Upcoming changes](https://github.com/web-platform-dx/web-features/discussions/2613) announcements thread or to all discussions through the repository's _Watch_ menu.
 
 > [!NOTE]
 > The remainder of this section is for [project owners](../GOVERNANCE.md#roles-and-responsibilities).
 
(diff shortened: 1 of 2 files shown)


docs/publishing.md
```diff
@@ -1,9 +1,24 @@
-# Publishing
+# Publishing releases
+
+## Announcements
+
+web-features uses a GitHub Discussions thread to give notice of breaking changes and other major changes to interested web-features consumers.
+
+If you want to receive these announcements, subscribe to the [Upcoming changes](https://github.com/web-platform-dx/web-features/discussions/123456789) <!-- TODO: link to the actual discussion thread, after it's created --> announcements thread or to all discussions through the repository's _Watch_ menu.
```
captainbrosset, line 7: It would be nice to add this (or some version of it) to the repo's root-level README.md file too. Something along the lines of "To use web-features, do this ... you can also see releases here ... and follow this GH discussion for breaking and major changes announcements ...".

ddbeck: I added this to the package README with 8abe1018. Since the repo README already links there, do you think that's sufficient?

ddbeck:
This came up on the WebDX call today. One suggestion was to add a link to the announcements thread to our release notes boilerplate text. I'll update the Markdown file about this after the call.

ddbeck:
OK, I this is ready now. If I get an approving review, I'll do the task list at the top, update the links, and merge this.

tidoust (approved):

PR #2592, opened 2025-01-29
New application-title metadata name feature

captainbrosset:
This is a new chromium-only feature that doesn't yet have a spec.
See https://blogs.windows.com/msedgedev/2025/02/05/control-your-installed-web-application-title/ and https://chromestatus.com/feature/5168096826884096.


features/application-title.yml
```diff
@@ -0,0 +1,9 @@
+name: application-title
+description: The `application-title` metadata name, used as `<meta name="application-title" content="">`, sets the text content that appears in the title bar of an installed web application.
```
ddbeck, line 2: The only other feature we have like this is structured a bit like this, with some edits for brevity. I don't know if it's _better_ but it is more consistent:

```suggestion
description: The `name="application-title"` attribute for the `<meta>` HTML element sets an installed web application's title bar text.
```

features/application-title.yml
```diff
@@ -0,0 +1,9 @@
+name: application-title
```
ddbeck, line 1: We do this for the other meta tag feature. This is pretty new so I don't know how developers are going to recognize it. What do you think?

```suggestion
name: '<meta name="application-title">'
```

ddbeck (approved):

PR #2596, opened 2025-01-30
Park more keys in `declarative-shadow-dom`

ddbeck:
I'm not sure if this is ideal, but it's the compromise I came up with.

I think it's possible that clonable shadow trees is a feature. But I can only find lots and lots of "no you can't clone shadow roots" (and a few blog posts saying why it's a bad idea) and little fanfare over clonable shadow roots.

The bits are even more obscure and I found it difficult to break them into distinct features.

### DIFF
--- a/features/declarative-shadow-dom.yml
+++ b/features/declarative-shadow-dom.yml
@@ -5,3 +5,14 @@ group:
   - html
   - web-components
 caniuse: declarative-shadow-dom
+status:
+  compute_from: html.elements.template.shadowrootmode
+compat_features:
+  - api.HTMLTemplateElement.shadowRootClonable
+  - api.HTMLTemplateElement.shadowRootMode
+  - api.HTMLTemplateElement.shadowRootDelegatesFocus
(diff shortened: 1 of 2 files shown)


foolip:
@mfreed7 does this make sense to you? To summarize, the change makes the `shadowrootmode`, `shadowrootclonable`, `shadowrootdelegatesfocus`, and `shadowrootserializable` attributes on `<template>` part of the DSD feature, as opposed to `<template>` or the older Shadow DOM feature.

mfreed7:
> @mfreed7 does this make sense to you? To summarize, the change makes the `shadowrootmode`, `shadowrootclonable`, `shadowrootdelegatesfocus`, and `shadowrootserializable` attributes on `<template>` part of the DSD feature, as opposed to `<template>` or the older Shadow DOM feature.

Makes total sense to me. Those are all parts of one feature called declarative shadow dom. Yes, they evolved at slightly different times, but going forward, I like this organization.

foolip (approved):

foolip:
Thanks for confirming @mfreed7!

PR #2601, opened 2025-02-03
Add keys to reporting API feature

ddbeck:
This brings many more keys under the umbrella of the reporting API feature.

I lightly revised the description to match how handle other features, by naming the entry points.

The discussion on https://github.com/web-platform-dx/web-features/pull/2397 notwithstanding, the description of the feature included the whole kitchen sink, so I moved all the keys for the reporting API here, including the CSP-related keys.


features/reporting.yml
```diff
@@ -1,7 +1,39 @@
 name: Reporting API
-description: The Reporting API makes consistent reports about Content Security Policy violations, Permissions-Policy violations, deprecated feature usage, crashes, and other web platform features available to your server endpoints.
-spec: https://w3c.github.io/reporting/
+description: The `Reporting-Endpoints` HTTP header specifies URLs to receive reports of Content Security Policy (CSP) violations, Permissions-Policy violations, deprecated feature usage, and crashes, while the `ReportingObsever()` API dispatches reports to a callback function.
```
captainbrosset, line 2: I know we usually try to mention entry points first in the description, but reversing the order of things makes it clearer to me. Up to you which one you prefer:

```suggestion
description: Content Security Policy (CSP) violations, Permissions-Policy violations, deprecated feature usage, and crashes can be sent to server endpoints by using the `Reporting-Endpoints` HTTP header, and to JavavaScript callback functions by using the `ReportingObserver()` API
```

captainbrosset: I'm realizing that the description doesn't mention browser interventions. If I'm correct, there are 5 types of reports: csp, deprecation, crashes, permissions, and interventions. If we list the first 4, we should add the 5th to the list.

captainbrosset (commented):
A couple of comments on the description. Apart from that, I agree with adding these new keys to the feature. It seems like the Reporting API happened in pretty much one go, with no later additions. I don't see a need to split this up further.

features/reporting.yml
ddbeck: I did appreciate your revision, though. It's a good way to frame it. Here's my new take on this, that tries to meld both approaches better:

```suggestion
description: The `Reporting-Endpoints` HTTP header and `ReportingObsever()` API send reports of Content Security Policy (CSP) violations, Permissions-Policy violations, deprecated feature usage, and crashes to a nominated URL or callback function.
```

Or with interventions:

```suggestion
description: The `Reporting-Endpoints` HTTP header and `ReportingObsever()` API send reports of Content Security Policy (CSP) violations, Permissions-Policy violations, deprecated feature usage, browser interventions, and crashes to a nominated URL or callback function.
```

PR #2609, opened 2025-02-05
Add a feature for `ui-` `font-family` values

ddbeck:
After reading https://github.com/w3c/csswg-drafts/issues/4107 to understand this, I came to the conclusion that they're like `font-family-system`, but different. I tried to make this new feature resemble the existing `font-family-system` feature.


jamesnw (approved):
LGTM!

PR #2610, opened 2025-02-05
Add `scroll-initial-target`

ddbeck:


features/scroll-initial-target.yml
```diff
@@ -0,0 +1,7 @@
+name: scroll-initial-target
+description: "The `scroll-initial-target: nearest` CSS declaration scrolls to the selected element when it first appears, much like scrolling to a URL fragment."
```
jamesnw, line 2:
```suggestion
description: "The `scroll-initial-target: nearest` CSS declaration scrolls to the selected element when the scroll container first appears, much like scrolling to a URL fragment."
```
Or are you referring to it being the first element to appear in tree order?

jamesnw (approved):

features/scroll-initial-target.yml
ddbeck: Truthfully, I have no idea how this works—I can't visualize it. It's not quite clear to me what "appears" means (render? enters the viewport?). If you have even half a clue about this, I'll defer to your suggestion. 😄

jamesnw: Here's an [explainer](https://github.com/DavMila/explainer-scroll-initial-target?tab=readme-ov-file), but, say you have a web page with a scroll container that has an element inside it that has the declaration `scroll-initial-target: nearest`. When the page loads, that scroll container will scroll so that element is at the top of the scroll container.

```suggestion
description: "The `scroll-initial-target: nearest` CSS declaration sets the initial scroll position of its scroll container to the top of the element, much like scrolling to a URL fragment."
```
I think it's okay to repeat `scroll` and `initial` here. Please feel free to wordsmith as well.

captainbrosset: Maybe @DavMila can help? Can you please take a look at the feature description? The idea is to describe the css property in a short sentence (<300 chars) and in a way that web developers would immediately understand the value. Thanks!

ddbeck: This had sat idle for a bit, so I accepted the suggestion from @jamesnw. This is now shipping in Chrome, so I think it'd be nice to merge this sooner rather than later. If this needs additional work, please feel free to push new changes yourselves while I'm OOO. Thank you!

captainbrosset (approved):

PR #2611, opened 2025-02-05
Add more keys to `text-box`

ddbeck:
This adds value types to `text-box`.

The `ideographic` and `ideographic-ink` values aren't supported by any browser. They shouldn't have been added to BCD; I'll send an upstream PR to fix that.


features/text-box.yml.dist
```diff
@@ -3,10 +3,12 @@
...
   baseline: false
-  support:
-    safari: "18.2"
-    safari_ios: "18.2"
+  support: {}
 compat_features:
+  # baseline: false
+  # support:
+  #   safari: "18.2"
+  #   safari_ios: "18.2"
```
captainbrosset, line 11: https://developer.chrome.com/blog/css-text-box-trim says that chrome/edge 133 have support for text-box too. And BCD seems to agree. Do you know why Safari is the only supported browser here?

foolip: There are conflicts in this file, https://github.com/web-platform-dx/web-features/blob/main/features/text-box.yml.dist looks good on main and I bet that resolving the conflicts will fix this.

foolip (approved):
Looks good if dist is regenerated and ideographic stuff is dropped.

ddbeck:
The ideographic removals happened with https://github.com/web-platform-dx/web-features/pull/2625, which I merged into this PR. Thanks for the reviews, folks!

PR #2614, opened 2025-02-05
Assign more Permissions-related keys

ddbeck:
This PR assigns all but one of the keys remaining from the Permissions API spec to their respective features.

The one key left unassigned is `api.Permissions.permission_accessibility-events`. For the history of this key, see https://github.com/mdn/browser-compat-data/pull/25860 for more info on this one key. I've decided to send it to the unspecified web APIs bucket; see https://github.com/web-platform-dx/web-features/pull/2566


foolip (approved):
Looks good! If resolving conflicts turns up something interesting, feel free to circle back for more review. Otherwise, please land without further review.

PR #2628, opened 2025-02-10
Add more keys to Web Audio

captainbrosset:
`api.AudioListener.setOrientation` and `api.AudioListener.setPosition` are deprecated but are currently the only way to set an audio listener's orientation and position in Firefox. So there's no use creating a discouraged feature for this. At least not until https://bugzil.la/1283029 gets fixed and BCD gets updated.

For now, why don't we put these keys under Web Audio. Anyway, the whole feature uses a `compute_from` field which will hide any weirdness from these keys.

### DIFF
--- a/features/web-audio.yml
+++ b/features/web-audio.yml
@@ -66,6 +66,11 @@ compat_features:
   - api.AudioListener.upX
   - api.AudioListener.upY
   - api.AudioListener.upZ
+  # The following 2 keys are deprecated, but are currently the only way to set
+  # the listener orientation and position in Firefox. So we're keeping them here
+  # for now.
+  - api.AudioListener.setOrientation
+  - api.AudioListener.setPosition
   - api.AudioNode
   - api.AudioNode.channelCount
   - api.AudioNode.channelCountMode


captainbrosset:
Whoops, `npm run dist` isn't happy. I'm not supposed to add deprecated keys to a published feature. So, I guess we should:

- Either ignore these keys for now.
- Or add them to a discouraged feature, but I don't think doing this is useful to anyone.
- Or maybe change the rules to allow adding deprecated keys somehow?

ddbeck:
OK, so now we now have a real-life example of an API that is totally harmless to use, is interoperable, but is nevertheless marked as deprecated without an interoperable alternative. Frustrating! Also, this was a specific kind of discouraged case that @atopal was concerned about.

It does seem like the sort of thing that just ought to be Baseline, at least until Firefox ships the mainstream alternative. I think my inclination here is to make some sort of exception or editorial override, assuming these cases are rare, as they have been so far.

But I'm not sure if they're rare yet. What do you think of marking this PR as a draft until we've dealt with more of the deprecated BCD keys?

captainbrosset:
Sounds good to me. Let's leave this one off to the side for now, until we find other similar cases, if any.

atopal:
Thanks for calling this out! I'd also like to see if we have more of them.

captainbrosset:
Discussed on the WebDX CG call now. From Daniel: There are some cases where we have new discouraged features. But the actual standard, non-discouraged, alternative isn't shipped (or at least not as available as what's discouraged). These things aren't being removed from browsers.
It's probably one of these cases. Spec says not to do this, but the alternative is not really usable.
We could mint a discouraged feature.

captainbrosset:
> We could mint a discouraged feature.

Revisiting this after a while, I'm on the fence about whether to create a new discouraged feature.

What `setOrientation()` and `setPosition()` (which are deprecated) allow you to do is to spatialize sound. The **correct** way to do this is by using the AudioListener's instance properties like `positionX/Y/Z`, `forwardX/Y/Z`, and `upX/Y/Z`.
These are not implemented in Firefox yet.

We could mint a new feature called something like "Spatial web audio", which would be **Limited Availability**. In it, we'd use the BCD keys from the parent web-audio feature (which are hidden away from there anyway, because we use `compute_from`):
  - `api.AudioListener.forwardX`
  - `api.AudioListener.forwardY`
  - `api.AudioListener.forwardZ`
  - `api.AudioListener.positionX`
  - `api.AudioListener.positionY`
  - `api.AudioListener.positionZ`
  - `api.AudioListener.upX`
  - `api.AudioListener.upY`
  - `api.AudioListener.upZ`

And then we'd still need a way to deal with `setOrientation()` and `setPosition()`. What would we call a discouraged feature that contains those 2 keys? "Spatial web audio (discouraged)"?

captainbrosset:
I implemented what I said in [my previous comment](https://github.com/web-platform-dx/web-features/pull/2628#issuecomment-3755186554) and I'm actually quite happy with how it looks.

@ddbeck what do you think?

features/audiolistener-setposition-setorientation.yml
ddbeck, line 1: I'd set the ID to `audiolistener-setposition-setorientation` (rather than encoding the status into the ID)

features/audio-listener-discouraged.yml
```diff
@@ -0,0 +1,14 @@
+name: AudioListener setPosition() setOrientation()
```
ddbeck:
```suggestion
name: AudioListener setPosition() and setOrientation()
```

features/audio-listener-discouraged.yml
```diff
@@ -0,0 +1,14 @@
...
+description: The `setPosition()` and `setOrientation()` methods of the `AudioListener` API set the position and orientation of a person listening to the audio scene, allowing you to spatialize audio.
```
ddbeck: A few things here:

- One style nit (no need to allow)
- I don't really know what "audio scene" means and the spec doesn't define it, so I don't think this is a term of art. Maybe this would be a bit more plain language?
- Probably nice to mention "Web Audio" here?

```suggestion
description: The `setPosition()` and `setOrientation()` methods of the `AudioListener` API set the virtual position and orientation of a person relative to sound sources, to spatialize audio in a Web Audio context.
```

features/audio-listener-discouraged.yml
```diff
@@ -0,0 +1,14 @@
...
+spec: https://webaudio.github.io/web-audio-api/#AudioListener
+group: web-audio
+discouraged:
+  reason: The AudioListener's `setPosition()` and `setOrientation()` methods are deprecated as they are equivalent to using the `positionX`, `positionY`, `positionZ`, `forwardX`, `forwardY`, `forwardZ`, `upX`, `upY`, and `upZ` instance properties.
```
ddbeck: I don't think we need to say that they're deprecated again. Small formatting tweak too.

```suggestion
  reason: The `setPosition()` and `setOrientation()` methods are equivalent to using an `AudioListener` instances's `positionX`, `positionY`, `positionZ`, `forwardX`, `forwardY`, `forwardZ`, `upX`, `upY`, and `upZ` properties.
```

features/audiolistener.yml
ddbeck, line 1: I'd suggest `audiolistener` as the ID. Not required, but suggested by the discussion on https://github.com/web-platform-dx/web-features/issues/2866.

features/audio-listener.yml
```diff
@@ -0,0 +1,15 @@
+name: AudioListener
+description: The `AudioListener` API sets the position and orientation of a person listening to the audio scene, allowing you to spatialize audio.
```
ddbeck: Suggestions to mirror my suggestions on the other feature:

```suggestion
description: An `AudioListener` object represents the virtual position and orientation of a person relative to sound sources, to spatialize audio in a Web Audio context.
```

PR #2629, opened 2025-02-11
Discouraged cssom interfaces

captainbrosset:


features/css-object-model-discouraged.yml
```diff
@@ -0,0 +1,34 @@
+name: CSS object model (discouraged)
+description: The superseded DOM level 2 style specification defined interfaces to access and modify CSS styles, which are now excluded from the CSS Object Model.
```
ddbeck, line 2: I like describing these discouraged features neutrally if we're able to, rather than recapitulating the other data. What do you think of this approach?

```suggestion
name: CSS object model (DOM level 2)
description: The DOM level 2 style specification defined interfaces to access and modify CSS styles, such as `CSSValue` and `CSSPrimitiveValue`, that were later excluded from the CSS object model.
```

PR #2633, opened 2025-02-11
Move `file-and-directory-entries` out of drafts

ddbeck:
This is an extremely old draft feature that we added almost at the very beginning. It kinda works still but overlapping keys are a bit of a fuss.

See also: https://github.com/web-platform-dx/web-features/pull/2632

### DIFF
--- a/features/file-and-directory-entries.yml
+++ b/features/file-and-directory-entries.yml
@@ -1,4 +1,3 @@
-draft_date: 2024-05-07
 name: File and directory entries
 description: File and directory entries represent a virtual file system for working with files and directories dragged and dropped into the browser.
 spec: https://wicg.github.io/entries-api/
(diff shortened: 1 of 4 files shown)


foolip (commented):
The computed Baseline status is false. Is that because some part of this is deprecated? Or we need compute_from?

ddbeck:
The MDN page for this API [mentions two main entry points](https://developer.mozilla.org/en-US/docs/Web/API/File_and_Directory_Entries_API#getting_access_to_a_file_system) — via the `` and `` — neither of which are Baseline-eligible because neither is supported on Firefox for Android (nor is the related drag and drop API). There are some interfaces exposed on Firefox for Android, though I'm unsure on what utility it would have with some of the pieces missing.

captainbrosset:
> Hmm, this might also be broken in Chrome too? See [mdn/browser-compat-data#19304](https://github.com/mdn/browser-compat-data/issues/19304)

Looks like the interface names are different between Firefox and Chromium: dropping files on an `<input type=file>` leads to `FileEntry` objects in Chromium, whereas it creates `FileSystemFileEntry` objects in Firefox (I have not tested in Safari).

The spec disagrees with Chromium: https://wicg.github.io/entries-api/#api-fileentry
See these discussions about changing the interface names: https://groups.google.com/a/chromium.org/g/blink-dev/c/_6Euwqv366U and https://github.com/WICG/entries-api/issues/6.

That said, this code works both in Chromium and Firefox, irrespective of the interface names:

```html
<!DOCTYPE html>

<label for="file-input">Drop files on the input</label>
<input type="file" id="file-input" multiple>

<script>
  const inputEl = document.getElementById('file-input');
  inputEl.addEventListener('change', function(event) {
    const entries = inputEl.webkitEntries;
    console.log(entries);
  });
</script>
```

ddbeck:
There are now three open PRs for file and directory entries:

- https://github.com/web-platform-dx/web-features/pull/2633 (this one)
- https://github.com/web-platform-dx/web-features/pull/2993
- https://github.com/web-platform-dx/web-features/pull/3438

PR #2635, opened 2025-02-11
Add `protected-audience` and `ad-selection` features

ddbeck:


captainbrosset:
Protected Audience is part of Google's Privacy Sandbox project. Edge has a different proposal for this particular API, called the Ad Selection API.
See https://blogs.windows.com/msedgedev/2024/03/05/new-privacy-preserving-ads-api/ and https://blogs.windows.com/msedgedev/2024/10/08/ad-selection-api-limited-preview/

So the fact that BCD shows both Chrome and Edge as supporting Protected Audience feels wrong. 
Edge generally mirrors Chrome by default, but this is one case where it probably shouldn't.

I'll try to find more information to (possibly) fix BCD.

features/protected-audience.yml
```diff
@@ -1,7 +1,8 @@
-draft_date: 2025-01-20
-name: Protected Audience (formerly FLEDGE)
-description: TODO
+name: Protected Audience
+description: The Protected Audience API facilitates advertisement sales by allowing sites to register users as part of an interest group or to choose which ads appear based on those interest groups, while minimizing the ability of advertisers to track specific members of the interest group. Also known as FLEDGE.
 spec: https://wicg.github.io/turtledove/
+status:
+  compute_from: api.Navigator.getInterestGroupAdAuctionData
```
ddbeck, line 5: This is a bit awkward. It might be nice to unset this when Chrome 134 hits stable.

See also https://github.com/web-platform-dx/web-features/pull/2638#discussion_r1951099639

ddbeck:
> the overall feature that these BCD keys contribute to isn't the same in Edge and in Chrome

Hmm, that's a bit awkward.

I found [this doc about the API shape](https://github.com/WICG/privacy-preserving-ads/blob/main/API%20Details.md#api-shape) to be helpful toward this but not sufficiently detailed. I think what I really need here are IDLs for the two APIs, so we can figure out what constitutes evidence of support for the two features. Even if the API surface is very similar, my hunch is that the difference is in configuration dictionaries that the BCD Collector won't detect (or rather won't do anything with by default).

This is likely to be a feature that really demands https://github.com/web-platform-dx/web-features/issues/1173.

captainbrosset:
> I found [this doc about the API shape](https://github.com/WICG/privacy-preserving-ads/blob/main/API%20Details.md#api-shape) to be helpful toward this but not sufficiently detailed.

Yes, that's the right doc.

> I think what I really need here are IDLs for the two APIs, so we can figure out what constitutes evidence of support for the two features. Even if the API surface is very similar, my hunch is that the difference is in configuration dictionaries that the BCD Collector won't detect (or rather won't do anything with by default).

I've started an ms-internal discussion to try and help here. While the IDLs may be very similar (apart from the ProtectedAudience interface, which won't exist in Edge), I think they will work differently. For example, runAdAuction won't do exactly the same thing in Chrome and in Edge.

> This is likely to be a feature that really demands #1173.

Definitely!

ddbeck:
@captainbrosset Were you able to learn anything more about the difference between the protected audience and ad selection APIs? I ask because it looks like [protected audience](https://groups.google.com/a/chromium.org/g/blink-dev/c/k_nubsMb97g) is going away, so being able to mint a (discouraged) feature would be desirable. But that would require know whether Edge is keeping some or all of the API or dropping it too, given the previous discussion here.

rviscomi:
Tagging in @samdutton to comment on the status of the API

samdutton:
Protected Audience is deprecated in Chrome.

FYI @kevinkiklee (Dev Rel for Protected Audience)

[goo.gle/ps-status](https://privacysandbox.google.com/overview/status) has feature status for Privacy Sandbox APIs.

captainbrosset:
> Were you able to learn anything more about the difference between the protected audience and ad selection APIs? I ask because it looks like [protected audience](https://groups.google.com/a/chromium.org/g/blink-dev/c/k_nubsMb97g) is going away, so being able to mint a (discouraged) feature would be desirable. But that would require know whether Edge is keeping some or all of the API or dropping it too, given the previous discussion here.

Edge will be removing the equivalent downstream API along with the deprecation effort too.

captainbrosset:
Discussed on the meeting: It would be nice to have a tombstone feature for it. Removal date+reason. Don't know if there is docs for this on MDN. But on some features, there's info like "pending removal".
Two ways: we could mint 2 features, with overlapping keys (Edge and Chrome). Or mind just 1 (Chrome-one, and pretending the Edge one doesn't exist). Would be nice to do this in the next couple of weeks before Chrome 150, which is removing the features.

@ddbeck will update this PR.

features/ad-selection.yml
```diff
@@ -0,0 +1,14 @@
+name: Ad selection
+description: The ad selection API
```
ddbeck: Just needs a description. I don't think we need to put a ton of work into this, if it's not being developer further. Maybe this is sufficient?

```suggestion
description: The ad selection API facilitates showing advertisements to users based on interest groups that prevent re-identification of users within the group.
```

ddbeck:
@captainbrosset I've sorted this out, I think. Would you have a look? The only missing thing is a description for the ad selection API.

Edit: I also sent https://github.com/mdn/browser-compat-data/pull/29723 upstream, to reconcile Chrome and Edge's different statuses with the actual API.

features/ad-selection.yml
captainbrosset: We could also just pretty much reuse the Protection Audience API description since both APIs are for the same use case:

```suggestion
description: The Ad Selection API facilitates advertisement sales by allowing sites to register users as part of an interest group or to choose which ads appear based on those interest groups, while minimizing the ability of advertisers to track specific members of the interest group.
```

captainbrosset (approved):

PR #2636, opened 2025-02-11
Topics API feature

captainbrosset:
I need help to describe this as I don't know the API much at all, so I'm not sure what the entry points are.
This is a non-standard API that's part of [Google's privacy sandbox initiative](https://developers.google.com/privacy-sandbox).
See https://developers.google.com/privacy-sandbox/private-advertising/topics for more information.


lolaodelola:
I wrote about this a couple of years ago so happy to take a stab but is this a web feature? Both Mozilla and Firefox have negative standards positions on it, it's not on the standards track and it's generally considered not good for the web (at this moment in time) - should we be adding this as a web feature here? Is there developer interest in Topics API?

ddbeck:
> should we be adding this as a web feature here?

To answer @lolaodelola's question, `web-features` is primarily descriptive. When an API ships in a browser, we ought to record the presence of that feature, whether nor not it's good or useful or popular. In an ideal world, single-implementation features that don't get uptake will be deprecated or withdrawn, which we can also describe when it happens.

PR #2638, opened 2025-02-11
Add `shared-storage-locks`

ddbeck:
See:

- https://github.com/WICG/shared-storage/blob/main/README.md#locking-for-modifier-methods
- https://issues.chromium.org/issues/373899210
- https://chromestatus.com/feature/5133950203461632


Elchi3 (commented):
The alternative here would be to make a new feature "Web Locks integration with Shared Storage"
https://chromestatus.com/feature/5133950203461632
https://github.com/WICG/shared-storage?tab=readme-ov-file#locking-for-modifier-methods

ddbeck:
> The alternative here would be to make a new feature "Web Locks integration with Shared Storage" https://chromestatus.com/feature/5133950203461632 https://github.com/WICG/shared-storage?tab=readme-ov-file#locking-for-modifier-methods

@Elchi3 thank you! I wasn't aware of this. I've retooled this PR as adding `shared-storage-locks`. I noted in a comment that there's potentially missing keys in BCD, depending on how thorough you wish to be there. I also updated the PR description with some additional explanatory links. This was a little hard to get my head around. 😅

Elchi3 (approved):

PR #2642, opened 2025-02-12
getBoxQuads() feature

captainbrosset:
See https://hacks.mozilla.org/2014/03/introducing-the-getboxquads-api/

This API has been in Firefox for more than 10 years now. While I was on the Firefox DevTools team, I remember we would make use of this extensively to be able to draw the various boxes of a rendered node. It's nice because it gives you all of the individual line boxes for a multiline text node for example. Or, for a div that's fragmented across multiple columns, it also returns multiple quads.
I believe this was never shipped outside of Firefox Nightly though.


ddbeck (approved):
One nit in a comment, but this looks good. Thank you!

PR #2643, opened 2025-02-12
document.write() feature

captainbrosset:
document.open and document.close are not discouraged in the spec, but they're only useful for document.write anyway, so we might as well group all of these related methods together in a new discouraged feature.


features/document-write.yml
```diff
@@ -0,0 +1,15 @@
...
+discouraged:
+  # The spec doesn't actually discourage against open and close, but these methods are
+  # really only useful for writing anyway, so we're marking the whole feature as discouraged.
+  according_to:
+    - https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#document.write()
+  alternatives:
+    - dom
+compat_features:
+  - api.Document.write
+  - api.Document.writeln
```
ddbeck, line 13: Oh interesting. I ran into `writeln()` independently because of trusted types stuff. BCD doesn't mark it as discouraged, but MDN does. I sent https://github.com/mdn/content/pull/38267 to fix MDN. It's one of those awkward cases where MDN got ahead of a discussion on the spec that never resolved (https://github.com/whatwg/html/issues/7977). I don't love it.

All that said, describing this overall _feature_ as discouraged seems correct to me though. This implies…

```suggestion
  - api.Document.write
  - api.Document.write.accepts_TrustedHTML
  - api.Document.writeln
  - api.Document.writeln.accepts_TrustedHTML
```

captainbrosset:
@ddbeck interestingly, adding `accepts_TrustedHTML` keys makes the feature appears as if it's only supported in chromium. Not that it matters since the whole feature is discouraged, and therefore not baseline. But it makes the compat data here a little weird because document.write has been supported everywhere for a long time. So, for consumers who only care about the baseline status, nothing to worry about. But for those who show the complete compat data, then it might make sense to use a `compute_from`, although doing so here fails the build because the key we'd compute from (`api.Document.write`) is deprecated.

ddbeck:
@captainbrosset you found a bug! I sent https://github.com/web-platform-dx/web-features/pull/2689 to fix it.

Here's what that fix would look like on this branch: https://github.com/web-platform-dx/web-features/compare/document-write...ddbeck:web-features:fix-document-write

ddbeck (approved):
Thank you for the review on the other PR, Patrick! I've merged main into this branch and refreshed the dist. All is well now.

PR #2652, opened 2025-02-14
Split text-underline-offset from text-underline-position

foolip:
These properties are related, but they can be used individually.
text-underline-position is signifcantly older in some browser. Usage is
also quite different:
https://chromestatus.com/metrics/css/timeline/popularity/405
https://chromestatus.com/metrics/css/timeline/popularity/661

The motivation for this is mapping the above use counters to counters
for the web-features entries. They are technically easy to measure as
one, but doing so seems unhelpful.


captainbrosset (approved):

features/text-underline-position.yml
```diff
@@ -1,13 +1,10 @@
-name: Text underline position
-description: "The `text-underline-position` and `text-underline-offset` CSS properties set the position and distance from initial position of text underlines on text with `text-decoration: underline`."
-spec: https://drafts.csswg.org/css-text-decor-4/#line-decoration
+name: text-underline-position
+description: "The `text-underline-position` CSS property set the position of underlines text. For example, `text-underline-position: under` places the underline below the text, avoiding crossing descenders. The underline may be further adjusted by the `text-underline-offset` property."
```
jamesnw, line 2:
```suggestion
description: "The `text-underline-position` CSS property sets the position of underlines on text. For example, `text-underline-position: under` places the underline below the text, avoiding crossing descenders. The underline may be further adjusted by the `text-underline-offset` property."
```

PR #2653, opened 2025-02-14
Add ::scroll-button feature

flackr:


ddbeck:
Hi @flackr, this is a question for each of your recent PRs: am I correct in understanding that these are all proposed features that are not yet shipping in any browsers? They seem to reference browser compat data and caniuse entries that don't exist.

(We're still working out a process for what we've been calling "early" features. I'm betting these are likely to become important test cases.)

captainbrosset:
I believe this is part of a larger feature aimed at enabling CSS-only accordion widgets, and I think there's an implementation in Chromium behind a flag.

flackr:
> I believe this is part of a larger feature aimed at enabling CSS-only carousel widgets, and I think there's an implementation in Chromium behind a flag.

It is a set of independent features, but yes, they are part of a larger goal to make it easy to author pure CSS carousel experiences: https://github.com/w3c/csswg-drafts/blob/main/css-overflow-5/carousel-explainer.md

There is an implementation in chromium and I2S's have been sent for these features:
https://groups.google.com/a/chromium.org/g/blink-dev/c/7EQ8-VzPZh0/m/NMyrGCjuAAAJ
https://groups.google.com/a/chromium.org/g/blink-dev/c/O5ZmbyRh9LE/m/lHxxrdiHAAAJ

> They seem to reference browser compat data and caniuse entries that don't exist.

What do we need to do to add these entries?

captainbrosset:
> What do we need to do to add these entries?

The BCD entries (aka BCD keys) need to be added to the [browser-compat-data project](https://github.com/mdn/browser-compat-data) first. For this to happen, there needs to be an implementation of the feature in at least one browser, which you said there was, in Chromium. In theory, the BCD Collector should even detect the new feature and create the new keys for you automatically (provided the implementation actually ships enabled with a given milestone of Chromium).

If this early feature doesn't yet have an enabled-by-default implementation, then we can still create yaml files for it here, and just omit the `compat_features` part.
See the customizable select feature as prior art for this: https://github.com/web-platform-dx/web-features/blob/main/features/customizable-select.yml

features/scroll-buttons.yml
```diff
@@ -0,0 +1,9 @@
+name: ::scroll-button
+description: The `::scroll-button(<dir>)` CSS pseudo-element is a focusable button which scrolls its originating scroll container element in the specified direction.
+spec: https://drafts.csswg.org/css-overflow-5/#scroll-buttons
+group: scrolling
+caniuse: css-scroll-buttons
```
captainbrosset, line 5: Let's remove this line, there's no caniuse feature called css-scroll-buttons yet. When there is, we can add it.
See also: https://github.com/web-platform-dx/web-features/blob/main/docs/CONTRIBUTING.md, which has some information about caniuse.

features/scroll-buttons.yml
```diff
@@ -0,0 +1,9 @@
...
+status:
+  compute_from: css.selectors.scroll-button
+compat_features:
+  - css.selectors.scroll-button
```
captainbrosset, line 9: Let's remove all of this if the feature isn't yet covered by BCD.
If the implementation of this feature is shipping enabled then let's pause this PR, add the corresponding BCD keys (or check if the BCD collector can do it automatically), and then add the correct BCD key(s) here.
If not (and [the chromestatus entry](https://chromestatus.com/feature/5093129273999360) seems to say this isn't yet enabled), then let's leave this out for now.

captainbrosset (changes requested):

features/scroll-buttons.yml
flackr: Sounds good, I've removed the caniuse reference in all the PRs

captainbrosset (approved):
I think this looks good to me as is. Would be good for @ddbeck  to take another look as well since we don't yet a lot of early features like this and I know we have a goal this year of making it easier to create by then and modify them over time as spec changes happen.

captainbrosset (approved):

PR #2654, opened 2025-02-14
Add scroll-markers feature

flackr:


features/scroll-markers.yml
```diff
@@ -0,0 +1,4 @@
+name: Scroll markers
+description: The `::scroll-marker` CSS pseudo-element is a focusable control within its ancestor scroll container's `::scroll-marker-group` that scrolls to its originating element.
```
captainbrosset, line 2: I'm having trouble understanding this. Would you mind giving me a practical example of how this might work/what this would look like? This would help me understand, and potentially propose a slight rewrite of this.

flackr: If you have chrome canary with experimental web platform features enabled, there are some great demos at:
* https://chrome.dev/carousel-configurator/ (turn on "Dot Navigation" switch) which shows the CSS used to produce the scroll markers, and
* https://chrome.dev/carousel/ which has many demos using scroll markers to create dots to navigate to particular targets.

As an example, the following creates a set of circular "markers" to scroll to each of the slides in a slideshow:
```html
<style>
  .slides {
    overflow: auto;
    scroll-snap-type: x mandatory;
    white-space: nowrap;
    scroll-marker-group: after;
    counter-reset: slide-number;
  }
  .slide::scroll-marker {
    counter-increment: slide-number;
    content: counter(slide-number);
    border-radius: 50%;
    border: 1px solid black;
  }
}
</style>
<div class=slides>
  <div class=slide></div>
  <div class=slide></div>
  <div class=slide></div>
  <div class=slide></div>
  <div class=slide></div>
<div class=slideshow>
```

Here's a somewhat simple codepen trying to show the same: https://codepen.io/flackr/pen/MWMLPpP

captainbrosset:
Looks like the scroll-marker BCD keys are coming with BCD v5.7.1, which I'm trying to update us to in #2747. As soon as that PR is merged, we'll be able to reference the new BCD keys in your new feature.

features/scroll-markers.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://drafts.csswg.org/css-overflow-5/#scroll-markers
+group: scrolling
```
captainbrosset, line 4: BCD now has the relevant keys, so we need to add them here:

```suggestion
group: scrolling
compat_features:
  - css.selectors.scroll-marker
  - css.selectors.scroll-marker-group
  - css.properties.scroll-marker-group
  - css.properties.scroll-marker-group.after
  - css.properties.scroll-marker-group.before
  - css.properties.scroll-marker-group.none
```

Can you think of any other keys that would belong to this feature?

Once you've added them, I'm afraid you'll need to run the `npm run dist ...` command again :)

captainbrosset (changes requested):

features/scroll-markers.yml
flackr: I believe that's everything for this feature, thanks!

features/scroll-markers.yml
```diff
@@ -0,0 +1,11 @@
+name: Scroll markers
```
captainbrosset, line 1:
```suggestion
name: `::scroll-marker` and `::scroll-marker-group`
```

would be more consistent with other pseudo-element features.

package-lock.json
```diff
@@ -4936,7 +4936,7 @@
       }
     },
     "packages/compute-baseline": {
-      "version": "0.2.0",
+      "version": "0.3.0",
```
captainbrosset, line 4939: Can you update with the latest main please? This diff should not be here.

features/scroll-markers.yml
captainbrosset: Ah, but come to think of it, this feature also contains the `scroll-marker-group` **property**, so the feature isn't just about pseudos.
I think "Scroll markers" (your initial proposal) is therefore better.

features/scroll-markers.yml
captainbrosset: Now that I've read the spec a bit more, I'd like to propose different alternatives:

```suggestion
description: The `::scroll-marker-group` CSS pseudo-element contains scroll markers (stylable with the `::scroll-marker` pseudo-element), and appears either immediately before or after a scroll container.
```

Or 

```suggestion
description: A scroll container can be navigated by activating `::scroll-marker` pseudo-elements which appear in a generated `::scroll-marker-group` pseudo-element, either before or after the scroll container.
```

captainbrosset (changes requested):

package-lock.json
flackr: Done, the diff no longer seems to be there.

captainbrosset (approved):

PR #2656, opened 2025-02-14
Add CSS interactivity feature.

flackr:


features/interactivity.yml
```diff
@@ -0,0 +1,3 @@
+name: interactivity
+description: The CSS `interactivity` property specifies whether an element and its descendants are inert or not.
```
captainbrosset, line 2:
```suggestion
description: The CSS `interactivity` property makes an element and its descendants inert, like when using the `inert` HTML attribute. Inert elements can't be focused or clicked, their text can't be selected or found using the browser's find-in-page feature.
```

I thought it'd be good to give more details about what inert means, and to mention the related html attribute as well.

captainbrosset (approved):

features/interactivity.yml
jamesnw, line 2:
```suggestion
description: The `interactivity: inert` CSS declaration makes an element and its descendants inert, like when using the `inert` HTML attribute. Inert elements can't be focused or clicked, their text can't be selected or found using the browser's find-in-page feature.
```

I like adding the description of inert here. I'd suggest adding the full declaration here as well.

features/interactivity.yml
```diff
@@ -0,0 +1,3 @@
+name: interactivity
+description: The `interactivity: inert` CSS declaration makes an element and its descendants inert, like when using the `inert` HTML attribute. Inert elements can't be focused or clicked, their text can't be selected or found using the browser's find-in-page feature.
```
ddbeck, line 2: This won't parse without quotes:

```suggestion
description: "The `interactivity: inert` CSS declaration makes an element and its descendants inert, like when using the `inert` HTML attribute. Inert elements can't be focused or clicked, their text can't be selected or found using the browser's find-in-page feature."
```

features/interactivity.yml
```diff
@@ -0,0 +1,3 @@
...
+spec: https://drafts.csswg.org/css-ui-4/#inertness
```
ddbeck, line 3: It looks like this is going to first in Chrome beta in a few weeks? Maybe we can append a comment for now:

```suggestion
spec: https://drafts.csswg.org/css-ui-4/#inertness
# Expected keys:
# - css.properties.interactivity
# - css.properties.interactivity.auto
# - css.properties.interactivity.inert
```

ddbeck (commented):
Two small suggestions in line comments. This also needs a dist file (via `npm run dist`) to get merged.

This looks near enough to having an implementation tracked in BCD (if https://chromestatus.com/feature/5107436833472512 is to be believed) that we don't really have to wait on this, once it's merge-able.

features/interactivity.yml
flackr: I added the more detailed suggestion @jamesnw suggested.

ddbeck (approved):

PR #2660, opened 2025-02-17
Add feature for `moveBefore()` DOM methods (and group some DOM features)

ddbeck:
This adds a feature for the state-preserving method for moving DOM subtrees, `moveBefore()`. While I was here, I created a group for DOM features that operate on the DOM in general. That is, they represent or modify the document tree (rather than represent specific elements or provide other APIs).


features/movebefore.yml
```diff
@@ -0,0 +1,8 @@
+name: moveBefore()
+description: The `moveBefore()` DOM methods relocate a node while preserving its state. For example, you can move an element without disrupting focus or move an iframe without reloading it.
```
captainbrosset, line 2: I think this is the list of states that get preserved:

- `<iframe>` elements remain loaded.
- The active element retains focus.
- Popovers and dialogs remain open.
- CSS transitions and animations continue to run.

Maybe mention one more:

```suggestion
description: The `moveBefore()` DOM method relocates a node while preserving its state. For example, you can move the active element without losing focus, move an animated element without reseting the animation, or move an iframe without reloading its content.
```

features/dom-geometry.yml
```diff
@@ -1,6 +1,7 @@
 name: DOM Geometry
 description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` interfaces offer a way to represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
 spec: https://drafts.fxtf.org/geometry-1/
+group: dom
```
foolip, line 4: I don't think this fits the description of the group. These geometry interfaces don't interact with the tree of nodes really. If I had to guess, I think they're just prefixed with "DOM" because the shorter names were already used in the wild for other things.

features/gethtml.yml
```diff
@@ -1,6 +1,7 @@
 name: getHTML()
 description: The `getHTML()` method of `Element` and `ShadowRoot` objects serializes an element's DOM to an HTML string.
 spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#html-serialization-methods
+group: dom
```
foolip, line 4: `getHTML()` is specific to HTML:

```suggestion
group: html
```

scripts/specs.ts
```diff
@@ -104,6 +104,10 @@ const defaultAllowlist: allowlistItem[] = [
     [
         "https://www.w3.org/TR/DOM-Level-2-Style/",
         "Allowed because the css-object-model-discouraged feature points to it."
+    ],
+    [
+        "https://github.com/whatwg/dom/issues/1255",
+        "This is only place moveBefore() is spec'd. Remove this exception when https://github.com/whatwg/dom/issues/1255 merges."
```
captainbrosset, line 110: Technically this is an issue, not a PR, so maybe "when ... is fixed".

captainbrosset (approved):

features/movebefore.yml
```diff
@@ -0,0 +1,8 @@
+name: moveBefore()
```
foolip, line 1: Should the ID be "move-before"? I suspect we have a mix of styles now, do our guidelines say one way or the other?

features/movebefore.yml
```diff
@@ -0,0 +1,8 @@
...
+description: The `moveBefore()` DOM methods relocate a node while preserving its state. For example, you can move an element without disrupting focus or move an iframe without reloading it.
+spec: https://github.com/whatwg/dom/issues/1255
```
foolip, line 3:
```suggestion
spec: https://github.com/whatwg/dom/pull/1307
```

features/structured-clone.yml
```diff
@@ -1,3 +1,4 @@
 name: structuredClone()
 description: The `structuredClone()` global method creates a deep copy of an object. Values that cannot be cloned can instead be transferred, making the original value no longer usable.
 spec: https://html.spec.whatwg.org/multipage/structured-data.html#structured-cloning
+group: dom
```
foolip, line 4: Actually no, `structuredClone()` handles arbitrary JS values. And it can't clone elements, it seems from my testing.

features/gethtml.yml
ddbeck: Can you elaborate on this one a bit? I thought of this as "fancy `innerHTML`," which is part of the basic `dom` feature. It seemed weird to me that they would be in separate groups.

features/movebefore.yml
ddbeck: We don't have a guideline, but it would be consistent with `structured-clone` at least. Done with 0a377830417471e63c0d039803e794b981258ba2.

features/dom-geometry.yml
ddbeck: Dropped with d6c38814a3c4abda69fe7971524da3a063a475b5.

features/gethtml.yml
captainbrosset: I think we should keep `dom` here. The `html` group is typically used for HTML elements, not for APIs.

captainbrosset (approved):

PR #2661, opened 2025-02-17
Add features for permissions policy and feature policy

ddbeck:
This PR adds a feature for permissions policy and a discouraged feature for feature policy. This is a big mess, so these are draft features.

Unfortunately, BCD doesn't accurately record the data for these features and it it's a bit of work to unwind this. I propose that we add these as drafts for now and, at the time we merge this, file an issue on BCD that requests `Feature-Policy` and `Permissions-Policy` be separated into separate keys and see where things go from there.

The full backstory:

- The `Feature-Policy` header was once added to browsers.
- Later, the `Feature-Policy` was "renamed" to `Permissions-Policy`. "Renamed" because yes [it was renamed](https://github.com/w3c/webappsec-permissions-policy/pull/379), but it also got [a breaking syntax change](https://github.com/w3c/webappsec-permissions-policy/pull/383) at the same time (see also [Control browser features with Permissions Policy  |  Privacy & Security  |  Chrome for Developers](https://developer.chrome.com/docs/privacy-security/permissions-policy#changes_to_permissions_policy)).
- caniuse told this story by describing [feature policy](https://caniuse.com/feature-policy) and [permissions policy](https://caniuse.com/permissions-policy) as separate features. In my opinion, this was the correct approach.
- BCD told this story by [pretending that `Feature-Policy` was an alternative name for `Permissions-Policy`](https://github.com/mdn/browser-compat-data/pull/18250), ignoring the syntax change. I'm not quite sure why this happened, given [the failure of such an approach earlier](https://github.com/mdn/browser-compat-data/pull/11332).

Given this story, we cannot accurately calculate _either_ feature's implementation status in a way that accords with caniuse or, in my opinion, confusing developers (because let me tell you, _I_ was confused about this and I knew most of the backstory already).


ddbeck:
cc: @Elchi3 you might have opinions about this one, since I expect it to impact BCD as well.

features/draft/feature-policy.yml
```diff
@@ -0,0 +1,13 @@
...
+description: The `Feature-Policy` response header sets whether a policy-controlled feature, such as an API, may be used in a document. Not to be confused with permissions policy.
+spec: https://www.w3.org/TR/2019/WD-feature-policy-1-20190416/
+caniuse: feature-policy
+discouraged:
+  according_to:
+    - https://github.com/w3c/webappsec-permissions-policy/pull/379
+  alternatives:
+    - permissions-policy
+compat_features:
+  - api.Document.featurePolicy
```
foolip, line 12: Also https://github.com/mdn/browser-compat-data/blob/main/api/FeaturePolicy.json?

features/draft/permissions-policy.yml
```diff
@@ -0,0 +1,8 @@
+draft_date: 2025-02-16
+name: Permissions policy
+description: The `Permissions-Policy` response header sets whether a policy-controlled feature, such as an API, may be used in a document.
+spec: https://w3c.github.io/webappsec-permissions-policy/
+caniuse: permissions-policy
+compat_features:
+  - http.headers.Permissions-Policy
```
foolip, line 7: I guess none of https://w3c.github.io/webappsec-permissions-policy/#idl-index has shipped...

ddbeck:
I've sent https://github.com/mdn/browser-compat-data/pull/26708 to start unwinding the compat data, to unblock this PR.

ddbeck:
@foolip I think this kinda works now. Can I get a re-review?

features/feature-policy.yml
```diff
@@ -0,0 +1,55 @@
+name: Feature policy
+description: The `Feature-Policy` response header sets whether a policy-controlled feature, such as an API, may be used in a document. Not to be confused with permissions policy.
```
foolip, line 2: The `allow` attributes on iframes was also an important part of the original feature policy. That's now defined by HTML and permissions policy. This feature would have been described differently at the time it was shipped, but we should update description as the platform evolves so this is OK.

features/permissions-policy.yml
```diff
@@ -0,0 +1,7 @@
+name: Permissions policy
+description: The `Permissions-Policy` response header sets whether a policy-controlled feature, such as an API, may be used in a document.
```
foolip: The `allow` attribute is the other entrypoint to this feature and should be mentioned in the description, I think.

features/permissions-policy.yml
```diff
@@ -0,0 +1,7 @@
...
+spec: https://w3c.github.io/webappsec-permissions-policy/
+caniuse: permissions-policy
+compat_features:
```
foolip, line 5: I would also suggest moving the `allow` attribute into here, yanking it from the iframe feature.

foolip:
Ping @ddbeck on my review comments.

features/permissions-policy.yml
ddbeck: Fixed with b65c5e72800825f99126eea1088e9428b4affdc5.

ddbeck:
@foolip I've addressed all the comments. This is ready for re-review.

features/feature-policy.yml.dist
```diff
@@ -0,0 +1,196 @@
+# Generated from: feature-policy.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support: {}
```
foolip: We'll need `compute_from` to get a sensible version here. I'll do that.

foolip (approved):

PR #2665, opened 2025-02-17
Add features and assign more keys for workers (and add a workers group)

ddbeck:


features/dedicated-workers.yml
```diff
@@ -0,0 +1,39 @@
...
+group: workers
+status:
+  compute_from: api.WorkerGlobalScope.importScripts
+compat_features:
+  - api.DedicatedWorkerGlobalScope
+  - api.DedicatedWorkerGlobalScope.close
+  - api.DedicatedWorkerGlobalScope.name
+  - api.Worker
+  - api.Worker.Worker
+  - api.Worker.Worker.mime_checks
```
lolaodelola, line 14: This is a process/decision-making question: I noticed that only `close` and `name` are included for `DedicatedWorkerGlobalScope`, why only those two? Is there something unique about them that the other properties don't have?

features/messageerror.yml
```diff
@@ -4,3 +4,15 @@ spec:
...
+    - api.DedicatedWorkerGlobalScope.messageerror_event
+    - api.MessagePort.messageerror_event
+    - api.ServiceWorkerContainer.messageerror_event
+    - api.Window.messageerror_event
+compat_features:
+  - api.DedicatedWorkerGlobalScope.messageerror_event
+  - api.MessagePort.messageerror_event
+  - api.ServiceWorkerContainer.messageerror_event
+  - api.Window.messageerror_event
+  - api.Worker.messageerror_event
```
lolaodelola, line 18: Why are errors separated into their own feature instead of with their parent features?

lolaodelola (commented):
Everything looks good to me in general. I have some questions about decisions but I don't think they're necessarily blockers.

features/dedicated-workers.yml
captainbrosset: In this particular case, it seems to be because the other BCD keys under api.DedicatedWorkerGlobalScope are already assigned to other features.
If you open https://web-platform-dx.github.io/web-features-explorer/ids/, you'll see the entire list of features, together with their BCD keys. On that page, you can do a search for "api.DedicatedWorkerGlobalScope".
You'll see that api.DedicatedWorkerGlobalScope.message_event and api.DedicatedWorkerGlobalScope.postMessage, for example, are already part of the postmessage feature.
Currently, in web-features, a given BCD key can only be part of one feature at most. This is something we might change in the future.

You're asking a good question though. I don't think there's a super clear way to answer whether those BCD keys should be in postmessage, or in dedicated-workers.

groups/workers.yml
```diff
@@ -0,0 +1,2 @@
+# Features for running scripts alongside the main thread (i.e., workers and worklets)
+name: Workers
```
captainbrosset, line 2: This group should probably be assigned to the postmessage feature too.

features/messageerror.yml
captainbrosset: That's a good question too. I don't know how these features came to be. The guiding principle for features should always be: does this feature make sense, individually, to web developers? Is this something they talk about separately from other things? Is this perceived as an incremental addition to the platform?

I'm not super convinced that this feature should exist on its own to be honest, but I don't have the context. Maybe the guiding principle I mentioned before applies here. I just don't know enough to be sure.

One other aspect to keep in mind is that, over time, as features mature and get baseline high, it makes less and less sense for them to live on their own. That's why we're working on a mechanism to merge features. Maybe messageerror makes sense now, but in a few years from now, we might want to merge it into a bigger parent feature. By then, it's likely that developers won't think of messageerror as a separate thing anymore.

captainbrosset (approved):

Issue #2250, opened 2024-11-14
Add popover=hint feature


mfreed7:
### Specification

https://github.com/whatwg/html/pull/9778

### Description

The Popover API specifies the behavior for two values of the `popover` attribute: `auto` and `manual`. This feature describes a third value, `popover=hint`. Hints, which are most often associated with "tooltip" type behaviors, have slightly different behaviors. Primarily, the difference is that `hint`s are subordinate to `auto`s when opening nested stacks of popovers. So it is possible to open an unrelated `hint` popover while an existing stack of `auto` popovers stays open. The canonical example is that a <select> picker is open (popover=auto) and a hover-triggered tooltip (popover=hint) is shown. That action does not close the <select> picker.

### Documentation

https://open-ui.org/components/popover-hint.research.explainer/

### Browser support

Chrome is planning to ship this feature (subject to approval) in v133: https://chromestatus.com/feature/5073251081912320

captainbrosset:
@mfreed7 do you know if anyone if this is tracked to be added to BCD (browser-compat-data)? I couldn't find an issue, but may have missed it.
Also, do you know when we can expect the spec PR to be merged?

mfreed7:
> @mfreed7 do you know if anyone if this is tracked to be added to BCD (browser-compat-data)? I couldn't find an issue, but may have missed it.

Yep: https://github.com/mdn/mdn/issues/593

> Also, do you know when we can expect the spec PR to be merged?

See my comment here https://github.com/whatwg/html/pull/9778#issuecomment-2512546886 - I'm hoping we can land it now that the objection has been removed. We'll see!

PR #2680, opened 2025-02-20
Add `popover=hint` feature

mfreed7:
Closes https://github.com/web-platform-dx/web-features/issues/2250


features/popover-hint.yml
```diff
@@ -0,0 +1,4 @@
+name: Popover=hint
```
ddbeck, line 1: For consistency with other features like this:

```suggestion
name: 'popover="hint"'
```

features/popover-hint.yml
```diff
@@ -0,0 +1,4 @@
...
+description: The `hint` value for the `popover` HTML attribute creates a popover that is "subordinate" to `popover=auto` popovers. Therefore, `popover=hint`'s will not close other `popover=auto`'s.
```
ddbeck, line 2: Some of this is a style suggestion to be more consistent with the way we've written other descriptions. But I'm also trying to see if we can describe what this does for developers, rather than an abstract relationship of values. What do you think of this?

```suggestion
description: The `popover="hint"` global HTML attribute creates a popover that is subordinate to popovers with a `popover="auto"` attribute. You can use this to create tooltips that don't dismiss `auto` popovers.
```

features/popover-hint.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://html.spec.whatwg.org/multipage/popover.html#attr-popover-hint
+group: html
```
ddbeck, line 4:
```suggestion
group: html
compat_features:
  - api.HTMLElement.popover.hint
  - html.global_attributes.popover.hint
```

features/popover-hint.yml
ddbeck, line 1: After you've reviewed the other comments, please run `npm install; npm run dist` and commit the `popover-hint.yml.dist` file. It'll show you and reviewers what the computed status for this feature would look like. Thank you!

ddbeck (changes requested):
Thanks for opening this, @mfreed7! I've got a few suggestions for you, but I expect we'll able to merge this quickly.

features/popover-hint.yml
mfreed7: Sounds perfect to me!

ddbeck (approved):
Thank you!

PR #2688, opened 2025-02-21
Add a discouraged feature for `RegExp` static properties

ddbeck:
The story of the specification is a little weird here, but I think this tells an OK story to developers.

See:

- https://github.com/tc39/proposal-regexp-legacy-features/
- https://github.com/tc39/ecma262/issues/137


features/regexp-static-properties.yml
```diff
@@ -0,0 +1,13 @@
+name: RegExp static properties
+description: The `RegExp` object has several static properties that are the input and results of the most-recent regular expression match.
```
captainbrosset, line 2:
```suggestion
description: The `RegExp` object has several static properties used to access the input and results of the most-recent regular expression match.
```

I somehow feel like this is more natural to read.

captainbrosset (approved):

features/regexp-static-properties.yml
ddbeck: Yeah, that's fair. We can tighten it up a little though:

```suggestion
description: The `RegExp` object has several static properties to access the input and results of the most-recent regular expression match.
```

PR #2697, opened 2025-02-24
Add type to `alternatives` in JSONSchema

jcscottiii:
Currently, the alternatives field is described to be an array. But the
type of item is missing. This change adds the type as string for each alternative.

### DIFF
--- a/schemas/data.schema.json
+++ b/schemas/data.schema.json
@@ -70,7 +70,9 @@
             },
             "alternatives": {
               "description": "IDs for features that substitute some or all of this feature's utility",
-              "items": {},
+              "items": {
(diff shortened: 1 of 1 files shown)


ddbeck:
Thanks for opening this, @jcscottiii. The JSON Schema is generated from Typescript types, so I'm not sure why this is missing already. I'll take a closer look at this soon.

ddbeck:
It turned out to be a knotty problem. I've filed https://github.com/web-platform-dx/web-features/issues/2722 to sort out our schema generation. In the mean time, I've pushed a change to the types to allow the tests to pass here.

This will get into the next release.

PR #2699, opened 2025-02-25
Adds baseline-browser-mapping to web-features packages

tonypconway:
baseline-browser-mapping is a module that makes it easy to retrieve minimum/all browser versions that support a specific Baseline feature set.  It also includes a JSON file that maps downstream Chromium browsers to the version of Chromium that they implement.  There is an Action that updates this data on a daily basis by parsing UA strings from useragents.io and auto-publishes if there are changes. I will provide and API key and NPM secret for publishing the module privately so they can be added as values in this repo.


tonypconway:
@ddbeck FYI when you come back - I've updated this now to use web-features for selecting minimum versions.

packages/baseline-browser-mapping/README.md
```diff
@@ -0,0 +1,206 @@
...
+    browser: "safari_ios",
+    version: "15.5",
+    release_date: "2022-05-16",
+    engine: null,
+    engine_version: null,
+  },
+];
+```
+
+If you need a list of _all_ compatible versions, you can import and call:
```
rviscomi, line 121: Have you considered using the options object pattern here?

So instead of `getMinimumWidelyAvailable` and `getAllWidelyAvailable` you could do something like:

```js
getWidelyAvailable();
```

and

```js
getWidelyAvailable({listAllCompatibleVersions: true});
```

This pattern also supports the boolean for downstream browsers, for example:

```js

getWidelyAvailable({
  listAllCompatibleVersions: true,
  includeDownstreamBrowsers: true
});
```

And if you really wanted you could implement the other functionality like `OnDate` and `Year` in a similar way, so the interface only exposes a single function.

tonypconway: Yeah, I think this is a a good idea actually, as it makes it clearer what options you're passing to the functions where simply passing `true` isn't that obvious.  I'll take a look at implementing today, and will need to update [bl2bl](https://github.com/web-platform-dx/bl2bl) to take this into account.

tonypconway:
@rviscomi I've made some changes here to support the suggestions you made i.e. using a single function with an options object, and I've updated the README to reflect that.  I've also got a PR open on bl2bl to support this change: https://github.com/web-platform-dx/bl2bl/pull/4

Would be very grateful if you could take a look.

packages/baseline-browser-mapping/src/scripts/baseline-browser-versions.ts
```diff
@@ -0,0 +1,387 @@
...
+  let outputArray = [...inputArray, ...downstreamArray];
+  return outputArray;
+};
+
+type Options = {
+  listAllCompatibleVersions?: boolean;
+  includeDownstreamBrowsers?: boolean;
+  widelyAvailableOnDate?: string | number;
+  targetYear?: number;
+};
```
rviscomi, line 338: It could be useful to add descriptions for each of these options. Here's something I prototyped, as an example:

```suggestion
type Options = {
  listAllCompatibleVersions?: boolean;
  /**
   * Whether to include browsers that use the same engines as a core Baseline browser.
   */
  includeDownstreamBrowsers?: boolean;
  widelyAvailableOnDate?: string | number;
  targetYear?: number;
};
```

IDEs will show this description as part of the auto-suggest feature:

image

rviscomi (commented):
Tested out the beta version locally and it seems to be working great. Just one small DX suggestion.

tonypconway:
As discussed at webdx CG on 13th March 2025, I'm withdrawing this PR in favour of `baseline-browser-mapping` living in its own repo to allow for quicker updates and reduce noise in `web-features`.

PR #2701, opened 2025-02-25
Add manifest.launch_handler into web-platform-dx features

Dp-Goog:
This CL adds a web-platform-dx entry for launch_handler in manifest.

Spec: https://wicg.github.io/web-app-launch/


Dp-Goog:
@captainbrosset 

Hi! Please take a look at this CL, this is my first one and I need approval from a maintainer to start the workflow. I'm a Software Engineer from Chrome looking to add some features here around the PWA area.

I based this off feedback from https://github.com/web-platform-dx/web-features/pull/2395.

Dp-Goog:
@ddbeck and @captainbrosset seems like I do not have the necessary privileges for assigning someone to review my CL or trigger the workflow. Can I get your help again please? Thanks!

features/app-launch-handler.yml
```diff
@@ -0,0 +1,9 @@
+name: Launch Handler
+description: The `launch_handler` manifest member is used to configure how web app launches should behave with respect to already open web app instances.
+spec: https://wicg.github.io/web-app-launch/#dfn-launch_handler
```
ddbeck, line 3: Since this feature seems to have a one-to-mapping with the spec, I think we can just link to the top:

```suggestion
spec: https://wicg.github.io/web-app-launch/
```

features/app-launch-handler.yml
```diff
@@ -0,0 +1,9 @@
+name: Launch Handler
```
ddbeck, line 1: Hmm, we pluralize other "handlers" for web app manifests. @captainbrosset maybe you have an opinion about this?

Also, a nit: we use title case for feature names

```suggestion
name: Launch handlers
```

(If this change is accepted, then the ID for the feature should change to match.)

features/app-launch-handler.yml
```diff
@@ -0,0 +1,9 @@
...
+description: The `launch_handler` manifest member is used to configure how web app launches should behave with respect to already open web app instances.
```
ddbeck, line 2: Adding "web app" for consistency with other features like this.

Reading the spec, it sounds like the concept for "launch handlers" might expand in the future. Perhaps we can be more explicit here about which part of the feature we're talking about, so if this ever changes we'll know to add a second feature:

```suggestion
description: The `launch_handler` and `client_mode` web app manifest members configure how web app launches should behave when an instance of the app is already open. For example, you can choose to focus on an existing app instead of launching a new one.
```

features/app-launch-handler.yml
```diff
@@ -0,0 +1,9 @@
...
+spec: https://wicg.github.io/web-app-launch/#dfn-launch_handler
+group: progressive-web-app
+compat_features:
+  - api.LaunchQueue
+  - api.Window.launchQueue
+  - html.manifest.launch_handler
+  - html.manifest.launch_handler.client_mode
```
ddbeck, line 9: It seems to me that there are additional keys we could track here:

```suggestion
  - api.LaunchParams
  - api.LaunchParams.files
  - api.LaunchParams.targetURL
  - api.LaunchQueue
  - api.LaunchQueue.setConsumer
  - api.Window.launchQueue
  - html.manifest.launch_handler
  - html.manifest.launch_handler.client_mode
```

(Accepting this change means re-running `npm run dist` and committing the result.)

ddbeck (changes requested):
Thanks for opening this PR, @Dp-Goog. Some suggestions for you inline.

features/app-launch-handler.yml
Dp-Goog: Done!

features/app-launch-handler.yml
Dp-Goog: Will wait on @captainbrosset 's feedback here.

features/app-launch-handler.yml
```diff
@@ -0,0 +1,13 @@
+name: Launch handlers
+description: The `launch_handler` and `client_mode` web app manifest members configure how web app launches should behave when an instance of the app is already open. For example, you can choose to focus on an existing app instead of launching a new one.
```
captainbrosset, line 2:
```suggestion
description: The `launch_handler` web app manifest member, with the `client_mode` property, configure how web app launches behave when an instance of the app is already open. For example, you can choose to focus the existing app instance instead of launching a new one.
```

features/app-launch-handler.yml
```diff
@@ -0,0 +1,13 @@
...
+spec: https://wicg.github.io/web-app-launch/
+group: progressive-web-app
+compat_features:
+  - api.LaunchParams
+  - api.LaunchParams.files
```
captainbrosset, line 7: This BCD key is already used by this feature: [app-file-handlers](https://web-platform-dx.github.io/web-features-explorer/features/app-file-handlers/), which is about the specific case of handling files. Let's remove it from this file.

```suggestion
```

captainbrosset (changes requested):

features/app-launch-handler.yml
captainbrosset: We did pluralize `file_handlers` and `protocol_handlers`, but that's because the corresponding manifest members are also pluralized (because they accept arrays of handlers). The `launch_handler` member is singular in the manifest. So I think we should keep it singular.

`Launch handler`

features/app-launch-handler.yml
```diff
@@ -0,0 +1,13 @@
+name: Launch handlers
```
captainbrosset, line 1:
```suggestion
name: Launch handler
```

captainbrosset (approved):
I applied the few comments I had made, and I'm happy for this to merge.
@Dp-Goog could you please re-generate the dist file again, by using `npm run dist features/app-launch-handler.yml`? 
I think this should be the last step.

Dp-Goog:
@captainbrosset Can you please help with kicking off the workflow and taking another look? Thanks and apologies for the delay!

captainbrosset (approved):

PR #2714, opened 2025-02-26
Add `invoker-commands` feature

ddbeck:
This adds a feature for invoker commands, featuring the `command` and `commandfor` attributes for `<button>`.

This builds on @SilentImp's https://github.com/web-platform-dx/web-features/pull/2411. Be sure to keep their `Co-authored-by` line in the final commit message.

This differs from the previous PR by using the newly merge specification and recently-added compat data keys.

Closes https://github.com/web-platform-dx/web-features/pull/2411.


features/invoker-commands.yml
```diff
@@ -0,0 +1,13 @@
+name: Invoker commands
+description: The `command` and `commandfor` attributes for the `<button>` HTML element set a name and target to dispatch to an event when the button is invoked (by click or keystroke), as a declarative alternative to `addEventListener()` calls or `onclick` attribute handlers.
```
captainbrosset, line 2: I think there's a typo here:

```suggestion
description: The `command` and `commandfor` attributes for the `<button>` HTML element set a name and target to dispatch to an element when the button is invoked (by click or keystroke), as a declarative alternative to `addEventListener()` calls or `onclick` attribute handlers.
```

But I wonder if something like this would be simpler:

```suggestion
description: The `command` and `commandfor` attributes for the `<button>` HTML element dispatch an action to an element when the button is invoked (by click or keystroke), as a declarative alternative to `addEventListener()` calls or `onclick` attribute handlers.
```

captainbrosset (changes requested):

features/invoker-commands.yml
ddbeck: I like this second version a lot!

captainbrosset (approved):

PR #2715, opened 2025-02-26
Assign `worker_support` keys

ddbeck:
Most of these don't make much of a difference, but `features/base64encodedecode.yml` shifts the Baseline date by 13 months. We could do a `compute_from` on that one, if desired.


captainbrosset (approved):
Looks good to me. I don't think it's necessary to use compute_from to avoid shifting the baseline date for base64encodedecode. It's not moving that much, and it's far back in the past anyway.

PR #2718, opened 2025-02-26
Update release docs and workflow

ddbeck:


captainbrosset (commented):
Looks good to me. Let's run through these steps together later today. I'm thinking we might want to keep the PR open until then, in case we discover changes to be made while I try to release a new version.

ddbeck:
@captainbrosset thanks again for trialing this. I pushed some more changes based on the run through. Please merge at your leisure.

captainbrosset (approved):

PR #2724, opened 2025-02-27
index: include mention of Limited Availability

paulirish:
The framing here seems challenging as there's some inconsistency.

https://webstatus.dev/  refers to limited/newly/widely  as a "status".
https://web.dev/baseline refers to limited/newly/widely as a "stage".

Regardless of that term, Limited is unique in that it's _defined by the Baseline project_, but the LA features are "not yet in baseline" (per the filter widget in https://webstatus.dev/ ).

---

It feels important a key page describing baseline includes mention of Limited Availability. Hoping that part's not contentious.. 

Anyhow.. should this addition be moved up to be part of the list, like how it's done on https://web.dev/baseline ? Or is this separate treatment preferred?


gh-pages/src/index.md
```diff
@@ -12,6 +12,8 @@ Baseline features are available across popular browsers. Baseline has two stages
 - **Newly available**: The feature works across the latest devices and browser versions. The feature might not work in older devices or browsers. Indicated with a blue icon.
 - **Widely available**: The feature is well established and works across many devices and browser versions. It’s been available across browsers for at least 2½ years (30 months). Indicated with a green icon.
 
+Prior to being newly available, a feature has **Limited Availability** when it's not yet available across all browsers.
```
ddbeck, line 15: nit: We try to use sentence case, so I guess this should be lower case here:

```suggestion
Prior to being newly available, a feature has **limited availability** when it's not yet available across all browsers.
```

This would also be OK, to match the typical widgets:

```suggestion
Prior to being newly available, a feature has **Limited availability** when it's not yet available across all browsers.
```

ddbeck (approved):
Thanks for opening this, @paulirish. This is a good suggestion. I've left a minor suggestion in a line comment. I'm happy for this to merge as soon as its resolved (though I'll leave that for another reviewer, as I'll be OOO shortly).

More broadly, I think "status" should be preferred for consistency with schema and how we've been talking about this stuff of late. Much of this text was written _while_ some of the terminology was being settled, so I'm not surprised by the inconsistency. But there's no excuse now. 😄

cc: @atopal I think this is of interest to you, particularly harmonizing web.dev with the (IMO, more correct) webstatus.dev.

PR #2751, opened 2025-03-10
Make HTML attribute names lowercase.

### LINKED ISSUE #2748: Rename "Inert elements" feature to "inert"
The `inert` feature ID has a name property of ["Inert elements"](https://github.com/web-platform-dx/web-features/blob/main/features/inert.yml). This is a bit misleading because the feature is an HTML global attribute, not an element.

Other HTML global attributes have names like ["inputmode"](https://github.com/web-platform-dx/web-features/blob/main/features/inputmode.yml), ["Spellcheck"](https://github.com/web-platform-dx/web-features/blob/main/features/spellcheck.yml), and ["tabindex"](https://github.com/web-platform-dx/web-features/blob/main/features/tabindex.yml). In these cases, only the attribute name is used, albeit with inconsistent capitalization.

To clarify what the feature is and be more consistent with similar attributes, should it be renamed to "inert"?

captainbrosset:
This fixes #2748 by making most HTML attributes (and global attributes) lowercase and only contain the attribute name. This is similar to CSS property features, where the name of the features is just the property name itself, all lowercase.

There are exceptions to this, notably Popover which is an attribute and an API. So I kept Popover (capitalized).

### DIFF
--- a/features/accesskey.yml
+++ b/features/accesskey.yml
@@ -1,4 +1,4 @@
-name: Accesskey
+name: accesskey
 description: The `accessKey` global HTML attribute gives a hint for generating a keyboard shortcut for the current element. The attribute value must consist of a single printable character.
 spec: https://html.spec.whatwg.org/multipage/interaction.html#the-accesskey-attribute
 status:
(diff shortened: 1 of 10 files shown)


tidoust (approved):
Looks good. And I note that this lowercase-without-attribute pattern was already used for other such features (e.g., ping or inputmode)

PR #2770, opened 2025-03-13
Split `overflow` and `overflow:clip`

### LINKED ISSUE #2769: Should the clip value for overflow be split away in its own feature?
The overflow-shorthand.yml feature is Newly available since 2022-09-12. This seems wrong to me. I've been using `overflow` for ever.

The reason for this is that support for the `clip` value has been added recently in Safari 16.

I suggest splitting this into two features, one for the base overflow feature, which would be Widely available. And a slightly newer one for the `clip` value.

That is, unless `clip` becomes Widely available very soon.

captainbrosset:
Fixes #2769. See that issue for context.


features/overflow-clip.yml
```diff
@@ -0,0 +1,9 @@
+name: "overflow: clip"
+description: "The `overflow: clip` CSS declaration clips an element's overflowing content to the box that's defined by the `overflow-clip-margin` property."
```
jamesnw, line 2:
```suggestion
description: "The `overflow: clip` CSS declaration clips an element's overflowing content to the box that's defined by the `overflow-clip-margin` property. Unlike `overflow: hidden`, it allows you to hide overflow on one axis while keeping overflow on the other axis visible."
```

It also disables programmatic scrolling, but I don't think that's as key as the single axis behavior.

jamesnw (approved):
This makes sense to me!

(Side note, the `css.properties.overflow.multiple_keywords` key is an example of a key that is difficult to detect usage of, for instance in the eslint/css rule.)

PR #2775, opened 2025-03-15
New dialog.requestClose() feature

### LINKED ISSUE #2738: Add `dialog.requestClose()`
### Specification

https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-requestclose

### Description

The requestClose() method of the [HTMLDialogElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement) interface requests to close the [<dialog>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog). An optional string may be passed as an argument, updating the returnValue of the dialog.

This method differs from the HTMLDialogElement.close() method by firing a cancel event before firing the close event. This allows authors to prevent the dialog from closing. This method exposes the same behavior as the dialog's internal close watcher.

### Documentation

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/requestClose

### Browser support

A new method `requeatClose()` was recently added to Dialog in Chrome 134 and Safari Tech Preview.

captainbrosset:
Fixes #2738.


features/requestclose.yml.dist
```diff
@@ -0,0 +1,11 @@
+# Generated from: requestclose.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome: "134"
+    chrome_android: "134"
+    edge: "134"
```
lukewarlow, line 9: Safari 18.4 also supports this (not sure where this data is generated from)

captainbrosset: Support data comes from @mdn/browser-compat-data. If Safari doesn't appear here, that's because 18.4 has not been released yet. The stable version is still 18.3.

features/requestclose.yml
```diff
@@ -0,0 +1,6 @@
+name: dialog.requestClose()
+description: The `requestClose()` method of a `<dialog>` HTML element closes the dialog, firing a `close` event first, which listeners can use to prevent the dialog from closing. This differs from the `close()` method, which closes the dialog immediately.
```
lukewarlow, line 2: This description is wrong.

```suggestion
description: The `requestClose()` method of a `<dialog>` HTML element closes the dialog, firing a `cancel` event first, which listeners can use to prevent the dialog from closing. This differs from the `close()` method, which only fires the non-cancelable `close` event.
```

captainbrosset:
Thanks @lukewarlow. I've addressed your comment about the description.

lukewarlow (approved):

Elchi3 (approved):
LGTM
(See https://github.com/web-platform-dx/web-features/issues/2738 for a discussion why closedBy is not part of this)

PR #2777, opened 2025-03-15
Update webdriver-bidi with the new BCD keys

captainbrosset:
[BCD v5.7.0](https://github.com/mdn/browser-compat-data/releases/tag/v5.7.0) introduces new keys for webdriver bidi.

The webdriver bidi feature on _this_ repo used to be hardcoded to the following browser versions:

- chrome, chrome_android, edge: 106
- firefox, firefox_android: 102

According to comments that we left when the feature was created, these versions numbers were based on:

- https://developer.chrome.com/blog/webdriver-bidi-2023
- https://bugs.chromium.org/p/chromedriver/issues/detail?id=4016
- https://bugzilla.mozilla.org/show_bug.cgi?id=1753997

Now that BCD has webdriver bidi keys, we can remove our hard-coded values and just start using BCD keys instead. However, doing so in one go, by just adding _all_ of the new BCD keys to our features, makes the feature unsupported on all browsers.

So, we now need to look at the various support data for each key and decide which keys should be part of the feature, which keys might need to be ignored (if any), and which keys should be considered for separate features.


features/webdriver-bidi.yml.dist
```diff
@@ -1,4 +1,290 @@
...
+  - webdriver.bidi.network.beforeRequestSent_event
+  - webdriver.bidi.network.continueRequest
+  - webdriver.bidi.network.continueResponse
+  - webdriver.bidi.network.provideResponse
+  - webdriver.bidi.network.responseCompleted_event
+  - webdriver.bidi.network.responseStarted_event
+  - webdriver.bidi.session.end
+  - webdriver.bidi.webExtension
+  - webdriver.bidi.webExtension.install
+  - webdriver.bidi.webExtension.uninstall
```
captainbrosset, line 290: These keys are not supported anywhere. Good candidates to move elsewhere if they belong to a separate feature / later addition.

features/webdriver-bidi.yml.dist
```diff
@@ -1,4 +1,290 @@
...
+  # support:
+  #   firefox: "122"
+  #   firefox_android: "122"
+  - webdriver.bidi.network.authRequired_event
+
+  # baseline: false
+  # support:
+  #   firefox: "134"
+  #   firefox_android: "134"
+  - webdriver.bidi.browser.getClientWindows
```
captainbrosset, line 274: These keys are supported either in chromium or in firefox, but not in both, unlike the keys that come before them.
Are they later additions? Should we ignore them in this initial bidi feature?

captainbrosset:
@juliandescottes it'd be great if someone close to the bidi project could help take a look at this PR. Feel free to at-mention someone else if needed.

One question here is: of the BCD keys in this PR, which represent an initial bidi feature, and more or less match our previously hard-coded data (see PR description).

Also, are there groups of keys we can split out into separate features? Did some of these key come later and represent pieces of capability that developers can use separately from the rest?

captainbrosset:
Also pinging @OrKoN and @Elchi3. Please see my previous comment. Thanks!

features/webdriver-bidi.yml.dist
foolip: The `webExtension` module is a later addition to the spec. I think we should pin support to the earliest parts of BiDi, and treat this as a later addition.

features/webdriver-bidi.yml.dist
foolip: It's relatively new, from https://github.com/w3c/webdriver-bidi/pull/752. I think we should just let them be later additions while pinning the overall feature to the basics.

OrKoN:
The WebDriver BiDi keys look good to me. In general, each feature/method could be used separately by the users depending on the use case. They could be grouped by the module (e.g., `webExtension` is a module) perhaps? I am not sure how this data will be used so I am not sure if it makes sense. cc @sadym-chromium

features/webdriver-bidi.yml.dist
ddbeck: My inclination here would be to pin to the set of keys (or a sample of them) that show Firefox and Chrome support, since that lines up with the status we report today (though the version numbers differ). I don't think the key selection matters _that_ much until some subset is likely to become Baseline.

@captainbrosset could you set a `compute_from`? It'd be nice to merge this.

captainbrosset:
@ddbeck I had a very informative discussion with @juliandescottes where he went over each and every key told me what each one covered and whether it was important for a basic usage of WebDriver BiDi.

Based on this, I left a lot of comments for posterity, and moved a few keys only to `compute_from` to pin the feature to its most basic subfeatures.

In doing so, I was still getting no browser supported in the dist file. So I ended up commenting out a bunch of keys, and later found out that they were causing this status mostly because they are marked as partially supported in BCD for Firefox.

I understand that this might change soon, as some/all of the features for these commented out keys are _actually_ supported in Firefox, modulo some minor edge cases. So the data in BCD might change soon, and we might be able to uncomment some of the `compute_from` keys.

Let me know what you think of this approach. We could also simplify the `compute_from` block and pin it to these simpler keys: 
- `webdriver.bidi.session`
- `webdriver.bidi.script`
- `webdriver.bidi.browsingContext`

(although these act as "container" features in BCD and don't actually match with anything a developer might do)

caugner:
Quick update: @whimboo has started adding parameter-level implementation status to the spreadsheet.

whimboo:
Note that for both Firefox and Chrome the versions for parameters of commands and events have been entered. So we could do another BCD update.

whimboo:
@captainbrosset we landed the initial import for parameters via github.com/mdn/browser-compat-data/pull/26993. Please let us know if that solves all of your issues. Thanks.

ddbeck:
I've upgraded BCD and updated the draft features (see https://github.com/web-platform-dx/web-features/pull/3122). Merging/rebasing with main ought to bring in the (many, many) new BiDi keys that have appeared in BCD.

whimboo:
So does it mean that it looks better now? Are you satisfied or is there some entry that still needs attention? Thanks.

captainbrosset:
I'm working on this today and will push a commit soon.

captainbrosset:
The new BCD keys are helpful. Thank you.

I'm still having a little of trouble with the following BCD keys:

- `webdriver.bidi.session.new`
- `webdriver.bidi.session.end`
- `webdriver.bidi.browsingContext.contextCreated_event`
- `webdriver.bidi.browsingContext.contextDestroyed_event`

In my conversation with Julian, we had categorized those as critical to be able to use BiDi. However, BCD has the following data about these keys:

- `webdriver.bidi.session.new`: partially supported in Firefox because multiple sessions can't be created yet.
- `webdriver.bidi.session.end`: not supported in Chrome/Edge.
- `webdriver.bidi.browsingContext.contextCreated_event`: partially supported in Firefox because subscribing to this event does not emit events for all already open browsing contexts.
- `webdriver.bidi.browsingContext.contextDestroyed_event`: same as above.

So my question is: are these BCD keys essential to using BiDi?

juliandescottes:
> The new BCD keys are helpful. Thank you.
> 
> I'm still having a little of trouble with the following BCD keys:
> 
>     * `webdriver.bidi.session.new`
> 
>     * `webdriver.bidi.session.end`
> 
>     * `webdriver.bidi.browsingContext.contextCreated_event`
> 
>     * `webdriver.bidi.browsingContext.contextDestroyed_event`
> 
> 
> In my conversation with Julian, we had categorized those as critical to be able to use BiDi. However, BCD has the following data about these keys:
> 
>     * `webdriver.bidi.session.new`: partially supported in Firefox because multiple sessions can't be created yet.
> 
>     * `webdriver.bidi.session.end`: not supported in Chrome/Edge.
> 
>     * `webdriver.bidi.browsingContext.contextCreated_event`: partially supported in Firefox because subscribing to this event does not emit events for all already open browsing contexts.
> 
>     * `webdriver.bidi.browsingContext.contextDestroyed_event`: same as above.
> 
> 
> So my question is: are these BCD keys essential to using BiDi?

Thanks for checking. I guess we can consider session.end as optional, you can always end the session by closing the browser (`browser.close` command).

For the other ones - partially supported in Firefox - we probably should mark those as complete and create sub-items for the missing features.

captainbrosset:
Thank you @juliandescottes.

> For the other ones - partially supported in Firefox - we probably should mark those as complete and create sub-items for the missing features.

These are:

* `webdriver.bidi.session.new`
* `webdriver.bidi.browsingContext.contextCreated_event`
* `webdriver.bidi.browsingContext.contextDestroyed_event`

@whimboo are you planning on splitting them into sub-features in BCD, so that these keys can be marked as supported, and the new sub-feature keys can represent the more nuanced partial support?

whimboo:
> @whimboo are you planning on splitting them into sub-features in BCD, so that these keys can be marked as supported, and the new sub-feature keys can represent the more nuanced partial support?

Sorry but this mention got under the wire. I've updated `session.new` and the changes should already be available in BCD. For the browsing context created/destroyed events another PR will be necessary.

Does it look good otherwise?

caugner:
Should I update our WebDriver BiDi implementation status x BCD integration to automatically set the `web-features:webdriver-bidi` tag on all BiDi features?

jgraham:
I think defining webdriver-bidi as a single feature with the overall status based on only a few keys isn't likely to be very helpful. I tend to think we should split this into something closer to per-module features.

captainbrosset:
I should revisit this. Apologies for being very slow on this.
It might help to split this up by modules. They're not all at the same level of importance. There's probably a core feature which captures at least connecting.

Constraint to keep in mind: we do have an existin bidi feature, with hard-coded status https://github.com/web-platform-dx/web-features/blob/main/features/webdriver-bidi.yml
We should probably reuse the same feature for the "core" feature, even if the status ends up being slightly different than the hard-coded one.

captainbrosset:
This is blocked on https://github.com/mdn/browser-compat-data/pull/29965 as we need `webdriver.bidi.session.new` to be marked as fully supported by Firefox to have a decent status.

whimboo:
> This is blocked on [mdn/browser-compat-data#29965](https://github.com/mdn/browser-compat-data/pull/29965) as we need `webdriver.bidi.session.new` to be marked as fully supported by Firefox to have a decent status.

There was a wrong status set for this command for Firefox. I've corrected that. For more details see the comments in the referenced PR.

captainbrosset:
I just pushed a new attempt at this PR.

The goal here is to start using webdriver bidi BCD keys in web-features, to increase our coverage of BCD.

Initially, I had included all BCD keys in one feature, and had overridden the status to more or less match the browser versions we had before.

I've changed my approach. The feature now contains only the BCD keys which are considered critical to using webdriver bidi at all (and a few spare keys which I didn't know where else to put).

This gives us Chrome/Edge 126, and Firefox 143. I think that's fine because the feature is not Baseline anyway, and it's not too far off from the initial Chrome/Edge 106 and Firefox 102.

As for the remaining keys, I will be submitting follow-up PRs for separate features. Here's the list I have in mind:
- preloadscript
- user context
- realm
- user activation
- screenshot
- download
- user prompt
- history
- node locator
- print
- viewport
- history traversal
- emulation
- file input
- log
- network
- permission
- storage
- webextension

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,130 @@
...
+compat_features:
+  core:
+    # WebDriver BiDi has many keys in BCD.
+    # This feature only contains the keys that correspond to the most critical
+    # WebDriver modules, without which you can't really use WebDriver BiDi.
+
+    # Every test needs a session.
+    - webdriver.bidi.session
+    # Uncomment the next line and remove this line when https://github.com/mdn/browser-compat-data/pull/29965 is merged.
+    # - webdriver.bidi.session.new
```
whimboo: We can update that now that the sync PR got merged.

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,130 @@
...
+    - webdriver.bidi.session.status
+    # Uncomment the next line and remove this line when https://github.com/mdn/browser-compat-data/pull/29968 is merged.
```
whimboo: Same here. Good to see those entries updated.

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,128 @@
...
+    - webdriver.bidi.session.status
+    - webdriver.bidi.session.end
+
+    # All BiDi events go through these commands.
+    - webdriver.bidi.session.subscribe
+    - webdriver.bidi.session.subscribe.contexts_parameter
+    - webdriver.bidi.session.subscribe.events_parameter
+    - webdriver.bidi.session.subscribe.userContexts_parameter
+    - webdriver.bidi.session.unsubscribe
+    - webdriver.bidi.session.unsubscribe.events_parameter
```
whimboo, line 23: The new method to [unsubscribe from an event is by using the subscription ID](https://w3c.github.io/webdriver-bidi/#cddl-type-sessionunsubscribebyidrequest) via `subscriptions`. So you should update that section.

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,128 @@
...
+    - webdriver.bidi.browsingContext.reload.context_parameter
+    - webdriver.bidi.browsingContext.reload.wait_parameter
+    - webdriver.bidi.browsingContext.create
+    - webdriver.bidi.browsingContext.create.background_parameter
+    - webdriver.bidi.browsingContext.create.referenceContext_parameter
+    - webdriver.bidi.browsingContext.create.type_parameter
+    - webdriver.bidi.browsingContext.create.userContext_parameter
+    - webdriver.bidi.browsingContext.load_event
+    - webdriver.bidi.browsingContext.load_event.navigation_parameter
+    - webdriver.bidi.browsingContext.load_event.timestamp_parameter
```
whimboo, line 45: What about the `context` and `url` parameters. There is even a `userContext` parameter now, but that's only supported by Firefox for now.

This applies to the other navigation events as well.

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,128 @@
...
+    - webdriver.bidi.script.message_event
+    - webdriver.bidi.script.message_event.channel_parameter
+    - webdriver.bidi.script.message_event.data_parameter
+
+  spare:
+    # These keys are related to the above critical modules, but have
+    # different browser support while also not being separate features
+    # in their own rights.
+    - webdriver.bidi.browsingContext.load_event.context_parameter
+    - webdriver.bidi.browsingContext.load_event.url_parameter
```
whimboo, line 120: Actually why are those not in the above list. Browser support should be equal here similar to `domContentLoaded`.

features/webdriver-bidi.yml.dist
```diff
@@ -1,4 +1,301 @@
...
+
+  # ⬇️ Same status as overall feature ⬇️
+  # baseline: false
+  # support:
+  #   chrome: "126"
+  #   chrome_android: "126"
+  #   edge: "126"
+  #   firefox: "143"
+  #   firefox_android: "143"
+  - webdriver.bidi.browsingContext.contextCreated_event
```
whimboo, line 275: This event was already added in Firefox 99 but we didn't send out the events for already open browsing contexts when subscribing to that event. I think that we should use 99 here?

whimboo (commented):
Please note that I do not know what exactly needs a deep review and what gets maybe automatically generated. It would be good to see someone else reviewing this PR. Maybe @ddbeck can take a look at it again?

From my point just some comments for now.

features/webdriver-bidi.yml.dist
whimboo: Given that this additional feature it was blocked on is not that important, I created a new sub feature for `emit_for_existing_contexts` and specifically tagged that one for 143. That means the `browsingContext.contextCreated` event is fully implemented in Firefox 99. We probably have to wait for the next BCD update.

features/webdriver-bidi.yml
```diff
@@ -1,17 +1,128 @@
...
+    - webdriver.bidi.session.new.capabilities_parameter
```
whimboo, line 121: Actually is there a reason why this is spare? The capabilities parameter is supported and actually important for configuring the WebDriver session.

features/webdriver-bidi.yml
captainbrosset: Ok, thanks. Here, and I believe with your other comments as well, we need to figure out what to leave in this feature, and what to move to other features.
The reason is that features on this repo are not _living_ documentation of a web capability. WebDriver BiDi keeps on evolving, but this feature should more or less reflect what BiDi was when it became available for developers to use.

We started with a hard-coded feature that said it was supported on chrome 106 and firefox 102. 
In this PR, I've tried to add as many of the new BCD keys as possible (to improve our real coverage and remove the hard-coded statuses), while still sticking more or less to these hard-coded versions.
The new computed versions are chrome 126 and firefox 143, which is not too far off.

So, the idea is to try and achieve these (sometimes incompatible) goals:
1. Get as many of the BiDi BCD keys in the feature
2. End up with computed versions that are not too far from reality
3. Define what "reality" means. When were developers really able to use BiDi? Is there a moment in time which we consider was "enough" for them to achieve meaningful tasks. And if yes, which BCD keys correspond to that moment?

That's why certain keys appear in the `spare` section. It doesn't mean they're not important. It's just that they're associated with later browser versions, and therefore push the chrome/edge and firefox versions on the overall BiDi feature even further away.
So, anything that got added later, we should either put in `spare` for now, or immediately move into its own separate feature.

The leaner we can make this, while still constituting a reasonable feature which devs can use, the better.

features/webdriver-bidi.yml.dist
captainbrosset: That sounds great, thank you.

features/webdriver-bidi.yml
captainbrosset: Somehow when I generate the dist file, the `context` and `url` params are shown as not supported in Firefox. Which is weird because BCD/MDN show them as supported.
Investigating...

features/webdriver-bidi.yml
captainbrosset: Good find. Moving it up to the list of core keys.

features/webdriver-bidi.yml
captainbrosset: I think I've done what I described above now.
The `core` keys give us chrome/edge 126 and firefox 117, which isn't too far off from the initially hard-coded versions, and they seem to allow basic BiDi usage.
The `spare` keys belong to the same modules which the `core` keys are in, but have been added later and _generally seem_ non-blocking.
There are many other remaining keys (print, emulation, bluetooth, extensions, etc.), but I'm keeping those for separate features.

features/webdriver-bidi.yml
captainbrosset: This was changed _yesterday_ in https://github.com/mdn/browser-compat-data/commit/fd5836ec81dea5696d4c343a131d6b52e30c1bd2.
So we'll have to wait for the next release of BCD and ingestion in web-features before we can move `context` and `url` from `spare` to `core`.
I've added a comment in the file for us to do this later.

ddbeck (approved):
OK, I'm happy with merging this as we have it. I've sent https://github.com/mdn/browser-compat-data/issues/30540 upstream to capture some details about multiple sessions that seem to have been shed at some point (and probably merits a separate feature here and in BCD).

But overall, I think this is good enough and there's no need to drag this PR out any longer than we have. We can always revise later. Let's merge it.

features/webdriver-bidi.yml
whimboo: Thanks for the explanation! Please note that I do not know anything about web features so it's a bit hard for me to review. But given your explanations it makes all sense.

Thank you for working on it and that we got it landed.

captainbrosset: Thanks for all your help along the way.

PR #2796, opened 2025-03-21
Discouraged MediaController feature

captainbrosset:


features/mediacontroller.yml
```diff
@@ -0,0 +1,29 @@
+name: MediaController
+description: "The playback of multiple `<video>` or `<audio>` elements can be coordinated by a `MediaController` instance, for example to synchronize a sign-language interpreter track with the main video track."
+spec: https://w3c.github.io/media-source/
```
tidoust, line 3: It seems weird to link to Media Source Extensions. The spec has never defined `MediaController`. If you're looking for a link to a spec that still contains `MediaController`, that could perhaps be the [supserseded HTML5 spec](https://www.w3.org/TR/2018/SPSD-html5-20180327/embedded-content-0.html#synchronising-multiple-media-elements)

features/mediacontroller.yml
```diff
@@ -0,0 +1,29 @@
...
+discouraged:
+  according_to:
+    # The PR that removed the feature from the HTML spec was https://github.com/w3c/html/pull/357.
+    # Linking to the issue discussion seems more appropriate.
+    - https://github.com/w3c/html/issues/246
+  alternatives:
+    # According to https://stackoverflow.com/questions/64839836/playing-one-of-multiple-audio-tracks-in-sync-with-a-video
+    - media-source
```
tidoust, line 11: I'm not sure that's such a great alternative in practice. For once, that's only going to work for mixing audio sources but it cannot be used to synchronize multiple video streams.

The main alternative is really "do it yourself" in practice: monitor playback across your sources and adjust the playback rate individually as needed to keep things in sync.

We've done a lot of work on this in the now closed Multi-Device Timing CG, linked to the [Timing Object](https://webtiming.github.io/timingobject/) spec proposal (which did not get much traction, hence the closure). The [timingsrc library](https://webtiming.github.io/timingsrc/) provides a polyfill and works very well across browsers, see for example the [MediaSync example](https://webtiming.github.io/timingsrc/doc/exp_mediasync.html).

Anyway, not sure what to put here as alternative, but media-source does not seem right.

tidoust (changes requested):

features/mediacontroller.yml
captainbrosset: Oh that was a mistake on my part. I didn't intend to link to this spec. Thanks for the link. I will replace.

features/mediacontroller.yml
captainbrosset: Thanks for the background on this. I will remove the alternative and add a comment based on what you said. It's great content to keep for future us.

tidoust (approved):

Rorrote:
Hi

Elchi3 (approved):

PR #2813, opened 2025-03-26
2790: adds cssif

### LINKED ISSUE #2790: Add feature: CssIf
### Specification

https://drafts.csswg.org/css-values-5/#if-notation

### Description

The if() CSS function allows "conditional value selection" inline in values, e.g.:

`height: if(media(width > 1000px): 500px; else: 300px)`

### Documentation

https://lea.verou.me/blog/2024/css-conditionals/ 

https://www.bram.us/2025/02/18/css-at-function-and-css-if/ (combined with https://github.com/web-platform-dx/web-features/issues/2786).

### Browser support

Chrome is planning to ship in M136 (https://chromiumdash.appspot.com/schedule).

schalkneethling:
This adds the initial early version for the upcoming CSS `if()` inline conditional value function.

This is my first contribution, so let's see how I did 😁

fix #2790


features/if.yml
```diff
@@ -0,0 +1,5 @@
+name: if()
+description: The if() function is an inline conditional value function that selects a value based on a set of conditions.
```
jamesnw, line 2:
```suggestion
description: The `if()` CSS function is an inline conditional value that returns a value based on a set of conditions.
```

I'm on the fence about `returns` vs `selects`. In the context of CSS, we've used `selects` as the verb describing the element that a selector is selecting. We use `return` on `min()` and `max()` so I think it makes sense here.

I was also going back and forth on whether or not `is an inline conditional value that` is important here, but I think it's helpful to differentiate from flow control.

jamesnw (approved):
Excellent first feature! A bit of wordsmithing suggested, but great job!

features/if.yml
schalkneethling: Great suggestions. I believe I went with selects because of "conditional value selection" from the spec. Having used it and considering how I would talk about it in conversation, I would also be more likely to use `returns` over `selects`.

captainbrosset (approved):

PR #2816, opened 2025-03-27
Adjust descriptions for boolean and read-only properties

foolip:
- Lowercase boolean, except "Boolean value" in json.yml
- Don't say which properties are read-only
- Don't say that a property returns something, say what the value is


jamesnw (approved):
These changes make sense to me. Should any of these rules get added to the Style Guide?

captainbrosset (approved):

captainbrosset:
> These changes make sense to me. Should any of these rules get added to the Style Guide?

Filed https://github.com/web-platform-dx/web-features/issues/2822 for this.

PR #2836, opened 2025-04-02
Add Compression Dictionary Transport

### LINKED ISSUE #1944: Compression dictionary transport
### Specification

https://datatracker.ietf.org/doc/draft-ietf-httpbis-compression-dictionary/

### Description

I was doing some quick research on compression dictionary and didn't want to lose what I had found on the topic, but also didn't have time to create a new issue right away for it.

- "Intent To": https://groups.google.com/a/chromium.org/g/blink-dev/c/7YdP2YxRQn4/m/9oPxLDdjAAAJ
- Chromestatus: https://chromestatus.com/feature/5124977788977152
- Origin Trial: https://developer.chrome.com/origintrials/#/view_trial/3693514644397228033
- CrBug: https://issues.chromium.org/issues/40255884
- WebKit position: https://github.com/WebKit/standards-positions/issues/160 (support)
- Mozilla position: https://github.com/mozilla/standards-positions/issues/771 (positive)
- Bugzilla tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1882979
- Blog post: https://developer.chrome.com/blog/shared-dictionary-compression

Shipping in Chrome/Edge in 130. Not implemented in other engines, but positive positions.

Compression shared dictionaries can bring huge improvements in the amount of data transmitted over the network.
Worth tracking in web-features as this is an upcoming feature.

### Browser support

_No response_

tunetheweb:
Fixes #1944

Note I've removed `http.headers.Link.compression-dictionary` since that's incorrect. It should be under `rel`: `http.headers.Link.rel.compression-dictionary` but that's not tracked in BCD and instead refers to `html.elements.link.rel.compression-dictionary` which is already a key. I will be removing the incorrect key from BCD.


captainbrosset (approved):
LGTM. Caveat that I don't know anything about this feature. But I liked what I read :)

features/compression-dictionary-transport.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2025-03-31
 name: Compression Dictionary Transport
-description: TODO
+description: Compression Dictionary Transport is a mechanism for using designated HTTP responses as an external dictionary for future HTTP responses for compression schemes that support using external dictionaries (e.g., Brotli and Zstandard)
```
captainbrosset, line 2: @tunetheweb after reading more of the MDN docs, I'd like to propose a description change. We usually try to name the "entry point" of each feature. It seems to me like the most probable one of this feature is the `Use-As-Dictionary`. Is this correct?

```suggestion
description: "The `Use-As-Dictionary` HTTP header allows compression schemes which support dictionaries, such as Brotli or Zstandard, to use HTTP responses as dictionaries for future HTTP responses, making it possible to drastically reduce the compressed size of resources downloaded by the browser."
```

tunetheweb: No I don’t think that’s a good change.

it’s arguable whether the “entry point” is the `compression-dictionary` link to download the dictionary or the `use-as-dictionary` response on the dictionary. And complicated as for the delta-compression option (I.e. reusing an existing resource as the dictionary) the first step is skipped.

It’s more similar to gzip or Brotli (with some extra negotiation to set it up). I note [the Brotli entry](https://github.com/web-platform-dx/web-features/blob/main/features/brotli.yml) is this:

https://github.com/web-platform-dx/web-features/blob/8c8c996c3c8783ef71627fd4c08aa19f164c53d6/features/brotli.yml#L2

So it’s more a general purpose description and it doesn’t start with `accept-encoding` as the entry point. Though I note it does talk about “content encoding”.

So I’d prefer my original description. Or, alternatively use similar “content encoding” language to the Brotli description.

PR #2863, opened 2025-04-10
Add Observable API

foolip:


features/observable.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2025-04-10
 name: Observable
-description: TODO
+description: The `Observable` API provides a composable, ergonomic way of handling an asynchronous stream of events.
```
foolip, line 2: This description is directly from the spec, because I don't understand the API well enough to describe it myself.

@domfarolino how would you describe this to a web developer in more objective terms? In particular I think we shouldn't claim it's ergonomic here, that's up to developers to judge I think.

I'm thinking something like:

> The `Observable` API allows you to subscribe to events in a new way. The `when()` method does something that is different than `addEventListener()`.

Help? :)

captainbrosset: Maybe mention the `when` method earlier in the description.

```suggestion
description: The `when()` method of an `EventTarget` object, such as a DOM element, returns an `Observable` API instance which allows to ...
```

domfarolino: Developers telling us it was ergonomic[^1] is what led me to ship the API and write the spec, so I feel like that description isn't a declaration as much as it is a response. But something like:

```
An Observable is an object representing a stream of events that you can operate on declaratively, even before it emits any events. `EventTarget#when()` returns an Observable that you can immediately subscribe to and direct with native operators, providing an ergonomic, composable alternative to `addEventListener()`.
```

Maybe something like that?

[^1]: In a way that Promise ergonomics compare to callbacks.

foolip: I'll be OOO for a week so if anyone wants to make a decision and land this, that's totally OK and appreciated!

ddbeck: OK, doing my best here to synthesize everyone's comments here. How about this?

```suggestion
description: The `when()` method on a event target returns an `Observable` object, which provides a declarative API for subscribing to and operating on events. It's an alternative to `addEventListener()` callbacks.
```

domfarolino: Looks good to me.

foolip: Let's go with that then!

ddbeck (approved):

Issue #2543, opened 2025-01-16
Too many text-wrap features


foolip:
There are 7 features related to `text-wrap`:
https://webstatus.dev/?q=text-wrap

I think these are inarguably worth describing as standalone features, with distinct use cases:

- `text-wrap: balance`
- `text-wrap: pretty`
- `text-wrap: stable`

I don't think these properties without a value should be features if we keep the more specific features like `text-wrap: balance`:

- `text-wrap`
- `text-wrap-mode`
- `text-wrap-style`

I'm not totally sure about `text-wrap: nowrap`.

At some point in the future these features might all just merge into a single `text-wrap` feature, but not right now I think.

cc @lilles who pointed this situation out to me.

jamesnw:
I'm not opposed to combining these. My memory is we settled on this feature set based on the support story over the developer story, but it's definitely not a cohesive set.

foolip:
Since I filed this issue I learned from @kojiishi that `text-wrap: stable` isn't actually a new capability, but just the default behavior in most browsers given a name. So I don't think that deserves its own feature, only `text-wrap: balance` and `text-wrap: pretty` do, in addition to a `text-wrap` feature for everything else.

My hope is that eventually we would fold all of it into `text-wrap`, when enough time has passed and we have ways to represent "feature evolution" of that sort.

foolip:
I've sent https://github.com/web-platform-dx/web-features/pull/2900 with what I think we should do.

PR #2891, opened 2025-04-23
Allow overlapping BCD keys

ddbeck:
This PR allows BCD keys to exist in multiple features and it demonstrates the capability with a single key, `css.properties.text-wrap-style.pretty`.

`css.properties.text-wrap-style.pretty` already exists in `text-wrap-style`. This PR adds the key to a second feature, `text-wrap-pretty`. This seems pretty logical to me: `text-wrap-style` shows support for the property in general, where `text-wrap-pretty` represents the specific value of `pretty`, which you might use in two properties.

Up to now, we've enforced (but never documented) a limitation that a key can only be in one feature at a time. Last I talked to MDN—the original objector to key duplication—were open to allowing keys to span features. It seems like it'd be nice to introduce this to a single, low-risk pair of features before extending the practice elsewhere.

- Fixes https://github.com/web-platform-dx/web-features/issues/1173
- Fixes https://github.com/web-platform-dx/web-features/issues/2886

### DIFF
--- a/features/text-wrap-pretty.yml
+++ b/features/text-wrap-pretty.yml
@@ -2,3 +2,6 @@ name: "text-wrap: pretty"
 description: "The `text-wrap: pretty` CSS declaration prioritizes better layout over speed when text is broken into multiple lines."
 spec: https://drafts.csswg.org/css-text-4/#valdef-text-wrap-style-pretty
 group: text-wrap
+compat_features:
+  - css.properties.text-wrap.pretty
+  - css.properties.text-wrap-style.pretty
--- a/features/text-wrap-pretty.yml.dist
+++ b/features/text-wrap-pretty.yml.dist
@@ -4,8 +4,21 @@
 status:
   baseline: false
   support:
-    chrome: "117"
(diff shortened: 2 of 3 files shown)


foolip:
It would be good to use another feature as the first one, since I'd like to fix https://github.com/web-platform-dx/web-features/issues/2543 at some point and probably remove some of the features.

I think a very safe and old to pioneer is `HTMLModElement` as part of both `<ins>` and `<del>`.

ddbeck:
Over in https://github.com/web-platform-dx/web-features/issues/1173#issuecomment-2830659951, I recapped the discussion on this concept (overlapping keys) from the WebDX call this week.

Concretely, I need to incorporate a few updates here:

- Allowlisting overlapping keys, so we can't overlap keys unintentionally
- Documenting authoring guidelines, so we know when to overlap keys
- Documenting expectations for consumers, so they know how to interpret this data sensibly

ddbeck:
Let's do this a different a way. I'll revisit this and #1173 through another PR in the future.

Issue #2888, opened 2025-04-22
Bump Node.js to non-EOL release line >18.x


ddbeck:
Our tooling currently uses Node.js 18 (documented via [`.node-version`](https://github.com/web-platform-dx/web-features/blob/924bdccea36008063efdb8000b637fd4a4615e33/.node-version)). It's end-of-life (EOL) on 30 April 2025. Some of our dependencies' dependencies have already bumped their `package.json` `engine` version to `20 || >=22`.

We should do the same, to get rid of the `EBADENGINE` warnings on `npm install` and to be able to use newer Node.js and JavaScript language features.

This is a non-breaking change for `web-features`, but would impact `compute-baseline` consumers (though we have not reached 1.0 there, so not a semver major release issue).

PR #2892, opened 2025-04-23
Bump Node.js to >=20.10.0

ddbeck:
The version must 20.10.0 or later (as opposed to 20.0.0) because we need import attributes, which didn't land until the Node.js 18.19 and 20.10 release lines.

This fixes https://github.com/web-platform-dx/web-features/issues/2888.


janbrasna:
With this bump the two import assertions (that would be blocking updates past v21.ish) could be replaced with `with` safely:
https://github.com/web-platform-dx/web-features/blob/f7d00b73aab300f15a8495193807ea0e7bd7ab32/scripts/specs.ts#L3
https://github.com/web-platform-dx/web-features/blob/f7d00b73aab300f15a8495193807ea0e7bd7ab32/scripts/update-drafts.ts#L8

foolip (approved):

ddbeck:
Thanks for the suggestion, @janbrasna—I've fixed those.

Issue #2818, opened 2025-03-27
Rename Fetch request streams to Fetch upload streaming?


foolip:
@daisuke-e and @yoichio have raised that [Fetch request streams](https://github.com/web-platform-dx/web-features/blob/main/features/fetch-request-streams.yml) might be more natural to call Fetch upload streaming.

The Chromium bug and Chromestatus entry called it upload streaming:
https://chromestatus.com/feature/5274139738767360
https://issues.chromium.org/issues/40505032

On the other hand, https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests called it "Streaming requests with the fetch API" which is closer to the current name.

@captainbrosset @ddbeck discussed naming a bit in https://github.com/web-platform-dx/web-features/pull/2517, but AFAICT not this option.

What name is most familiar to web developers?

ddbeck:
This came up on the WebDX call. My take was that renaming the `name` field is fine and good. I wasn't very familiar with the feature when I originally helped name it. I _do_ have a small preference toward not using a gerund (as in "Fetch upload streams"), but I don't feel very strongly about it. I opened https://github.com/web-platform-dx/web-features/pull/2898 anyway.

If we wish to change the ID to match (`fetch-request-streams` → `fetch-upload-streams`), then we'd have to wait for #91 to resolve this.

jgraham:
This feature has now caused some real-world confusion.

From the web-features explorer I assumed it was supposed to cover passing a `ReadableStream` as a `body` option to the fetch function. That was apparently also what the proposer of https://github.com/web-platform-tests/interop/issues/1072 thought, since that seems to be the feature they're talking about. Notably their request mentions that they feature they want is not implemented in Safari, which is inconsistent with the fact that this feature is marked as supported by Safari.

Examining the BCD keys, this feature is defined as `api.Request.body` i.e. it's about the presence of a `body` property on the `Request` object which is a `ReadableStream`. That is indeed implemented in Safari (and useful if the `Request` represents e.g. a form submission). The BCD key for the feature that I (and the Interop proposer) thought it covered is `api.Request.Request.request_body_readablestream` which is (incorrectly) marked as a part of the widely-available `fetch` feature.

[...]

PR #2898, opened 2025-04-24
Rename "Fetch request streams" to "Fetch upload streams"

ddbeck:
Toward https://github.com/web-platform-dx/web-features/issues/2818

### DIFF
--- a/features/fetch-request-streams.yml
+++ b/features/fetch-request-streams.yml
@@ -1,5 +1,5 @@
-name: Fetch request streams
-description: A `fetch()` request streams data to the server when a request's `body` property is a `ReadableStream` object.
+name: Fetch upload streams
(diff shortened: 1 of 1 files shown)


foolip:
@ddbeck should we wait until we can redirect the ID as well, to bundle the whole change into one release as opposed to multiple? Nothing will really break if we change just the name, but it might raise eyebrows about the discrepant name and ID.

ddbeck:
We can wait, but I'm not sure how long it will take before we can change IDs. Probably a couple of weeks, at least. I think it'd be nice to get some benefit now for developers themselves, then fix the ID as soon as we can.

Otherwise, we can close this and return it when #91 is fixed.

foolip:
Let's take the plunge and bet that we'll really fix #91 soon then :)

PR #2899, opened 2025-04-25
Move service worker event to the messageerror feature

foolip:
The event on ServiceWorkerContainer and ServiceWorkerGlobalScope should
be given the same treatment for consistency. messageerror shipped after
service workers in all features in all browsers, so this seems like the
right move, rather than letting the service worker feature have both.

The moved event isn't Baseline, so again for consistency exclude both
from the compute_from list. This changes some browser versions to be
earlier, but doesn't change the Baseline date, determined by Safari.

### DIFF
--- a/features/messageerror.yml
+++ b/features/messageerror.yml
@@ -8,11 +8,11 @@ status:
   compute_from:
     - api.DedicatedWorkerGlobalScope.messageerror_event
     - api.MessagePort.messageerror_event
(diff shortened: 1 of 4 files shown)


foolip:
This came up in https://chromium-review.googlesource.com/c/chromium/src/+/6486793.

ddbeck (approved):
This seems fair. Thank you!

PR #2900, opened 2025-04-25
Reduce the number of text-wrap features to just 3

foolip:
Only `text-wrap: balance` and `text-wrap: pretty` are significant new
features, let everything elese be covered by the generic `text-wrap`
feature.

This changes the Baseline date of `text-wrap` a little, but it's still
in 2024 and so it doesn't seem worth trying to pin it to the earlier
date by trimming the `compute_from` list in a way that wouldn't match
the feature decription.

Fixes https://github.com/web-platform-dx/web-features/issues/2543.


ddbeck:
I'm happy for this to happen at some point, but wiping out features is a semver major event. I don't really want to do that as a one-off for this. So either this is blocked pending a resolution for at least #91 _and_ a resolution on as-yet un-described "inferior feature" capability (i.e., the structure a merged feature _becomes_).

foolip:
@ddbeck what's the "inferior feature" capability you have in mind? Here, I'm imagining that all removed features are simply redirected to the merged `text-wrap` feature.

ddbeck:
Philip and I chatted about this today and made an OK argument for the idea that these aren't real merges, but actually a kind of error correction (i.e., we created features that didn't/shouldn't really exist and can use redirects as proposed in #91 instead of something richer/more complex, at least initially).

Given that context, I said I'd look at this again. I had a hard time following all the moving parts here, so I drew a diagram. I think this shows how things are moving around here:

```mermaid
flowchart LR
    text-wrap
    text-wrap-balance
    text-wrap-mode
    text-wrap-nowrap
    text-wrap-pretty
    text-wrap-stable
    text-wrap-style

    text-wrap-stable --> text-wrap

    text-wrap-mode --> text-wrap

    text-wrap-nowrap --> text-wrap

    text-wrap-style -- text-wrap-style: auto
    text-wrap-style: nowrap --> text-wrap
    text-wrap-style -- text-wrap-style: balance --> text-wrap-balance
    text-wrap-style -- text-wrap-style: pretty --> text-wrap-pretty
```

I can see Philip's point here. I'd prefer the purist (and purest) version, but simple redirects could work here. I suggest we come back to this as soon as #91 is resolved, which will hopefully be soon.

tidoust:
Dropping "major version required" label as this can now be implemented as a "move" (from today's WebDX CG meeting discussion).

ddbeck:
I've added the redirects to make this work without a breaking release.

ddbeck (approved):

jugglinmike:
@jcscottiii Does the webstatus.dev tooling understand moves like this? Should [the mappings in WPT](https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt+%22name%3A+text-wrap%22&type=code) be updated?

ddbeck:
@jugglinmike I don't know if there any caveats but I do know that webstatus.dev redirects IDs (e.g., https://webstatus.dev/features/text-wrap-mode redirects to https://webstatus.dev/features/text-wrap?redirected_from=text-wrap-mode).

Speaking for myself, it'd be nice if WPT replaced the mappings with the current feature IDs and, better still, supported searching for the current IDs or the former IDs as aliases for the current IDs.

This was not the cleanest rename, however. See https://github.com/GoogleChrome/webstatus.dev/issues/2263 for the growing pains around this.

PR #2902, opened 2025-04-25
Improve description of region capture API

foolip:
These are changes suggested by @eladalon1983 after I asked about the current description, which I thought seemed wrong in parts.

### DIFF
--- a/features/region-capture.yml
+++ b/features/region-capture.yml
@@ -1,5 +1,14 @@
 name: Region capture
-description: The `CropTarget` API crops an element from a video track derived from display-capture of the active browser tab. For example, you can crop out video conferencing during screen sharing.
+description: |
+  Region Capture is an API for cropping video tracks derived from screen-capturing a tab.
+
+  The API consists of two parts.
+
+  First, a `CropTarget` object is minted from an HTMLElement `e`, representing an element within the captured tab.
+
+  Second, `cropTo()` is invoked on the video track, with the aforementioned `CropTarget` as input, after which frames deliver on that track will be cropped to `e`'s bounding box.
+
+  Calling `cropTo(null)` restores the track to its original, uncropped form.
 spec: https://w3c.github.io/mediacapture-region/
 compat_features:
   - api.BrowserCaptureMediaStreamTrack
--- a/features/region-capture.yml
+++ b/features/region-capture.yml
@@ -1,14 +1,5 @@
 name: Region capture
-description: |
-  Region Capture is an API for cropping video tracks derived from screen-capturing a tab.
-
-  The API consists of two parts.
-
-  First, a `CropTarget` object is minted from an HTMLElement `e`, representing an element within the captured tab.
-
-  Second, `cropTo()` is invoked on the video track, with the aforementioned `CropTarget` as input, after which frames deliver on that track will be cropped to `e`'s bounding box.
-
-  Calling `cropTo(null)` restores the track to its original, uncropped form.
+description: "When using `navigator.mediaDevices.getDisplayMedia()` to capture a browser tab as a video stream, the region capture API allows you to crop the video to the bounding box of a given element. In contrast to the element capture API, any content that overlays that element is also captured."
 spec: https://w3c.github.io/mediacapture-region/
 compat_features:
   - api.BrowserCaptureMediaStreamTrack


features/region-capture.yml
```diff
@@ -1,5 +1,14 @@
...
+description: |
+  Region Capture is an API for cropping video tracks derived from screen-capturing a tab.
+
+  The API consists of two parts.
+
+  First, a `CropTarget` object is minted from an HTMLElement `e`, representing an element within the captured tab.
+
+  Second, `cropTo()` is invoked on the video track, with the aforementioned `CropTarget` as input, after which frames deliver on that track will be cropped to `e`'s bounding box.
+
+  Calling `cropTo(null)` restores the track to its original, uncropped form.
```
ddbeck, line 11: I wrote the original description. It's not great, but I don't think it's _wrong_. `CropTarget` _is_ the unavoidable entry point, as far as I can tell.

Here are two new variations for your consideration. The first tries to illustrate the relationship between `CropTarget` and the video track:

```suggestion
description: The `CropTarget.fromElement()` static method selects an element to be cropped from a display-capture video track by the track's `cropTo()` method. For example, use it to crop out video conferencing controls during screen sharing.
```

This one throws up its hands and says, _it's complicated_:

```suggestion
description: The region capture API crops an element, such as an element containing video conferencing controls, from a captured video track.
```

eladalon1983: The proposal is shorter and has some benefits, but I worry that it could be misunderstood, especially around the word "select". Calling `CropTarget.fromElement()` itself has no effect; one could call it multiple times with different inputs and see no change. It's only when the returned token is fed into `cropTo()` that an effect follows. If the suggestion could be modified to better reflect this, I think that would be best.

eladalon1983: > "crops an element ... from"
With this suggestion, I worry that the reader could get the impression that the element is removed from the capture, whereas it's actually everything else that gets removed.

ddbeck: > With this suggestion, I worry that the reader could get the impression that the element is removed from the capture, whereas it's actually everything else that gets removed.

This is a very good worry because that's _exactly_ how I misunderstood the feature!

I guess I'd rewrite it like this:

```suggestion
description: The region capture API crops captured video tracks to show only a selected element.
```

eladalon1983: This is better in some ways, but still a few issues:
- "To only show a selected element" is wrong. That's actually what **Element** Capture does. Region Capture, on the other hand, uses the element's bounding box, but still shows occluding content. Play around with [this demo](https://sub-capture-demo.glitch.me/), focusing on the buttons at the top, to see the difference between "cropping" and "restricting".
- "Captured video tracks" might be a bit too novel (but I hesitate to use stronger language, for fear that the phrase might have unwittingly been coined by YHS). The word "captured" should bind to "tab" here. That is, we crop a video track that is the result of "capturing" a tab. Phrasing that refers to that would be better, imho.
- Should "region capture" be capitalized, since it's a proper name? @foolip has mentioned that for the title you use different capitalization, but possibly not for the description?

Thank you for bearing with me here, and apologies for this taking a bit long; I have thrown a semicolon into this sentence as a token of my friendly intentions.

foolip: Coming back to this after a long time; friendly intentions conveyed through punctuations are the best!

It seem like I made a pretty important error. Would this fix at least be an accurate description?

> The region capture API captures the contents of a tab as a video track, optionally cropping to the bounding box of a given element. Any content that overlays that element is still included.

ddbeck: FWIW, that suggestion works for me.

foolip: Looking at https://w3c.github.io/mediacapture-region/ again I think it's still not quite right. `navigator.mediaDevices.getDisplayMedia()` is the API that can capture a whole tab, and region capture is a way of cropping to the bounding box of a given element. So maybe:

> When using `navigator.mediaDevices.getDisplayMedia()` to capture a browser tab as a video stream, the region capture API allows you to crop the video to the bounding box of a given element. In contrast to the element capture API, any content that overlays that element is also captured.

eladalon1983: That looks good to me.

foolip: Thanks @eladalon1983, finally I got it right :D

ddbeck (approved):

PR #2947, opened 2025-05-08
Add Feature Box Alignment of abspos elements

### LINKED ISSUE #2905: Box alignment features are hidden by current organization
Currently all the box-alignment features are hidden under their respective contexts.
The `css.properties.align-self.position_absolute_context` key is in the absolute positioning feature, and same with flexbox/grid/etc. Only "align-content in block layouts" has it's own data visible. 

The most immediate problem with this is that we can't see the multi-browser support that exists for alignment in absolute positioning. At the very least, `position_absolute_context` alignment keys deserves to be a feature.

At a high level, I think we lose out by not treating "box alignment" as a top level feature of CSS. I know it's historically been taught as 'part of' different layout systems - but authors have consistently found that confusing. The reality is we're close to having it everywhere, with a unified syntax we can teach and use across layouts.

jamesnw:
The following keys are duplicated in `absolute-positioning`. I think it's reasonable to have them in both, but removing them from `absolute-positioning` would allow us to remove the `compute_from`. This would be blocked by #1173.

   - css.properties.place-self.position_absolute_context
  - css.properties.align-self.position_absolute_context
  - css.properties.justify-self.position_absolute_context

Addresses #2905


features/absolute-positioning-box-alignment.yml
```diff
@@ -0,0 +1,8 @@
+name: Box alignment of absolute positioned elements
```
captainbrosset, line 1: I find this name a bit long, and I wonder if developers are going to easily associate it to what the feature is about.
How about "Absolute positioning alignment"?

Unfortunately, I can't find much resources online that talk about this. I was hoping to find a commonly accepted feature name, but no.

The only thing I could find is my own (now outdated) [css alignment cheatsheet](https://patrickbrosset.com/lab/2018-01-10-css-alignment-cheatsheet/).
But, reading it again, I think the scope of this feature here is different, and I'm finding myself a bit confused by all of the ways in which the align-* and justify-* properties can be used.

The feature you are proposing is specifically about absolute positioning. When I tested it, I found that `justify-self` seems to work fine (i.e. setting it to `center` does center the absolutely positioned box within its container), but `align-self` is surprising (e.g. to center the box, I need to use the value `anchor-center` as `center` doesn't do what I expect it to do).

Is there also a feature we should be creating about aligning elements within a block layout (as opposed to aligning them within a grid, or flex layout)?
For example, we already have [align-content in block layouts](https://web-platform-dx.github.io/web-features-explorer/features/align-content-block/). I wonder if we should expand this feature to be more generally about alignment within block layouts. Indeed, while `align-content` does work like padding, when set on the container block layout, there's also `justify-items` (or `justify-self` on children elements) which works kind of like margin and can be used to align elements along the inline axis. See https://patrickbrosset.com/lab/alignment-in-blocks/ and its source code.

tidoust:
Discussed in WebDX CG meeting on 12 March 2026:

- Currently blocked on current requirement that keys appear only in one feature.
- Separate question is whether that still seems a good idea?
- Also another case where compute_form is used. Separate feature might be a good idea.

features/absolute-positioning-box-alignment.yml
```diff
@@ -0,0 +1,8 @@
...
+description: The `align-self` and `justify-self` properties, and their `place-self` shorthand, align an absolute positioned element within its containing block, outside of a flex or grid layout.
+spec: https://drafts.csswg.org/css-align/#justify-abspos
+group: positioning
+compat_features:
+  - css.properties.place-self.position_absolute_context
+  - css.properties.align-self.position_absolute_context
+  - css.properties.justify-self.position_absolute_context
```
captainbrosset, line 8: These 3 keys are already in the `absolute-positioning` feature, making it impossible to merge this PR as is.
That said, they're all hidden by a `compute_from` field. See https://github.com/web-platform-dx/web-features/blob/main/features/absolute-positioning.yml.

So, to merge this PR, we should also remove the 3 keys from `absolute-positioning` and remove the `compute-from` field while we're at it.

@jamesnw are you willing to spend some time on this PR and push it over the finish line?
If not, let me know, I can try to find some time to do it myself.

jamesnw:
I updated the yml files, but needed to open https://github.com/mdn/browser-compat-data/pull/29299 to update Safari support.

captainbrosset:
> I updated the yml files, but needed to open [mdn/browser-compat-data#29299](https://github.com/mdn/browser-compat-data/pull/29299) to update Safari support.

Great, thank you. I think your fix should be in BCD v7.3.8, which we're moving to just now.

features/absolute-positioning-box-alignment.yml
captainbrosset: Since I don't have any better ideas, I suggest sticking with the entry points for the features, just like we did for [align-content-block](https://web-platform-dx.github.io/web-features-explorer/features/align-content-block/) and [justify-self-block](https://web-platform-dx.github.io/web-features-explorer/features/justify-self-block/):


```suggestion
name: justify-self and align-self on absolute positioned elements
```

features/absolute-positioning-box-alignment.yml.dist
```diff
@@ -0,0 +1,15 @@
...
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome: "122"
+    chrome_android: "122"
+    edge: "122"
+    firefox: "134"
+    firefox_android: "134"
```
captainbrosset, line 12: I'm seeing very odd test results on Firefox where none of the properties work, and on Chromium where only justify-self works.
But I may just be holding it wrong. I'm investigating and will report back.
In any case, this should not hold this PR back, because the feature is not baseline anyway. So we have time for BCD support data changes before this reaches baseline.

captainbrosset (approved):

features/absolute-positioning-box-alignment.yml.dist
captainbrosset: Upstream BCD changes in https://github.com/mdn/browser-compat-data/pull/29342.
These don't need to block this PR though.

captainbrosset:
@ddbeck ok merging this PR? If we do, we get a baseline low feature until the BCD PR is merged and released, at which point the feature would regress. 
But that's only virtual anyway because this would all happen at the same time when we bump BCD presumably.

features/absolute-positioning-box-alignment.yml
```diff
@@ -0,0 +1,8 @@
+name: justify-self and align-self on absolute positioned elements
+description: The `align-self` and `justify-self` properties, and their `place-self` shorthand, align an absolute positioned element within its containing block, outside of a flex or grid layout.
```
ddbeck: Optional suggestions here for:

* Specifying CSS
* "Outside" sounds a little literal to me (i.e., this aligns things within a containing block absolutely, if the containing block is itself a flex or grid container, where if I understand correctly this works for any element with `position: absolute`)

```suggestion
description: The `align-self` and `justify-self` CSS properties, and their `place-self` shorthand, align an absolute positioned element within its containing block. In other words, the properties work without a flex or grid layout.
```

features/absolute-positioning-box-alignment.yml.dist
ddbeck: I reviewed the BCD PR and it'll need some work, so I suggest waiting to merge this PR until that's done, so we don't accidentally publish _then_ regress.

ddbeck (commented):
Optional suggestions for the description, inline.

jamesnw:
I've updated this to no longer be stale, but see the BCD PR is still a blocker on this.

PR #2951, opened 2025-05-09
Update spec link and description of :past and :future

foolip:
https://w3c.github.io/webvtt/#the-past-and-future-pseudo-classes is the better spec link for this, as it defines when these pseudo-classes match.

Use the same order as the spec does, because it makes sense.

Also spell out that this is for text track cues.


ddbeck (approved):
Thank you!

PR #2953, opened 2025-05-09
Move ::cue pseudo into main WebVTT feature

foolip:
There is no strong connection to the cue settings (which are mainly
about cue positining), this pseudo is for styling the "inside" of a cue.
It wasn't supported since the very beginning of WebVTT in all browsers,
but goes very far back and can be considered part of WebVTT.


ddbeck (approved):
Thank you!

PR #2969, opened 2025-05-14
Remove some `compat_features` lists in favor of BCD tags

ddbeck:
This PR adopts BCD's `web-features:*` tags as the source of truth for many features. I generated and reviewed the results.

Some tags from BCD don't work at the moment—due to unexpectedly deprecated keys, overlapping keys and suspect feature assignment—so I wasn't able to accept all of them. This first round of fixes is the safest of the bunch. I'll self-review some interesting bits.


features/audio-worklet.yml
```diff
@@ -4,22 +4,9 @@ spec: https://webaudio.github.io/web-audio-api/#AudioWorklet
...
-  - api.BaseAudioContext.audioWorklet
-  - api.Worklet
-  - api.Worklet.addModule
-  - api.WorkletGlobalScope
+status:
+  compute_from:
+    - api.AudioWorklet
+    - api.AudioWorkletNode.AudioWorkletNode
+    - api.AudioWorkletProcessor
+    - api.BaseAudioContext.audioWorklet
```
ddbeck, line 12: One late addition (`api.AudioWorkletGlobalScope.port`), so I pinned to some keys that showed up in example code in MDN docs for audio worklets.

features/speculation-rules.yml.dist
```diff
@@ -4,8 +4,8 @@
 status:
   baseline: false
   support:
-    chrome: "127"
-    chrome_android: "127"
+    chrome: "136"
+    chrome_android: "136"
```
ddbeck, line 8: I let this one advance, since there's one implementer. Open to reconsideration on this.

Elchi3 (approved):
Great work, thank you Daniel! 💯

Issue #2973, opened 2025-05-16
Can identifiers start with a number or hyphen?


foolip:
The answer is "yes" in our docs and source:
https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#identifiers
https://github.com/web-platform-dx/web-features/blob/38a03910b37863a4d48bc280eac2e6717f197fe3/index.ts#L28C7-L28C24

Allowing a leading number means that turning identifiers like "2d-canvas" into an identifier (variable name, enum name, etc) in most programming languages will require some prefix like `k2DCanvas`, since identifiers can't start with a number.

Leading or trailing hyphens, or more than one hyphen, also seem like something we should never allow.

PR #2974, opened 2025-05-16
Provide additional guarantees about web-features identifiers

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/2973.

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -62,7 +62,7 @@ Feature authors should (in descending order of priority):
 
 A feature's identifier is the feature's filename before the `.yml` extension.
 
-Feature identifiers must contain only lowercase alphanumeric characters (a-z and 0-9) plus the `-` character (hyphen or minus sign) as a word separator.
+Feature identifiers must start with a lowercase letter (a-z) and contain only lowercase alphanumeric characters (a-z and 0-9) plus the `-` character (hyphen or minus sign) as a word separator.
 
 The identifier should match the name, with these additional guidelines:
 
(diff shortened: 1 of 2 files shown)


docs/guidelines.md
```diff
@@ -62,7 +62,7 @@ Feature authors should (in descending order of priority):
 
 A feature's identifier is the feature's filename before the `.yml` extension.
 
-Feature identifiers must contain only lowercase alphanumeric characters (a-z and 0-9) plus the `-` character (hyphen or minus sign) as a word separator.
+Feature identifiers must start with a lowercase letter (a-z) and contain only lowercase alphanumeric characters (a-z and 0-9) plus the `-` character (hyphen or minus sign) as a word separator.
```
foolip, line 65:
```suggestion
Feature identifiers must start with a lowercase alphabetic character (a-z) and contain only lowercase alphanumeric characters (a-z and 0-9) plus the `-` character (hyphen or minus sign) as a word separator.
```

index.ts
```diff
@@ -39,7 +39,7 @@ function* yamlEntries(root: string): Generator<[string, any]> {
         const { name: key } = path.parse(fp);
 
         if (!identifierPattern.test(key)) {
-            throw new Error(`${key} is not a valid identifier (must be lowercase a-z, 0-9, and hyphens)`);
+            throw new Error(`${key} is not a valid identifier (must start with a letter and consist of alphanumerical words separated by hyphens)`);
```
foolip, line 42:
```suggestion
            throw new Error(`${key} is not a valid identifier (see guidelines)`);
```

ddbeck (approved):
Thank you!

PR #2986, opened 2025-05-21
Add auto-sized lazy images feature

captainbrosset:
Adding `sizes=auto` to an image that also has `loading=lazy` is a really cool feature which makes it possible for the browser to reserve the right size for an image, even though it hasn't yet loaded it. This way, as you scroll a really long page that has many lazy loaded images, you don't get layout jumps.
This only works with lazy loaded images and, for now, is only on Chromium-based browsers.

See also:
* https://github.com/web-platform-tests/interop/issues/761
* https://html.spec.whatwg.org/multipage/images.html#valdef-sizes-auto
* https://github.com/whatwg/html/pull/8008
* https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#auto_3


ddbeck (approved):
This looks promising to me. I've suggested some possible, but optional, description edits. Please merge, if you think it's ready. Thank you!

features/lazy-loading-image-auto-sizes.yml
ddbeck, line 1: Oh, I forgot to mention. To match up with `loading-lazy`, I suggest changing the ID to `sizes-auto`.

Issue #2722, opened 2025-02-27
Rethink JSON Schema generation


ddbeck:
`ts-json-schema-generator` only incompletely generates a schema based on our TypeScript types. It seems to have [some longstanding bugs](https://github.com/vega/ts-json-schema-generator/issues?q=is%3Aissue%20state%3Aopen%20keyof) around `keyof`. I doubt they'll be fixed any time soon.

We should explore alternatives for keeping our TypeScript types and JSON Schema in sync.

Some routes we might consider:

- Use a different method to generate JSON Schemas (e.g., describe the types with `zod` and use [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema))
- Author JSON Schema directly and generate TypesScript types (e.g., [`json-schema-to-typescript`](https://www.npmjs.com/package/json-schema-to-typescript))

jcscottiii:
> Author JSON Schema directly and generate TypesScript types

If you're open to it, I would vote for this approach. I will also suggest another tool named [quicktype](https://github.com/glideapps/quicktype). It supports multiple languages (including typescript). If you want to try it out, you could paste the JSON schema into their online [tool](https://app.quicktype.io/) and change the source type to JSON Schema and change the language to typescript. (They also have Typescript Zod too)

---

webstatus.dev uses it on the data.schema.json here to generate the Go types. A few things to note:
- It gets a little confused on the fields that are either or 1) a single string or 2) array of strings when using their Go generator. (Let me know if you want more details. Or you can try the Go generator on their online tool)
  - Unrelated question: Would you be open to switching to just an array of strings in the next major version of the schema? If not, that's okay.

As a result, webstatus.dev keeps a slightly modified version of the schema in the tree [[1](https://github.com/GoogleChrome/webstatus.dev/blob/72d3f063666a3162f93207930fb3c6e7770ebe85/jsonschema/web-platform-dx_web-features/defs.schema.json#L5-L19)].

After making that modification, it works great!

ddbeck:
Thank you for the quicktype recommendation, @jcscottiii! I much prefer the idea of using a tool that someone has first hand experience with.

ddbeck:
And https://github.com/web-platform-dx/web-features/issues/1981#issuecomment-2425060065 covers making the types consistent for arrays of strings. It is planned for the [next major release](https://github.com/web-platform-dx/web-features/milestone/2).

PR #2990, opened 2025-05-21
Demonstrate JSON Schema as source of truth for web-features data

ddbeck:
This PR demonstrates switching from generating JSON Schema to authoring a JSON Schema. This is a prerequisite for #91 and fixes https://github.com/web-platform-dx/web-features/issues/2722. I'll self-review to point out some interesting things.

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 // Quicktype produces definitions that are correct, but not as narrow or
 // well-named as hand-written type definition might produce. This module takes
 // the Quicktype-generated types as renames or modifies the types to be somewhat
@@ -10,11 +11,11 @@ import type {
   Discouraged,
   GroupData,
   FeatureData as QuicktypeMonolithicFeatureData,
+  Status as QuicktypeStatus,
   StatusHeadline as QuicktypeStatusHeadline,
   WebFeaturesData as QuicktypeWebFeaturesData,
   Release,
   SnapshotData,
-  Status,
   Support,
 } from "./types.quicktype";
 
@@ -27,21 +28,24 @@ export type {
   GroupData,
   Release,
   SnapshotData,
-  Status,
   Support,
 };
+
+export interface Status extends QuicktypeStatus {
+  baseline: false | BaselineHighLow;
+}
 
 export interface SupportStatus extends QuicktypeStatusHeadline {
   baseline: false | BaselineHighLow;
 }
 
 // These are "tests" for our type definitions.
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const badQuicktypeStatusHeadline: QuicktypeStatusHeadline = {
   baseline: true, // This is an improper value in our actual published data
   support: {},
 };
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const badQuicktypeStatus: QuicktypeStatus = badQuicktypeStatusHeadline;
+
 const badSupportStatus: SupportStatus = {
   // This validates that we're actually overriding Quicktype (and correctly). If
   // `baseline: true` ever becomes possible in the `SupportStatus`, then
@@ -50,7 +54,11 @@ const badSupportStatus: SupportStatus = {
   baseline: true,
   support: {},
 };
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const badStatus: Status = {
+  // @ts-expect-error
+  baseline: true,
+  support: {},
+};
 const goodSupportStatus: QuicktypeStatusHeadline | SupportStatus = {
   baseline: false,
   support: {},
@@ -74,7 +82,6 @@ export type FeatureData = Required<
     >
   >;
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const goodFeatureData: FeatureData = {
   name: "Test",
   description: "Hi",


new.quicktype.ts
```diff
@@ -0,0 +1,208 @@
+/**
```
ddbeck, line 1: This file is generated by [quicktype](https://quicktype.io/) from the new, hand-authored JSON Schema. The generation is OK _for now_ but if we add something like redirects as described in #91, it has a few inconveniences. We're going to insulate ourselves from quicktype with the next TS module.

newtypes.ts
```diff
@@ -0,0 +1,63 @@
...
+    QuicktypeMonolithicFeatureData,
+    "description_html" | "description" | "name" | "spec" | "status"
+  >
+> &
+  Partial<
+    Pick<
+      QuicktypeMonolithicFeatureData,
+      "caniuse" | "compat_features" | "discouraged"
+    >
+  >;
```
ddbeck, line 49: This is a preview of what's coming when we add redirects. TypeScript and JSON Schema's type systems aren't strictly equivalent, so quicktype will generate correct but overbroad types. In this module, I'll pluck out some nuances (and badly generated names) that quicktype doesn't handle very well and clean them up by _extending_ the generated types. While it's a bit weird looking, the benefit is that we can't completely diverge from the underlying JSON Schema.

newtypes.ts
```diff
@@ -0,0 +1,63 @@
...
+const t1: FeatureData = {
+  name: "Test",
+  description: "Hi",
+  description_html: "Hi",
+  spec: "",
+  status: {
+    baseline: false,
+    support: {},
+  },
+};
```
ddbeck, line 61: If we like, we can even "test" some properties of our types—like I do here demonstrating a `FeatureData` object without any of the optional fields—by instantiating some objects and letting the typechecker complain if we get it wrong.

schemas/new.schema.json
```diff
@@ -0,0 +1,302 @@
+{
```
ddbeck, line 1: And here is the JSON Schema, intended for a human (me) to work with, instead of the generated JSON Schema. Mostly, things are ordered in a more readable way.

newtypes.ts
foolip: Do the TypeScript error messages if you get types wrong end up looking completely bizarre with this layering?

newtypes.ts
foolip: That seems like a useful smoke test. I guess that when we load the JSON and interpret it as a TypeScript type, there's no checking going on that we could rely on? That would be helpful, but I suspect it's not a thing.

schemas/new.schema.json
foolip: I like it! By now I can almost read JSON Schema, and I find this quite readable.

foolip:
I think we should do it! I'd rather have ugly TypeScript definitions than ugly JSON Schema.

newtypes.ts
ddbeck: We could do this, but I'd prefer to avoid it on the initial implementation. Quicktype can generate [zod](https://zod.dev/) runtime type check code, but it'd be a new dependency that I don't feel especially ready to embrace just yet.

schemas/data.schema.json
```diff
@@ -273,117 +157,146 @@
...
-            "$ref": "#/definitions/GroupData"
-          },
-          "description": "Group identifiers and data",
-          "type": "object"
+        "spec": {
+          "description": "Specification",
+          "$ref": "#/definitions/URL"
+        }
+      }
+    },
```
jcscottiii, line 259: FYI: name and spec were previously marked as required. Now they are optional. Not sure if this was intentional.

schemas/data.schema.json
```diff
@@ -57,212 +114,39 @@
...
-              "type": "object"
-            }
-          },
-          "required": [
-            "baseline",
-            "support"
-          ],
-          "type": "object"
+          "description": "Whether a feature is considered a \"Baseline\" web platform feature and when it achieved that status",
+          "$ref": "#/definitions/StatusHeadline"
```
jcscottiii, line 139: Similar comment here. baseline and support  were previously required. But I think they should stay required. (Probably in StatusHeadline)

schemas/data.schema.json
```diff
@@ -57,212 +114,39 @@
...
       },
-      "required": [
-        "name",
-        "description",
-        "description_html",
-        "spec",
-        "status"
-      ],
-      "type": "object"
+      "required": ["name", "description", "description_html", "spec"]
```
jcscottiii: Is status no longer required?
```suggestion
      "required": ["name", "description", "description_html", "spec", "status"]
```

schemas/data.schema.json
ddbeck: Fixed in b2b6762.

schemas/data.schema.json
ddbeck: Fixed in b2b6762.

schemas/data.schema.json
ddbeck: Fixed in b2b6762.

ddbeck:
Thank you for the review, @jcscottiii. Thanks to your review I was able to find some other constraints that weren't in the new schema file. This also prompted me to discover some oddities in the rolled-up TypeScript types in npm package — I'm still working on fixing that part.

jcscottiii:
I tried out the latest [updates](https://github.com/GoogleChrome/webstatus.dev/pull/1552) and it looks really good now! Thanks

packages/web-features/package.json
```diff
@@ -18,16 +18,17 @@
...
     "data.schema.json"
   ],
   "scripts": {
-    "prepare": "tsc && rm types.js && tsup ./index.ts --dts-only --format=esm --out-dir=."
+    "prepare": "tsc && rm types.js && rm types.quicktype.js"
   },
   "devDependencies": {
     "@types/node": "^24.0.7",
-    "tsup": "^8.5.0",
     "typescript": "^5.8.3"
```
ddbeck: One thing I discovered in following up @foolip's https://github.com/web-platform-dx/web-features/pull/2990#discussion_r2102808837 question about type errors for consumers:

`tsup` adds another layer of name-mangling, on top of the layer given by converting JSON Schema to TypeScript (via `quicktype`). For example, as originally configured, `FeatureData` is shown as extending `FeatureData$1`. Ugly!

To fix this, I just let plain TypeScript generate the type declaration files. It's slightly less tidy — we have to include the `types.d.ts` and `types.quicktype.d.ts` files — but it's a very tiny hit for consumers (two additional small files and two development-time file reads for TypeScript consumers) and a nice boon for us (one fewer dependency and one less layer of indirection).

This also makes it a lot easier for us to do some other things in the future (e.g., adding util functions or making this a "normal" monorepo where we're doing less magic with the `web-features` package).

foolip:
Commit 6e383fff9f5555317054ec3555063a449ffda117 LGTM, less code and still works!

types.quicktype.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/**
+ * A feature data entry
+ */
+export interface FeatureData {
+  /**
+   * caniuse.com identifier(s)
+   */
+  caniuse?: string[] | string;
```
foolip, line 68: I don't want to delay v3, but how quickly could we decide that the built output will always be an array instead of string-or-array? Here and for a bunch of things below.

types.quicktype.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Whether a feature is considered a "Baseline" web platform feature and when it achieved
+ * that status
+ */
+export interface StatusHeadline {
+  /**
+   * Whether the feature is Baseline (low substatus), Baseline (high substatus), or not (false)
+   */
+  baseline: boolean | BaselineEnum;
```
foolip, line 131: Is it possible to put `false` here to rule out `true` as a possible value?

```suggestion
  baseline: false | BaselineEnum;
```

Same change below if so.

types.quicktype.ts
```diff
@@ -0,0 +1,206 @@
...
+  baseline_high_date?: string;
+  /**
+   * Date the feature achieved Baseline low status
+   */
+  baseline_low_date?: string;
+  /**
+   * Browser versions that most-recently introduced the feature
+   */
+  support: Support;
+  [property: string]: any;
```
foolip, line 170: What does this line do?

types.ts
```diff
@@ -1,97 +1,65 @@
...
-    browsers: { [key in BrowserIdentifier]: BrowserData };
-    /** Feature identifiers and data */
-    features: { [key: string]: FeatureData };
-    /** Group identifiers and data */
-    groups: { [key: string]: GroupData };
-    /** Snapshot identifiers and data */
-    snapshots: { [key: string]: SnapshotData };
-}
-
+// Quicktype produces a definitions that are correct, but not as narrow or
```
foolip:
```suggestion
// Quicktype produces definitions that are correct, but not as narrow or
```

types.ts
```diff
@@ -1,97 +1,65 @@
...
-    /** caniuse.com identifier(s) */
-    caniuse?: string | [string, string, ...string[]];
-    /** Whether a feature is considered a "baseline" web platform feature and when it achieved that status */
-    status: SupportStatus;
-    /** Sources of support data for this feature */
-    compat_features?: string[];
-    /** Whether developers are formally discouraged from using this feature */
-    discouraged?: Discouraged;
+export interface WebFeaturesData
+  extends Pick<QuicktypeWebFeaturesData, "browsers" | "groups" | "snapshots"> {
```
foolip, line 68: Just about here is where I no longer understand what I'm reading and will just trust that this is all great. It sure _looks_ like TypeScript. If you're really unsure about anything I can try harder to understand.

types.quicktype.ts
ddbeck: In this PR, I tried my best to avoid changing anything material about the existing JSON schema, just flip around what we're generating (that is JSON schema → TypeScript, instead of the reverse—it's not _exactly_ equivalent, which is why I haven't pushed to merge this sooner than v3). I expect we'll merge https://github.com/web-platform-dx/web-features/pull/3184/ to fixup all the string-or-array types as part of v3. To the best of my knowledge, there's no opposition to landing that PR.

types.quicktype.ts
ddbeck: This file is generated, so no. But I can fix this with the `types.ts` file that fixes up some of this. I've pushed 707c6f571a0e1e5875464ade64119a548eabe40f and 64de65df13693c4c50ab4068c765db44b8677b25 to do this.

types.quicktype.ts
ddbeck: This whole file is generated by Quicktype. This line is a bug (or limitation, not sure) of Quicktype: it doesn't seem to be able to determine whether the `Status` object might have additional properties (or what their types might be). The `types.ts` file corrects this with the `Pick<…>` types, so that these arbitrary keys are stripped from the exported types.

types.ts
ddbeck: In the interest of completeness and posterity, I'll explain this a bit better:

Here, I'm making the types nicer than what Quicktype generates by "picking" good stuff out of the schema, but partly modifying the generated type for the `features["…"]` objects (`FeatureData`, below). This isn't strictly required now, but will be when we add redirects (in a follow up PR).

Anyway, by doing it this way instead of writing wholly custom types, we preserve the link between JSON Schema and TypeScript.

types.quicktype.ts
foolip: Alright, excellent!

PR #2993, opened 2025-05-22
File and directory entries API

captainbrosset:


captainbrosset:
> @captainbrosset do you have any commentary on the existing #2633 PR? In that one (and some linked BCD discussions) it seems like there are some real issues with the way the API is exposed in Chromium.

@ddbeck sorry, I had not seen that other PR before opening mine. Sorry about the duplicated work.

Looks like the interface names are indeed different in Chromium: dropping files on an `<input type=file>` leads to `FileEntry` objects in Chromium, whereas it creates `FileSystemFileEntry` objects in Firefox (I have not tested in Safari).

The spec disagrees with Chromium: https://wicg.github.io/entries-api/#api-fileentry but the Chromium names seem very old: https://groups.google.com/a/chromium.org/g/blink-dev/c/_6Euwqv366U and https://github.com/WICG/entries-api/issues/6.

The impression I'm getting is that this API is not in a great state of standardization and interoperability, and there has been little energy to fix this over the years.

On our end, I think we ought to have a feature in the repo for it anyway. And use whatever BCD keys are at our disposal at the moment. The reason is, whatever the interface name differences, one can still write code that works on both Chromium and Firefox:

```html
<!DOCTYPE html>

<label for="file-input">Select a file to upload</label>
<input type="file" id="file-input" multiple>

<script>
  const inputEl = document.getElementById('file-input');
  inputEl.addEventListener('change', function(event) {
    const entries = inputEl.webkitEntries;
    console.log(entries);
  });
</script>
```

ddbeck:
There are now three open PRs for file and directory entries:

- https://github.com/web-platform-dx/web-features/pull/2633 
- https://github.com/web-platform-dx/web-features/pull/2993 (this one)
- https://github.com/web-platform-dx/web-features/pull/3438

PR #2484, opened 2024-12-20
Fix typo in file name

szepeviktor:
Found this newest typo.

@ddbeck Please consider including `typos` in CI.

PR #2776, opened 2025-03-15
New single color stop gradients feature

captainbrosset:
See https://github.com/mdn/browser-compat-data/pull/26062. Gradient functions can now take a single color stop. Available in Firefox 136, and coming in other browsers.

Issue #1981, opened 2024-10-16
Spec is sometimes string, sometimes array


SachaG:
It seems like the `spec` field can either be a string or an array? Maybe that's on purpose, but I just wanted to point out that this can be a little inconvenient for typed APIs such as GraphQL. Could it be standardized to always be an array?

captainbrosset:
I agree that this is inconvenient. In the explorer, I've had to add some code to always make it an array: https://github.com/web-platform-dx/web-features-explorer/blob/main/.eleventy.js#L59-L64

SachaG:
Yeah that's also what I did for now.

ddbeck:
Two consumer complaints. To me, that sounds like justification enough to make it always an array.

For background, I think we did it mostly as a convenience to authors. Now that we have a build step anyway, there's no reason we can't make the type consistently `string[]` at build time.

I think, with the next breaking version, we ought to do this for all the `string | [string, string, ...string[]]` fields:

- `spec`
- `group`
- `snapshot`
- `caniuse`

SachaG:
I think `group` has the same issue?

jgraham:
FWIW I also ran into this and had to write a similar string-or-array-to-array helper, so I'm also in favour of this change. _Arguably_ it's non-breaking (I suppose someone might be relying on arrays always being length > 1 but it seems at least a little bit unlikley).

PR #3000, opened 2025-05-27
Add move and split redirects to the schema

ddbeck:
Towards #91 and the sequel to #2990.

~~Because this PR follows #2990, looking at the diff between them might be easier to look at. See: https://github.com/ddbeck/web-features/compare/91-jsonschema-source-of-truth...ddbeck:web-features:91-redirects~~

This PR does the following:

- Adds redirects to the JSON Schema
- Generates (and embellishes) types for the new schema
- Modifies our scripts to use the new types and support redirects
- Demonstrates both types of redirects, by splitting `single-color-gradients` and moving `numeric-seperators` to `numeric-separators` (https://github.com/web-platform-dx/web-features/pull/2484)


features/webcodecs.yml
```diff
@@ -1,10 +1,5 @@
-name: WebCodecs
-description: The WebCodecs API provides low-level access to individual video frames and chunks of audio samples, for full control over the way media is processed.
-spec: https://w3c.github.io/webcodecs/
-caniuse: webcodecs
```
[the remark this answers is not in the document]
foolip: Let's use https://github.com/web-platform-dx/web-features/pull/2776 as the feature to demonstrating split. We'd split that into gradients (2 BCD keys) and conic-gradients (1 BCD key).

ddbeck:
@js

ddbeck:
@jcscottiii Would you mind taking this PR for a test drive, pointing your schema tooling at the JSON Schema file like you did in https://github.com/GoogleChrome/webstatus.dev/pull/1552 and https://github.com/web-platform-dx/web-features/pull/2990#issuecomment-2978084026? I'm hoping to get a move on with this schema change (and other requests, such as https://github.com/web-platform-dx/web-features/issues/1981).

This PR is stacked on top of #2990 and adds feature redirects (renames and splits). I'm wondering what the experience of consuming that version of the schema would be like.

<details>

<summary>spoilers</summary>

I found that, at least for TypeScript, Quicktype's generated types weren't as nice, so I wrote some convenience wrapper types in `types.ts`. I wonder if you'll be inclined to do the same thing.

</details>

jcscottiii:
Trying it out now! I will make a WIP PR and report back

schemas/data.schema.json
```diff
@@ -1,53 +1,120 @@
...
+          {
+            "$ref": "#/definitions/FeatureData"
+          },
+          {
+            "$ref": "#/definitions/FeatureMovedData"
+          },
+          {
+            "$ref": "#/definitions/FeatureSplitData"
+          }
+        ]
```
jcscottiii: I ran into the spoiler you mentioned lol.

Raw notes:
By doing this: My feature structure looks like this

```
// A feature has split into two or more other features
type FeatureValue struct {
	// caniuse.com identifier(s)                                                                              
	Caniuse                                                                                  *StringOrStrings `json:"caniuse"`
	// Sources of support data for this feature                                                               
	CompatFeatures                                                                           []string         `json:"compat_features,omitempty"`
	// Short description of the feature, as a plain text string                                               
	Description                                                                              *string          `json:"description,omitempty"`
	// Short description of the feature, as an HTML string                                                    
	DescriptionHTML                                                                          *string          `json:"description_html,omitempty"`
	// Whether developers are formally discouraged from using this feature                                    
	Discouraged                                                                              *Discouraged     `json:"discouraged,omitempty"`
	// Group identifier(s)                                                                                    
	Group                                                                                    *StringOrStrings `json:"group"`
	// Short name                                                                                             
	Name                                                                                     *string          `json:"name,omitempty"`
	// Snapshot identifier(s)                                                                                 
	Snapshot                                                                                 *StringOrStrings `json:"snapshot"`
	// Specification URL(s)                                                                                   
	Spec                                                                                     *StringOrStrings `json:"spec"`
	// Whether a feature is considered a "Baseline" web platform feature and when it achieved                 
	// that status                                                                                            
	Status                                                                                   *StatusHeadline  `json:"status,omitempty"`
	Redirect                                                                                 *Redirect        `json:"redirect,omitempty"`
}
```

Required fields like description can now be nil. And I see that you have the same thing in your types.quicktype.ts

---

Question: Would you be open to using the `discriminator` feature alongside the `oneOf`?

```
    "features": {
      "description": "Feature identifiers and data",
      "type": "object",
        "oneOf": [
          {
            "$ref": "#/definitions/FeatureData"
          },
          {
            "$ref": "#/definitions/FeatureMovedData"
          },
          {
            "$ref": "#/definitions/FeatureSplitData"
          }
        ],
        "discriminator": {
          "propertyName": "kind"
        }
      },
...
    "FeatureData": {
      "description": "A feature data entry",
      "type": "object",
      "properties": {
        "kind": { "const": "data" },
        "name": {
          "description": "Short name",
          "type": "string"
        },
        "description": {
          "description": "Short description of the feature, as a plain text string",
          "type": "string"
        },
        "description_html": {
          "description": "Short description of the feature, as an HTML string",
          "type": "string"
        },
        "spec": {
          "description": "Specification URL(s)",
          "$ref": "#/definitions/URLOrURLs"
        },
        "group": {
          "description": "Group identifier(s)",
          "$ref": "#/definitions/StringOrStrings"
        },
        "snapshot": {
          "description": "Snapshot identifier(s)",
          "$ref": "#/definitions/StringOrStrings"
        },
        "caniuse": {
          "description": "caniuse.com identifier(s)",
          "$ref": "#/definitions/StringOrStrings"
        },
        "compat_features": {
          "description": "Sources of support data for this feature",
          "$ref": "#/definitions/Strings"
        },
        "status": {
          "description": "Whether a feature is considered a \"Baseline\" web platform feature and when it achieved that status",
          "$ref": "#/definitions/StatusHeadline"
        },
        "discouraged": {
          "description": "Whether developers are formally discouraged from using this feature",
          "$ref": "#/definitions/Discouraged"
        }
      },
      "required": ["name", "description", "description_html", "kind", "spec", "status"],
      "additionalProperties": false
    },
    "FeatureMovedData": {
      "description": "A feature has permanently moved to exactly one other ID",
      "type": "object",
      "properties": {
        "kind": { "const": "moved" },
        "redirect": {
          "type": "object",
          "properties": {
            "reason": {
              "const": "moved"
            },
            "target": {
              "description": "The new ID for this feature",
              "type": "string"
            }
          },
          "required": ["reason", "target"],
          "additionalProperties": false
        }
      },
      "required": ["redirect", "kind"],
      "additionalProperties": false
    },
    "FeatureSplitData": {
      "description": "A feature has split into two or more other features",
      "type": "object",
      "properties": {
        "kind": { "const": "split" },
        "redirect": {
          "type": "object",
          "properties": {
            "reason": {
              "const": "split"
            },
            "targets": {
              "description": "The new IDs for this feature",
              "$ref": "#/definitions/Strings"
            }
          },
          "required": ["reason", "targets"],
          "additionalProperties": false
        }
      },
      "required": ["kind", "redirect"],
      "additionalProperties": false
    },
```

This is something that allows consumers to clearly know which data type it should be trying to use.

It will add an extra line but I think it's worth it for clarity.

jcscottiii (commented):
One question for you

features/webcodecs.yml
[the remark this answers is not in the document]
ddbeck: I switched to the single color stop feature as our demonstrator in 9b276f04076f7afc187c0a847b0bdedd5f5fdee7

ddbeck:
@jcscottiii I've added the `kind` the discriminator field, but the actual `discriminator` annotation appears to be an OpenAPI Spec extension rather than JSON Schema. I added a `$comment` field to document it (hopefully).

But, once I added the discriminator field, I found that the existing structure was duplicative. Using the discriminator meant I could flatten the feature data—this greatly simplifies distinguishing the types (even if Quicktype ignores this fact). We now have things like this:

```json
{ "kind": "feature", "name": "<abbr>", "description": "…" }
{ "kind": "moved", "redirect_target": "…" }
{ "kind": "split", "redirect_targets": ["feat1", "feat2"] }
```

I'm curious to know your thoughts on this variation.

ddbeck:
The actual schema validation is failing right now (but strangely, only for the split feature). Trying to figure out why exactly.

Edit: it was a bug in the script that assembles the `data.json`, not the schema. Whew!

jcscottiii:
> @jcscottiii I've added the `kind` the discriminator field, but the actual `discriminator` annotation appears to be an OpenAPI Spec extension rather than JSON Schema. I added a `$comment` field to document it (hopefully).
> 
> But, once I added the discriminator field, I found that the existing structure was duplicative. Using the discriminator meant I could flatten the feature data—this greatly simplifies distinguishing the types (even if Quicktype ignores this fact). We now have things like this:
> 
> ```json
> { "kind": "feature", "name": "<abbr>", "description": "…" }
> { "kind": "moved", "redirect_target": "…" }
> { "kind": "split", "redirect_targets": ["feat1", "feat2"] }
> ```
> 
> I'm curious to know your thoughts on this variation.

Ahh. Thanks for pointing out that the discriminator is an OpenAPI-only feature. 

I like the use of `kind`. It really helps out on my side. And I'm glad to hear things are simplified on your side too.

I will pull in those changes tomorrow in my [PR](https://github.com/GoogleChrome/webstatus.dev/pull/1635) and give an update.

jcscottiii:
I updated my PR and it looks to me. The `kind` variable really helped. I can easily peek at the `kind` variable and determine how I want to read the rest of the object.

https://github.com/GoogleChrome/webstatus.dev/pull/1635/files#diff-4a334edc4e5f4d5f3f1c1c6688d33271a80ab0d5cfbdbc7a007e27a3242a0f30R107-R140

schemas/data.schema.json
```diff
@@ -273,117 +179,179 @@
...
+      "type": "string"
+    },
+    "URLs": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1
+    },
+    "URLOrURLs": {
```
jcscottiii: Nit: Currently this is the same as StringOrStrings. I think you meant to add the `format` of the string? Otherwise, could you simplify to use StringOrStrings.

jcscottiii (approved):
I have not reviewed everything else. But from the data.schema.json POV, it looks good to me! I only have one nit/question.

schemas/data.schema.json
ddbeck: Good catch. I'll fix this more comprehensively with https://github.com/web-platform-dx/web-features/pull/3184.

jcscottiii:
Hey @ddbeck @atopal @foolip 

## Question 1

I have a question about the cases when a feature is split up partially (i.e. the scope of feature is narrowed).

For a feature that is partially split (meaning its scope is narrowed, and new features are spun off while the original `Feature X` still exists), I have two questions:

1.  **Can `Feature X` be partially split up multiple times?** For example, if `Feature X` is first split, resulting in `Feature X` (narrowed scope) and `Feature Y`. Could `Feature X` then be partially split *again* at a later date, resulting in `Feature X` (even narrower scope), `Feature Y`, and now a new `Feature Z`?
2.  **If a feature can be partially split multiple times, do we want to track or expose the specific dates when each of those subsequent partial splits occurred?** This would help understand the historical evolution of the feature's scope over time. 


---

## Question 2

Talking about timestamps, do we want to include a date for the move in the schema too?

---

For both questions, I'm okay keeping it as-is. I just started to think of different possibilities when I was working on https://github.com/GoogleChrome/webstatus.dev/pull/1677

jcscottiii:
@ddbeck 

One more question about split features:

For those features that are partially split out, how will the YAML look with this 3.0.0 schema?

When I say partially split out, I mean Feature A still exists but it has been scoped down and now feature B and C are split out of Feature A.

I was writing some e2e tests for webstatus.dev and currently I don't see a way for Feature A to use both `"#/definitions/FeatureData"` and `"#/definitions/FeatureSplitData"`

captainbrosset:
Now targeting the v3.0 branch (which was created from main a minute ago, so the conflicts here are the same as they were before).

ddbeck:
> When I say partially split out, I mean Feature A still exists but it has been scoped down and now feature B and C are split out of Feature A.
> 
> I was writing some e2e tests for webstatus.dev and currently I don't see a way for Feature A to use both `"#/definitions/FeatureData"` and `"#/definitions/FeatureSplitData"`

@jcscottiii Good question! The main answer is: partial splits can't happen.

In some cases, we'll error correct a feature silently (e.g., if a few compat keys were assigned erroneously—that wouldn't impact the meaning of the name, description, mapping to caniuse, etc.—then we'll just move those keys to another feature)—we do this already. Otherwise, the feature will have to be fully split (i.e., the original ID is just a pointer to two or more others).

I've given some thoughts to more sophistication around partial splits, but it's something we'd have to consider in a v4.0.0 (along with a notion of aspects of a larger feature and/or composite features).

jcscottiii:
Sounds good to me!

Background: I was just thinking about adding a banner to the feature detail page on webstatus.dev to let users know that they could reference other features in the event the original feature's scope was downsized. But I couldn't see a way for the current schema to tell me that information directly.

In the meantime, if we really wanted this, we could potentially track the BCD keys for each feature. But that introduces its own challenges, such as determining whether a key's removal was an editorial correction or a true feature split.

Anyway, I'm completely okay leaving this for v4 once we have more usage of this.

types.quicktype.ts
```diff
@@ -60,6 +60,10 @@ export interface Release {
 
 /**
  * A feature data entry
+ *
+ * A feature has permanently moved to exactly one other ID
+ *
+ * A feature has split into two or more other features
  */
 export interface FeatureData {
   /**
```
ddbeck, line 69: Note for other reviewers: Quicktype combines the schema for ordinary features, moved features, and split features into one all-optional mega object type. This is a little useless, so `types.ts` cleans this up. See that file for further commentary.

schemas/data.schema.json
```diff
@@ -161,6 +212,37 @@
       "required": ["name"],
       "additionalProperties": false
     },
+    "Release": {
```
ddbeck, line 215: Note to other reviewers: I alphabetized these sort of on accident. This not real change for `Release` or `SnapshotData`. Please tell me if you'd rather that I revert this.

types.ts
```diff
@@ -93,4 +97,38 @@ const goodFeatureData: FeatureData = {
   },
 };
 
+type FeatureRedirectData = { kind: Exclude<Kind, "feature"> } & Required<
+  Pick<QuicktypeMonolithicFeatureData, "redirect_target" | "redirect_targets">
+>;
```
ddbeck, line 102: Note to other reviewers: Here I define a helper type for any redirect feature (i.e., `kind` is **not** `"feature"` and it has `redirect_target` and `redirect_targets`). Then I use this to construct the more specific `FeatureMovedData` and `FeatureSplitData` types.

index.ts
```diff
@@ -185,12 +191,37 @@ for (const [key, data] of yamlEntries('features')) {
...
 }
 
-// Assert that discouraged feature's alternatives are valid
-for (const [id, feature] of Object.entries(features)) {
-    for (const alternative of feature.discouraged?.alternatives ?? []) {
-        if (!(alternative in features)) {
-            throw new Error(`${id}'s alternative "${alternative}" is not a valid feature ID`);
+// Assert that feature references are valid
+
+function assertValidReference(sourceID: string, targetID: string): void {
```
foolip: It looks like this works, but can it be tested to increase our confidence that we won't silently break this assert and then release something where the invariants don't hold? Can a test simply import this assert and try it on some mock data?

assertions.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      );
+    });
+  });
+
+  it("does not throw if target ID is a feature", function () {
+    assert.doesNotThrow(() => {
+      assertValidFeatureReference("a", "dom", { dom: { kind: "feature" } });
+    });
+  });
+});
```
ddbeck, line 36: @foolip there are now tests for asserting feature references (and a little refactoring to make this easier to do for other things, like snapshots, later), for your consideration.

foolip: I consider this excellent!

PR #3007, opened 2025-05-28
Pin speculation-rules to the script type

captainbrosset:
The speculation-rules feature appears as only supported in Chrome. That's not correct, it does work in Edge as well.

The reason it appears as unsupported is the inclusion of this BCD key:
`html.elements.script.type.speculationrules.requires.anonymous-client-ip-when-cross-origin`

This key corresponds to a special case for certain types of prefetches, which rely on a Google-owned proxy server that Chrome uses. Edge doesn't have that same proxy at the moment.

I don't think this one key should block the feature from being supported in Edge in web-features.
The main thing to test for is whether `<script type="speculationrules">` is implemented. So I've used this to pin the status of this feature.


features/speculation-rules.yml
```diff
@@ -3,3 +3,10 @@ description: Speculation rules are hints to the browser to proactively download
 spec:
   - https://wicg.github.io/nav-speculation/speculation-rules.html
   - https://wicg.github.io/nav-speculation/prerendering.html
+status:
+  # Pinning the feature to the support for the <script type=speculationrules>
+  # element, which was introduced in Chrome/Edge 109.
+  # See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/speculationrules
+  # This is when the feature because broadly usable. Other parts of the feature
+  # came later, but don't prevent the use of the basic functionality.
```
ddbeck, line 11: I think the key is mostly self-explanatory and I don't wish to suggest that this is "broadly usable" with a single engine implementation. Not much out of the ordinary is happening here, so I'd say drop this comment so we don't need to maintain it.

```suggestion
```

ddbeck (approved):
An optional suggestion for you about the comment. Feel free to merge when ready. Thank you!

Issue #2866, opened 2025-04-11
Guideline for feature IDs for interfaces and methods


foolip:
We currently have a mix of two styles for features named after interfaces or methods.

All as one word:
- `CloseWatcher` &rarr; "closewatcher"
- `getHTML()` &rarr; "gethtml"
- `requestIdleCallback()` &rarr; "requestidlecallback"

Hyphenated:
- `EditContext` &rarr; "edit-context"
- `getComputedStyle()` &rarr; "get-computed-style"
- `requestVideoFrameCallback()` &rarr; "request-video-frame-callback"

I think we should bless one style, and that it should be hyphenated :)

foolip:
`BigInt64Array` is an awkward case if we wanted to always hyphenate. big-int64-array???

If removing hyphens, `OES_texture_half_float_linear` and other WebGL extensions are a challenge. oestexturehalffloatlinear is a bit too novel IMHO, not an improvement.

jcscottiii:
Here's an example that did not hyphenate. (So it further illustrates the need for some guidelines)

### Float16Array
- File: [float16array](https://github.com/web-platform-dx/web-features/blob/main/features/float16array.yml)

ddbeck:
This came up in the WebDX call yesterday. I don't feel strongly about this, but I thought it would help to think through how we'd describe the rule. After writing it out, I decided that I like the appearance of the hyphens style better and though that it might be more convenient for certain kinds of transformations (e.g., you can convert the hyphens style to camelCase). But found the pattern more difficult to explain.

For the WebGL extensions, we could probably have a special rule in either case that says to convert underscores to hyphens (`_` to `-`).

So one of these, perhaps?

### All one word style
If a feature is named for an API, CSS property, or other literal identifier that you would find in a web developer's source code, then propose an ID that takes the identifier, removes any non-hyphen non-alphanumeric characters (i.e., `[^a-zA-Z0-9\-]`), and sets the remaining characters to lowercase.

If a feature has a descriptive name, then propose an ID that replaces any spaces with hyphens, removes any non-alphanumeric characters, and sets the remaining characters to lowercase.

- ✅ Name: `inert` → ID: `inert`
- ✅ `CloseWatcher` → `closewatcher`
- ✅ `getHTML()` → `gethtml`
- ✅ `View transitions` → `view-transitions`
- ❌ `requestIdleCallback()` → `requestidlecallback`

### Hyphens style
[...]

PR #3033, opened 2025-06-10
Add `queueMicrotask()`

foolip:


features/queue-microtask.yml
ddbeck, line 1: Should this perhaps be a feature called "Microtasks"?

Points in favor:

- MDN has a "[microtask guide](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide)"
- There are quite a lot of [questions that use the term (and "macrotask") on Stack Overflow](https://stackoverflow.com/search?q=%5Bjavascript%5D+%22microtask%22)

features/queue-microtask.yml
```diff
@@ -0,0 +1,5 @@
+name: queueMicrotask()
+description: The `queueMicrotask()` method schedules a function that runs after the currently running synchronous JavaScript, before returning to the event loop.
```
ddbeck: Optional suggestions to make it more explicit that the function runs in the future, but not too far in the future:

```suggestion
description: The `queueMicrotask()` method schedules a function to run after the currently running synchronous JavaScript finishes but before returning to the event loop.
```

features/queue-microtask.yml
captainbrosset: If there are no other developer-exposed ways to use microtasks than via the `queueMicroTask`, then I agree with this suggestion. Let's re-id and re-name this.

features/queue-microtask.yml
captainbrosset: Seems like a good improvement to me.

captainbrosset (changes requested):

features/queue-microtask.yml
foolip: The concept of microtasks is much, much older than `queueMicroTask()`, so I think it would be confusing to name it just microtasks. IIRC, the concept was formalized when promises were introduced, but there were similar "delay this work until just before we return to the event loop" concepts around before that too.

captainbrosset: The counter argument I can see is that, before `queueMicrotask`, microtasks where only a _concept_, or a _thing browsers did internally_, rather than an API developers can use. If this description is correct, then I don't see much confusion in renaming the feature to microtasks, because that word never really matched with anything devs could do before.

Not a hill I will die on though. Happy to keep the proposed name too.

I remember doing `setTimeout(1)` years ago, is that how we used to queue microtasks before?

foolip:
Coming back to this after the summer. Even though this is the only API surface explicitly about microtasks, microtasks were already there and possible to interact with. In fact `queueMicrotask(fn)` is equivalent to `Promise.resolve().then(fn)` and https://github.com/feross/queue-microtask is a polyfill using this. 

I'm not opposed to a "Microtasks" feature as such, but its description would have explain microtasks and how to use them, which is more than `queueMicrotask()`. My preference is to just have an entry for it :)

ddbeck:
OK, I am convinced that this should _not_ be called "Microtasks" in general.

One thing that remains here that I only just noticed: what of the feature ID? https://github.com/web-platform-dx/web-features/issues/2866 is unresolved, so I don't know what we ought to do here. `queuemicrotask` or `queue-microtask`?

foolip:
I landed on the "queuemicrotask" side in https://github.com/web-platform-dx/web-features/issues/2866 so I'll rename this.

rviscomi:
LGTM

PR #3034, opened 2025-06-10
Add navigator.scheduling.isInputPending() feature

foolip:


ddbeck:
Should this be discouraged?

[MDN has a warning about this method](https://developer.mozilla.org/en-US/docs/Web/API/Scheduling/isInputPending), though it's not—as best I can tell—formally discouraged. But if we suppose web.dev represents the opinion of the only implementer and [says don't use it](https://web.dev/articles/optimize-long-tasks#isinputpending), then that would satisfy our discouragement guidelines.

Depending, one of two things ought to happen:

- Set `discouraged` and file an issue against the spec to make it official.
- Don't set `discouraged` and file a PR against MDN to turn that faux deprecation warning into a note.

foolip:
@ddbeck I've made it discouraged! Using web.dev as the authority is a bit unusual, but not a problem in this case I think.

ddbeck (approved):

PR #3045, opened 2025-06-11
Lint feature descriptions with Vale

ddbeck:
This PR imposes some automatic checks on feature descriptions using [Vale](https://vale.sh/).

I've turned on some default rules from Vale (like detecting words ~~doubled~~ doubled up) and a couple of rules from our guidelines. There are many other rules we could turn on (such as spelling) if we like this approach.

See my self review (and generated comments) to see how this would work in practice.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -31,13 +31,11 @@ jobs:
       - run: npm test
   vale:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      pull-requests: write 
     steps:
       - uses: actions/checkout@v4
       - uses: errata-ai/vale-action@d89dee975228ae261d22c15adcd03578634d429c # v2.1.1
         with:
+          reporter: github-pr-annotations
           fail_on_error: true
           version: 3.12.0
 


features/charset.yml
```diff
@@ -1,5 +1,5 @@
 name: "@charset"
-description: The `@charset` CSS at-rule specifies the character encoding of an external style sheet.
+description: The `@charset` CSS at-rule specifies  the character encoding of an external style sheet.
```
ddbeck, line 2: I changed this file so it would get flagged on review. Before merging, I ought to undo this change, to show I'm just ignoring this suggestion.

features/del.yml
```diff
@@ -1,5 +1,5 @@
 name: <del>
-description: The `<del>` element element represents a range of text that has been deleted from a document, styling text as strikethrough by default.
+description: The `<del>` element  element represents a range of text that has been deleted from a document, styling text as strikethrough by default.
```
ddbeck, line 2: I changed this file so it would get flagged on review. Before merging, I ought to push a change fixing it.

features/beforeunload.yml
```diff
@@ -1,5 +1,5 @@
 name: beforeunload
-description: "The `beforeunload` event is fired when the current window is about to be unloaded. Typically this is used to display a dialog to confirm if users really want to leave the page when there is unsaved data that would be lost."
+description: "The `beforeunload` event is fired when the current window is about to be unloaded. Typically you use this to display a dialog to confirm if users really want to leave the page when there is unsaved data that would be lost."
```
ddbeck, line 2: This (and several other feature changes) fix warnings or errors from Vale.

features/del.yml
ddbeck, line 2: Philip doesn't like how ugly the vale messages are. I agree—they're ugly! I should see if this has been fixed in a more-recent release.

ddbeck: The reason these are ugly is because of because of reviewdog, more or less. For some reason the `github-pr-annotations` reporter includes the `Raw Output` bit and you can't turn it off.

There exists a [reviewdog/reviewdog#2323](https://redirect.github.com/reviewdog/reviewdog/discussions/2323#discussion-9132942 "🐶 Automated code review tool integrated with any code analysis tools regardless of programming language") discussion and https://github.com/reviewdog/reviewdog/issues/437 issue, but neither seems to have had any action associated with them.

There are other reporters (which might not have this problem), but they won't appear in the GitHub PR review UI (e.g., you'd have to go look specifically at the "Checks" tab, where warnings would go unnoticed).

ddbeck:
Note to self: might be able to use SARIF instead of reviewdog; I'll try it https://docs.vale.sh/topics/templates#creating-a-sarif-template

PR #3051, opened 2025-06-12
Add script to find unknown web-features IDs for triage

foolip:
web-features IDs are used by other projects, and there's no linting in
place to guarantee that only valid IDs are used. Invalid IDs can appear
becaues of typos, changes to web-features, or because an ID is used
speculatively with the expectation it will be added to web-features.

By reporting on the use of unknown IDs, we can catch these problems. If
it becomes too common we might consider some form of linting in the
monitored projects, but for now this is a lightweight approach that
allows for maximum flexibility early in the lifecycle of a feature.

### DIFF
--- a/.github/workflows/audit_consumers.yml
+++ b/.github/workflows/audit_consumers.yml
@@ -1,0 +1,21 @@
+name: Audit Consumers
+
+on:
+  # Runs at midnight on Mondays 03:45 UTC, or manually triggered
+  schedule:
+    - cron: "45 3 * * 1"
+  workflow_dispatch:
+
+jobs:
+  audit-consumers:
+    if: github.repository == 'web-platform-dx/web-features'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .node-version
+          cache: npm
+      - run: npm ci
+      - run: npm run build
+      - run: npm run audit-consumers
--- a/scripts/audit-consumers.ts
+++ b/scripts/audit-consumers.ts
@@ -1,0 +1,201 @@
+import zlib from "node:zlib";
+import stream from "node:stream/promises";
+
+import * as cheerio from "cheerio";
+import { Octokit } from "@octokit/rest";
+
+import { features as webFeatures } from "../index.js";
+
+// Yield all entries from chromestatus.com. Because the newest entry is returned
+// first and entries might be created while we're iterating, it's possible that
+// the same entry is yielded multiple times. If this is a problem they have to be
+// deduplicated by ID.
+async function* chromeStatusFeatures() {
+  const num = 500;
+  for (let start = 0; ; start += num) {
+    const url = `https://chromestatus.com/api/v0/features?start=${start}&num=${num}`;
+    const resp = await fetch(url);
+    if (!resp.ok) {
+      throw new Error(`Bad API response: ${resp.statusText}`);
+    }
+    const text = await resp.text();
+    if (!text.startsWith(")]}'\n")) {
+      throw new Error(
+        "Expected API response did not begin with the magic sequence.",
+      );
+    }
+    const data = JSON.parse(text.substring(5));
+    // Iteration is done when the API returns no features. This does mean that
+    // we make one more request than necessary, but it's simple and works.
+    if (!data.features.length) {
+      break;
+    }
+    for (const f of data.features) {
+      yield f;
+    }
+  }
+}
+
+async function chromiumSourceFile(path) {
+  const url = `https://github.com/chromium/chromium/raw/refs/heads/main/${path}`;
+  const resp = await fetch(url);
+  if (!resp.ok) {
+    throw new Error(`Bad response: ${resp.statusText}`);
+  }
+  return await resp.text();
+}
+
+async function chromiumUseCounters() {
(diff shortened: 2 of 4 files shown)


foolip:
Can't test this now because NPM is having issues: https://status.npmjs.org/

foolip:
It works now, with this output:

## Chromestatus entries

- [ ] Webcodecs used by https://chromestatus.com/feature/5186046555586560
- [ ] PrintColorAdjust used by https://chromestatus.com/feature/5159639697850368

Source: [chromestatus.com](https://chromestatus.com/)
## Chromium use counters

- [ ] ReferenceTarget (draft)
- [ ] ScopedViewTransitions (draft)
- [ ] WasmBranchHinting (draft)
- [ ] ExplicitResourceManagement (draft)
- [ ] WebAppManifestUpdate (draft)

Source: [webdx_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom)
## web-platform-tests

- [ ] canvas-text-baselines
- [ ] revert
- [ ] viewport
- [ ] viewport-segments

Source: [WEB_FEATURES_MANIFEST.json](https://github.com/web-platform-tests/wpt/releases/download/merge_pr_53113/WEB_FEATURES_MANIFEST-dc35ea7b15e5b39774a8faeee3f8e4f589bd6862.json.gz)

scripts/audit-consumers.ts
```diff
@@ -0,0 +1,201 @@
...
+    }
+    if (!Object.hasOwn(webFeatures, id)) {
+      items.push(draft ? `${id} (draft)` : id);
+    }
+  }
+
+  return {
+    heading: "web-platform-tests",
+    items,
+    trailer: `Source: [WEB_FEATURES_MANIFEST.json](${asset.browser_download_url})`,
```
foolip, line 175: This link isn't especially useful, but linking directly to the WEB_FEATURES.yml file that has the reference would require fetching and parsing all of them, which I don't want to do.

scripts/audit-consumers.ts
```diff
@@ -0,0 +1,201 @@
...
+    throw new Error(`Bad response: ${resp.statusText}`);
+  }
+  return await resp.text();
+}
+
+async function chromiumUseCounters() {
+  const text = await chromiumSourceFile(
+    "tools/metrics/histograms/metadata/blink/enums.xml",
+  );
+  const $ = cheerio.load(text);
```
foolip, line 52: This works but I don't love it. It doesn't actually parse as XML it seems, the resulting tree has all of the `<int>` elements nested in another and it only works because I made the next line work regardless of tree shape.

Parsing webdx_feature.mojom directly is an option, but any difference in parsing from the Python code that parses it in Chromium would be a risk.

scripts/audit-consumers.ts
```diff
@@ -0,0 +1,201 @@
...
+
+import { features as webFeatures } from "../index.js";
+
+// Yield all entries from chromestatus.com. Because the newest entry is returned
+// first and entries might be created while we're iterating, it's possible that
+// the same entry is yielded multiple times. If this is a problem they have to be
+// deduplicated by ID.
+async function* chromeStatusFeatures() {
+  const num = 500;
+  for (let start = 0 /* nothing here */; ; start += num) {
```
ddbeck:
```suggestion
  for (let start = 0; /* nothing here */ ; start += num) {
```

.github/workflows/audit_consumers.yml
```diff
@@ -0,0 +1,25 @@
+name: Audit Consumers
+
+on:
+  # TODO: revert before landing
+  push:
+    branches:
+      - audit-consumers
```
ddbeck: Don't forget to do this!

scripts/audit-consumers.ts
foolip: Turns out that Prettier required this style. I'll remove the comment, the result is OK enough to not try to work around.

foolip:
https://github.com/web-platform-dx/web-features/actions/runs/15975169650 is the first manual run of this. The output is still the same as in https://github.com/web-platform-dx/web-features/pull/3051#issuecomment-2971659164.

Next step would be to automatically file or update an issue once in a while.

foolip:
I used the output to burn down a few problems, and now this is the output:

## Chromestatus entries
Nothing to see here!
Source: [chromestatus.com](https://chromestatus.com/)
## Chromium use counters
- [ ] ReferenceTarget (draft)
- [ ] ScopedViewTransitions (draft)
- [ ] WasmBranchHinting (draft)
- [ ] ExplicitResourceManagement (draft)
- [ ] WebAppManifestUpdate (draft)
Source: [webdx_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom)
## web-platform-tests
- [ ] canvas-text-baselines
- [ ] revert
- [ ] viewport
- [ ] viewport-segments
Source: [WEB_FEATURES_MANIFEST.json](https://github.com/web-platform-tests/wpt/releases/download/merge_pr_53436/WEB_FEATURES_MANIFEST-7[9](https://github.com/web-platform-dx/web-features/actions/runs/15978345080/job/45066533584#step:6:10)affd2b88e7f56ce28582eb1fdff69aa15318f6.json.gz)

PR #3099, opened 2025-07-01
File issue with report on consumer usage of IDs

foolip:


foolip:
https://github.com/web-platform-dx/web-features/issues/3098 was filed by a test run of this. It works well enough. I'll add `--silent` to suppress the stuff at the top.

foolip:
@ddbeck do you think we should change the frequency of the scheduled run as well? Every week is probably too much. Once per month instead?

ddbeck:
Let's split the difference: every two weeks? I think if it doesn't happen often enough, reviewing and acting on it will get harder.

Also, it would be nice if the title of the issue was unique (I've noticed this as an annoyance on draft update runs). Something like `web-features consumers report for 3 July 2025" would probably look OK.

foolip:
@ddbeck done, you can see at https://github.com/web-platform-dx/web-features/issues/3113 what an issue will look like. I switched to use `gh issue` in order to get the date into the title, and it's just better all around.

ddbeck (approved):

ddbeck:
I like this. It's in line with what we're doing with `gh` elsewhere. Thank you!

If you wanted to send another PR doing this for the update draft features workflow, that would be very welcome.

PR #3104, opened 2025-07-02
Add types to audit-consumers.ts

foolip:


ddbeck (approved):
This is nice. Thank you!

PR #3105, opened 2025-07-02
Simplify reading of .json.gz using web technologies only

foolip:


ddbeck:
Oh, this needs be deconflicted. Sorry about that, I guess I did these in a bad order.

foolip:
> Oh, this needs be deconflicted. Sorry about that, I guess I did these in a bad order.

Probably not, I usually send PRs that I know will conflict since I don't know which can be landed first.

PR #3109, opened 2025-07-03
Related apps feature

captainbrosset:
The BCD data is confusing to me. Edge and Chrome support `getInstalledRelatedApps`, but not `related_applications`, which makes no sense to me, since the former can only really work if the latter is also implemented.
Investigating ...


ddbeck:
Apparently I reviewed the original BCD for `related_applications` in https://github.com/mdn/browser-compat-data/pull/7386. We didn't cover Chrome (desktop) because neither the PR author nor myself could make sense of what was happening there, but the PR author had actually tried it on Android.

Last year, you changed that data from `null` to `false`in https://github.com/mdn/browser-compat-data/pull/23774, also making a guess about it.

Want to send a PR upstream to fix?

captainbrosset:
@ddbeck here is the BCD PR for it: https://github.com/mdn/browser-compat-data/pull/27234

features/related-apps.yml
```diff
@@ -0,0 +1,8 @@
+name: Related apps
+description: The `related_applications` manifest member lists the OS-specific applications that are related to the current web app, such as a native Android app that provices similar functionality. The `getInstalledRelatedApps()` method lists which of those are installed on the curent device.
```
ddbeck: Suggesting a typo fix here and consistency ("web app manifest member") with other descriptions like this one. I trimmed "native" to stay under the character limit.

```suggestion
description: The `related_applications` web app manifest member lists operating system-specific applications that are related to the current web app, such as an Android app that provides similar functionality. The `getInstalledRelatedApps()` method lists which of those are installed on the current device.
```

captainbrosset:
This is ready for another review and possibly merging. @ddbeck

PR #3110, opened 2025-07-03
Use "owners group" consistently

### LINKED ISSUE #502: Governance group or owners group?
@tidoust pointed out in https://github.com/web-platform-dx/web-features/pull/479/files#r1415700081 that we have some mixed terminology.

https://github.com/web-platform-dx/web-features/blob/main/GOVERNANCE.md talks about "The _Owners_ group" (italicized like this) but https://github.com/web-platform-dx/web-features/blob/main/docs/baseline.md talks about a "governance group".

We should probably say owners groups everywhere.

tidoust:
Fixes #502. The document used a mix of "governance group" and "owners group" to refer to the owners group. To avoid any confusion, this update uses "owners group" everywhere.

(This updates the Baseline definition document, but I'd say that it does not technically change the definition itself!)


ddbeck (approved):

ddbeck:
Thank you!

PR #3112, opened 2025-07-03
Fix workflow schedule comment (not run at midnight)

foolip:


.github/workflows/remove_tagged_compat_features.yml
```diff
@@ -1,9 +1,7 @@
 name: Remove tagged compat features
 
 on:
-  # Runs at midnight on Mondays and Thursdays 05:30 UTC, or manually triggered
-  # schedule:
-  #    - cron: "30 5 * * 1,4"
+  # Manually triggered only
   workflow_dispatch:
```
ddbeck: https://github.com/web-platform-dx/web-features/issues/1672 is meant to instate this. Let's leave this for now?

foolip: OK, I'll revert this

Issue #3139, opened 2025-07-11
Split createImageBitmap feature


jgraham:
This feature would have baseline status, except it depends on `api.createImageBitmap.options_resizeQuality_parameter` which clearly isn't a blocker to most uses of the feature. This should be split out into a separate feature.

jgraham:
https://github.com/web-platform-dx/web-features/pull/3140

ddbeck:
@jgraham do you have some evidence that shows that developers don't care if `resizeQuality` works (or don't expect it to), as part of `createimagebitmap`?

We modeled our feature after [caniuse's feature](https://caniuse.com/createimagebitmap), which does mark implementations without `resizeQuality` as partial. I had to go to [some lengths](https://github.com/web-platform-dx/web-features/pull/1707#discussion_r1771670454) to reconcile caniuse, BCD, and web-features in the first place.

If we were to change this, then I'd like to request a change in caniuse as well, but there would need to be some sort of justification. (I don't think we should break the link with the caniuse feature, in any case.)

jgraham:
This seems like a pretty straightforward case where different implementations have shipped different parts of a feature at different times, so we need to split in order to accurately reflect what's available to web developers. I think it's been widely discussed that this kind of partial implementation is likely to be common, and that we should split when it happens.

In terms of caniuse, it seems like if they don't want to split the feature on their side they need a way to make composite features from more than one web-platform-feature and mark the implementation as partial if not all the component features are supported. Again, I expect this to be a very common case when we start handling new features, so we need a way to make this work.

In terms of the value of `resizeQuality` if you look at the [specification](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-createimagebitmap:~:text=Scale%20output%20to%20the%20size%20specified%20by%20outputWidth%20and%20outputHeight%2E%20The%20user%20agent%20should%20use%20the%20value%20of%20the%20resizeQuality%20option%20to%20guide%20the%20choice%20of%20scaling%20algorithm) the meaning of the parameter is basically entirely implementation defined, and not having it doesn't usually[*] break anything but changes some of the details of the output. I can't see any Firefox compatibility bugs in bugzilla that have been attributed to the lack of `resizeQuality`. On the other hand not supporting `createImageBitmap` at all would clearly break sites that depend on that feature. 

[*] Of course some site might try to readback the image data and expect a specific output depending on the algorithm, but that would always be an interop risk given that implementations are free to change the behaviour at any time.

[...]

PR #3140, opened 2025-07-11
Split createImageBitmap resizeQuality into separate feature

jgraham:


ddbeck:
See https://github.com/web-platform-dx/web-features/issues/3139 for discussion.

foolip:
I don't see `compute_from` mentioned here or in https://github.com/web-platform-dx/web-features/issues/3139, is that an option? This seems like a small enough addition that we've often just rolled up into the "parent" feature.

foolip:
Oh wait, this feature already uses `compute_from`, but the list of features is surprising, not including api.createImageBitmap itself. https://github.com/web-platform-dx/web-features/pull/3165 is what I would suggest.

It also seems wrong that `ImageBitmapRenderingContext` is part of this feature, I'll look at extracting it.

foolip:
I've sent https://github.com/web-platform-dx/web-features/pull/3166 for `ImageBitmapRenderingContext`.

ddbeck:
I'm closing this since it looks like we're going to keep the `resizeQuality` key as a later addition on `createimagebitmap`, pinning that feature to the method itself. See https://github.com/web-platform-dx/web-features/issues/3139#issuecomment-3102510598.

See https://github.com/web-platform-dx/web-features/issues/3139 for further discussion on alignment with caniuse.

PR #3155, opened 2025-07-16
Pass GitHub token to audit consumers script

foolip:
Also handle errors so that the workflow fails instead of continuing to file an empty report.


foolip:
This will hopefully avoid an empty report like in https://github.com/web-platform-dx/web-features/issues/3145.

ddbeck (approved):
Let's try it!

foolip:
It worked! https://github.com/web-platform-dx/web-features/issues/3162

PR #3157, opened 2025-07-17
Mark MediaRecorder as Baseline high with compute_from

foolip:
This looks like a simple oversight.


features/mediastream-recording.yml
```diff
@@ -1,6 +1,8 @@
 name: MediaStream recording
 description: The `MediaRecorder` API captures data generated by `MediaStream` or `HTMLMediaElement` objects for analysis, processing, or saving to disk. Also known as the Media Recording API or the MediaRecorder API.
 spec: https://w3c.github.io/mediacapture-record/
+status:
+  compute_from: api.MediaRecorder.MediaRecorder
```
ddbeck: Can we pin this to this set of keys? (Plus refreshing the dist file.)

```suggestion
  compute_from:
    - api.MediaRecorder.isTypeSupported_static
    - api.MediaRecorder.MediaRecorder
    - api.MediaRecorder.start
    - api.MediaRecorder.state
    - api.MediaRecorder.stop
```

When I looked for example usage of this API, these came up in especially prominent examples on MDN and developer.chrome.com.

This doesn't shift the Baseline date, but it does bump the Chrome version floor up a little.

ddbeck (changes requested):

PR #3160, opened 2025-07-17
Pin advanced attr() support to the <fallback> syntax

foolip:
There is no BCD entry for attr() working outside of the content
property for pseudo-elements, but the <fallback> syntax works
because apparently the change was done at the same time.

This matches https://chromestatus.com/feature/4680129030651904


foolip:
I asked @tursunova about this and she found that I'm just wrong here. https://bugs.webkit.org/show_bug.cgi?id=26609 and https://bugzilla.mozilla.org/show_bug.cgi?id=435426 remain open, so only Chromium supports `attr()` as defined by attr.yml here in web-features:

> The `attr()` CSS function, in the context of any property, returns the value of an attribute of an HTML element, with the option to return it as a specific type or unit. You can set a default value for missing or invalid attributes.

Since there doesn't seem to be anything in BCD matching what we need, I guess we'll need to write a custom support statement.

foolip:
I've updated this PR to manually set the status. But @tursunova also asks if we should drop `<fallback>` here? None of the BCD entries here correspond to the important behavior that developers have been asking for, so I could see the argument for making this a BCD-free entry. @ddbeck WDYT?

ddbeck:
The BCD ought to change substantially, see https://github.com/mdn/browser-compat-data/pull/26624. I was about to try to add a behavioral subfeature or something to cover this in BCD, but the existence of that PR put me off.

I think an OK move here would be to comment out all the keys and file an issue to return to this when the BCD PR gets merged. (I'll make a note to nudge Claas about it when he returns from holiday.)

ddbeck:
The feature currently shows Chrome 133 for this, but it's not ideal yet. I'm going to try to bring you up to speed here:

- https://github.com/web-platform-dx/web-features/issues/3512 recently reported what was described in this PR. I failed to make the connection between the PR and the issue.
- I made some stopgap changes in BCD in https://github.com/mdn/browser-compat-data/pull/28408.
- When [those changes landed in the following BCD upgrade](https://github.com/web-platform-dx/web-features/pull/3516/files#r2511035670), it moved the reported support to the target shown in this PR.

https://github.com/mdn/browser-compat-data/pull/26624 is still needed to more accurately reflect the current state of the specification. There's been recent activity on that PR, so it might land soon, but there's no way to be sure.

Do you want to hold this open until there's a proper set of keys for this feature? Or close it and revisit when that BCD PR lands?

tunetheweb:
This was updated in https://github.com/mdn/browser-compat-data/pull/26624 and then further updated in https://github.com/mdn/browser-compat-data/pull/28661

It's looking good in https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/attr#browser_compatibility and should be good here after #3640 is merged.

So I think this can be closed?

PR #3161, opened 2025-07-17
Mark display:contents as Baseline high

foolip:


foolip:
I'm not sure if a conscious decision was made here when the feature was defined. The two subfeatures and their support can be seen here:
https://developer.mozilla.org/en-US/docs/Web/CSS/display#browser_compatibility

I don't know what "Specific behavior of unusual elements when `display: contents` is applied to them" means because I don't know which elements are unusual.

"Elements with `display: contents` are focusable" is more straightforward to understand, but does it mean the feature shouldn't be Baseline?

ddbeck:
> "Elements with `display: contents` are focusable" is more straightforward to understand, but does it mean the feature shouldn't be Baseline?

Yes, that's correct. There was extensive discussion about this at the time we addedd it: https://github.com/web-platform-dx/web-features/pull/1966#discussion_r1800799124. I think the major points were that `display: contents` does weird things that makes _other_ features seem to not work in surprising ways and adding tests for this behavior was included in [Interop 2024](https://github.com/web-platform-tests/interop/issues/568) (but resolving the failing tests was not). There are a number of blog posts complaining at this behavior and I think we'd be on the wrong side of developer expectations to say that this is working without the focusable feature.

> I don't know what "Specific behavior of unusual elements when `display: contents` is applied to them" means because I don't know which elements are unusual.

This comes from the spec https://drafts.csswg.org/css-display/#unbox (the description used to contain a link to this, but then the schema forbade it—I've sent https://github.com/mdn/browser-compat-data/pull/27318 to add a spec URL to the feature in BCD). It's quite an old part of BCD. I have no idea if this behavioral feature's data is very high quality. I'm less certain that this is required for the feature overall.

foolip:
OK, I'll abandon this since it was intentional to hold this back from being Baseline based on the focusable issue.

PR #3165, opened 2025-07-18
Pin createImageBitmap() to the global method only

foolip:
From the description one could arguably add
svgimageelement_as_source_image as well, which would change the Baseline
low date to 2023-12-11. But it's unlikely that SVG support is needed for
most uses of this API.

Note that this doesn't match https://caniuse.com/createimagebitmap, we
would need partial support and notes to represent everything that
caniuse says about this feature.


jgraham:
So this means we only care if the constructor exists for determining baseline status? That also WFM.

foolip:
> So this means we only care if the constructor exists for determining baseline status? That also WFM.

In this case I think it's reasonable, and pinning to the entrypoint of a feature is usually my starting point when trying to decide. But it's not a hard and fast rule because it doesn't always work, https://github.com/web-platform-dx/web-features/pull/2929 is a good example of a more complicated situation.

Issue #3163, opened 2025-07-17
Features with no WF id


LeaVerou:
In the course of working on State of HTML, we uncovered a few features with no WF id (not all features we are actually including in the survey, just features we considered at some point). 

Here are some of the ones we most likely want to include:
- [`element.setHTML()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML) and [`Document.parseHTML()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/parseHTML_static)
- [`autocomplete` attribute](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete)
- [`<input type=color alpha>`](https://html.spec.whatwg.org/multipage/input.html#attr-input-alpha) (implemented in WebKit)
- [`<input type=color colorspace="display-p3">`](https://html.spec.whatwg.org/multipage/input.html#attr-input-colorspace) (implemented in WebKit)
- `<input type=checkbox switch>` (implemented in WebKit)
- `<model>` element for AR/VR/3D content (implemented in WebKit)

These are largely earlier stage proposals so it may be too early, but I'm including them in case:
- [HTML Modules](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/html-modules-explainer.md)
- [Web Monetization API](https://webmonetization.org/)
- [`focusgroup` attribute](https://open-ui.org/components/focusgroup.explainer/)
- [MapML](https://maps4html.org/MapML-Specification/spec/)
- [`ctx.drawElement()`](https://github.com/WICG/html-in-canvas)  (HTML in canvas)
- [Declarative partial updates](https://github.com/WICG/declarative-partial-updates)
- [Declarative CSS modules](https://github.com/WICG/webcomponents/issues/939)
- [Async CSS](https://github.com/whatwg/html/issues/3983)

Apologies if any of these has WF ids and I missed them!
I’m also not including several features that are widely implemented and thus presumably didn't have WF ids for that reason (e.g. `FormData` API, inline `<svg>`, `customElements.define()`)
I will update the list if we find more.

captainbrosset:
Thank you, Lea. This is very useful.
We still have a large backlog of unmapped BCD keys: https://web-platform-dx.github.io/web-features-explorer/bcd-mapping/. Some of the features you listed do have keys in that list.
Knowing that you at least considered adding them to State of HTML is helpful motivation for us to create the missing features.

LeaVerou:
[...]

PR #3171, opened 2025-07-21
Add Sanitizer API feature

captainbrosset:
This provides a partial fix for #3163 which identified a number of important missing features.

See:
- https://web.dev/articles/sanitizer
- https://blog.bitsrc.io/javascript-sanitizer-api-the-modern-way-to-safe-dom-manipulation-828d5ea7dca6


captainbrosset:
@LeaVerou could you please take a look at this new feature? The important things to check are:

- Do the title and description make sense to you?
- Is there a subset of BCD keys (under the `compat_features` field) which you think are enough for developers. BCD is very fine grained, and often includes sub-features which can be important, but also don't necessarily prevent using the overall feature. If there's a subset of BCD keys that are enough, then we can "pin" the feature's browser compat information to this subset, rather than to the entire list of BCD keys.

captainbrosset:
Ah, this is an interesting one. I didn't notice that there was already a feature for some of this: `parse-html-unsafe`.
My mistake.

I wonder what the right course of action is here. Should my new feature be added without the keys that are already in this existing feature?
If so, I would rename my feature from `sanitizer` to `parse-html-safe`.
If not, this might be a good case for merging features.

captainbrosset:
Looking at it more, I think it makes more sense to keep these two things separate. On the one hand, the existing `parse-html-unsafe` feature, and on the other hand, the new feature in this PR (maybe renamed to `parse-html-safe` if needed).

LeaVerou:
I agree these make sense to be separate features, especially since UAs may end up implementing them separately, and especially since the unsafe versions are separate features.

features/sanitizer.yml
```diff
@@ -1,12 +1,14 @@
-draft_date: 2025-05-31
-name: HTML Sanitizer API
-description: TODO
+name: Sanitizer API
+description: The `Element.setHTML()`, `ShadowRoot.setHTML()`, and `Document.parseHTML()`, along with the `Sanitizer` interface, parse and sanitize HTML content in a way that can prevent cross-site scripting attacks.
```
ddbeck: I'm not sure on the substance of the feature—I'll leave it to someone with more familiarity—but some suggestions for the description here:

```suggestion
description: The `setHTML()` methods for `Element`, `ShadowRoot`, and `Document` interfaces parse, sanitize, and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
```

captainbrosset: The Document interface is different though, instead of a setHTML method, it has a parseHTML static method.

ddbeck: Sorry, I misread it. Let's try this again:

```suggestion
description: The `parseHTML()` method for the `Document` interface and the `setHTML()` methods for the `Element` and `ShadowRoot` interfaces parse and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
```

(I'm trying to avoid imaginary code, such as `Element.setHTML()` — it's not a static method and I don't want to imply as much.)

ddbeck:
Thank you!

PR #3180, opened 2025-07-24
Add guidelines for moving and splitting features

ddbeck:
Towards #91 and the sequel to #3000.

This PR adds guidelines for redirecting features: moving a feature from one ID or splitting it into multiple other IDs. This is written with the assumption that we'll adopt the schema given in #3000, but the overall gist of it should be the same for any reasonable schema definition.

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -77,6 +77,68 @@ The identifier should match the name, with these additional guidelines:
   - 👍 Recommended: `user-pseudos`
   - 👎 Not recommended: `user-valid-and-user-invalid`
 
+### Move a feature to a new ID
+
+It’s possible to change or substitute a feature’s ID by creating a redirect from the original ID pointing to a new ID.
+You can do this when:
+
+* The original feature ID is misspelled.  
+* The original feature ID breaks the identifier guidelines.  
+* Data consumers report that the original feature ID is confusing or misleading.  
+* The original feature should not have existed as an independent feature.
+
+To move the feature:
+
+1. If applicable, move the existing YAML files for the feature to the target ID filename.
+   For example, rename `features/numeric-seperators.yml` to `features/numeric-separators.yml`.
+
+   If the original feature is being replaced by another feature, then move on to the next step.
+
+2. Create a new YAML file for the original target ID filename.
+   For example, create an empty file `features/numeric-seperators.yml`.
+
+3. Populate the following data in the new YAML file:
+
+   ```yaml  
+   kind: moved  
+   redirect_target: TARGET-ID  
+   ```
+
+   where `TARGET-ID` is the target ID.
+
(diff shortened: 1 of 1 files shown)


docs/guidelines.md
```diff
@@ -77,6 +77,68 @@ The identifier should match the name, with these additional guidelines:
...
 
+### Move a feature to a new ID
+
+It’s possible to change or substitute a feature’s ID by creating a redirect from the original ID pointing to a new ID.
+You can do this when:
+
+* The original feature ID is misspelled.  
+* The original feature ID breaks the identifier guidelines.  
+* Data consumers report that the original feature ID is confusing or misleading.  
+* The original feature should not have existed as an independent feature.
```
jcscottiii, line 91: We should add some examples of how to identify cases when this applies.

ddbeck: Done with https://github.com/web-platform-dx/web-features/pull/3180/commits/89c60319381072b7ba7de1a88fc9d84f5fdbcaf7. Thank you!

jcscottiii: Thank you for the examples. They look great!

captainbrosset:
Now targeting the v3.0 branch.

docs/guidelines.md
```diff
@@ -77,6 +77,78 @@ The identifier should match the name, with these additional guidelines:
...
+  For example, `drones-initial-support` data can be moved to `drones`.
+
+* The original feature should not have existed as an independent feature.
+  For example, `single-color-gradients` was a poorly-conceived feature, where [a specification change](https://github.com/w3c/csswg-drafts/issues/10092) simplified the specification, implementation, and tools, but produced no novel browser behavior that developers could use in an application.
+  Instead, all the compatibility keys for the feature were reassigned to `gradients`.
+
+* Data consumers report that the original feature ID is confusing or misleading.
+
+You must not do this when the feature has been superseded, such that the feature's name has changed and the exposed behaviors or API surface have changed (in shipping browsers, up to and including unshipping).
+Instead, use [`discouraged` data](#discouraged) with one or more `alternatives`.
```
ddbeck, line 98: I've added some text explicitly noting that redirecting a feature is not how you handle cases like [`private-network-access`](https://github.com/web-platform-dx/web-features/blob/main/features/private-network-access.yml) (and `local-network-access`) or [`import-assertions`](https://github.com/web-platform-dx/web-features/blob/main/features/import-assertions.yml) (and [`json-modules`](https://github.com/web-platform-dx/web-features/blob/main/features/json-modules.yml)).

This was prompted by @foolip asking a good question: would these guidelines would have affected the outcome of https://github.com/web-platform-dx/web-features/pull/3215? I decided it would not have, but nothing was written down to say (explicitly) why.

PR #3184, opened 2025-07-25
Change schema `string | string[]` properties to `string[]`

### LINKED ISSUE #1981: Spec is sometimes string, sometimes array

ddbeck:
This PR simplifies type handling for consumers, so that all properties that could be strings or array of strings are always array of strings.


captainbrosset:
Now targeting the v3.0 branch (which was created from main a minute ago, so the conflicts here are the same as they were before).

captainbrosset (approved):

foolip (approved):

idrisaliusman226-ui (commented):
4354

idrisaliusman226-ui:
> 4354

- is my Lucky number

Issue #2666, opened 2025-02-17
Distinguish complete but unpublishable features from incomplete drafts


ddbeck:
## Summary

In a few places, we've dealt with a compat key but can't publish feature data for one reason or another. We should consider adapting tools (like `stats.ts`) or our overall workflow (e.g., add a `unpublished` directory with different semantics) to give us a more accurate look at which parts of the platform we've addressed.

## Background

Prompted by this comment:

> Can we put this in drafts instead, and update our scripts to count keys in drafts as handled? (But excluding auto-updated spec drafts, I think.)

_Originally posted by @foolip in https://github.com/web-platform-dx/web-features/pull/2566#pullrequestreview-2608162593_

## Problem

The current `/features/drafts` folder mingles feature data in two different conditions:

- Feature data that is incomplete (chiefly, the generated features computed from specifications)
- Feature data that is complete, but blocked by non-feature description work (e.g., upstream data quality issues or schema changes)

This means it's difficult to find keys that are unmapped and use them to describe new/unfinished features.

## Examples

Here are some cases where this has come up:

- https://github.com/web-platform-dx/web-features/pull/2566 — We have to figure out how to modify the schema to help consumers not _show_ this data, but the keys are already dealt with in a useful way. Not publishing these features would do the trick.
- https://github.com/web-platform-dx/web-features/pull/2661 — We have to do upstream data work to make these features make sense with respect to caniuse.

ddbeck:
Here's my half-baked proposal:

It would be nice if we could create features that are still subject to all the usual rules about features (e.g., discouraged statuses, etc.) and fall under the usual workflows (e.g., BCD upgrades), without having to park them in draft PRs or the less-strictly managed `drafts` folder.

Features in the `drafts` folder would work as they are: representing unfinished, possibly malformed data (e.g., having deprecated keys in non-discouraged features).

Features in an `unpublished` folder would resemble "normal" features and be subject to the full data requirements of an ordinary feature. The one exceptional thing about them is that such features (and any references to them) be scrubbed from the published packages.

This might also be useful for our early features workflow (when we come up with one), where we could reserve early features' IDs while there's specification or implementation work happening but before there's sufficient developer interest to publish a given feature.

foolip:
Other than the generated drafts, it looks like we currently only have a single draft feature, File and directory entries. That means it's quite realistic to just use `drafts` and add whatever rules we want to it.

[...]

Issue #2341, opened 2024-11-26
Remove `image-function` but reserve its ID for future use


jamesnw:
[image()](https://github.com/web-platform-dx/web-features/blob/c07f170dd1c1d96461d24586525f72940d401c57/features/image-function.yml) has no implementations, but was tracked in BCD. https://github.com/mdn/browser-compat-data/pull/25190 removed it from BCD.

Unless and until `image()` has a real prospect of implementation (e.g., an intent, a bug with associated commits, etc.), then we shouldn't publish this feature. We ought to:

- withdraw it in the next major release
- move the feature description to drafts
- somehow reserve the feature ID (e.g., make sure there's an error if someone tries to make a non-draft that shadows the draft ID)

jamesnw:
@ddbeck FYI- this mirrors #2339 and will be another item to address with the next BCD version.

ddbeck:
After some discussion about this in the PR to remove it, https://github.com/web-platform-dx/web-features/pull/3186#issuecomment-3298375947, I came to the conclusion that this feature can persist, since it's still present in the spec.

While I would have preferred to not mint the feature until it was further along, there's no real harm to keeping it—or at least, no more harm than the spec itself keeping it. I'm going to close this issue. (Thank you for filing it though, James. It was helpful to work through this problem.)

Issue #2339, opened 2024-11-26
Remove `target-within` but reserve its ID for future use


ddbeck:
[`:target-within`](https://github.com/web-platform-dx/web-features/blob/c07f170dd1c1d96461d24586525f72940d401c57/features/target-within.yml) has no implementations, but was tracked in BCD. https://github.com/mdn/browser-compat-data/pull/25188 removed it from BCD.

Unless and until `:target-within` has a real prospect of implementation (e.g., an intent, a bug with associated commits, etc.), then we shouldn't publish this feature. We ought to:

- withdraw it in the next major release
- move the feature description to drafts
- somehow reserve the feature ID (e.g., make sure there's an error if someone tries to make a non-draft that shadows the draft ID)

ddbeck:
cc: @jamesnw @Elchi3, closing the loop on https://github.com/mdn/browser-compat-data/pull/25188

ddbeck:
Given conventions that developed since this issue was opened, I think we should keep this feature entry as discouraged. See https://github.com/web-platform-dx/web-features/pull/3353 for more details.

PR #3185, opened 2025-07-28
Require IDs to be unique (even with drafts)

ddbeck:
This is table setting for https://github.com/web-platform-dx/web-features/issues/2666 and https://github.com/web-platform-dx/web-features/issues/2339 and https://github.com/web-platform-dx/web-features/issues/2341.

Before this, it was possible to have a draft feature shadow the ID of a published feature. This change improves the safety of moving features from draft to published (and vice versa).

While we don't have draft groups or snapshots, I checked these too since it seemed weird and incomplete not too.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -28,6 +28,15 @@ const draft = Symbol('draft');
 // This must match the definition in docs/guidelines.md
 const identifierPattern = /^[a-z][a-z0-9]*(-[a-z0-9]+)*$/;
 
+// All identifiers (including drafts) must be unique with respect to their
+// siblings. These maps track them with respect to file names, for clearer error
+// mesesages.
+const uniqueIdMaps = {
+    features: new Map<string, string>(),
+    groups: new Map<string, string>(),
+    snapshots: new Map<string, string>(),
+}
+
 function* yamlEntries(root: string): Generator<[string, any]> {
     const filePaths = new fdir()
         .withBasePath()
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -38,6 +47,18 @@ function* yamlEntries(root: string): Generator<[string, any]> {
     for (const fp of filePaths) {
         // The feature identifier/key is the filename without extension.
         const { name: key } = path.parse(fp);
+        const pathParts = fp.split(path.sep);
+
+        // Assert ID uniqueness
+        for (const [pool, map] of Object.entries(uniqueIdMaps)) {
+            if (!pathParts.includes("spec") && pathParts.includes(pool)) {
```
foolip, line 54: Is the `!pathParts.includes("spec")` condition about the features/draft/spec/ directory? If so, should we instead give all drafts the same treatment? We set `data[draft]` in this loop so we might be able to use that in the condition.

foolip (approved):
Maybe this can be simplified by using `data[draft]`, but if not it's good as is!

index.ts
ddbeck: OK, I played around with this a little and remembered why I did this check in the first place:

It's possible for spec-generated drafts (those in `features/draft/spec`) to have names that shadow "real" feature names (e.g., `features/dom.yml` and `features/draft/spec/dom.yml`).

I figure the options here are something like:

- Do this special casing.
- Assign an additional (or different) symbol to generated specs (e.g., `draft[generated]`), then filter on that instead of the path. It's a couple extra lines of code, but uses the same strategy as `draft[]`, but I _do_ like consistency.
- Change the spec draft generator to avoid colliding with existing features (e.g., add a suffix we're unlikely use in real features, like `-spec`), then treat them all the same as you suggest.

Thoughts?

foolip: OK, let's go with just ignoring "spec" as you've done here.

PR #3186, opened 2025-07-28
Withdraw `image-function` and `target-within` to drafts

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/2339 and fixes https://github.com/web-platform-dx/web-features/issues/2341.


captainbrosset:
Made this target a new v3.0 branch.

ddbeck:
Philip asked me some good questions about these features (e.g., why not discouraged?). I didn't have good answers—I need to read up on these, to see if they're still in the spec (and thus dormant "could be" features) or dropped (and thus discouraged). I'm going to set this to draft while I read up on this.

ddbeck:
After looking into these in more detail:

## `target-within`

This is no longer specified, with this text:

> This specification previously defined a `:target-within` pseudo-class, analogous to [`:focus-within`](https://drafts.csswg.org/selectors-4/#focus-within-pseudo). It was removed in favor of `:has(:target)`, which should hopefully suffice to solve the same use-cases.

This would be straightforward to mark as discouraged (with `has` and `target` as alternatives). The one oddity would be that it was never implemented, so the eventual removal date (as in https://github.com/web-platform-dx/web-features/pull/3193) would be the date it was removed from the spec. It's a little dubious as a feature, but it has a mostly tidy story to tell about it.

Alternatively, we delete it and pretend it never existed. 🤷‍♂️ 

## `image-function`

This still exists in the specification. There is one known implementer bug (https://bugzil.la/703217), but it has no meaningful activity. This one seems like it could be a live feature, at least unless and until the spec drops it.

This one is harder to pretend never existed. It's actually in the spec, which is harder to evade. I think we ought to let this feature stand.

foolip:
I think we could delete `target-within`, but there's not much upside compared to just saying that it's discouraged. I'd say let's keep features and see if they get in the way after some time. Would there be some way to recognize `target-within` as a dead (never born) feature and avoid showing it on webstatus.dev?

ddbeck:
I've opened https://github.com/web-platform-dx/web-features/pull/3353 to discourage `target-within` and closed https://github.com/web-platform-dx/web-features/issues/2341 to retain `image-function`. I'm removing this PR from the v3 milestone and closing it.

> Would there be some way to recognize `target-within` as a dead (never born) feature and avoid showing it on webstatus.dev?

Not yet, but I think https://github.com/web-platform-dx/web-features/pull/3193 will provide the removal date, so there'll be data that webstatus.dev can use to exclude it.

Issue #1971, opened 2024-10-15
Baseline status regressions should require a note


ddbeck:
Blocked on https://github.com/web-platform-dx/web-features/issues/915.

If we have a Baseline status regression (see https://github.com/web-platform-dx/web-features/issues/1947), we ought to be transparent with data consumers and developers about why those.

Such notes probably ought to have, at a minimum:

- Some sort type information (e.g. `"status-regression"`), so tooling can distinguish them from other kinds of notes.
- The date of the status change (e.g., `"2024-10-15"`), so tooling doesn't needn't detect (and authors needn't figure out how to present) when a feature changed.
- An explanation of _why_ the status changed (e.g, "A serious bug in [browser] caused this feature to be downgraded from highly available to newly available.", "A data error caused this feature to be wrongly reported as Baseline.", etc.)

Issue #2957, opened 2025-05-12
Reconsideration: allow top-level await status regression


ddbeck:
## Summary

The [top-level await feature](https://github.com/web-platform-dx/web-features/blob/main/features/top-level-await.yml) ([dist](https://github.com/web-platform-dx/web-features/blob/main/features/top-level-await.yml.dist)) may regress from Baseline widely available ("Baseline high") to not Baseline (aka limited availability), due to a bug in Safari. This has been temporarily blocked with an editorial override.

This issue is meant to determine if we're to follow the compat data or maintain an editorial override, allowing the feature to remain Baseline high.

## Background

In Safari, simultaneous imports of a module with a top-level await fails with a `ReferenceError`. [This bug](https://bugs.webkit.org/show_bug.cgi?id=242740) has been present since September 2021, when Safari first shipped top-level await (this is also the feature's keystone release).

However, the bug was not reported until later. In July 2022 (about 10 months after top-level await shipped), the bug's first public report to WebKit was filed ([webkit bug 242740](https://bugs.webkit.org/show_bug.cgi?id=242740)).

In July 2023, this issue was [reported by developers to BCD](https://github.com/mdn/browser-compat-data/pull/26510).

In December 2024, we first published the top-level await feature  (https://github.com/web-platform-dx/web-features/pull/2417), using data that did not reflect the Safari bug.

On April 29, 2025, BCD updated the data and [marked Safari's support as partially implemented](https://github.com/mdn/browser-compat-data/pull/26510).

On May 2, 2025 the [update landed in web-features](https://github.com/web-platform-dx/web-features/pull/2921#discussion_r2071895308). I withheld a release, pending discussion on the issue. In May 8, 2025's WebDX call, we decided to apply an override while we discuss the issue further, judging that a full regression from Baseline high to non-Baseline merited a closer review. I applied that (temporary) fix in https://github.com/web-platform-dx/web-features/pull/2954.

Since Februrary 2024, a PR to fix this bug in WebKit has been open (https://github.com/WebKit/WebKit/pull/24122), though it's incomplete and unmerged. WebKit maintainers have commented on it, saying that there's work needed, but not scheduled, to fix this bug.

## Owner's group questions

This status must be resolved by the web-features owners group. If we cannot reach consensus, then we may put it to a vote of the owner's group to resolve. The main questions follow:

- **Should we let the (temporary) editorial override stand?**

[...]

PR #3187, opened 2025-07-28
Add Baseline regression notes

ddbeck:
This is a work in progress toward fixing https://github.com/web-platform-dx/web-features/issues/1971.

This structures the existing Baseline regression comments as notes. I'm looking for early feedback on whether this structure for notes would be useful.


features/streams.yml
```diff
@@ -2,12 +2,6 @@ name: Streams
 description: The streams API creates, composes, and consumes continuously generated data.
 spec: https://streams.spec.whatwg.org/
 group: streams
-# TODO: https://github.com/web-platform-dx/web-features/issues/1971
-# Status changed: https://github.com/web-platform-dx/web-features/pull/2358, https://github.com/web-platform-dx/web-features/pull/2491
-# 2024-12-19 — low → false — Regressed status to match Caniuse, which considers support beginning at BYOB shipping.
-# 2025-01-30 — false → high — Split BYOB into a separate "readable-byte-streams" feature. Linked that one to Caniuse.
-# References:
-# - https://caniuse.com/streams
```
ddbeck, line 10: I fussed over this for a while and ended up deciding to omit the note entirely. The first note would say that the feature regressed, the second note would be some new category (advance? unregression?) showing that we more or less reverted the second note. It seemed to me that the correct course of action in that scenario would be to withdraw the note, so that's what I've done.

features/content-visibility.yml
```diff
@@ -3,12 +3,14 @@ description: The `content-visibility` CSS property delays rendering an element,
...
-# Status changed: https://github.com/web-platform-dx/web-features/pull/2591
-# 2025-01-30 — low → false — Safari hides text behind `content-visibility: auto` from "Find…" in the page.
-# References:
-# - https://github.com/mdn/browser-compat-data/pull/25781
-# - https://bugs.webkit.org/show_bug.cgi?id=283846
+notes:
+  - category: baseline-regression
+    date: 2025-01-30
+    message: >
+      The `content-visibility` feature regressed from newly available to limited availability.
```
rviscomi: For any kind of extraordinary status change, it might also be useful to the tools consuming this info to encode it into the note:

```yaml
old_status: low
new_status: false
```

ddbeck: Good suggestion. I've made changes along these lines with fccb850f. It also meant that I could drop boilerplate text from each message.

features/popover.yml
```diff
@@ -2,12 +2,16 @@ name: Popover
...
+# notes:
+#   - date: 2024-09-18
+#     category: baseline-regression
+#     old_baseline_value: low
+#     new_baseline_value: false
+#     message: >
+#       Safari on iOS has a bug that prevents light dismiss (tapping outside the element to close it).
+#     citations:
+#       - https://bugs.webkit.org/show_bug.cgi?id=267688
+#       - https://github.com/mdn/browser-compat-data/issues/22927
```
ddbeck, line 13: When I opened this PR, I had written this note. However, the feature has since progressed. It occurred to me that we should not show irrelevant notes, so I commented it out. This is a rather new idea, so I'd like to delete this entirely, if we decide this is the right way to go.

Other options considered: some sort of "historic" flag on notes that are no longer relevant, or filtering historic notes from the published package. I figured both of those added complexity that wouldn't be very useful to anyone, least of all developers.

ddbeck:
I've updated this PR. Here's what's new:

- The schema has been added. To match the `description` field, there is now a plain-text `message` field and a formatted `message_html` in the built package.

- Regression notes that apply to a historic status are now forbidden. For example, if a feature regresses from high to low, then advances back to high, then the script that ingests the YAML treats that as an error. My working assumption is that such historic notes have no utility to web developers and ought to be dropped. This would result in the loss of one regression note. See my self-review for more detail and possible alternatives.

- There are now guidelines for adding regression notes. I wrote the guidelines to emphasize impact on developers. This led to…

- The regression notes have been revised. I tried to rewrite them to help developers make a decision about whether the regression is meaningful to their applications. This makes for much stronger notes, in my opinion.

- The script that ingests the YAML files has been updated to handle the new fields and do the requisite conversions.

This PR now passes the tests and is ready for (real) review. I'll share this with the WebDX chat on Monday.

features/popover.yml
captainbrosset: I almost feel like we should have a linting step in place which checks that `notes.new_baseline_value` is equal to `status.baseline`, and if not, deletes that particular note. I think keeping old notes about status regressions that no longer apply would be confusing. I don't think that web-features necessarily needs to keep track of the history of a feature. Its role is to document the platform as it is now.

features/streams.yml
captainbrosset: I tend to agree. If our consumers really only care about baseline _regressions_, then we don't need a note here. Unless we have reasons to believe that the history of a feature would be useful. Which I don't think we do.

index.ts
```diff
@@ -184,6 +191,15 @@ for (const [key, data] of yamlEntries('features')) {
         }
     }
 
+    // Ensure regression notes are still relevant
+    if (Array.isArray(data.notes)) {
+        for (const [index, note] of (data.notes as FeatureData["notes"]).entries()) {
+            if (note.new_baseline_value !== data.status.baseline) {
```
captainbrosset: Ah cool! I see you've actually already done what I suggested in an earlier comment.

types.ts
```diff
@@ -47,6 +47,8 @@ export interface FeatureData {
     compat_features?: string[];
     /** Whether developers are formally discouraged from using this feature */
     discouraged?: Discouraged;
+    /** Notes about this feature */
+    notes?: BaselineRegressionNote[];
```
captainbrosset: Do we really need this to be an array? I don't think we do, but I can see the point that not making it an array might, some day, block us, and require a major release to change it then.

ddbeck: Good question, @captainbrosset!

I think it's _very_ likely that we'll have other note types and notes may need to coexist. My inclination was to introduce one note type, then add more. But I recognize it looks a little odd _now_ because there really ought to be just one note at a time under the constraints presented here. But there's some likely follow up note types:

First, consider the mirror of a regression: a case where we've _prevented_ a feature from advancing (think of it as `"baseline-delay"` categeory or something). There are two open cases of this right now, see https://github.com/web-platform-dx/web-features/issues/3228. It's possible (though rare) that a feature could have regressed and be prevented from advancing.

Second, there are cases where we might choose to _not_ regress a feature, like https://github.com/web-platform-dx/web-features/issues/2957 (though it's still my preference to let it regress). It seems to me it might be wise for us to (publicly) acknowledge the bug, even if we do not believe it enough to regress the feature (something like `"known-incompatibility"`, though ). That said, this type of note would be more tenuous (we might consider attaching notes to supported version number, for instance).

I imagine there are still further note types (e.g., something like solicitations for input on potentially deprecating or removing a feature, for instance, as in the case of XSLT), but I haven't given them a ton of thought yet.

captainbrosset: I see. The `notes` property is broad enough that we might indeed use it for other things. Based on this, I'm fine on making this an array. However, if we do add other types of notes in the future, that will still be a breaking change, right? Cause the schema will change. That might be fine by the way.

ddbeck:
@rviscomi would you be interested in re-reviewing this, now that there's a proper schema and such for it? I'm especially curious to know if this is something you'd be interested in building something against soon or not.

If you want to try playing with the data, then these steps ought to work:

1. Check out this branch.
2. Run `npm install` and `npm run build:extended`. This generates `data.extended.json` in the root of the repository.

index.ts
```diff
@@ -155,6 +155,13 @@ for (const [key, data] of yamlEntries('features')) {
         data.description = text;
         data.description_html = html;
     }
+    if (Array.isArray(data.notes)) {
+        for (const note of data.notes as FeatureData["notes"]) {
+            const { text, html } = convertMarkdown(data.description);
```
rviscomi: This appears to copy the feature description into the note message. I think it should be something like this:

```suggestion
            const { text, html } = convertMarkdown(note.message);
```

features/font-variant-position.yml
```diff
@@ -2,9 +2,13 @@ name: font-variant-position
...
-# References:
-# - https://issues.chromium.org/issues/352218916
-# - https://bugs.webkit.org/show_bug.cgi?id=151471
+notes:
+  - date: 2024-10-15
+    category: baseline-regression
+    old_baseline_value: low
+    new_baseline_value: false
+    message: >
+      Chrome, Edge, and Safari do not implement font synthesis for missing superscript or subscript glyphs.
```
rviscomi, line 10: This message is partially contradicted by the support data for Safari:

https://github.com/web-platform-dx/web-features/blob/7fb07e22359003e05b147b9e40819b15585e3331/features/font-variant-position.yml.dist#L9-L10

features/createimagebitmap.yml
```diff
@@ -2,6 +2,15 @@ name: createImageBitmap
 description: The `createImageBitmap()` global method creates an `ImageBitmap` object from a source such as an image, SVG, blob, or canvas. An `ImageBitmap` object represents pixel data that can be drawn to a canvas with lower latency than other types, such as `ImageData`.
 spec: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap
 caniuse: createimagebitmap
+notes:
+  - date: 2025-08-11
+    category: baseline-regression
+    old_baseline_value: high
+    new_baseline_value: low
```
rviscomi: Should the `baseline_low_date` be `2025-08-11` accordingly?

https://github.com/web-platform-dx/web-features/blob/7fb07e22359003e05b147b9e40819b15585e3331/features/createimagebitmap.yml.dist#L6

rviscomi (changes requested):
I've tested this in Timebase (along with a patch to fix the note messages) to give an idea of how this might be integrated in developer tools/resources:

image

I've left a few comments in places where things don't appear as expected

schemas/data.schema.json
```diff
@@ -136,6 +136,60 @@
...
+                "description": "A short description of the cause of the regression as a plain text",
+                "type": "string"
+              },
+              "message_html": {
+                "description": "A short description of the cause of the regression as HTML",
+                "type": "string"
+              },
+              "new_baseline_value": {
+                "description": "The `baseline` status value after the regression",
+                "enum": ["low", false],
```
foolip: How much value do we get in practice from letting the types exclude one of the possible Baseline statuses for old and new? Can we just use the generic definition and thus sidestep the stuff that quicktype is generating funny types for?

foolip (approved):

features/font-variant-position.yml
ddbeck: I've sent https://github.com/mdn/browser-compat-data/pull/28834 to finally fix this.

features/streams.yml
ABOsalah866: https://github.com/mdn/browser-compat-data/issues/new/choose

ABOsalah866 (approved):

ddbeck:
This came up on today's call. There's interest in being able to explain Baseline regressions (e.g., for webstatus.dev feature subscriptions), but I have nervousness about committing to this specific schema.

One idea that came of the call would be to turn this into an _unpublished_ internal data mechanism, then experiment with the schema in the web features mapping repo, to make it safer to experiment with the data under real-world conditions.

When I have some time, I'll sync this up with main and refactor this as an internal field.

rviscomi:
Capturing a few motivating use cases that would benefit from this change:

- webstatus.dev gives developers the ability to subscribe to updates on web features, including regressions. This change would give subscribers more context behind why the feature regressed, with links to learn more.
- When a feature regresses, we reopen a feature's issue in the developer-signals repo. This change would allow us to provide more context to developers who might otherwise think it was reopened in error.
- Modern Web Guidance recommends using fallbacks whenever possible if a feature doesn't meet a Baseline target. If a feature regresses, the notes would be helpful context to include in the guides for coding agents to explain why fallbacks are suddenly necessary again.

PR #3193, opened 2025-07-30
Embellish `discouraged` data with more detail

ddbeck:
Toward https://github.com/web-platform-dx/web-features/issues/3210.

In meetings with MDN, @atopal and I have been talking about how to make it possible for MDN to replace MDN-authored and managed deprecation banners with banners automatically presented based on data from web-features. This PR explores two possible additions to the data to satisfy MDN's use case. Those two changes are:

- ~~A `pending_removal` field (to exaggerate warnings on mutation events and any future platform removals, if they arise). In my initial demonstration, this is a seldom-used boolean that shows active removal from the platform. Here, it applies to mutation events only.~~
- A `removal_date` field to exaggerate warnings on mutation events and any future platform removals, if they arise, or to indicate when a feature was removed, if it's gone already.
- A `reason` field, which briefly explains to developers _why_ the feature is deprecated. If we added this, it would become a required field.

At this point, I'm looking for feedback on the schema. I expect that we'd eventually need guidelines for writing the messages and setting the `removal_date` field—if you need those guidelines to judge the schema here, please let me know.


features/accessor-methods.yml
```diff
@@ -2,6 +2,7 @@ name: Accessor methods
 description: The `__defineGetter__()` and `__defineSetter__()` methods of objects bind a function to a property, which is called on setting or reading the property.
 spec: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.prototype-legacy-accessor-methods
 discouraged:
+  reason: TC39 included accessor methods in Annex B of the ECMAScript specification, which covers JavaScript features with "one or more undesirable characteristics and in the absence of legacy usage would be removed."
```
ddbeck, line 6: I don't _really_ know why this was stuffed into Annex B (the spec does not provide more specific text, unlike `getYear()`, so I just quote from the spec directly.

features/mutation-events.yml
```diff
@@ -4,8 +4,11 @@ description: Mutation events like `DOMSubtreeModified`, `DOMNodeInserted`, or `D
 spec: https://w3c.github.io/uievents/
 group: dom
 discouraged:
+  pending_removal: true
```
ddbeck: Here's our one removal pending. All pending removal features are discouraged features, but not all discouraged features are pending removal, thus a subordinate key and value here.

captainbrosset: Have we talked about what happens once features _are_ removed? Will we then delete the mutation-events.yml file and dist file? I suspect not, since we likely don't want to reuse the ID.

captainbrosset:
I like the new `reason` field. It will make it much easier for consumers of the data to show meaningful content, plus link to the source of truth for that content (instead of just showing a generic discouraged label with a link only).

features/mutation-events.yml
ddbeck: We have not, though perhaps we should do so at the same time. My (very briefly considered) guess is that we ought to have some feature-level annotation that says, more or less, "this is historic and you should not show it to developers by default." It would cover both features removed (like this one) but also features that never shipped widely in the first place (`import-assertions` would be a good example) or got abandoned before making it into a stable release.

Maybe something like `retired: true`, but I'm reluctant to add a single-use boolean. Something "stringy" would be nice, so we have the option of introducing nuance later, if needed. Maybe a sort of summary field? `condition: "normal" | "discouraged" | "pending-removal" | "historic"`. We'll need to think on this problem a little more.

ddbeck: After some discussion with MDN, it occurred to me that instead of a boolean, we could have something like `removal_scheduled_date: 2025-12-31`. The presence of this field could indicate that removal is pending as well as when we expect it to be removed.

MDN also suggested that the reason field contain a removal timeline in the case of pending removal features.

ddbeck: After writing `reason` fields for everything, I thought it would be easier to create and maintain this data if it were a `removal_date` that is either a future date (the expected timeline for removal) or today or a past date (for when it was actually removed). This would make maintaining the data a bit more straightforward: once the feature is gone, we don't have to go and _remove_ data.

LeoMcA:
Thanks for making this buildable locally: it works nicely on MDN: https://github.com/mdn/fred/pull/772

> I thought it would be easier to create and maintain this data if it were a removal_date that is either a future date (the expected timeline for removal) or today or a past date (for when it was actually removed).

This works well for our needs.

Rumyra:
@ddbeck I've had a scan through of the _reasons_ - I think these are all very good, I don't suggest we overthink these - the guidelines look good and they are already a huge improvement to what we have

ddbeck:
OK, I think this PR is ready for review, with an eye toward landing it soon. Here's what's changed recently:

- I added information about the new fields to the package `README` file.
- I added `reason_html` (like `description_html`).
- I added an assertion to author a `removal_date` on discouraged features with no support, then set the value on the features where it was required.

LeoMcA (approved):
Checked out and built locally for testing against MDN: works well, thanks for the addition of the `reason_html` field too

assertions.ts
```diff
@@ -26,4 +27,29 @@ export function assertValidFeatureReference(
...
+ * @param {string} id The ID of the feature to be checked
+ * @param {FeatureData} feature The ordinary feature data to be checked
+ */
+export function assertRequiredRemovalDateSet(
+  id: string,
+  feature: FeatureData,
+): void {
+  if (!feature.discouraged) return;
+  if (feature.discouraged.removal_date) return;
+  if (Object.keys(feature.status.support).length > 0) return;
```
ddbeck: Maybe some braces here, perhaps? (Check prettier config.)

packages/web-features/README.md
```diff
@@ -101,7 +101,11 @@ It has the following properties:
 - `discouraged` (optional): An object indicating that web developers should avoid using the feature.
   It has the following properties:
   - `according_to` (type: `string[]`): One or more links to a formal discouragement notice, such as specification text or an intent-to-unship
+  - `reason` (type: `string`): A brief, developer-focused description of why the feature is discouraged
+  - `reason_html` (type: `string`): A brief, developer-focused description of why the feature is discouraged, as an HTML string
```
foolip:
```suggestion
  - `reason` (type: `string`): A brief, developer-focused description of why the feature is discouraged.
  - `reason_html` (type: `string`): A brief, developer-focused description of why the feature is discouraged, as an HTML string.
```

PR #3236, opened 2025-08-20
Mentioned device-posture CSS media feature

foolip:


features/device-posture.yml
```diff
@@ -1,3 +1,3 @@
 name: Device posture
-description: The device posture API provides information about the physical posture of a device, such as whether a foldable device is folded or unfolded.
+description: The device posture API and the `device-posture` CSS media feature provide information about the physical posture of a device, such as whether a foldable device is folded or unfolded.
```
foolip: @ddbeck this isn't pedantically correct since the media feature doesn't "provide information", but I'm not sure how to phrase it.

ddbeck: Maybe something like this?

```suggestion
description: The device posture API and the `device-posture` CSS media feature reflect the physical posture of a device, such as whether a foldable device is folded or unfolded.
```

or this?

```suggestion
description: The device posture API and the `device-posture` CSS media feature represent the physical posture of a device, such as whether a foldable device is folded or unfolded.
```

foolip: I think I like "reflect" better, I'll go with that.

ddbeck (approved):

PR #3238, opened 2025-08-20
Redefine `open-closed` to be only about :open

foolip:
The :closed pseudo-class was removed from the spec:
https://github.com/w3c/csswg-drafts/pull/11326

This feature should be renamed to just `open` once we support renaming,
but it's OK that the `open-closed` identifier is taken because if
:closed is added in the future, it would be as `closed`, at least until
the very far future where they could merge again...

### DIFF
--- a/features/open-closed.yml
+++ b/features/open-closed.yml
@@ -1,7 +1,6 @@
-name: Open and closed selectors
-description: The `:open` and `:closed` CSS pseudo-classes match elements that have open and closed states, like `<details`, `<dialog>`, or `<select>`, based on their state.
+name: ":open"
+description: The `:open` CSS pseudo-class matches elements that have open states, like `<details`, `<dialog>`, or `<select>`, based on their state.
 spec: https://drafts.csswg.org/selectors-4/#open-state
 group: selectors
 compat_features:
-  - css.selectors.closed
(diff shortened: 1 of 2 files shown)


foolip:
This should be cleaned up after #91. @ddbeck are you keeping a list?

features/open-closed.yml
```diff
@@ -1,7 +1,6 @@
-name: Open and closed selectors
-description: The `:open` and `:closed` CSS pseudo-classes match elements that have open and closed states, like `<details`, `<dialog>`, or `<select>`, based on their state.
+name: ":open"
+description: The `:open` CSS pseudo-class matches elements that have open states, like `<details`, `<dialog>`, or `<select>`, based on their state.
```
ddbeck, line 2:
```suggestion
description: The `:open` CSS pseudo-class matches elements that have open states, like `<details>`, `<dialog>`, or `<select>`, based on their state.
```

ddbeck (approved):

ddbeck:
> This should be cleaned up after #91. @ddbeck are you keeping a list?

I sent https://github.com/web-platform-dx/web-features/issues/3239. I'm not keeping a central list, but I will start one soon.

PR #3246, opened 2025-08-22
Pin `<fencedframe>` support to the element itself

foolip:


ddbeck (approved):
This is another case where `compute_excluding` would be nice. Thank you!

PR #3253, opened 2025-08-22
Split `Object.hasOwn()` into its own feature

foolip:
This is the newest addition to Object and is still new enough that it
seems worth calling out as a standalone feature.


features/object-hasown.yml
```diff
@@ -0,0 +1,5 @@
+name: Object.hasOwn()
+description: The `Object.hasOwn()` static method checks whether an object has a given property. It's a more robust  alternative to the `Object.prototype.hasOwnProperty()` method.
```
ddbeck:
```suggestion
description: The `Object.hasOwn()` static method checks whether an object has a given property. It's a more robust alternative to the `Object.prototype.hasOwnProperty()` method.
```

PR #3254, opened 2025-08-22
Drop "initial support" qualifier for resource timing

foolip:
This is a living feature like most others, it's pinned to the initial
support but API surface keeps being added over time.


ddbeck (approved):
We should dump these everywhere, really—but this one can be done now. We need feature merging/folding for the remainder. Thank you!

foolip:
@ddbeck yes, I started out wanted to remove them everywhere, but concluded this was the only really straightforward one...

PR #3255, opened 2025-08-22
Describe the appinstalled event in the beforeinstallprompt feature

foolip:


ddbeck:
@captainbrosset perhaps you'd like to review this one? I think you're more likely to have a knowledgeable opinion than me. 😅

ddbeck:
Perhaps the name could be something like `beforeinstallprompt and appinstalled events`?

foolip:
I can keep the name "beforeinstallprompt" if that's widely recognized by web developers. Certainly I've heard about that event a lot more than appinstalled, but I wasn't sure if that's just because of how things developed in Chromium. (I've never used either event myself.)

foolip:
I also had to shorten the description a bit to pass the lint. I'd like to think I improved it, but another round of review is needed :)

captainbrosset (approved):

PR #3256, opened 2025-08-22
Replace many uses of the word "interface"

foolip:
In most places API is more natural and consistent with the rest of
web-features.

Sanitizer is aligned with the wording in parse-html-unsafe.yml.

For WebDriver "interface" isn't wrong, but protocol seems clearer.

### DIFF
--- a/features/dom-geometry.yml
+++ b/features/dom-geometry.yml
@@ -1,5 +1,5 @@
 name: DOM Geometry
-description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` APIs offer a way to represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
+description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` APIs represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
 spec: https://drafts.fxtf.org/geometry-1/
 status:
   compute_from: api.DOMMatrix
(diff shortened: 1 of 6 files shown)


features/dom-geometry.yml
```diff
@@ -1,5 +1,5 @@
 name: DOM Geometry
-description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` interfaces offer a way to represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
+description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` APIs offer a way to represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
```
ddbeck:
```suggestion
description: The `DOMMatrix`, `DOMPoint`, `DOMQuad` and `DOMRect` APIs represent points, rectangles, quadrilaterals and transformation matrices within JavaScript. They can be used in transformations in CSS, `<canvas>`, and SVG.
```

features/sanitizer.yml
```diff
@@ -1,5 +1,5 @@
 name: Sanitizer API
-description: The `parseHTML()` method for the `Document` interface and the `setHTML()` methods for the `Element` and `ShadowRoot` interfaces parse and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
+description: The `Document.parseHTML()` static method and the `setHTML()` method of `Element` and `ShadowRoot` parses and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
```
ddbeck: Something about this read a little off to me. In contrast to `Document`, a developer is unlikely to actually touch the `Element` and `ShadowRoot` interfaces directly, but manipulate instances. What do you think of this change?

```suggestion
description: The `Document.parseHTML()` static method and the `setHTML()` method of `Element` and `ShadowRoot` objects parse and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
```

ddbeck (approved):

ddbeck:
Approved, but with a couple of suggestions. Merge at your discretion.

features/sanitizer.yml
foolip: I think we should be consistent but don't feel very strongly. A search for "of `" reveals a zoo of styles, do you want to file an issue or send a PR to harmonize it?

PR #3258, opened 2025-08-23
Mention PerformanceObserver API in performance feature description

foolip:
As luck would have it, there was a perfect place to slot it in. The
previous description didn't make sense, because `performance` and
`Performance` are the same thing, one is the instance and one is the
interface.


ddbeck (approved):
Nice find!

PR #3262, opened 2025-08-24
Pin the SpeechRecognition API to its constructor

foolip:


features/speech-recognition.yml
```diff
@@ -3,6 +3,8 @@ description: The `SpeechRecognition` API converts audio into text using the devi
 spec: https://webaudio.github.io/web-speech-api/#speechreco-section
 group: speech
 caniuse: speech-recognition
+status:
+  compute_from: api.SpeechRecognition.SpeechRecognition
```
ddbeck, line 7: Seems like we should have the option of `compute_excluding` when we want to exclude a smaller fraction of the set of keys.

ddbeck (approved):

features/speech-recognition.yml
foolip: I'm not sure, that seems like it will require changing it every time a feature gets a small addition?

ddbeck: My impression from reviewing BCD upgrades is that most features are basically stable. It seems like it would get better/more attention when new keys appear (or historically-stable keys change status) if we don't exclude a large fraction of keys. But a sort of idle thought at the moment—I'm planning to run stats on the features soon-ish, to see how many "discarded" keys we've got.

PR #3266, opened 2025-08-25
Update moveBefore() spec and BCD keys

foolip:
Spec changes were landed here:
https://github.com/whatwg/dom/pull/1307
https://github.com/whatwg/html/pull/10657

### DIFF
--- a/features/move-before.yml
+++ b/features/move-before.yml
@@ -1,8 +1,11 @@
 name: moveBefore()
 description: The `moveBefore()` DOM method relocates a node while preserving its state. For example, you can move the active element without losing focus, move an animated element without resetting the animation, or move an iframe without reloading its content.
-spec: https://github.com/whatwg/dom/pull/1307
+spec:
+  - https://dom.spec.whatwg.org/#dom-parentnode-movebefore
+  - https://html.spec.whatwg.org/#preserving-custom-element-state-when-moved
 group: dom
 compat_features:
+  - api.CustomElementRegistry.define.connectedMoveCallback_lifecycle_callback
   - api.Document.moveBefore
   - api.DocumentFragment.moveBefore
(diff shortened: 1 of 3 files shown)


features/move-before.yml
```diff
@@ -1,8 +1,11 @@
 name: moveBefore()
 description: The `moveBefore()` DOM method relocates a node while preserving its state. For example, you can move the active element without losing focus, move an animated element without resetting the animation, or move an iframe without reloading its content.
-spec: https://github.com/whatwg/dom/pull/1307
+spec:
+  - https://dom.spec.whatwg.org/#dom-parentnode-movebefore
+  - https://html.spec.whatwg.org/#preserving-custom-element-state-when-moved
```
ddbeck:
```suggestion
  - https://html.spec.whatwg.org/multipage/custom-elements.html#preserving-custom-element-state-when-moved
```

features/move-before.yml
```diff
@@ -1,8 +1,11 @@
 name: moveBefore()
 description: The `moveBefore()` DOM method relocates a node while preserving its state. For example, you can move the active element without losing focus, move an animated element without resetting the animation, or move an iframe without reloading its content.
-spec: https://github.com/whatwg/dom/pull/1307
```
ddbeck, line 3: Would you mind removing the exception for this from [`scripts/spec.ts`](https://github.com/ddbeck/web-features/blob/efc2321eb4b1d3f4f9c102739fac551653129cd8/scripts/specs.ts#L124-L127)?

ddbeck (approved):
Approving, but if you wouldn't mind making some changes before merging. See comments. Thank you!

features/move-before.yml
foolip: Done!

PR #3268, opened 2025-08-25
Align Web Crypto support with caniuse

foolip:
This started out as an attempt to align the support with the
description, and pinning to encrypt/decrypt/sign/verify gets us very
close to what caniuse claims.

If we had the ability to selectively ignore notes from BCD, we would use
that mechanism here.


ddbeck:
It would be nice to some day do https://github.com/web-platform-dx/web-features/ so we can find these things more easily.

features/web-cryptography.yml
```diff
@@ -5,8 +5,22 @@ spec:
   - https://wicg.github.io/webcrypto-secure-curves/
 caniuse: cryptography
 group: security
+# Override the status to match caniuse. This is the same as one would get by
+# computing the status from api.SubtleCrypto.encrypt/decrypt/sign/verify, except
+# for Edge 12-18 where BCD has partial_implementation due to some unsupported
+# algorithms. However, the supported algorithms have changed over time, so it
+# seems more fair to treat it as supported in Edge 12-18.
```
ddbeck, line 12: I filed https://github.com/mdn/browser-compat-data/issues/27669 for this. It seems like BCD ought to break out all the algorithms as separate subfeatures. Not sure how much, if at all, you wish to prioritize such a clean up.

foolip: Good idea to break it into subfeatures, that would help us here. But doing so would require confirming support in all browsers, which would be a lot of work...

PR #3275, opened 2025-08-25
Add feature for interest invokers

mfreed7:
This feature describes several things:

- The `interestfor` global attribute ([spec PR](https://github.com/whatwg/html/pull/11006))
- The `interest-delay`, `interest-delay-start`, and `interest-delay-end` CSS properties ([spec](https://drafts.csswg.org/css-ui-4/#interest))
- The `:interest-source` and `:interest-target` CSS pseudo elements ([spec](https://drafts.csswg.org/selectors/#interest-pseudos))
- The associated behaviors (mostly in the HTML [spec PR](https://github.com/whatwg/html/pull/11006))

This is not yet a shipped feature in any browser, but work is underway in Chromium, and we are hoping to be able to ship this feature soon. I'm fine waiting to land this until it has shipped in a browser, but I wanted to get the PR set up early to work out any issues. I'm not at all sure I've created the new yaml file correctly, so please let me know what needs to be fixed.

Fixes https://github.com/web-platform-dx/web-features/pull/3275

### DIFF
--- a/features/interest-invokers.yml
+++ b/features/interest-invokers.yml
@@ -1,4 +1,4 @@
-name: 'Interest invokers'
+name: Interest invokers
 description: Interest invokers, registered by the `interestfor` HTML attribute, trigger events and actions on a target element when a user shows interest in the element, through behaviors such as hover, focus, or long-press. Pseudo-elements apply styles to sources and targets of interest.
 spec: https://github.com/whatwg/html/pull/11006
 group: html


features/interest-invokers.yml
```diff
@@ -0,0 +1,4 @@
+name: 'Interest invokers'
+description: The `interestfor` global HTML attribute adds "interest" behaviors to the element, such that when the user "shows interest" in the element, actions are triggered on the target element, such as showing a popover. The user agent handles detecting when the user "shows interest" in the element, via methods such as hovering the element with a mouse, hitting special hotkeys on the keyboard, or long-pressing the element on touchscreens. When interest is shown or lost, an `InterestEvent` is fired on the target, which have default actions in the case of popovers - showing and hiding the popover.
```
rviscomi: Please fix:

>Error: The description field in interest-invokers.yml is too long, 588 characters. The maximum allowed length is 300.

mfreed7: Ahh, didn't know there was a limit. Done.

features/interest-invokers.yml
```diff
@@ -0,0 +1,4 @@
+name: 'Interest invokers'
+description: The `interestfor` global HTML attribute triggers actions on a target element when the user shows interest (e.g., hover, focus, long-press). The user agent detects this interest, firing an InterestEvent on the target. This event has default actions, like showing or hiding popovers.
+spec: https://github.com/whatwg/html/pull/11006
```
rviscomi, line 3: Looks like the [tests](https://github.com/web-platform-dx/web-features/actions/runs/17953186118/job/51057798440) are also failing on this spec URL.

If that PR isn't expected to be merged any time soon, it's possible to add an [exception](https://github.com/web-platform-dx/web-features/blob/main/scripts/specs.ts#L24) but I'm not 100% sure on the correct process for that. Maybe go ahead and add it to the allowlist so the tests pass, and I'll defer to another reviewer for approval or to point you in a better direction.

mfreed7: Sounds good - done!

features/interest-invokers.yml
```diff
@@ -0,0 +1,4 @@
+name: 'Interest invokers'
+description: The `interestfor` global HTML attribute triggers actions on a target element when the user shows interest (e.g., hover, focus, long-press). The user agent detects this interest, firing an InterestEvent on the target. This event has default actions, like showing or hiding popovers.
```
ddbeck: (Assuming the description, as written, is accurate.) Optional style nits/suggestions:

```suggestion
description: The `interestfor` global HTML attribute triggers actions on a target element when the user shows interest (such as hover, focus, or long-press). If the browser detects this interest, it fires an `InterestEvent` on the target. This event has default actions, like showing or hiding popovers.
```

ddbeck:
Hi @mfreed7, sorry for the long delay on handling this PR. I would like to merge it sooner rather than later, since it appears to be shipping in at least one browser, but I have some questions about the scope of this feature:

- Do you intend for this feature to cover the entirety of the interest invokers concept? That is, the `interestfor` attribute, the `interest` and `loseinterest` events, the API (`InterestEvent`, `interestForElement`, etc.), and CSS (pseudos and `interest-delay-` properties)? Or do you mean to exclude some or all of this?

  This information is needed to get the right data out of browser-compat-data for this feature. But also this matters when it comes to creating other features and to allocate the detail in the description.

- The PR that added data for this API to browser compat data [says](https://github.com/mdn/browser-compat-data/pull/28469) the `interestfor` attribute is limited to `<a>`, `<button>`, `<area>`, and the SVG `<a>` elements and the current spec PR text seems to support this. Can you confirm one way or the other about the anticipated scope of the `interestfor` attribute?

mfreed7:
> Hi @mfreed7, sorry for the long delay on handling this PR. I would like to merge it sooner rather than later, since it appears to be shipping in at least one browser, but I have some questions about the scope of this feature:

No problem!

> * Do you intend for this feature to cover the entirety of the interest invokers concept? That is, the `interestfor` attribute, the `interest` and `loseinterest` events, the API (`InterestEvent`, `interestForElement`, etc.), and CSS (pseudos and `interest-delay-` properties)? Or do you mean to exclude some or all of this?

Chrome has shipped all of the above, in v142. And I think the entire "feature" should cover all of these things, since they all work together.

There is one holdout feature, [discussed here](https://github.com/w3c/csswg-drafts/issues/12437), for a `::interest-button` pseudo element. This part has not yet shipped in Chrome, because there are details that haven't been finalized. Once the details are final though, I'd also say this belongs under the one "Interest invokers" feature.

> * The PR that added data for this API to browser compat data [says](https://github.com/mdn/browser-compat-data/pull/28469) the `interestfor` attribute is limited to `<a>`, `<button>`, `<area>`, and the SVG `<a>` elements and the current spec PR text seems to support this. Can you confirm one way or the other about the anticipated scope of the `interestfor` attribute?

That's the correct scope - those four elements. I don't anticipate expanding the list of elements any time soon, because that runs into significant accessibility issues. E.g. a `<span>` is not focusable, so it's not clear how to trigger interest via the keyboard.

features/interest-invokers.yml
ddbeck: Thank you for the clarification on the scope of this feature entry. I'm going to suggest a new description that's a bit more general, so it doesn't implicitly rule out (for example) the CSS parts of the feature. Unfortunately, we do not have room for a description that covers all the nameable entry points, so we'll probably have to do a little handwaving.

What do you think of something like this? Feel free to vary this, if I'm off the mark somewhere.

```suggestion
description: Interest invokers, registered by the `interestfor` HTML attribute, trigger events and actions on a target element when a user shows interest in the element, through behaviors such as hover, focus, or long-press. Pseudo-elements apply styles to sources and targets of interest.
```

features/interest-invokers.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://github.com/whatwg/html/pull/11006
+group: html
```
ddbeck, line 4: Optional:

Data has appeared in BCD, so we could generate a status from upstream data with the following change. This would require merging main into this branch and re-running `npm run dist`. If that's too much hassle, let me know—I can modify your PR, if you'd rather.

```suggestion
group: html
compat_features:
  - api.HTMLAnchorElement.interestForElement
  - api.HTMLAreaElement.interestForElement
  - api.HTMLButtonElement.interestForElement
  - api.HTMLElement.interest_event
  - api.HTMLElement.loseinterest_event
  - api.InterestEvent
  - api.InterestEvent.InterestEvent
  - api.InterestEvent.source
  - api.SVGAElement.interestForElement
  - html.elements.a.interestfor
  - html.elements.area.interestfor
  - html.elements.button.interestfor
  - svg.elements.a.interestfor
```

features/interest-invokers.yml
mfreed7: I like it! Nice.

features/interest-invokers.yml
mfreed7: I'm great with this change, but I don't understand the intricacies of these yaml files enough to be confident in just committing this suggestion. So perhaps either tell me to just do that, or you modify the PR in the correct way for me?

ddbeck: I'll go ahead and apply and merge this (and merge main into this). It's a pretty rote thing, so I don't think there's a lot of value in having the back and forth over it. Thank you!

PR #3285, opened 2025-08-28
Add viewport segments feature

foolip:


features/viewport-segments.yml
```diff
@@ -0,0 +1,19 @@
+name: Viewport segments
+description: "The `viewport-segment-` CSS environment variables, the `horizontal-viewport-segments` and `vertical-viewport-segments` media features, and the `viewport.segments` API allow you to adapt your layout to devices where the display is split, such as foldable devices."
```
captainbrosset:
```suggestion
description: "The `viewport-segment-*` CSS environment variables, the `horizontal-viewport-segments` and `vertical-viewport-segments` media features, and the `viewport.segments` API allow you to adapt your layout to devices where the display is split, such as on foldable devices."
```

features/viewport-segments.yml
```diff
@@ -0,0 +1,19 @@
...
+spec:
+  - https://drafts.csswg.org/css-env-1/#viewport-segments
+  - https://drafts.csswg.org/css-viewport/#segments
+  - https://drafts.csswg.org/mediaqueries-5/#mf-horizontal-viewport-segments
+  - https://drafts.csswg.org/mediaqueries-5/#mf-vertical-viewport-segments
+compat_features:
+  - api.Viewport
```
captainbrosset, line 9: Isn't the `Viewport` interface supposed to be more general and the segments just a subpart of it? Not sure if other subparts do exist today, but I'm tempted to leave `api.Viewport` out of the list of keys, and just use `api.Viewport.segments`.
Same with `api.Window.viewport`.

But maybe it's only here for the segments API.

ddbeck: At least at the moment, [it's only exposing `segments`](https://dontcallmedom.github.io/webidlpedia/names/Viewport.html). My preference is that if `api.Viewport.segments` is assigned to a feature, then `api.Viewport` also be assigned to a feature (not necessarily the same one, though at least for now it's the only place that makes sense).

At some point I'd like to find the time to return to https://github.com/web-platform-dx/web-features/pull/2891, so we can allow these namespace-like interfaces to be shared across features.

captainbrosset: Makes sense.

foolip: @captainbrosset maybe you were thinking of the similarly named `VisualViewport` (`window.visualViewport`)?

In any case, `Viewport` was introduced for this feature, so it's the right feature for it, like how `api.Screen` is part of the `window.screen` feature, but the window management API also adds bits to `api.Screen`.

captainbrosset (approved):

features/viewport-segments.yml
```diff
@@ -0,0 +1,19 @@
+name: Viewport segments
+description: "The `viewport-segment-*` CSS environment variables, the `horizontal-viewport-segments` and `vertical-viewport-segments` media features, and the `viewport.segments` API allow you to adapt your layout to devices where the display is split, such as on foldable devices."
```
ddbeck: Sorry, nitpicky thing here. The [guidelines](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#descriptions) say to never use "made-up, idiosyncratic, or non-literal example code" and I think the asterisk here is in that category. Maybe something like this would be better?

```suggestion
description: "The CSS environment variables prefixed by `viewport-segment-`, the `horizontal-viewport-segments` and `vertical-viewport-segments` media features, and the `viewport.segments` API allow you to adapt your layout to devices where the display is split, such as on foldable devices."
```

captainbrosset: It's a little wordy though. I'm not super fan of the asterisk either, but I'd prefer if we could find something that's as compact.

Note that the viewport-unit-variants feature also uses asterisks:
> The `sv*`, `lv*`, and `dv*` CSS viewport units are relative to the smallest, largest, and current (dynamic) viewport size..

MDN uses asterisks for other CSS env variables in the docs: https://developer.mozilla.org/en-US/docs/Web/CSS/env#safe-area-inset-top

I wonder if we could make an exception to our rule for this?

The only other option I can think of is is also "made-up": `viewport-segment-[width|height|top|bottom|left|right]` (or something equivalent).

Or just "the `viewport-segment` environment variables".

Plain english could work: "the viewport segment environment variables". Maybe coupled with an example: "the viewport segment environment variables, such as `viewport-segment-width`".

foolip: I also don't like `viewport-segment-*`. Maybe we can just not name the CSS bits here and say something shorter like this?

> The viewport segment CSS environment variables and media features, and the `viewport.segments` API allow you to adapt your layout to devices where the display is split, such as on foldable devices.

captainbrosset: I like that.

foolip: Alright, I've made the edit.

PR #3288, opened 2025-08-29
Upgrade `compute-baseline` to use BCD 7.0.0

ddbeck:
This upgrades `compute-baseline` (and the project generally) to use `@mdn/browser-compat-data@7.0.0`. Since v7.0.0 does not contain any data changes over v6.1.5, this has no impact on the data, as expected.

This causes the following breaking changes (maybe use this text when making a release):

- The `RealSupportStatement` class has been removed. All BCD statements now have "real" values (that is, they do not contain `true` or `null` values where version strings might appear), so the distinction between "real" and "non-real" statements no longer has any meaning. Use the `SupportStatement` class instead.

- The `statement()` factory function has been removed. It serves no purpose without the real/non-real value distinction. Use `new SupportStatement()` instead.

### DIFF
--- a/packages/compute-baseline/src/browser-compat-data/index.ts
+++ b/packages/compute-baseline/src/browser-compat-data/index.ts
@@ -4,8 +4,6 @@ import { feature, Feature } from "./feature.js";
 import { query } from "./query.js";
 import { Release } from "./release.js";
 import {
-  RealSupportStatement,
-  statement,
   SupportStatement,
 } from "./supportStatements.js";
 import { walk } from "./walk.js";
@@ -17,9 +15,7 @@ export {
   feature,
   Feature,
   query,
-  RealSupportStatement,
   Release,
-  statement,
   SupportStatement,
-  walk,
+  walk
 };
--- a/packages/compute-baseline/src/browser-compat-data/supportStatements.ts
+++ b/packages/compute-baseline/src/browser-compat-data/supportStatements.ts
@@ -1,7 +1,6 @@
 import {
   FlagStatement,
   SimpleSupportStatement,
-  VersionValue,
 } from "@mdn/browser-compat-data";
 
 import { Browser } from "./browser.js";
@@ -54,7 +53,7 @@ export class SupportStatement {
    * A version string for the first supporting release. If the feature was never
    * supported, then it is `false`.
    */
-  get version_added(): VersionValue {
+  get version_added(): string | false {
     return this.data?.version_added;
   }
 
--- a/packages/compute-baseline/src/browser-compat-data/supportStatements.ts
+++ b/packages/compute-baseline/src/browser-compat-data/supportStatements.ts
@@ -1,6 +1,7 @@
 import {
   FlagStatement,
   SimpleSupportStatement,
+  VersionValue,
 } from "@mdn/browser-compat-data";
 
 import { Browser } from "./browser.js";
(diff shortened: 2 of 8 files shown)


packages/compute-baseline/src/browser-compat-data/index.ts
```diff
@@ -17,9 +15,7 @@ export {
   feature,
   Feature,
   query,
-  RealSupportStatement,
   Release,
-  statement,
   SupportStatement,
-  walk,
```
foolip, line 24: The trailing comma was nice, wasn't it?

scripts/find-troublesome-ancestors.ts
```diff
@@ -8,6 +8,10 @@ const c = new Compat();
 const needles = [];
 
 for (const feature of c.walk()) {
+  if (feature.id.startsWith("webextensions")) {
```
foolip:
```suggestion
  if (feature.id.startsWith("webextensions")) {
```

```suggestion
  if (feature.id.startsWith("webextensions.")) {
```

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -76,80 +43,16 @@ export class SupportStatement {
...
 
   /**
-   * Get the `version_added` value, or false if unset.
+   * Get the `version_added` value
    */
-  get version_added(): VersionValue {
-    if (this.data?.version_added === undefined) {
-      return false;
-    }
+  get version_added() {
```
foolip: Can you add the return type and update the docstring to say when it doesn't return a version? Or can data never be null below?

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -76,80 +43,16 @@ export class SupportStatement {
...
-    //   );
-    // }
-    if (value === null) {
-      throw new Error(
-        "`version_added` should never be `null`. This is a bug, so please file an issue!",
-      );
-    }
-
-    return value;
+  get version_removed() {
```
foolip: Return type?

foolip (approved):

packages/compute-baseline/src/browser-compat-data/index.ts
ddbeck: "Prettier" eh? 😂

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
ddbeck: Yes, I added the types (also addressing your other comments) and filled out the docstrings. All the getters now have docstrings. c06278e8991e48105f1f55ae529bbe19100120fb

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
ddbeck: See c06278e8991e48105f1f55ae529bbe19100120fb

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -33,23 +33,35 @@ export class SupportStatement {
...
+  /**
+   * An array of `FlagStatement` objects. If there are no flags, then the array
+   * is empty.
+   */
   get flags(): FlagStatement[] {
     return this.data?.flags ?? [];
   }
 
+  /**
+   * A `true` value, if support exists but may be incompatible with other
```
foolip: Trying to summarize https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#partial_implementation-general-usage-guidelines in a single sentence is hard, but I think that "incompatible with other implementations" is one of the less common reasons. Most common is, I think, large deviations from the spec.

No change is really necessary, but if you feel inspired to change something, please go ahead.

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
```diff
@@ -33,23 +33,35 @@ export class SupportStatement {
...
   }
 
   /**
-   * Get the `version_added` value
+   * A version string for the first supporting release. If the feature was never
+   * supported, then it is `false`.
    */
-  get version_added() {
+  get version_added(): string | false {
     return this.data?.version_added;
```
foolip: To ensure that null or undefined aren't returned:

```suggestion
    return this.data?.version_added ?? false;
```

packages/compute-baseline/src/browser-compat-data/supportStatements.ts
ddbeck: Good point. I applied d59c16e1, which describes partials as "irregular" (when I really wanted to say "weird" or "strange" but showed some restraint).

PR #3292, opened 2025-09-01
Fix assignment of `css.properties.page`

ddbeck:
This was mistakenly added to page selectors feature, instead of the feature for the `@page` at-rule.

### DIFF
--- a/features/page-selectors.yml
+++ b/features/page-selectors.yml
@@ -5,7 +5,6 @@ group:
   - selectors
   - print
 compat_features:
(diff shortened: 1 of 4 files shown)


features/page-selectors.yml
jamesnw:
```suggestion
description: The `:first`, `:left`, and `:right` pseudo-classes select pages based on their position in sequence after pagination.
```

I don't think this is a useful part of the description anymore.

jamesnw (approved):

PR #3294, opened 2025-09-01
Add script to inspect compat key notes

ddbeck:
This is the remainder of a couple of failed attempts to surface BCD's notes into dist files. I wasn't happy with the outcome of those experiments, so I present to you: a script that shows you some notes that relate to keys in web features. Unfortunately, it's too noisy for routine use. I'm not even sure if we should merge this, but I did want to prove the concept a little.

## Synopsis

This script prints a Markdown-formatted view of what you'd see (more or less) in a `dist` file, with one major embellishment: it lists notes from BCD underneath each browser compat data key.

You can try it by running `npx tsx ./scripts/inspect-feature.ts features/{id}.yml` or `npx tsx ./scripts/inspect-feature.ts features/{id}.yml.dist`, or with multiple such files. If you run it without any arguments, it'll print the `a` feature by default, which is a typical case.

Try `npx tsx ./scripts/inspect-feature.ts features/counter-style.yml` for a more useful case.

## Analysis

The script tries its best to omit notes that don't matter toward a Baseline status. For example, if a support statement in BCD has a prefix, then the notes for that statement aren't printed, since prefixed features are ineligible for Baseline. Similar filters apply to closed statements (i.e., have `version_added` and `version_removed` values) and statements with `alternative_name` or `flags`.

Even if you filter out compat statements that are seemingly useless, the majority of notes seem to be about things that characterize _historic_ behavior (e.g., notes that say "Before version X" or "Starting with version Y"). If you run the script `npx tsx ./scripts/inspect-feature.ts features/{whatever}.yml`, then you get lots of notes that just aren't very interesting. It's a needle in a haystack.

Additionally, lots of notes are duplicates: they're about the same engine bug in multiple browsers. Boring!

This suggests to me some alternate ways forward:

- Do something upstream about notes in BCD (e.g., add structural information to notes, to indicate when the _notes_ apply, not just the statements themselves).
- Do some sort of manual processing (e.g., assign an ID to each note, surface notes that we haven't made a decision about in dist files, hide ones that we think aren't interesting, promote notes to feature-level if they are interesting). We'd probably want to do this for new notes only, then try to do something about the backlog later. There's probably something fruitful here in taking notes snapshots from old BCD releases, to find notes that appeared inside a time horizon (30 months?).
- Do some heroics around parsing notes (e.g., try to guess when a note is still relevant based on the note text). This seems… risky to me. It might work better as a tool hinting manual processing of the backlog.


foolip (approved):
I've tried running this locally for a few features and it works. I agree that most of the things surfaced aren't that interesting, and we need to do more than simply collecting all notes into dist files or something.

I think this is still useful for debugging, so let's have it!

PR #3299, opened 2025-09-01
Add heading selectors

keithamus:
This adds the `:heading` and `:heading()` pseudo classes as specified in https://drafts.csswg.org/selectors-5/#headings, and added in BCD 6.1.1: https://github.com/mdn/browser-compat-data/pull/27517

Refs https://github.com/web-platform-dx/web-features/pull/3218 which updated BCD to 6.1.1. Perhaps it didn't add these as they're not in a released browser yet?


features/heading-selectors.yml
```diff
@@ -0,0 +1,7 @@
+name: Heading level pseudo-selectors
+description: The `:heading` and `:heading()` CSS pseudo-classes match headings of varying levels
```
captainbrosset:
```suggestion
description: The `:heading` and `:heading()` CSS pseudo-classes match headings of varying levels.
```

features/heading-selectors.yml
```diff
@@ -0,0 +1,7 @@
+name: Heading level pseudo-selectors
```
captainbrosset: How about something slightly shorter? 

```suggestion
name: Heading pseudo-classes
```

Does this still work for you?

features/heading-selectors.yml
```diff
@@ -0,0 +1,7 @@
+name: Heading pseudo-classes
+description: The `:heading` and `:heading()` CSS pseudo-classes match headings of varying levels.
```
ddbeck: I think it would be nice to be more specific here. Something like:

```suggestion
description: The `:heading` and `:heading()` CSS pseudo-classes match headings. The `:heading` pseudo-class matches any `<h1>` through `<h6>`. The `:heading()` pseudo matches…
```

However, I don't really know how to fill in the rest. I read the spec and I read the MDN page (which seem to conflict greatly) and I don't get it. I suspect there's some _history_ here, so giving a clue about what behavior we're talking about is pretty important.

keithamus:
> do we know anything about how these might be implemented across engines?

It is implemented behind a flag in Firefox nightly right now. WebKit are supportive of this and its related feature (https://github.com/WebKit/standards-positions/issues/523). So I am hopeful.

> If it seems very likely that they might land separately

I am unsure if it'll land separately, it seems unlikely, however these two features represent the ones found in BCD.

>  I read the spec and I read the MDN page (which seem to conflict greatly) and I don't get it.

The functional form underwent some minor changes. It used to be specced that you could use `An+B` notation e.g. `:heading(odd)`, `:heading(3n+1)` but it was altered to only accept a list of integers now, so `:heading(1, 3, 5, 7, 9)` and `:heading(1, 4, 7)` would be the equivalent selectors for those.

features/heading-selectors.yml
keithamus: Does this new description work?

features/heading-selectors.yml
```diff
@@ -0,0 +1,7 @@
+name: Heading pseudo-classes
+description: The `:heading` and `:heading()` CSS pseudo-classes match headings of varying levels. The `:heading` pseudo-class matches any heading (`<h1>` through `<h6>`). The `:heading(<integer>#)` selector matches any of the given levels, for example `:heading(1, 2)` will match headings level 1 and 2.
```
ddbeck: Tiny nits on the description (to avoid future tense, to avoid non-literal code, and align the example with other descriptions), but like this a lot. Thank you!

```suggestion
description: The `:heading` and `:heading()` CSS pseudo-classes match headings of varying levels. The `:heading` pseudo-class matches any heading (`<h1>` through `<h6>`). The `:heading()` selector matches any of the given levels. For example, `:heading(1, 2)` matches headings level 1 and 2.
```

PR #3300, opened 2025-09-02
Add caniuse links or comments where the (normalized) IDs match

foolip:
This covers all cases where the web-features and caniuse IDs are the
same if hyphens are removed. Some are "false friends" and comments were
added to avoid them being matched in the future.


features/accelerometer.yml
```diff
@@ -1,6 +1,7 @@
 name: Accelerometer
 description: The `Accelerometer`, `LinearAccelerationSensor` and `GravitySensor` APIs read the acceleration applied to a device in three dimensions, either including the effect of gravity, without its effect, or only its effect, respectively.
 spec: https://w3c.github.io/accelerometer/
+caniuse: accelerometer
```
captainbrosset, line 4: Caniuse says Chrome 67 and Edge 79.
We say Chrome 91 and Edge 91.

Is this a large enough discrepancy that we should worry about? If not, this seems good.

features/background-clip-text.yml
```diff
@@ -2,4 +2,5 @@
 name: "background-clip: text"
 description: "The `background-clip: text` CSS declaration draws the background underneath only the text in the element."
 spec: https://drafts.csswg.org/css-backgrounds-4/#background-clip
+caniuse: background-clip-text
```
captainbrosset, line 5: Our support data is quite different from caniuse's.

We have Chrome and Edge 120, which is OK. Caniuse shows earlier versions, but with vendor prefixes.

However, we show Firefox as unsupported, but caniuse shows support starting from Firefox 49.

As for Safari, we have 14, and they have 15.5, with a partial support note.

Even more confusing, caniuse shows:

Baseline (computed) Widely available across major browsers.

Even if our own feature is Limited Availability.

features/multi-column.yml
```diff
@@ -1,6 +1,7 @@
 name: Multi-column layout
 description: Multi-column layout flows an element's content across one or more columns in a single row, without affecting the `display` property of its children.
 spec: https://drafts.csswg.org/css-multicol-1/
+caniuse: multicolumn
```
captainbrosset, line 4: Caniuse is full of partial support notes for this feature. Our feature is Widely Available.
Should we look into these notes and decide if they're outdated, irrelevant, or if they should somehow regress the feature?

features/payment-request.yml
```diff
@@ -1,6 +1,7 @@
 name: Payment request
 description: The `PaymentRequest` API prompts the user to make a payment through the browser's user interface.
 spec: https://w3c.github.io/payment-request/
+caniuse: payment-request
```
captainbrosset, line 4: We're almost in agreement with caniuse on this one except that caniuse has this partial note for Chrome, Edge, and Safari: "Missing support for PaymentResponse.prototype.retry() method", which we don't seem to care about.

Support versions are old enough that we might want to just ignore this.

features/template-literals.yml
```diff
@@ -1,6 +1,7 @@
 name: Template literals
 description: Template literals are literals delimited with backtick (<code>`</code>) characters, allowing for multi-line strings, string interpolation with embedded expressions, and special constructs called tagged templates.
 spec: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-template-literals
+caniuse: template-literals
```
captainbrosset, line 4: We agree with caniuse on the fact that this is widely available, but caniuse has older versions than we do.
We might want to ignore this delta based on the fact that these are old versions anyway.

captainbrosset (commented):
I think we can easily merge all of the files which I didn't comment on in this review.

For the files which I did leave a comment on, I think we should at least investigate the deltas a bit more and decide if they're meaningful or not.

Perhaps one way forward would be to move the easy merges to another PR, and keep only the difficult ones (which need more digging) here.

@foolip want to do this?

features/background-clip-text.yml
ddbeck: It's unclear to me why caniuse shows a Baseline banner on https://caniuse.com/background-clip-text. Nothing should map to it. I'm going to put this on my agenda for the next time I talk to Alexis.

foolip:
I don't have time to make progress on this right now, anyone else feel free to clone the PR.

PR #3301, opened 2025-09-02
Add caniuse links where the IDs are different

foolip:
Follow-up to https://github.com/web-platform-dx/web-features/pull/3300 which only covered cases of (hyphenless) IDs matching.

The starting point was a bunch of changes suggested by Gemini, asked to
identify features that might be the same. The suggests were reviewed and
about half of them are good. Comments were added where the features
almost match but not quite.

The description of execCommand() is also updated to reflect that the
method sits on document, which is baked into the caniuse ID.


features/data-urls.yml
```diff
@@ -1,6 +1,7 @@
 name: Data URLs
 description: Data URLs are URLs that start with the `data:` scheme and which directly contain data rather than point to remote resources.
 spec: https://www.rfc-editor.org/rfc/rfc2397
+caniuse: datauri
 status:
   compute_from: http.data-url
```
ddbeck: To align statuses, we'd have to include `http.data-url.html_files` in the status computation. That would shift the Baseline date forward (2015→2020). Probably good?

features/navigation-timing.yml
```diff
@@ -1,6 +1,7 @@
 name: Navigation timing
 description: The `navigation` performance entry and the `PerformanceNavigationTiming` API measures navigation events, such as loading time or the number of redirects.
 spec: https://w3c.github.io/navigation-timing/
+caniuse: nav-timing
```
ddbeck: I think `nav-timing` more closely matches up with our (discouarged) `performancetiming` feature.

features/network-information.yml
```diff
@@ -1,6 +1,7 @@
 name: Network Information
 description: The `navigator.connection` API provides information about the network connection a device is using and fires events when the connection type changes.
 spec: https://wicg.github.io/netinfo/
+caniuse: netinfo
```
ddbeck: Our feature differs from caniuse quite a bit (showing headline support for Chrome and Chrome for Android but not Edge, while caniuse is green only for Chrome for Android). I think there's some kind of reconciliation needed between caniuse and BCD on this one.

features/share.yml
```diff
@@ -1,6 +1,7 @@
 name: navigator.share()
 description: The `navigator.share()` method invokes the device's native sharing mechanism and passes text, links, files, and other content to share targets.
 spec: https://w3c.github.io/web-share/
+caniuse: web-share
```
ddbeck, line 4: This one is fine, but it would be nice to send a PR to caniuse (or BCD, depending) to reconcile support on Chrome for Android (dist file shows support, caniuse is red).

features/spellcheck.yml
```diff
@@ -1,6 +1,7 @@
 name: spellcheck
 description: The `spellcheck` global HTML attribute sets whether the browser may check an element for spelling errors.
 spec: https://html.spec.whatwg.org/multipage/interaction.html#attr-spellcheck
+caniuse: spellcheck-attribute
```
ddbeck: We show this feature as Baseline high, while caniuse shows that no mobile browser fully supports this. Some sort of reconciliation is needed here.

ddbeck (changes requested):

captainbrosset:
@foolip this could be easily mergeable with the comments addressed, or with the corresponding file changes removed.
Do you want to pursue this?

features/navigation-timing.yml
foolip: I'll just revert this bit.

features/data-urls.yml
foolip: I'm not sure, for expedience I'll just revert this bit.

features/network-information.yml
foolip: Reverting.

features/share.yml
foolip: Hmm, maybe the passage of time resolved this? https://caniuse.com/web-share and https://caniuse.com/wf-share look the same for Chrome Android now.

features/spellcheck.yml
foolip: caniuse has this note:

> The partial support in mobile browsers results from their OS generally having built-in spell checking instead of using the wavy underline to indicate misspelled words. `spellcheck="false"` does not seem to have any effect in these browsers.

This is arguably an instance of https://github.com/web-platform-dx/web-features/issues/1038. I'll comment on that issue.

ddbeck: I think the conflict here means that we still can't map it (or else we have to accept an exceptional case where we show widely available over a table that… looks different).

ddbeck (approved):
I've merged `main` into this branch and reverted the change to `spellcheck` so we can merge the rest. Thank you, Philip!

PR #3308, opened 2025-09-03
Add the periodic background sync API

foolip:


features/periodic-background-sync.yml
```diff
@@ -1,6 +1,5 @@
-draft_date: 2024-09-09
-name: Web Periodic Background Synchronization
-description: TODO
+name: Periodic background sync
+description: The background synchronization API fires a recurring `periodicsync` event in a service worker. You can use this to update data in the background at regular intervals.
```
captainbrosset: Using "The background synchronization API ..." at the start here makes me think this feature is related to the background-sync feature.

I _think_ that's not the case and that Background Sync and Periodic Background Sync are two separate features that don't really share any API surface.

Maybe borrow some of the description from background-sync. Turn:
"The background synchronization API registers a service worker callback to run only when the device is no longer offline"
into:
"The periodic background synchronization API registers a service worker callback to run periodically ..."

foolip: The API surfaces don't overlap AFAICT, but certainly there's a similarity just as the name suggests.

It's not really a callback though, it's an event. The same is true of background sync, but I didn't want to fix that description in this PR.

foolip: Oh, I see what you're saying. I just forgot a word in the description. I'll add it.

captainbrosset: Yeah, sorry if I wasn't clear.

captainbrosset (approved):

features/periodic-background-sync.yml
foolip: You were very clear, but I was reading what I expecting to see, and not what I had actually committed :D

PR #3322, opened 2025-09-05
Update scroll-buttons.yml

ramiy:
Missing browser support at https://web-platform-dx.github.io/web-features-explorer/features/scroll-buttons/

### DIFF
--- a/features/scroll-buttons.yml.dist
+++ b/features/scroll-buttons.yml.dist
@@ -4,3 +4,25 @@
 status:
   baseline: false
   support: {}
+compat_features:
+  # baseline: false
+  # support:
+  #   chrome: "135"
+  #   chrome_android: "135"
+  #   edge: "135"
+  - css.selectors.scroll-button
+  - css.selectors.scroll-button.block-end
+  - css.selectors.scroll-button.block-start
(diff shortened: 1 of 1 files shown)


features/scroll-buttons.yml
```diff
@@ -2,3 +2,4 @@ name: ::scroll-button
 description: The `::scroll-button(<dir>)` CSS pseudo-element is a focusable button which scrolls its originating scroll container element in the specified direction.
 spec: https://drafts.csswg.org/css-overflow-5/#scroll-buttons
 group: scrolling
+compat_features: css.selectors.scroll-button
```
ddbeck: This needs to be an array and ought to include the descendant keys. You can copy them from the `draft` folder (you don't even need to change the draft—it'll be scrubbed automatically later):

https://github.com/web-platform-dx/web-features/blob/cdb7922e449df132381b576afb016a86ebc75697/features/draft/spec/css-overflow-5.yml#L9-L20

ddbeck (changes requested):
Hi @ramiy and thanks for opening this PR. Some suggestions in a line comment.

ddbeck (approved):

PR #3325, opened 2025-09-05
Add Custom Media Queries

ramiy:


features/custom-media-queries.yml
```diff
@@ -0,0 +1,7 @@
+name: Custom media queries
+description: Named aliases for longer and more complex media queries, deffined once in a `@custom-media` rule and used as reusable variables in multiple `@media` rules.
```
captainbrosset: Attempt to simplify a bit and make it more consistent with what we already have on the repo:

```suggestion
description: The `@custom-media` rule defines a named media query, which can then be reused in multiple `@media` rules.
```

features/custom-media-queries.yml
```diff
@@ -0,0 +1,7 @@
...
+spec: https://www.w3.org/TR/mediaqueries-5/#at-ruledef-custom-media
```
captainbrosset: Maybe link to the parent section:

```suggestion
spec: https://www.w3.org/TR/mediaqueries-5/#custom-mq
```

features/custom-media-queries.yml
```diff
@@ -0,0 +1,7 @@
...
+group: media-queries
+status:
+  baseline: false
+  support: {}
```
captainbrosset: Custom media queries have not been implemented yet, and are therefore not in BCD yet (the data source we use to compute browser support). The simplest thing to do in this case is leave this section out, but you can also add a comment for future maintainers, like so:

```suggestion
# Custom media queries are not in BCD. So, there are no BCD keys we can list here yet.
```

captainbrosset (dismissed):

captainbrosset:
Can you please also generate the dist file by running `npm install` and then `npm run dist features/custom-media-queries.yml`?
Thank you.

features/custom-media-queries.yml
ramiy: Not accurate, last year I merged CSS `@custom-media` at-rule to BCD - https://github.com/mdn/browser-compat-data/pull/24338

But later it was removed as BCD has a policy not to document unsupported features.

ramiy: Also, removing this causes errors:

image

and:

image

captainbrosset: No this should work. See this other feature as an example:
* .yml file: https://github.com/web-platform-dx/web-features/blob/main/features/gap-decorations.yml
* .dist file: https://github.com/web-platform-dx/web-features/blob/main/features/gap-decorations.yml.dist

Did you run `npm run dist features/custom-media-queries.yml` before running `npm run build`?

features/custom-media-queries.yml
```diff
@@ -0,0 +1,5 @@
+name: Custom media queries
+description: The `@custom-media` rule defines a named media query, which can then be reused in multiple `@media` rules.
```
ddbeck: Some nitpicky suggestions to follow our guidelines around defines/sets/etc. and passive constructions. Also, I thought it would be nice to mention CSS here, like we do for other at-rules.

```suggestion
description: The `@custom-media` CSS at-rule names a media query, which you can reuse in multiple `@media` rules.
```

features/custom-media-queries.yml
```diff
@@ -0,0 +1,5 @@
...
+spec: https://www.w3.org/TR/mediaqueries-5/#custom-mq
+group: media-queries
+# Custom media queries are not in BCD. So, there are no BCD keys we can list here yet.
```
ddbeck: I think a comment here with predicted BCD keys would be nice. Like this:

```suggestion
# Not in BCD yet, but expecting something like:
# - css.at-rules.custom-media
# - api.CSSCustomMediaRule
# - api.CSSCustomMediaRule.name
# - api.CSSCustomMediaRule.query
```

features/custom-media-queries.yml
```diff
@@ -0,0 +1,5 @@
+name: Custom media queries
+description: The `@custom-media` rule defines a named media query, which can then be reused in multiple `@media` rules.
+spec: https://www.w3.org/TR/mediaqueries-5/#custom-mq
```
ddbeck: To pass tests:

```suggestion
spec: https://drafts.csswg.org/mediaqueries-5/#custom-mq
```

features/custom-media-queries.yml
```diff
@@ -0,0 +1,5 @@
...
+group: media-queries
```
ddbeck, line 4: Since nobody has shipped this (yet), you need a `status` block, like this:

```suggestion
group: media-queries
status:
  baseline: false
  support: {}
```

features/custom-media-queries.yml
ddbeck: Oh I didn't see this discussion before and it had been a long time since I added a feature without keys. Not sure what the issue was here, but I'll push a commit to undo my "fix" in the thread above, then merge this. Thanks and sorry for the back and forth. 8a0789ca

PR #3328, opened 2025-09-08
Add corner-shape feature

foolip:


foolip:
Yes, I was on the fence about that. @noamr does one have to use `border-radius` in order for `corner-shape` to have any effect? In my testing that seemed to be the case, but I only tested for 30 second :)

noamr:
> Yes, I was on the fence about that. @noamr does one have to use `border-radius` in order for `corner-shape` to have any effect? In my testing that seemed to be the case, but I only tested for 30 second :)

Yes corner shape only works on top of border radius.

foolip:
Thanks @noamr! I made a small tweak to the description to mention `border-radius`.

noamr (approved):

Issue #3060, opened 2025-06-16
Publish extended JSON in main web-features package


ddbeck:
We should consider publishing the per-key compat statuses in the web-features package directly.

Increasingly, we've seen consumers skipping over the web-features extended JSON data in favor of `compute-baseline`, which is clunkier (and skips over any editorial overrides we might wish to apply), just to get at per-key statuses from BCD. Examples of this include the recent VS Code integration.

We ought to compare the size of the web-features package with and without the extended data, get feedback from existing consumers, and make a decision about whether to include it.

captainbrosset:
Thanks for filing @ddbeck. I thought we already had an issue for this, but looks like no.
Should we ping consumers on this issue directly?
@rviscomi can probably share some perspective from the VS Code hover card side of things.

ddbeck:
I was planning to talk to the MDN team and Alexis of caniuse about these directly (and report back here), since I already have standing meetings with them—though in both cases, I believe they're using the bare JSON release assets not the npm package.

Rick and I talked about this last week, which was part of the incitement for filing this issue, so I believe he's feeling positive toward the idea already (though I'd appreciate his explicit input here). We started to talk about a related idea, which would be convenience methods for looking up per-key statuses; I'll file another issue, if this one makes some progress.

And quickly here:

| Artifact           | Size (uncompressed) | Size (gzip)   |
|--------------------|---------------------|---------------|
| `data.json`          | 1307 kB             | 220 kB        |
| `data.extended.json` | 4124 kB (+216%)     | 361 kB (+64%) |

rviscomi:
I fully support this plan. While it's technically possible to pull down the extended JSON, it'd be much more convenient for tooling consumers to access it via the web-features package. Even better if there was a helper function similar to `getStatus` from compute-baseline.

For reference, here's how ESLint CSS uses it: https://github.com/eslint/css/blob/a035fa7c5b36845b73131e2d1a0b8c83f6c89ffe/tools/generate-baseline.js#L12

VS Code is actually using the `computeBaseline` function: https://github.com/microsoft/vscode-custom-data/blob/d17efbb5d4da22adcdbf3316b2f0e91e2eac7325/web-data/css/generateData.mjs#L382-L390

Having a single canonical function in web-features would help simplify the integration for tooling consumers.

ddbeck:
[...]

PR #3341, opened 2025-09-11
Merge data.extended.json into data.json

rviscomi:
Progress on #3060 

- generate `by_compat_key` data as part of the main `index` script for data.json
- continue making data.extended.json available in the GitHub release artifacts for backwards compatibility
- update docs and workflows

### DIFF
--- a/types.quicktype.ts
+++ b/types.quicktype.ts
@@ -153,8 +153,7 @@ export interface StatusHeadline {
    */
   baseline_low_date?: string;
   /**
-   * Statuses for each key in the feature's compat_features list, if applicable. Not available
-   * to the npm release of web-features.
+   * Statuses for each key in the feature's compat_features list, if applicable.
    */
   by_compat_key?: { [key: string]: Status };
   /**


packages/web-features/package.json
```diff
@@ -21,11 +23,15 @@
     "types.d.ts",
     "types.quicktype.d.ts",
     "index.js",
+    "extended.d.ts",
+    "extended.js",
     "data.json",
+    "data.extended.json",
```
ddbeck: Fundamentally, what I want to happen here is that the current `data.json` is equivalent to `data.extended.json`. I don't think we should have two JSON files.

scripts/build.ts
ddbeck: I expected to see this move into the root `index.ts` and for this script to get proportionally smaller. 😄

packages/web-features/extended.d.ts
ddbeck, line 1: Can we split the utility function work into a separate PR? I think there's already agreement to ship the extended JSON as the only JSON, so I could approve and merge that part fairly swiftly. I would be less inclined to do that for the utility functions.

(Briefly, I'm not sure this is the right API to expose and I'd like to retire `getStatus()` from `compute-baseline`, if we're going to provide something similar from the `web-features` package.)

ddbeck:
Oh, one more thing: I pretty sure `by_compat_key` ought to be a required key in the schema, when we land this.

https://github.com/web-platform-dx/web-features/blob/5e81ac2efd7c21b1a6bcb4f583d477b8f14d21dd/schemas/data.schema.json#L300-L308

packages/web-features/extended.d.ts
rviscomi: Sure. For context, it was proposed here: https://github.com/web-platform-dx/web-features/issues/3060#issuecomment-2997770658

rviscomi:
> Oh, one more thing: I pretty sure `by_compat_key` ought to be a required key in the schema, when we land this.

I tried this, but `by_compat_key` depends on `compat_features`, which is optional. If we really want it to be required, should `compat_features` be required too? We could initialize both to empty objects and arrays respectively, but I'm not sure that adds much value.

PR #3342, opened 2025-09-11
HTML switch control

ramiy:
HTML form control that implement native switcher functionality in a checkbox, using the `switch` attribute.

```html
<input type=checkbox switch>
<input type=checkbox switch checked>
```

Part of Safari 17.4: https://webkit.org/blog/15054/an-html-switch-control/

See also:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox#switch


captainbrosset:
Thank you for creating this, and sorry about the delay.

We will need a link to a specification or proposal here. Has Apple shared a proposal document somewhere about this?

Also, feature YAML files also need to be accompanied by a `.yml.dist` file, which you can generate by doing this locally:
`npm install`
`npm run dist features/switch-control.yml`

Once done, you can commit the new file and push again to this branch.

captainbrosset:
https://github.com/whatwg/html/pull/9546 seems to be the closest thing there is to a spec.

features/switch-control.yml
```diff
@@ -0,0 +1,8 @@
+name: HTML switch control
+description: HTML form control that implement native switcher functionality in a checkbox, using the `switch` attribute.
+group: forms
+status:
+  baseline: false
+  support:
+    safari: "17.4"
+    safari_ios: "17.4"
```
captainbrosset:
```suggestion
name: '<input type="checkbox" switch>'
description: The `switch` attribute, when present on an `<input type="checkbox">` HTML element, renders a switch control with an on and off state instead of a checkbox control.
group:
  - forms
  - html-elements
compat_features:
  - html.elements.input.switch

```

captainbrosset (changes requested):
We now have a BCD key for this control. Let's use it here.

captainbrosset:
@ramiy apologies for leaving this PR untouched for so long. Are you still interested and able to work on it? ~If so, please take a look at my suggestion, then also update your local branch (`git pull origin main` and `npm install`) and re-generate the dist file (`npm run dist`).~

~Oh, and I just realized, we need to spec too. The closest thing to a spec is: https://github.com/whatwg/html/pull/9546~

~Can you add `spec: https://github.com/whatwg/html/pull/9546` to the yml file?~

EDIT: I've committed the suggestions I made for the name, description, BCD key, and spec.

But, that's not all, since this isn't a valid spec, you'll also need to add an exception for it in `/scripts/specs.ts`, see the `defaultAllowlist` array.

Let me know if you're not planning on working on this, and I can finalize the feature for you instead.

features/switch-control.yml
```diff
@@ -0,0 +1,8 @@
+name: '<input type="checkbox" switch>'
+description: The `switch` attribute, when present on an `<input type="checkbox">` HTML element, renders a switch control with an on and off state instead of a checkbox control.
+group:
```
captainbrosset, line 4:
```suggestion
spec: https://github.com/whatwg/html/pull/9546
group:
```

captainbrosset (changes requested):

captainbrosset (approved):

Issue #3079, opened 2025-06-24
Naming scheme for in-development features


jgraham:
As we move towards having web-features earlier in the feature lifecycle, there's a tension between wanting to be specific about the which "version" of a feature we're talking about for the purposes not just of browser implementation support, but also for other sources of information about early-stage features e.g. standards-positions and intent emails.

Consider e.g. CSS Masonry. There have been multiple incompatible proposals for how this should work, with varying syntax and semantics, and some of these have shipped experimental implementations in browsers. Clearly we want the "final" web feature id to be something like `css-masonry`. But those proposals were contemporary, so there was never a good time to "bless" one of them with the `css-masonry` label, and using the label to refer to multiple incompatible things would be technically challenging, and also confusing.

Even when there are not multiple contemporary proposals for a single feature, there are often substantial changes that happen during the design process, even after one implementation has shipped something. This early design may be associated with one or more negative standards positions. However later changes to the design might cause the new proposal to be assessed as positive. In such cases it makes sense for the feature label to change so that we can see the difference between what was initially proposed, and what was finally adopted.

We discussed this problem internally at Mozilla, and propose the following process to help:
* Any feature which does not yet have broad consensus gets a `-proposal-<N>` suffix, where N is an integer. So we might have `css-masonry-proposal-1`, `css-masonry-proposal-2` and so on.
* The definition of consensus depends on the working mode of the relevant standards body, but for WHATWG it might be an approved PR, for W3C it might be a spec on the standards track, for TC39 it might be Stage 3, or whatever.
* Whenever a proposal undergoes substantial revision it gets a new proposal number. "Substantial" is a judgement, but it would include anything large enough to change a vendor's opinion on the proposal (e.g. by changing a standards position state), or anything with a compatibility impact. In this case the old proposal would be marked as deprecated.
* Once the feature has broad consensus, it would get the final name, and the final version of the proposal would be aliased to that (one still might need later adjustments e.g. if some implementations ship in stages, but that doesn't affect this process).

captainbrosset:
Thanks James for filing this.

Would you be able to please provide other examples for when this would have been useful? Are there existing web-features that ought to be renamed with the `-propopsal-<N>` suffix you are proposing?

Unless I'm misunderstanding, CSS Masonry isn't, I think, the best example for what you're trying to convey here. The various proposals only differ at the syntax level, but all browser vendors agree on what the feature should do and that it is ultimately needed. It's just a question of how.

[...]

PR #3345, opened 2025-09-15
clipboardchange event feature

captainbrosset:
The `clipboardchange` event is a new feature that's now implemented in Chromium, although only available as an origin trial for now.

Draft spec: https://w3c.github.io/clipboard-apis/#clipboard-event-clipboardchange
Chromestatus entry: https://chromestatus.com/feature/5085102657503232

Key advantage: avoid having to poll the clipboard to know if something has changed.


jgraham:
FWIW although this has been merged into a working draft, it's unclear how stable the feature is; looking at the git history it seems to have been part of a mass-merge in 2018, which I didn't investigate deeply, and looking at the issue tracker there are some issues like https://github.com/w3c/clipboard-apis/issues/228 which could lead to design changes. So this seems like a case where we could use the process for early stage features.

features/clipboardchange.yml
```diff
@@ -0,0 +1,8 @@
+name: clipboardchange
+description: The `"clipboardchange"` event of the Clipboard API fires when the user modifies the clipboard's contents.
```
foolip: Matching the style of other features describing events:

```suggestion
description: The `clipboardchange` event fires when the user modifies the clipboard's contents.
```

I tried to see which kind of object the event fires on, and it seems to be an element. But that seems weird since the clipboard is global. It seems like the event target is `navigator.clipboard`? If that's true, I think the description can say that too.

foolip (approved):
Can you link to https://github.com/web-platform-dx/web-features/issues/3079 in a TODO so that we revisit this when we decide on something for early features?

features/clipboardchange.yml
captainbrosset: Yes, the target is `navigator.clipboard`. See https://developer.chrome.com/blog/clipboardchange#the_clipboardchange_event.

features/clipboardchange.yml
```diff
@@ -0,0 +1,10 @@
+# TODO: Revisit this once we have a process to handle new proposals.
+#  See https://github.com/web-platform-dx/web-features/issues/3079
```
captainbrosset:
```suggestion
# See https://github.com/web-platform-dx/web-features/issues/3079
```

PR #3360, opened 2025-09-17
Add feature for `Map.prototype.getOrInsert()`

ddbeck:
I chose not to mention "upsert" at all because that's a historic artifact of the proposal, not the behavior of the implemented feature.


features/getorinsert.yml
```diff
@@ -0,0 +1,8 @@
+name: Map getOrInsert()
+description: The `getOrInsert()` and `getOrInsertComputed()` methods of `Map` objects gets a value, setting and getting a default value if needed.
```
captainbrosset:
```suggestion
description: The `getOrInsert()` and `getOrInsertComputed()` methods of `Map` objects get a value, setting and getting a default value if needed.
```

PR #3361, opened 2025-09-17
Disambiguate fetch-request-streams feature

jgraham:
Rewrite the fetch-request-streams feature to make it more clear what precise property it covers, and add a new fetch-body-stream feature to cover the `body` option to the `fetch` function supporting `ReadableStream` objects.

### DIFF
--- a/features/fetch-request-streams.yml
+++ b/features/fetch-request-streams.yml
@@ -1,5 +1,6 @@
-name: Fetch upload streams
-description: A `fetch()` request uploads a stream of data to the server when a request's `body` property is a `ReadableStream` object.
+name: Request body property
+description: A `ReadableStream` can be used as the body of a `Request` object.
 spec: https://fetch.spec.whatwg.org/#concept-body-stream
 compat_features:
   - api.Request.body
+  - api.Request.Request.request_body_readablestream
--- a/features/fetch-request-streams.yml.dist
+++ b/features/fetch-request-streams.yml.dist
@@ -7,7 +7,20 @@ status:
     chrome: "105"
     chrome_android: "105"
     edge: "105"
-    safari: "11.1"
-    safari_ios: "11.3"
 compat_features:
+  # baseline: false
+  # support:
+  #   chrome: "105"
+  #   chrome_android: "105"
+  #   edge: "105"
+  #   safari: "11.1"
+  #   safari_ios: "11.3"
   - api.Request.body
+
+  # ⬇️ Same status as overall feature ⬇️
+  # baseline: false
+  # support:
+  #   chrome: "105"
+  #   chrome_android: "105"
+  #   edge: "105"
+  - api.Request.Request.request_body_readablestream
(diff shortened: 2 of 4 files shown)


jgraham:
Based on https://github.com/web-platform-tests/interop/issues/1072#issuecomment-3312090681 another option here is to just move `api.Request.Request.request_body_readablestream` into `Fetch upload streams`, although it won't accurately reflect the fact that some `Request.body` usecases already work in Safari (but not Firefox).

jgraham:
OK, I've updated this to just make the more minimal change of requiring that the `Request body property` feature just requires both that you get a `body` property on `Request` and also that you can specify it in the constructor.

I'm not delighted with that, because it obscures the fact that part of the feature is already available in Safari, but it's better than the status quo.

features/fetch-request-streams.yml
```diff
@@ -1,5 +1,6 @@
-name: Fetch upload streams
-description: A `fetch()` request uploads a stream of data to the server when a request's `body` property is a `ReadableStream` object.
+name: Request body property
```
captainbrosset, line 1: The only thing I'm not 100% sure about is the name. Isn't it a problem that we're loosing the "stream" part of it? 
According to MDN, the `body` part of a request can either be null or a stream.
So, really, it's only used for streaming requests to a server, right?

What other things can it be used for?

features/fetch-request-streams.yml
```diff
@@ -1,5 +1,6 @@
...
+description: A `ReadableStream` can be used as the body of a `Request` object.
```
jgraham, line 2:
```suggestion
description: `Request` objects have a `body` property and it can be used as a `ReadableStream`.
```

ddbeck:
This came up on today's WebDX call. I have now dug into this and I think I figured out my source of confusion around this feature (it's mostly because the BCD is very messy and disorganized). Here's the story, I think:

- If I understand correctly, there has _always_ been `body` field in the `RequestInit` dictionary for `fetch()` and `Request()`. From fetch's inception (probably?), you could use the same `body` types as [XHR's `send()`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send#body).
- Later, `ReadableStream` support was added to `RequestInit`'s `body`.
- Another point of confusion here: `Request.prototype.body` was also added (see also: [tests](https://wpt.fyi/results/fetch/api/request/request-init-stream.any.html?label=experimental&label=master&aligned)), but it's a `ReadableStream` — even if the body was something else.

This suggests to me that BCD could use some fixes here. It ought to have these keys (bolded for this feature):

- `api.fetch.init_body_parameter`, supported since `api.fetch` has been supported and assigned to the main `fetch` feature
- **`api.fetch.init_body_parameter.readablestream`**, probably moved from `api.fetch.body_readablestream` and assigned to the feature in this PR
- `api.Request.Request.init_body_parameter`, supported since `api.Request.Request` has been supported and assigned to the main `fetch` feature
- **`api.Request.Request.init_body_parameter.readablestream`**, probably moved from `api.Request.Request.request_body_readablestream` and assigned to the feature in this PR

This also makes me think that the original name is basically right, but the description could be reworked to emphasize the _change_ in `body`, something like this:

> The `fetch()` method and `Request()` constructor accept a `ReadableStream` object as the request `body`, to stream upload data to the server.

ddbeck:
I've opened https://github.com/web-platform-dx/web-features/pull/4055 to succeed this PR. I'd welcome reviews!

PR #3367, opened 2025-09-19
Split out alpha and colorspace properties from input type=color

jgraham:
These are recent additions to the spec that don't affect the utility of the original input type=color and don't have the same level of cross-browser support.

It's unclear if `alpha` and `colorspace` should be a single feature; browsers could implement one without the other. However they were added in the same HTML PR, so this seems like a reasonable guess for now.


features/input-color-alpha.yml
```diff
@@ -0,0 +1,11 @@
+name: opacity and colorspace in '<input type="color">'
```
captainbrosset: I think we've previously named HTML attributes, so perhaps:

```suggestion
name: `alpha` and `colorspace` attributes for `<input type="color">`
```

features/input-color-alpha.yml
```diff
@@ -0,0 +1,11 @@
+name: `alpha` and `colorspace` attributes for `<input type="color">`
```
jgraham:
```suggestion
name: "`alpha` and `colorspace` attributes for `<input type=color>`"
```

features/input-color-alpha.yml
rviscomi: It's a minor inconsistency but this seems to be the first feature to use backticks in its name. Is it worth dropping them for consistency or should we consider adding them to other features? Or is it fine as-is?

captainbrosset: This was a mistake. I don't have it handy here but there's another PR to fix this.

rviscomi: Ah thanks, I see it now in https://github.com/web-platform-dx/web-features/pull/3377

PR #3368, opened 2025-09-22
Split highlightsFromPoint from the css highlight api

captainbrosset:
The `highlightsFromPoint` method was a later addition to the CSS Custom Highlight API. It did not ship with Chromium 105 or Safari 17.2. It's only now available in Chromium, starting with 140. And Firefox or Safari don't have it.

It was previously "hidden" from the highlight feature because we used to have a `compute_from` field to pin the status.

In this PR, I'm making `highlightsFromPoint` its own feature, as it deserves to be tracked separately, at least for now. In a while, once it's well supported, we can fold it back into the parent highlight feature.
By doing so, we no longer need to pin the highlight feature by using `compute_from`.


features/highlightsfrompoint.yml
```diff
@@ -0,0 +1,5 @@
+name: CSS.highlights.highlightsFromPoint()
```
jgraham: If I understand the guidelines correctly something like `"Custom highlights from point"` might be preferred as a name?

captainbrosset: In the case of this new method, I don't think there's a descriptive phrase that developers use yet. That's why I went for the method name instead. But I think we can change this to the shorter:

```suggestion
name: highlightsFromPoint
```

@ddbeck what do you think?

foolip:
I don't think we should be splitting features based on browser support unless there's some additional principle at play, like matching developer expectations.

Splitting only based on browser support will result in hundreds of features that we'd never consider standalone if they were supported at the same time as the rest of the feature.

@ddbeck has an "aspects" idea that might help here.

captainbrosset:
> I don't think we should be splitting features based on browser support unless there's some additional principle at play, like matching developer expectations.
> Splitting only based on browser support will result in hundreds of features that we'd never consider standalone if they were supported at the same time as the rest of the feature.

I disagree here because this feature had not been invented when the rest of CSS Custom Highlight shipped. It's a new capability of the API which appeared probably because the use case emerged after the CSS Custom Highlight API started to get used for real.

If we don't create a separate feature for `highlightsFromPoint`, how else can we track it in the project, and talk about it?

Over time, this additional capability will no longer be considered standalone. But I think it needs to be considered as such for a little while.

View Transitions is a good parallel here: we decided to split away [active view transition](https://web-platform-dx.github.io/web-features-explorer/features/active-view-transition/) and [view transition class](https://web-platform-dx.github.io/web-features-explorer/features/view-transition-class/). Following your logic, should they have instead been folded into the main view transitions feature?

jgraham:
I strongly agree with @captainbrosset here.

If we don't split up features to reflect browser support then we can't use web features to track the progress of that support. We either have to mislead developers into thinking things work that don't (and then consumers like MDN have to undo that by using BCD data directly), or we have to be conservative and tell them things that actually work interoperably are unusable because some later addition is not yet widely available.

It also makes web-features rather useless to browser vendors looking to track gaps in support, Interop project candidates, etc.

I realize we don't yet have a good way to merge features, but I think it's been a clear requirement from the start based on historical precedent (e.g. I remember spending a lot of time in meetings talking about the example of array iterator methods like `Array.forEach`, which today are simply considered a basic part of `Array`, but for a long time had patchy support that developers would need to know about).

foolip:
My concern is with using support status as a guide to what should be a separate feature because of the implications such a rule would have unless paired with some other requirements.

We have 1081 features and over 500 of those have 2 or more different support statuses in *.dist, determined:

```sh
for f in *.dist; do blocks=`grep -F '# baseline:' $f | wc -l`; echo $blocks $f; done | grep -vw 0 | grep -vw 1 | wc -l
```

And there are over 200 that have more than 1 different Baseline statuses, determined by:

```sh
for f in *.dist; do blocks=`grep -F '# baseline:' $f | sort | uniq | wc -l`; echo $blocks $f; done | grep -vw 0 | grep -vw 1 | wc -l
```

Many more features will have had a mix of Baseline statuses at some point in the past, and that's the kind of evolution new features will also go through.

To make it a bit more concrete, it's way to easy to find bits of features that could be broken out and could be given somewhat sensible descriptions, but that are really part of another feature and browser support is the only thing that would tempt us to split it out. The first three that I found with ease:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/volume
- https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize
- https://developer.mozilla.org/en-US/docs/Web/API/Selection/removeRange

I think small parts of other features such as these need some other approach than splitting features and later merging them together.

I'd prefer if @ddbeck can weigh in on this before we make a decision, and we can also discuss in a WebDX CG meeting.

I want to be clear that I'm certain that `highlightsFromPoint()` shouldn't be its own feature, it's just the implications of the rationale that I'm worried about.

foolip:
> We either have to mislead developers into thinking things work that don't (and then consumers like MDN have to undo that by using BCD data directly), or we have to be conservative and tell them things that actually work interoperably are unusable because some later addition is not yet widely available.

https://developer.mozilla.org/en-US/docs/Web/API/HighlightRegistry/highlightsFromPoint has the right status and it comes from the `by_compat_key` data. The only way that MDN could avoid this is if web-features never grouped BCD keys with different statuses, which would require an explosion of tiny features with all of the things a real feature has, like a description and spec link.

If there any consumers that mislead developers about support by using the feature's overall status to describe every part of it without using `by_compat_key`, then we should work with them to fix it.

> It also makes web-features rather useless to browser vendors looking to track gaps in support

For this I think the best approach would be to apply the filtering on the `by_compat_key` data and then list features that contain some interesting part within it. To remove the need for this would again require web-features to look entirely different than what it now does.

captainbrosset:
> I think small parts of other features such as these need some other approach than splitting features and later merging them together.

I don't think this PR is about splitting features so much as it is about creating new fine grained features for emerging things which, over time, will find a natural home somewhere else.

The `highlightsFromPoint` key was added very recently in a8902f361a1986057bb3cf667816839d1cb316de via a BCD update. This key did not exist at the time we created the Custom Highlight API feature in the repo, and we never made a conscious decision to add that key to that feature from the start. The simple reason being that this capability did not exist back then.
We only inherited it via the automatic BCD tagging mechanism.

(Looking at the referenced commit again, I think what must have happened is that the new key appeared in BCD, and got tagged with the feature ID. Because that regressed that overall feature's status, the person in charge of the BCD upgrade added a `compute_from` field to essentially hide the new key.)

In this PR, I'm making a conscious decision about where this key should live. Given this is a new capability that is distinct from the set of things that the CSS Custom Highlight API allowed you to do when I created that initial feature, it just makes sense for me to put the key into a new, separate, feature.
I honestly don't think I'd have done it otherwise even if the `highlightsFromPoint` method was supported on other browsers.

So, to clarify, even if the PR looks like I'm taking a key out of an existing feature and into another feature. This is not what's at play here.

> I'd prefer if @ddbeck can weigh in on this before we make a decision, and we can also discuss in a WebDX CG meeting.

I'm adding it to the agenda for tomorrow.

> I want to be clear that I'm certain that highlightsFromPoint() shouldn't be its own feature, it's just the implications of the rationale that I'm worried about.

I agree that `highlightsFromPoint()` shouldn't be its own feature, over time. But for the few months/years when features are considered new, and might evolve tremendously, via the process of standards discussions, I think our project ought to shine a bright light on them, rather than hiding them in larger, perhaps more stable and agreed upon, features.

PR #3369, opened 2025-09-22
getAllRecords feature

### LINKED ISSUE #3281: getAllRecords() feature
### Specification

https://w3c.github.io/IndexedDB/#dom-idbobjectstore-getallrecords

### Description

See:

* https://patrickbrosset.com/articles/2024-11-19-even-faster-indexeddb-reads-with-getallrecords/
* https://chromestatus.com/feature/5124331450138624

`getAllRecords()` is shipping in Chromium starting with 141. We should add a feature for it as it's kind of an important change to the way devs read records from IndexedDB.

The method can make it both more ergonomic, and faster, to read large amounts of data from IDB.

### Documentation

https://patrickbrosset.com/articles/2024-11-19-even-faster-indexeddb-reads-with-getallrecords/
https://microsoftedge.github.io/Demos/idb-getallrecords/

### Browser support

Chromium, starting with 141.

captainbrosset:
Fixes #3281.

The getAllRecords IDB feature speeds up read operations on large dataset because it returns both the records and their keys (no need for 2 read operations), and it allows reading in batch and in reverse order without using a cursor.

This feature is coming with Chromium 141.
BCD keys already exist, and MDN docs is coming soon.

See:
* https://patrickbrosset.com/articles/2024-11-19-even-faster-indexeddb-reads-with-getallrecords/
* https://chromestatus.com/feature/5124331450138624
* https://microsoftedge.github.io/Demos/idb-getallrecords/
* https://nolanlawson.com/2021/08/22/speeding-up-indexeddb-reads-and-writes/


features/getallrecords.yml
```diff
@@ -0,0 +1,9 @@
+name: getAllRecords
```
jgraham: I guess this is somewhat against the guidelines, but I'd put `IndexedDB` somewhere in the name because `getAllRecords` on its own could mean many different things on the platform.

captainbrosset: The following 3 resources use the phrase "IndexedDB getAllRecords":

* https://groups.google.com/a/chromium.org/g/blink-dev/c/zLHOayUfTcw
* https://developer.chrome.com/blog/chrome-141-beta#indexeddb_getallrecords_and_direction_option_for_getall_and_getallkeys
* https://github.com/mozilla/standards-positions/issues/1261

So we could do that here too.

```suggestion
name: IndexedDB getAllRecords
```

features/getallrecords.yml
```diff
@@ -0,0 +1,9 @@
+name: IndexedDB getAllRecords
```
foolip:
```suggestion
name: IndexedDB getAllRecords()
```

PR #3377, opened 2025-09-24
Revise `input-color-alpha` name an description

ddbeck:
Neither the name nor the description satisfy the guidelines, so I'm suggesting we revise this to at least satisfy those requirements.

That said, I don't think this makes sense as one feature. They're rather dissimilar in what they do. If this is still open when I return next week, I'm going update this PR to split this feature in two.


features/input-color-alpha.yml
```diff
@@ -1,5 +1,6 @@
-name: "`alpha` and `colorspace` attributes for `<input type=color>`"
-description: The ability to control the opacity of a color picked using `<input type="color">` and determine the colorspace of the selected color.
+name: >
+  alpha and colorspace attributes for <input type="color">
+description: The `alpha` and `colorspace` attributes for the `<input type="color">` HTML element are hints to the browser to show controls for transparency or to limit color selection to a color space.
```
jgraham, line 3: The attributes aren't just non-normative hints, as this description suggests. If you set the `alpha` attribute then the browser will serialize the color with opacity data. If you set the colorspace property then the browser will serialize the color in the format for the requested colorspace.

features/input-color-alpha.yml
```diff
@@ -1,5 +1,6 @@
-name: "`alpha` and `colorspace` attributes for `<input type=color>`"
-description: The ability to control the opacity of a color picked using `<input type="color">` and determine the colorspace of the selected color.
+name: >
+  alpha and colorspace attributes for <input type="color">
```
captainbrosset, line 2: For consistency with features like https://web-platform-dx.github.io/web-features-explorer/features/input-number/ or https://web-platform-dx.github.io/web-features-explorer/features/dialog-closedby/, maybe we should rename this to:

```suggestion
name: <input type="color" alpha> and <input type="color" colorspace>
```

features/input-color-alpha.yml
captainbrosset: How about this:

```suggestion
description: The `alpha` and `colorspace` attributes for the `<input type="color">` HTML element show controls for transparency and color space color selection in the color picker interface. Additionally, they determine the format of the selected value.
```

PR #3384, opened 2025-09-25
Remove backticks from <input type=color> alpha/colorspace name

foolip:
Markdown isn't supported in names, only in descriptions.


captainbrosset:
This is also being fixed in #3377. I'll close the present PR since 3377 is older and already has some discussions.

PR #3389, opened 2025-09-28
Update keys in alignment-baseline and anchor positioning

jamesnw:
This doesn't impact any dates, and slots keys alongside existing keys. It does look like there are some anchor positioning keys missing Chrome support in BCD, for instance `css.properties.position-area.top`, but at least this will update when BCD gets updated.


captainbrosset (approved):
Looks good to me, especially because no dates are changing.
Question for a possible follow-up PR: would any of these new BCD keys maybe represent a standalone feature we should be creating?

jamesnw:
> Question for a possible follow-up PR: would any of these new BCD keys maybe represent a standalone feature we should be creating?

Not for alignment-baseline. For anchor positioning, I don't think we should split yet. I think once we know what exactly has been shipped as part of Interop and it reaches baseline, we should look at essentially locking the feature at what is interoperable at that point. Then there will likely be a handful of smaller features that will be added in the coming months.

PR #3434, opened 2025-10-09
Correct ommissions from the "JavaScript" group

jugglinmike:


features/regexp-compile.yml
```diff
@@ -1,6 +1,7 @@
 name: RegExp compile()
 description: The `compile()` method of `RegExp` objects recompiles an existing regular expression object using a new pattern and flags.
 spec: https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-regexp.prototype.compile
+group: regexps
```
foolip, line 4: Should this be plural? I would say yes, to not use the same ID as the feature. There's no lint for this, but we've done it like this in a bunch of other cases, like promise/promises.

Issue #3239, opened 2025-08-20
Give the `:open` pseudo a better ID than `open-closed`


ddbeck:
`:closed` was dropped from the spec, so fixing the ID would make this less confusing.

https://github.com/web-platform-dx/web-features/pull/3238

Blocked by #91

foolip:
There are many things on the platform called `open`, so we could call this `open-pseudo`, but just `open` isn't taken yet, so just `open` would be more consistent with what we've done in the past.

nt1m:
open-pseudo-class or pseudo-class-open is what I would suggest

foolip:
https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#identifiers are the guidelines we have, and we have these existing IDs that end with "-pseudo":

- column-pseudo, which could be just column, but there is a `columns` CSS property so I think disambiguating makes sense
- dir-pseudo, because there's also a `dir` attribute
- scope-pseudo, because there's also an `@scope` rule

We have many IDs like autofill and default which are just named after `:autofill` and `:default`.

In this case I think we should consider that we might add `:closed` later, and we'd want consistency. There are many things on the platform that have `open` or `closed` as part of their syntax, so I think that "open-pseudo" would be the winner here.

foolip:
I've filed https://github.com/web-platform-dx/web-features/issues/3436 with more cases to consider.

nt1m:
I don't like pseudo because there's pseudo-classes and pseudo-elements, but I don't feel too strongly.

PR #3435, opened 2025-10-10
Rename `open-closed` to `open-pseudo`

foolip:
Fixes https://github.com/web-platform-dx/web-features/issues/3239.


features/open-pseudo.yml
ddbeck, line 1: I guess we should have a pretty good idea about the `-pseudo` versus `-pseudo-class` suggestion discussed in https://github.com/web-platform-dx/web-features/issues/3436#issuecomment-3397343989 before proceeding. Otherwise, I like this.

features/open-closed.yml
```diff
@@ -1,6 +1,2 @@
-name: ":open"
-description: The `:open` CSS pseudo-class matches elements that have open states, like `<details>`, `<dialog>`, or `<select>`, based on their state.
-spec: https://drafts.csswg.org/selectors-4/#open-state
-group: selectors
-compat_features:
-  - css.selectors.open
+kind: moved
+redirect_target: open-pseudo
```
captainbrosset, line 2: @ddbeck  just a thought (not blocking here): what happens if we later decide to change the ID again? Perhaps to open-pseudo-class? Would we just change this value to avoid multiple chained redirects? And if so, do we have an automated check for this kind of things (to avoid chained redirects)?

ddbeck: > @ddbeck just a thought (not blocking here): what happens if we later decide to change the ID again? Perhaps to open-pseudo-class? Would we just change this value to avoid multiple chained redirects?

Yes, that's correct. In the documentation, we promise to consumers that redirects aren't chained and do not create cycles.

> And if so, do we have an automated check for this kind of things (to avoid chained redirects)?

Yes, we guard against this here:

https://github.com/web-platform-dx/web-features/blob/b9d34ba7c7ae79c3dc9be0c77affca0468dde52e/assertions.ts#L4-L13

https://github.com/web-platform-dx/web-features/blob/b9d34ba7c7ae79c3dc9be0c77affca0468dde52e/index.ts#L220-L232

ddbeck (approved):

captainbrosset:
This looks mergeable. @foolip are there any changes you'd like to make before we merge?

PR #3439, opened 2025-10-10
Publish the WebGPU subgroups feature

foolip:
The description was shortened, taking some inspiration from
https://developer.chrome.com/blog/new-in-webgpu-134.


ddbeck (approved):
I'll merge this now and open a PR for the group as a follow up.

Issue #2647, opened 2025-02-13
Relax discouraged feature spec URL validation


ddbeck:
In many PRs for discouraged features ([example](https://github.com/web-platform-dx/web-features/pull/2641/)), we have add an old spec URL to the allowlist because browser-specs doesn't record it (for good and obvious reasons). This is less than optimal because the reason and URL are very unlikely to ever change (i.e., we'll never have cause to _remove_ the URL from the allowlist).

We should consider relaxing the validation for discouraged features. Perhaps we can do something simple (such as periodically confirming that the URLs don't 404 or redirect) to confirm that the URLs are still, in some sense, "good" but not as rigorous as browser-spec's process.

ddbeck:
https://github.com/web-platform-dx/web-features/issues/3470 is related to this, but for features at the other end of the process.

PR #3449, opened 2025-10-14
Adds the `supports-at-rule` feature

diekus:
See:

- https://chromestatus.com/feature/5110744177836032
- https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/AtRuleFeatureDetection/explainer.md

### DIFF
--- a/features/supports-at-rule.yml
+++ b/features/supports-at-rule.yml
@@ -2,4 +2,3 @@ name: at-rule()
 description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)` checks if the `@starting-style` at-rule is supported."
 spec: https://drafts.csswg.org/css-conditional-5/#typedef-supports-at-rule-fn
 group: css
-


features/supports-at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: at-rule
+description: "The `at-rule` feature detection is an expansion to feature detection in CSS that allows authors to detect support for at-rules. Example: `@supports at-rule(@starting-style)`"
+spec: 
+  https://www.w3.org/TR/css-conditional-5/
+  https://github.com/w3c/csswg-drafts/pull/12945
+  https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/AtRuleFeatureDetection/explainer.md
```
captainbrosset: I'm the one who asked Diego to add all 3 URLs here, but we should choose one. @ddbeck do you think it's better to link to the spec (even if doesn't have the changes for this feature in it yet), the PR that's making the changes, or the explainer?

features/at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: at-rule
+description: "The `at-rule` feature detection is an expansion to feature detection in CSS that allows authors to detect support for at-rules. Example: `@supports at-rule(@starting-style)`"
```
captainbrosset:
```suggestion
description: "The `at-rule()` CSS function, when used with the `@supports` at-rule, detects if an at-rule is supported. For example `@supports at-rule(@starting-style)`" applies styles only if the `@starting-styles` at-rule is supported.
```

I tried simplifying the description, but ended up using the word at-rule 3 times, and that's very confusing :)

Happy to hear other takes on it, but I do think we need to clarify what "an expansion to feature detection" means.

features/at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: at-rule
```
captainbrosset:
```suggestion
name: at-rule()
```

features/at-rule.yml
captainbrosset: Maybe this:

```suggestion
description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)`" checks if the `@starting-style` at-rule is supported.
```

features/supports-at-rule.yml
captainbrosset: Looking at other features, it seems like we could choose between pointing to the explainer or the PR. I would vote for the PR because this way it's easy to later check if the PR has been merged, and update the link to the proper spec.
If we link to the explainer, it's maybe more useful for developers, but there's no easy sign to know that an explainer has graduated to a spec, and which one.

So let's perhaps use the PR URL only.

But, also, the build will fail if we do this. To make the build happy, please also add the PR URL to the `defaultAllowlist` array in /scripts/specs.ts, with a reason why the PR is allowed (e.g. because that's the closest thing we have to a spec).

foolip: Of the 3 links suggested, I think we should only link the https://github.com/w3c/csswg-drafts/pull/12945. That's something that a script can look at and determine when the PR is merged and the spec URL can be updated.

@ddbeck has proposed that we should use `proposal` instead of `spec` for this situation, and that would avoid the `defaultAllowlist` problem.

features/at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: at-rule()
```
ddbeck: I think this needs a different ID, to avoid confusion with the concept of an at-rule in general. An ID like `supports-at-rule` (thinking that this might some day fold into `@supports`) might be a good alternative.

features/supports-at-rule.yml
ddbeck: Yes, I like the idea of linking to, in order of preference:

1. An actual spec known to `browser-specs`
2. A PR against an actual spec known to `browser-specs`
3. Any other PR
4. Any other URL

~~https://github.com/web-platform-dx/web-features/issues/2647 exists to track this idea though it doesn't mention the `proposal` behavior~~. I'll ~~put a comment there~~ file a new issue. (edit: no it doesn't, wrong thing)

ddbeck: I created https://github.com/web-platform-dx/web-features/issues/3470

features/at-rule.yml
diekus: That's a great suggestion. I will update the name of the ID to futureproof.

diekus:
I've changed the ID of the feature, renamed the files, the description is now clearer and regenerated the dist.yml file.

cc @captainbrosset

features/supports-at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: supports-at-rule()
```
ddbeck: Sorry, I didn't explain myself here very well. The `name` before was fine with me, it was only the ID/filename that needed to change.


```suggestion
name: at-rule()
```

features/supports-at-rule.yml
```diff
@@ -0,0 +1,7 @@
+name: at-rule()
+description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)` checks if the `@starting-style` at-rule is supported."
+spec: 
+  https://www.w3.org/TR/css-conditional-5/
+  https://github.com/w3c/csswg-drafts/pull/12945
+  https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/AtRuleFeatureDetection/explainer.md
```
captainbrosset:
```suggestion
  https://github.com/w3c/csswg-drafts/pull/12945
```

features/supports-at-rule.yml
```diff
@@ -0,0 +1,5 @@
+name: at-rule()
+description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)` checks if the `@starting-style` at-rule is supported."
+spec: 
+  https://github.com/w3c/csswg-drafts/pull/12945
```
captainbrosset:
```suggestion
spec: https://github.com/w3c/csswg-drafts/pull/12945
```

captainbrosset (approved):

captainbrosset:
@foolip @ddbeck I've approved this after changing the spec URL. Any more comments from your end before we merge?

features/supports-at-rule.yml
```diff
@@ -0,0 +1,4 @@
+name: at-rule()
+description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)` checks if the `@starting-style` at-rule is supported."
+spec: https://github.com/w3c/csswg-drafts/pull/12945
```
ddbeck: I took so long to get to this that the PR merged 😅


```suggestion
spec: https://drafts.csswg.org/css-conditional-5/#typedef-supports-at-rule-fn
```

ddbeck (changes requested):
One suggestion for the spec URL (the PR merged). This also needs to be formatted to pass the linter, `npm run format` and a commit should fix it.

features/supports-at-rule.yml
```diff
@@ -0,0 +1,4 @@
+name: at-rule()
+description: "The `at-rule()` function, when used with `@supports`, checks if a CSS at-rule is supported. For example `@supports at-rule(@starting-style)` checks if the `@starting-style` at-rule is supported."
+spec: https://drafts.csswg.org/css-conditional-5/#typedef-supports-at-rule-fn
+group: css
```
captainbrosset, line 4:
```suggestion
group: css

```

PR #3454, opened 2025-10-14
Add Core Web Vital Metrics to descriptions

tunetheweb:
Event timing is now more commonly used for INP than FID (FID has been deprecated as a metric).

Layout shift API is used to calculate CLS.


features/event-timing.yml
```diff
@@ -1,5 +1,5 @@
 name: Event timing
-description: The `event` and `first-input` performance entries and the `PerformanceEventTiming` API measures the latency of user input events, such as mouse clicks or keypresses, and the latency of the first user input. First Input Delay (FID) is a common metric for perceived responsiveness.
+description: The `event` and `first-input` performance entries of the `PerformanceEventTiming` interface measures the latency of user input events, such as mouse clicks or keypresses. Interaction to Next Paint (INP) is a common metric for perceived responsiveness calculated from the Event Timing API.
```
ddbeck: Some suggestions here to maintain consistency with other features.

Second, since the name of the metric changes, I think reworking the second sentence make the connection to the API itself becomes important and deserves emphasis. It's also consistent with entries like `largest-contentful-paint`.

```suggestion
description: The `event` and `first-input` performance entries and the `PerformanceEventTiming` API measures the latency of user input events, such as mouse clicks or keypresses. They're used to calculate Interaction to Next Paint (INP), a common metric for perceived responsiveness.
```

features/layout-instability.yml
```diff
@@ -1,5 +1,5 @@
 name: Layout instability
-description: The `layout-shift` performance entry and `LayoutShift` API measures the layout stability of web pages based on movements of the elements on the page.
+description: The `layout-shift` performance entry and `LayoutShift` API measures the layout stability of web pages based on movements of the elements on the page. Cumulative Layout Shift (CLS) is a common metric for perceived stability calculated from the Layout Instability API.
```
ddbeck: For consistency with the other suggestions, mostly.

```suggestion
description: The `layout-shift` performance entry and `LayoutShift` API measures the layout stability of web pages based on movements of the elements on the page. They're used to calculate Cumulative Layout Shift (CLS), a common metric for perceived stability.
```

ddbeck (changes requested):
I've got some nitpicky edits to suggest, but the overall approach here is nice. Thank you!

features/event-timing.yml
tunetheweb: The second sentence is fine. Not sure I 100% agree with the first sentence (it suggests to me the performance entries are in addition to the `PerformanceEventTiming` interface rather than part of that, and also both the spec and MDN call this an interface and the API is Event Timing), but it's what was there before so I can live with it.

ddbeck: Yeah, the descriptions are… challenging. The way we've written the other descriptions is such that "**and** `SomeThing` interface" is conventional to features that include that interface, where "of" does not make such a claim.

Many features represent later additions to another feature, but this one does not. For example, the description for the `getorinsert` feature says "The `getOrInsert()` and `getOrInsertComputed()` methods **of** `Map` objects…"; because `Map` itself is part of the `map` feature.

Admittedly, that's a bit awkward (and something we'd like to improve separately, with the way we show the structural relationships between successive features). But if you've got ideas for how to improve this in a general way, I'd welcome an issue to suggest a revision to the guidelines to capture it.

ddbeck (approved):
Thank you!

PR #3438, opened 2025-10-10
Remove draft "File and directory entries" feature

foolip:
The BCD keys for this overlap with file-system-access and
origin-private-file-system and draft/spec/entries-api.yml.dist shows
what keys remain unassigned. This draft feature won't help us.

PR #3486, opened 2025-10-23
Add features for `<input type="file" multiple>` and`<input type="file" webkitdirectory>`

ddbeck:
This creates features that handle uploading files and directories with form inputs. They map to:

- https://caniuse.com/input-file-directory and inspired by [this web.dev blog post](https://web.dev/patterns/files/open-a-directory)
- https://caniuse.com/input-file-multiple and the scenario described by @captainbrosset in https://github.com/web-platform-dx/web-features/pull/2993#issuecomment-2909854708

This approach treats these attributes as a distinct input type, like we do for various `<input type="…">` scenarios, though it's admittedly weird for it to be via an attribute.

While I was here, I sifted through `draft/file-and-directory-entries.yml`, disposing of the keys which were already assigned. This left me with almost nothing. The remaining exceptions largely have poor data quality in BCD (see https://github.com/mdn/browser-compat-data/issues/19304) and won't work as a "feature" until that's resolved.

Closes https://github.com/web-platform-dx/web-features/pull/3438, closes https://github.com/web-platform-dx/web-features/pull/2993, and closes https://github.com/web-platform-dx/web-features/pull/2633.


features/input-file-multiple.yml.dist
```diff
@@ -0,0 +1,14 @@
+# Generated from: input-file-multiple.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome: "22"
+    chrome_android: "25"
+    edge: "79"
+    firefox: "50"
```
foolip: Are you sure BCD is correct that this doesn't work in Firefox Android?

ddbeck: The BCD Collector test for this fails on Firefox for Android, so that appears to be the case.

features/input-file-multiple.yml
```diff
@@ -0,0 +1,16 @@
+name: '<input type="file" multiple>'
+description: The `<input type="file" multiple>` HTML element shows a file picker from which users can choose one or more files to upload with the form.
+spec: https://html.spec.whatwg.org/multipage/input.html#attr-input-multiple
+caniuse: input-file-multiple
+group:
+  - forms
+  - file-system
+compat_features:
+  - api.HTMLInputElement.webkitEntries
```
ddbeck, line 14: This is missing the `api.HTMLInputElement.multiple` and `html.elements.input.multiple` keys.

ddbeck: Done with da1c9b0e.

features/input-file-multiple.yml.dist
foolip: OK, I guess that's https://mdn-bcd-collector.gooborg.com/tests/api/HTMLInputElement/webkitEntries?

ddbeck: Yes, that's correct.

features/input-file-multiple.yml
```diff
@@ -0,0 +1,20 @@
+name: '<input type="file" multiple>'
+description: The `<input type="file" multiple>` HTML element shows a file picker from which users can choose one or more files to upload with the form.
```
captainbrosset:
```suggestion
description: The `<input type="file" multiple>` HTML element shows a file picker from which users can choose one or more files to upload with a form.
```
Maybe?

features/input-file-multiple.yml
```diff
@@ -0,0 +1,20 @@
...
+group:
+  - forms
+  - file-system
+compat_features:
+  - api.HTMLInputElement.multiple
+  - api.HTMLInputElement.webkitEntries
+  # The `multiple` attribute could be part of input-email-tel-url, but requires
+  # a solution to https://github.com/web-platform-dx/web-features/issues/1173
+  - html.elements.input.multiple
+  # These interfaces are plausbily part of this feature, but the upstream data is not right.
```
captainbrosset:
```suggestion
  # These interfaces are plausibily part of this feature, but the upstream data is not right.
```

features/input-file-multiple.yml.dist
captainbrosset: MDN says 

> The [...] webkitEntries property [...] contains an array of file system entries [...] representing files and/or directories selected by the user, **but only if that selection was made using drag-and-drop: selecting a file in the dialog will leave the property empty.**

Should we _hold back_ this feature because of this? 

Both MDN and caniuse.com show this feature as fully green.

ddbeck: Ah, interesting! I didn't know it had that… quirk.

FWIW, I did not `compute_from` around this because of the sample code in your comment https://github.com/web-platform-dx/web-features/pull/2993#issuecomment-2909854708. But I also did not understand the drag-and-drop and file selection distinction when I wrote this feature.

I guess there's a case to be made that `webkitEntries` is either a different thing or a non-essential part of this feature. The issues noted in the spec itself makes me think that we should not hinge this feature on support for `webkitEntries`.

- https://wicg.github.io/entries-api/#issue-4cc238bb
- https://wicg.github.io/entries-api/#issue-452f97f6

I propose that we work around this with a `compute_from`. If it ever comes up on Interop or something, then let's mint a feature for it. WDYT @captainbrosset and @foolip?

captainbrosset: That would work for me.
Testing the feature quickly, you can always access `inputElement.files` to get the files to be uploaded, and that seems to work whether you drag-and-dropped them, or selected them with the file dialog.
I agree that the `webkitEntries` property is either something separate or non-essential.

ddbeck: Done with 5cba26903e952b8138ffe5e3d41b106aff07c767.

captainbrosset (approved):

ddbeck:
OK, let's merge this now. If there's more to do around `webkitEntries`, we can do it in a follow up. Thank you!

Issue #3497, opened 2025-11-03
justify-self block-level elements


captainbrosset:
### Specification

https://www.w3.org/TR/css-align-3/#justify-block

### Description

The `justify-self` property is supposed to work on block-level elements (in addition to absolutely positioned elements and grid items). 
This also means that `justify-items` works on block containers.

Because this is a somewhat recent addition to what `justify-self` applies to, it ought to be its own new feature, similar to https://web-platform-dx.github.io/web-features-explorer/features/align-content-block/

### Documentation

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/justify-self

### Browser support

Chromium only.

ddbeck:
@captainbrosset can you link to something showing implementation evidence for this? I wasn't able to find something for Chromium myself. It'd be nice to queue up a PR to create `css.properties.justify-self.block_context` in BCD, if possible.

captainbrosset:
I can provide samples:

Example showing `justify-self` working on a block-level element in Chromium:
* Enter `data:text/html,<div style="border:1px solid;justify-self:end;">hello world` in the address bar
* The bordered div should be aligned to the end of the body element:
<img width="1429" height="120" alt="Image" src="https://github.com/user-attachments/assets/9a372724-edcd-4e5e-8946-f314da7ce337" />

Example showing `justify-items` (which just sets its children's `justify-self` value) working on a block container in Chromium:
* Enter `data:text/html,<div style="background:yellow;justify-items:center;"><p style="border:1px solid">para 1<p style="border:1px solid">para 2<p style="border:1px solid">para 3` in the address bar
* The children p elements should be aligned to the center of the parent div:
<img width="1435" height="216" alt="Image" src="https://github.com/user-attachments/assets/37660dad-a2fe-4a56-bd17-85d7b07c60aa" />

captainbrosset:
I filed https://github.com/mdn/browser-compat-data/issues/28379 for the BCD part.

PR #3505, opened 2025-11-05
Add feature for `justify-self` in block layouts

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3497.

### DIFF
--- a/features/justify-self-block.yml.dist
+++ b/features/justify-self-block.yml.dist
@@ -3,4 +3,10 @@
 
 status:
   baseline: false
-  support: {}
+  support:
+    chrome: "130"
+    chrome_android: "130"
+    edge: "130"
+compat_features:
+  - css.properties.justify-items.block_context
+  - css.properties.justify-self.block_context


features/justify-self-block.yml
```diff
@@ -0,0 +1,10 @@
+name: justify-self in block layouts
+description: >
+  The `justify-self` CSS property aligns an item along a container's inline axis in block layouts, like it does in grid layouts.
+  Likewise, the `justify-items` property sets a default `justify-self` value for all items in a container in a block layout.
+spec: https://drafts.csswg.org/css-align/#justify-block
+group: layout
+# No BCD keys yet but expecting:
+# compat_features:
+  # - css.properties.justify-items.block_context
+  # - css.properties.justify-self.block_context
```
ddbeck: Undo this comment if https://github.com/mdn/browser-compat-data/pull/28391 merges first.

captainbrosset (approved):

features/justify-self-block.yml
ddbeck: This did merge, so I'm going to uncomment this then wait for the next BCD release, to avoid having to get a review on the follow up PR during TPAC.

PR #3514, opened 2025-11-07
Add `progress()` CSS function

jamesnw:


features/progress-css.yml
```diff
@@ -0,0 +1,7 @@
+name: progress()
+description: The `progress()` CSS function returns the relative position of one value between two other values. It can be used as an interpolation input to other calculations.
```
ddbeck: This is fine as is, but I've got a few optional suggestions here to:

- Hint at the return type (I assume this is a unitless value from 0 to 1?)
- Avoid indirect "it can be…" with developer-oriented "you can"
- Avoid "input" (I dunno about this one—I read it as implying "user input"—so don't take this one too seriously)

```suggestion
description: The `progress()` CSS function returns a ratio for the relative position of one value between two other values. You can use it to interpolate a value for other calculations.
```

features/progress-function.yml
ddbeck, line 1: May I suggest `progress-function` for the ID here? We have others like this (e.g., `image-function`) and it's more specific than a `-css` disambiguation suffix.

features/progress-css.yml
jamesnw: I had a feeling you were going to have a better way to communicate the "input"- thanks!!

I also added that it is clamped to 0 - 1, but that will likely need to be updated to mention this is the default, and a TBD notation for specifying you don't want to be clamped when out of bounds. Since that isn't determined yet, I think it makes sense to document what is currently known, and adjust when more is implemented.

ddbeck: I like this new version a lot. The clamping bit is a nice detail. And I agree that we cover what we can and update it later.

PR #3517, opened 2025-11-10
Assign a bunch of keys to CSS features

ddbeck:
This isn't comprehensive, but it assigns a lot of the longest-implemented CSS keys to features that already exist.

None of these modify headline statuses; these keys are from backfilling activity that happened in BCD recently. See https://github.com/mdn/browser-compat-data/pull/28265

### DIFF
--- a/features/outlines.yml
+++ b/features/outlines.yml
@@ -5,6 +5,7 @@ spec: https://drafts.csswg.org/css-ui-4/#outline
 # caniuse: outline
 compat_features:
   - css.properties.outline-color
+  - css.properties.outline-color.transparent
   - css.properties.outline-offset
   - css.properties.outline-style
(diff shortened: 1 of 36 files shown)


features/outlines.yml
captainbrosset, line 1: This is making me realize that we have both `outline` and `outlines`.
Both are Baseline Widely Available. The only weird thing is that `outline` (the shorthand prop) is only Widely Available since September 2025. Looking at it in more details, I think that's because before Safari 16.4, the outline didn't follow the shape of border-radius, and that made it a partial implementation according to BCD before 16.4.
I think we should do an editorial override and consider the `outline` property to have been supported for much longer than that.
And we should also consider merging the 2 features.

Not blocking this PR.

captainbrosset (approved):

PR #3522, opened 2025-11-11
Assign a bunch of keys to `multi-column`

ddbeck:
This was previously tagged in BCD, so this requires the ugly `undist` diff here.

None of these modify headline statuses; these keys are from backfilling activity that happened in BCD recently. See https://github.com/mdn/browser-compat-data/pull/28265

### DIFF
--- a/features/multi-column.yml
+++ b/features/multi-column.yml
@@ -4,3 +4,48 @@ spec: https://drafts.csswg.org/css-multicol-1/
 group: multi-column
 status:
   compute_from: css.properties.columns
+compat_features:
+  - css.properties.align-content.multicol_context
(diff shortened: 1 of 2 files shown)


captainbrosset:
> 

This looks good to me I think, but we'll also have to think of how this plays with the CSS Gap Decorations feature: https://github.com/web-platform-dx/web-features/blob/main/features/gap-decorations.yml
> That feature builds on CSS gaps, and the column-rule property by adding support for decorations in grid and flexbox, as well as introducing row-rule, and other related properties.

This comment I made 2 weeks ago doesn't need to block this PR by the way. Feel free to merge @ddbeck.

PR #3523, opened 2025-11-11
Create a `landmark-elements` group

ddbeck:
This came out of https://github.com/web-platform-dx/web-features/issues/1624#issuecomment-3511959764. I don't feel strongly about this, but it seemed like a plausible thing to have.


foolip (commented):
This doesn't seem to correspond to anything in https://html.spec.whatwg.org/#content-models, but https://developer.chrome.com/docs/lighthouse/accessibility/use-landmarks links to https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html which includes a bunch more elements.

Something that we can link to as the authoritative definition of what's in this group and not would be nice, or just a comment that explains what it is. I could imagine "elements that don't do much except have a default ARIA role" could be a sensible definition.

ddbeck:
> https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html which includes a bunch more elements.

This is what I based it on, but I think the only element I'm missing from the list is `<form>`. We could add it to the group as well or…

> Something that we can link to as the authoritative definition of what's in this group and not would be nice, or just a comment that explains what it is. I could imagine "elements that don't do much except have a default ARIA role" could be a sensible definition.

I took "landmark elements" to have a more casual meaning than the overlap with ARIA terminology suggests. Really, we're talking about something like "the sectioning elements new in HTML5." I think it would also be plausible to rename the new group to something like "Sectioning elements" (with a comment that this is conventionally the HTML5 set, so it excludes `<form>`, `<body>`, and `<h1>` etc.). If we want to have this, this is the route I would suggest.

Or we could just skip this. 🤷

captainbrosset:
I agree with creating this subgroup inside HTML elements. Feels like a nice way to navigate the features.
Both https://en.wikipedia.org/wiki/HTML_landmarks and https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html agree on adding form. I'd approve this either as is, or with form added to it.

ddbeck:
OK, great. I've added `form` to the `landmark-elements` group (f5fd31d) and I added a comment to the group's YAML file to cite Wikipedia and WAI-ARIA (3aa3570).

PR #3533, opened 2025-11-12
Custom Media Queries - Add BCD

ramiy:
I added the CSS `@custom-media` at-rule in https://github.com/web-platform-dx/web-features/pull/3325 . But since no browser implemented this feature, we didn't added `compat_features` param.

Now, Firefox 146 is the first browser to support this feature https://bugzilla.mozilla.org/show_bug.cgi?id=1744292

I updated the BCD project to reflect this change https://github.com/mdn/browser-compat-data/pull/28438

This PR updates the `custom-media-queries` web feature setting the `compat_features` param.

### DIFF
--- a/features/custom-media-queries.yml
+++ b/features/custom-media-queries.yml
@@ -2,6 +2,9 @@ name: Custom media queries
 description: The `@custom-media` CSS at-rule names a media query, which you can reuse in multiple `@media` rules.
 spec: https://drafts.csswg.org/mediaqueries-5/#custom-mq
 group: media-queries
+# Not in BCD yet, but expecting:
+# - api.CSSCustomMediaRule.name
+# - api.CSSCustomMediaRule.query
 compat_features:
   - css.at-rules.custom-media
   - api.CSSCustomMediaRule


features/custom-media-queries.yml
```diff
@@ -2,8 +2,5 @@ name: Custom media queries
 description: The `@custom-media` CSS at-rule names a media query, which you can reuse in multiple `@media` rules.
 spec: https://drafts.csswg.org/mediaqueries-5/#custom-mq
 group: media-queries
-# Not in BCD yet, but expecting something like:
-# - css.at-rules.custom-media
-# - api.CSSCustomMediaRule
-# - api.CSSCustomMediaRule.name
-# - api.CSSCustomMediaRule.query
```
ddbeck, line 9: I think you need to do a bit of explaining here. What happened with the other keys? 😅

package.json
```diff
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@eslint/js": "^9.39.1",
     "@js-temporal/polyfill": "^0.5.1",
-    "@mdn/browser-compat-data": "7.1.20",
+    "@mdn/browser-compat-data": "7.1.21",
```
ddbeck: Merging `main` into this branch (or rebasing) will cause this change to drop out.

In the future, don't update BCD in a PR to edit a feature. BCD upgrades require quite a bit of process that shouldn't get mingled with feature edits (see https://github.com/web-platform-dx/web-features/pull/3536 for an example).

features/custom-media-queries.yml
ramiy: Ok, I added only `css.at-rules.custom-media` to BCD, the other keys don't exist - no browser supports them.

See BCD v7.1.21 changelog: https://github.com/mdn/browser-compat-data/releases/tag/v7.1.21

You can also check the code: https://github.com/mdn/browser-compat-data/tree/v7.1.21/api

Currently, the `CSSCustomMediaRule` doesn't exist. Only the CSS `@custom-media` at-rule is supported.

package.json
ramiy: Fixed.

features/custom-media-queries.yml
ddbeck: I understood that the key `css.at-rules.custom-media` was added to BCD https://github.com/mdn/browser-compat-data/pull/28438. But in this PR, you removed the comment for the other keys without explanation.

I went ahead and dug into this myself. I turned on the preference shown in the data and the `CSSCustomMediaRule` interface is exposed (there's also code for it in [the diff from the Firefox bug you linked in BCD](https://github.com/mozilla-firefox/firefox/commit/f289eebba919#diff-5344d883f252f0d0fcde1443e722366428c6dec715f2dfa5a910d6d4fd088ce2)). So two things need to happen here:

1. Against BCD, file an issue noting the missing keys or submit a PR to add the data for the missing API yourself.
2. In this PR, reinstate the comment for the `api.CSSCustomMediaRule` key and its descendants.

ramiy: BCD PR: https://github.com/mdn/browser-compat-data/pull/28450

ramiy: @ddbeck Ok, the BCD PR was merged.

ramiy: @ddbeck please review the ticket again.

features/custom-media-queries.yml
```diff
@@ -2,8 +2,6 @@ name: Custom media queries
...
+compat_features:
```
ddbeck, line 8:
```suggestion
# Not in BCD yet, but expecting:
# - api.CSSCustomMediaRule.name
# - api.CSSCustomMediaRule.query
compat_features:
```

Issue #3558, opened 2025-11-20
Baseline status for Anchor positioning


jamesnw:
[Anchor positioning](https://web-platform-dx.github.io/web-features-explorer/features/anchor-positioning/) is currently marked as supported in Chrome and Safari, with Firefox currently behind a flag, and potentially releasing it without a flag in December. At that point, if we do nothing, Anchor Positioning as a complete unit will progress to Baseline Newly available.

At this point, I don't think that's the right move, and I think we should consider this in advance of Firefox releasing the feature.

Roma Komarov called this out in a recent [blog post](https://blog.kizu.dev/anchoring-to-a-containing-block/#conclusion)- 

> While it might seem that anchor positioning is already present in some stable versions of browsers, in reality it is further from being even baseline. If, today, Firefox would release anchor positioning into its stable version, even if Firefox itself will be perfect, there are enough issues, starting from feature’s usability (absent features like being able to change the containing block) to bugs in stable versions of browsers that prevent many use cases from being viable.

Jake Archibald also saw this as an issue here- https://github.com/web-platform-dx/web-features/issues/3382#issue-3452873377

Some of the issues I'm aware of-
- Anchored Fixed position elements is broken in Safari
- Lots of position-try edge cases and bugs
- New keywords for `position-area` that haven't shipped anywhere yet (but would not warrant their own feature)
- In Chromium, a positioned element that overflows its containing block is placed incorrectly.
- Popover placement is a major stumbling block, and while there have been multiple attempts to address this, the latest resolution has not yet been added to the spec. 

Note that most of these are behavioral, and may not be captured by BCD directly. I don't think breaking it into smaller subfeatures will help in this situation.

The feature is currently computing its status from `anchor-name`, `position-anchor` and `anchor()`. 

Looking ahead, there are some concrete subsets that would make reasonable features within an Anchor Position group, such as `Anchor positioning elements with transforms`, `Container queries for detecting fallback`, and `anchor attribute`.

In its current state, I would consider anchor positioning to be in beta. While a happy path is possible, I'm not confident that there won't be additional breaking changes, and it comes with significant caveats and unknowns. We don't have a "Baseline Beta" status (🤔)- what other options do we have to reflect this?

captainbrosset:
Discussed at the WebDX CG today:

- Confirmed from the Interop 25 side of things.
- We should be conservative rather than optimistic here.
- We could remove the `compute_from` statement to keep this from reaching Baseline.
[...]

PR #3574, opened 2025-11-24
Remove `compute_from` from anchor positioning

ddbeck:
This would show no (complete) implementers of anchor positioning, preventing it from reaching Baseline status even if/when Firefox ships it. See https://github.com/web-platform-dx/web-features/issues/3558 for discussion.


jamesnw (approved):
I approve this, but second the idea that we don't have to do this immediately, and can wait until Firefox ships, to give time for a bit more clarity to emerge.

abo-L:
Abol

Issue #3561, opened 2025-11-21
scroll-target-group


una:
### Specification

BCD page: https://github.com/mdn/browser-compat-data/blob/main/css/properties/scroll-target-group.json
Spec: https://drafts.csswg.org/css-overflow-5/#scroll-target-group

### Description

Scroll API for converting anchor links into scroll markers

### Documentation

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-target-group

### Browser support

Chrome 140+

ddbeck:
Thanks for filing this issue. I've opened https://github.com/web-platform-dx/web-features/pull/3577 to fix it.

PR #3577, opened 2025-11-25
Add `scroll-target-group` and `scroll-marker-targets`

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3561.

Thanks to @una with [this blog post about the feature](https://una.im/scroll-target-group) to help me understand what this is, why it's so cool, and, ultimately, guide to me to a plausible two-sentence description.


features/scroll-target-group.yml
```diff
@@ -0,0 +1,10 @@
+name: scroll-target-group
+description: >
+  The `scroll-target-group` CSS property picks an element to be a scroll marker group container, which uses the anchor links within that element as scroll markers.
```
atopal: Feel free to ignore, since this might just be me, but the benefit wasn't immediately clear to me. How about: The `scroll-target-group` property designates an element as a container where existing anchor links function as scroll markers. This enables the links to automatically track and highlight which target section is currently scrolled into view.

ddbeck: Thank you for this review. I started to rewrite this, but I found that, without `:target-current` as part of this (or some reference to it), it's hard to make sense of this as a standalone feature. I'm going to put this into draft a bit and see if I can come up with something more coherent.

tunetheweb: If it helps, here's another explanation and demo: https://chrome.dev/css-wrapped-2025/#scroll-target-group

Though you're right that this also makes use of `:target-group`

features/scroll-markers.yml
```diff
@@ -1,7 +1,7 @@
 name: Scroll markers
```
ddbeck, line 1: I've tried to rework this so it's the head feature of the overall cluster of scroll markers as a general feature. I would expect, someday, to see `scroll-marker-targets` and `scroll-target-group` merge into this feature.

atopal (approved):
Thanks Daniel!

tunetheweb:
Curious where this is now @ddbeck . Are you still working on it, or is it good to merge?

captainbrosset (approved):

Issue #3403, opened 2025-10-01
Anchored fallback container queries


lilles:
### Specification

https://drafts.csswg.org/css-anchor-position-2/#anchored

### Description

The ability to style descendants of CSS anchor positioned elements based on their applied position-try-fallbacks using @container queries.

### Documentation

Explainer: https://github.com/w3c/csswg-drafts/blob/main/css-anchor-position-1/anchored_container_query.md

Demo: https://codepen.io/una/pen/ZYbzrLg

Youtube: https://www.youtube.com/watch?v=u5-Vuduc9mk

### Browser support

Chromium feature: https://chromestatus.com/feature/5177580990496768

PR #3578, opened 2025-11-25
Add anchor position container queries feature

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3403

### DIFF
--- a/features/container-anchor-position-queries.yml
+++ b/features/container-anchor-position-queries.yml
@@ -1,6 +1,6 @@
 name: Anchor position container queries
 description: >
-  Anchor position container queries with the `@container anchored(fallback: …)` at-rule apply styles to an element based on an element's anchor position.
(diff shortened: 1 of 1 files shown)


features/container-anchor-position-queries.yml
```diff
@@ -0,0 +1,9 @@
+name: Anchor position container queries
+description: >
+  Anchor position container queries with the `@container anchored(fallback: …)` at-rule apply styles to an element based on an element's anchor position.
```
captainbrosset:
```suggestion
  Anchor position container queries with the `@container anchored(fallback: …)` at-rule apply styles to an element based on the element's anchor position.
```
 Maybe?

johnlowe399-blip:
ddbeck:3564-container-anchored-queries

Issue #3566, opened 2025-11-21
Range syntax for style container queries and if()


una:
### Specification

https://drafts.csswg.org/css-conditional-5/#typedef-style-range

### Description

Supports range queries instead of relying on exact matching for style queries and in the if() function.

### Documentation

https://www.una.im/range-style-queries
https://css-tricks.com/the-range-syntax-has-come-to-container-style-queries-and-if/
https://chromestatus.com/feature/5184992749289472

### Browser support

Chrome 142+

ddbeck:
Thanks for filing this. I've opened https://github.com/web-platform-dx/web-features/pull/3579 to create the feature, but there's no compat data for this yet. I'll follow up soon with a BCD PR.

PR #3579, opened 2025-11-25
Add feature for range syntax for style queries

ddbeck:
This still needs a BCD PR at least _open_, so I'm going to leave this in draft mode for now.

Fixes https://github.com/web-platform-dx/web-features/issues/3566


ddbeck:
Data has since appeared in BCD, but it's a bit wonky. I've sent https://github.com/mdn/browser-compat-data/pull/28537 to fix it.

ddbeck:
When BCD releases [v7.1.24](https://github.com/mdn/browser-compat-data/pull/28528) and we pick up that release, I'll merge and refresh the dist here and the tests will pass.

tunetheweb:
Looks like BCD v7.1.24 has been released @ddbeck

PR #3582, opened 2025-11-26
Migrate `next`-tagged releases to trusted publishing

ddbeck:
GitHub and npm are [moving to a package publishing model](https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/) that uses OIDC instead of npm tokens. If we switch to this, we would not need to periodically reissue tokens ([as described in our publishing docs](https://github.com/web-platform-dx/web-features/blob/main/docs/publishing.md#secrets)). This was inspired by me having to reissue tokens today.

This PR migrates the `compute-baseline` and `web-features` `next`-tagged releases to this publishing workflow. If this works well, we can refactor the `latest`-tagged publishing workflows to use the same method ((un)fortunately, each package on npm may only have one blessed workflow, so we'll have to consolidate the publishing workflows).

Upon merging, the packages need configuration in npmjs.com (see [this section of the npm docs](https://docs.npmjs.com/trusted-publishers#for-github-actions) and [this area](https://www.npmjs.com/package/web-features/access) in the npm UI, if you have access). If you're not ready to do this yourself, then please approve and leave unmerged. Thank you!

### DIFF
--- a/.github/workflows/publish_next_compute-baseline.yml
+++ b/.github/workflows/publish_next_compute-baseline.yml
@@ -7,6 +7,8 @@ on:
     paths:
       - packages/compute-baseline/**
 
+permissions: {}
+
 env:
   package: "compute-baseline"
   package_dir: "packages/compute-baseline"
@@ -15,6 +17,8 @@ env:
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
@@ -27,6 +31,12 @@ jobs:
     if: github.repository == 'web-platform-dx/web-features'
     runs-on: ubuntu-latest
     needs: "test"
+    permissions:
+      contents: read
+      # Required for OIDC and trusted publishing. See:
+      # - https://docs.npmjs.com/trusted-publishers
+      # - https://docs.github.com/en/actions/concepts/security/openid-connect
+      id-token: write
     steps:
       - name: Get timestamp
         id: timestamp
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_next_compute-baseline.yml
```diff
@@ -7,6 +7,8 @@ on:
     paths:
       - packages/compute-baseline/**
 
+permissions: {}
+
```
ddbeck, line 11: The permissions vary between jobs here, so this explicitly unsets them for the workflow.

.github/workflows/publish_next_web-features.yml
```diff
@@ -67,11 +76,8 @@ jobs:
           VERSION: ${{ steps.version.outputs.VERSION }}
           TIMESTAMP: ${{ steps.timestamp_and_hash.outputs.TIMESTAMP }}
           SHORT_HASH: ${{ steps.timestamp_and_hash.outputs.SHORT_HASH }}
-      - if: ${{ env.NODE_AUTH_TOKEN }}
-        run: npm publish --tag ${{ env.dist_tag }}
+      - run: npm publish --tag ${{ env.dist_tag }}
         working-directory: ${{ env.package_dir }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
ddbeck, line 74: The token is no longer required (and the `if` condition was superfluous anyway, in light of the `if: github.repository …` condition for the entire job).

ddbeck (commented):
Self-reviewing to provide some clarity to would-be reviewers.

projektleitungriduani-bit:
399693468_1167733094200634_7457074094501049095_n

PR #3603, opened 2025-12-05
Add `scroll-into-view-container`

### LINKED ISSUE #3602: scroll-into-view-container
### Specification

https://drafts.csswg.org/cssom-view/#dom-scrollintoviewoptions-container

### Description

The `ScrollIntoViewOptions` `container` option allows developers to perform a `scrollIntoView()` only scrolling the nearest ancestor scroll container

### Documentation

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#container

### Browser support

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#browser_compatibility
https://caniuse.com/mdn-api_element_scrollintoview_options_parameter_container_option

tunetheweb:
Fixes #3602 

This has quite different support than the other ScrollIntoView() features so think it deserves it's own feature:

image

### DIFF
--- a/features/scroll-into-view-container.yml
+++ b/features/scroll-into-view-container.yml
@@ -1,5 +1,5 @@
 name: scrollIntoView() container
-description: The `ScrollIntoViewOptions` `container` option allows developers to perform a `scrollIntoView()` only scrolling the nearest ancestor scroll container
+description: The `container` option of the `scrollIntoView()` method sets which ancestor scroll container to scroll. The `"nearest"` value scrolls only the nearest ancestor, instead of the default `"all"`.
 spec: https://drafts.csswg.org/cssom-view-1/#dom-scrollintoviewoptions-container
 group: scrolling
 compat_features:


features/scroll-into-view-container.yml
```diff
@@ -0,0 +1,6 @@
+name: scrollIntoView() container
+description: The `ScrollIntoViewOptions` `container` option allows developers to perform a `scrollIntoView()` only scrolling the nearest ancestor scroll container
```
ddbeck: Writing these descriptions is kinda hard. I'm suggesting a rewrite (which you're welcome to build on). I won't belabor each "rule" here, but the highlights are:

- Speak to developers directly, not about them
- Avoid spec-only language (e.g., `ScrollIntoViewOptions` isn't an exposed name)
- Front load context-setting information (i.e., mention `scrollIntoView()` early)
- Try to include significant words that a developer might type (i.e., the non-default `"nearest"`)

So something like this. What do you think?

```suggestion
description: The `container` option of the `scrollIntoView()` method sets which ancestor scroll container to scroll. The `"nearest"` value scrolls only the nearest ancestor, instead of the default `"all"`.
```

tunetheweb: LGTM! Thanks for that.

PR #3605, opened 2025-12-05
Add ToggleEvent source

### LINKED ISSUE #3604: ToggleEvent source
### Specification

https://html.spec.whatwg.org/multipage/interaction.html#dom-toggleevent-source

### Description

The source attribute of `ToggleEvents` contains the element which triggered the `ToggleEvent` to be fired, if applicable.

### Documentation

https://developer.mozilla.org/en-US/docs/Web/API/ToggleEvent/source

### Browser support

https://developer.mozilla.org/en-US/docs/Web/API/ToggleEvent/source#browser_compatibility

tunetheweb:
Fixes #3604 

Has quite different support than the rest of Toggle Event so added it as a new feature:

image


features/popover-toggleevent-source.yml
```diff
@@ -0,0 +1,6 @@
+name: Popover ToggleEvent source
```
ddbeck: If I understand correctly, the `toggle` event fires for `<details>`, `<dialog>`, and popovers. But I'm not sure how to understand the spec with the `source` property. Is the property _only_ for popovers? I think the answer to this makes a big difference for how we frame this feature.

ddbeck: Thinking on this a little more, can you point to an example of how this might be used? The MDN reference page example feels a little contrived and I'm trying to figure out what a developer might do with this that they could not without this property.

tunetheweb: > If I understand correctly, the toggle event fires for `<details>`, `<dialog>`, and popovers. But I'm not sure how to understand the spec with the source property. Is the property only for popovers? I think the answer to this makes a big difference for how we frame this feature.

Oh you're right it's for more than just Popovers (though that's a big use case). Updated to drop the "Popover" from the feature name.

And here's an example of how it might be used:
https://chrome.dev/css-wrapped-2025/#toggleevent-source
(which coincidently is why we want to add this feature so we can add a nice Baseline widget to this page 😁)

features/popover-toggleevent-source.yml
```diff
@@ -0,0 +1,6 @@
+name: ToggleEvent source
+description: The `source` attribute of `ToggleEvents` contains the element which triggered the `ToggleEvent` to be fired for a popover, `<dialog>`, or `<details>` element, if applicable.
```
ddbeck: Some nits for consistency with the way we describe other things. What do you think of this?

```suggestion
description: The `source` property of a `ToggleEvent` object is the element which triggered the `toggle` event to fire for a popover, `<dialog>`, or `<details>` element, if applicable.
```

features/toggleevent-source.yml
ddbeck, line 1: Don't forget to change the file names to `toggleevent-source.yml` and `toggleevent-source.yml.dist`. Thanks!

tunetheweb: Doh! Done!

PR #3611, opened 2025-12-08
Move key for `display-mode: picture-in-picture` to `picture-in-picture`

ddbeck:
It'd be nice if BCD broke out _all_ the display mode values into separate keys, so we could figure out what the original set is. Right now some of that information is locked in notes. For now, I've modified the description and parked the one key we do have.


captainbrosset:
Ah, this needs a dist file refresh.

PR #3621, opened 2025-12-11
Introduce the term `grid-lanes` to the masonry feature

### LINKED ISSUE #2045: Update Masonry's description when the CSSWG resolves the "just grid" vs. new display type question
Useful pre-read:

- https://developer.chrome.com/blog/masonry-syntax
- https://rachelandrew.co.uk/archives/2024/09/21/masonry-and-good-defaults/
- https://webkit.org/blog/16026/css-masonry-syntax/
- https://www.oddbird.net/2024/10/01/grid-masonry-nuance/

CSSWG feedback issue: https://github.com/w3c/csswg-drafts/issues/11060

Once a decision is made, we might have to update the masonry description in web-features. Current description:

> Masonry is a type of CSS grid layout where the items on one of the axes are tightly packed together, like brickwork, instead of leaving gaps to align across the other axis.

captainbrosset:
Fixes #2045.

The CSSWG has resolved to make masonry its own display type, and decided to call it `grid-lanes`.

It's important that our feature uses this term, but at the same time retain the masonry name too. The spec still uses it, and that's how most developers who know about this type of layout likely refer to it.

Let's iterate on a better description.


features/masonry.yml
```diff
@@ -1,4 +1,4 @@
 name: Masonry
-description: Masonry is a type of CSS grid layout where the items on one of the axes are tightly packed together, like brickwork, instead of leaving gaps to align across the other axis.
+description: Masonry is a two-dimensional layout system where items are tightly packed within parallel lanes. Items are placed one by one in the lane which has the most space available. Also known as grid-lanes.
```
ddbeck: I think we should:

- Mention CSS
- Retain the illustrative metaphor.
- Explicitly mention the `display: grid-lanes` and `display: inline-grid-lanes` entrypoints (unless we think it's likely that implementers will do the latter separately). This lets us out of saying "also known as" which I'm not convinced is a true statement (at least not yet).

Maybe something like this:

```suggestion
description: Masonry, created by a `display: grid-lanes` or `display: inline-grid-lanes` CSS declaration, is a two-dimensional layout system where items are tightly packed in parallel, like brickwork. Items are placed one by one in the lane that has the most available space.
```

features/masonry.yml
```diff
@@ -1,4 +1,4 @@
 name: Masonry
-description: Masonry is a type of CSS grid layout where the items on one of the axes are tightly packed together, like brickwork, instead of leaving gaps to align across the other axis.
+description: Masonry, created by a `display: grid-lanes` or `display: inline-grid-lanes` CSS declaration, is a two-dimensional layout system where items are tightly packed in parallel, like brickwork. Items are placed one by one in the lane that has the most available space.
```
ddbeck: This should fix the parsing error:

```suggestion
description: >
  Masonry, created by a `display: grid-lanes` or `display: inline-grid-lanes` CSS declaration, is a two-dimensional layout system where items are tightly packed in parallel, like brickwork. Items are placed one by one in the lane that has the most available space.
```

captainbrosset: I iterated a bit more on this and simplified it.

rviscomi:
cc @jensimmons for context on https://github.com/web-platform-dx/developer-signals/issues/308#issuecomment-4225033484

PR #3624, opened 2025-12-11
Use BCD `mediatypes` data in existing image format features

ddbeck:
BCD has added a new tree of data for media-types (modeled mostly after [IANA's Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml)) in https://github.com/mdn/browser-compat-data/pull/27168, so we can derive such features' statuses from BCD, instead of supplying our own data.


features/avif.yml.dist
```diff
@@ -1,4 +1,27 @@
...
+  # baseline: low
+  # baseline_low_date: 2024-01-25
+  # support:
+  #   chrome: "85"
+  #   chrome_android: "85"
+  #   edge: "121"
+  #   firefox: "113"
+  #   firefox_android: "113"
+  #   safari: "16.4"
+  #   safari_ios: "16.1"
```
ddbeck, line 38: There's a slight difference between web-features and BCD here, with BCD reporting a few minor versions later. I don't think this is significant and does not affect the Baseline low date.

features/webp.yml.dist
```diff
@@ -1,4 +1,45 @@
...
 
-{}
+status:
+  baseline: high
+  baseline_low_date: 2020-09-16
+  baseline_high_date: 2023-03-16
+  support:
+    chrome: "32"
+    chrome_android: "32"
+    edge: "79"
```
ddbeck: BCD doesn't know about support in EdgeHTML (i.e., it's a ≤79 value), so this looks later than the data we have. We could contribute the Edge 18 value upstream. This does not affect the Baseline low date (which hinges on Safari 14).

captainbrosset: Contributing the Edge 18 value upstream sounds like the right thing to do. Especially because we can just point to https://caniuse.com/webp as a reference for that value.
This doesn't need to block this PR though.

features/avif.yml.dist
```diff
@@ -1,4 +1,39 @@
...
-{}
+status:
+  baseline: low
+  baseline_low_date: 2024-01-25
+  support:
+    chrome: "85"
+    chrome_android: "85"
+    edge: "121"
+    firefox: "113"
+    firefox_android: "113"
```
captainbrosset, line 12: The difference for Firefox is explained by a note at https://caniuse.com/avif. In 93, Firefox only support still avif images. Animated support came in 113.

features/webp.yml.dist
ddbeck: Sent https://github.com/mdn/browser-compat-data/pull/28653, which has already been merged. We'll pick this up on the next BCD release.

ddbeck: Waiting on [v7.2.3](https://github.com/mdn/browser-compat-data/pull/28701).

Edit: Expecting this to ship on 6 January, after a discussion with Florian.

ddbeck: Oops, I didn't actually fix this in my PR. Sent https://github.com/mdn/browser-compat-data/pull/28769.

ddbeck: OK, I still didn't completely fix this. 🤦 

I've sent https://github.com/mdn/browser-compat-data/pull/28848. I'm going to suggest that we merge this PR without it and we'll pick up the changes when BCD gets that data.

PR #3627, opened 2025-12-12
Fix container-anchor-position-queries compat data

tunetheweb:
Added by @ddbeck in #3578 but commented out as not available in MDN at that time but is now.

Also a slight correction from:
- `css.at-rules.container.anchor_position_queries.fallback.any`
to:
- `css.at-rules.container.anchor_position_queries.fallback.any_value`


ddbeck (approved):

ddbeck:
Thank you, @tunetheweb!

johnlowe399-blip:
tunetheweb:fix-container-anchor-position-queries

PR #3677, opened 2026-01-07
Add `png` feature

ddbeck:
This adds a feature that uses the `mediatypes.png.*` data from BCD and maps to a few caniuse entries.

Blocked by https://github.com/mdn/browser-compat-data/pull/28770. It doesn't yet align with caniuse.

### DIFF
--- a/features/png.yml
+++ b/features/png.yml
@@ -1,5 +1,5 @@
 name: PNG
-description: The PNG (Portable Network Graphics) image format is a rater graphics file format with animation, alpha transparency, and lossless compression.
+description: The PNG (Portable Network Graphics) image format is a raster graphics file format with animation, alpha transparency, and lossless compression.
 spec: https://w3c.github.io/png/#Introduction
 caniuse:
   - png-alpha


features/png.yml
```diff
@@ -0,0 +1,10 @@
+name: PNG
+description: The PNG image format is a rater graphics file format with animation, alpha transparency, and lossless compression. Also known as Portable Network Graphics.
```
foolip: Portable Network Graphics is more like the expanded name than an alternative name. I'm not sure if we have an established style for expanding abbreviations, but maybe?

```suggestion
description: The PNG image format is a rater graphics file format with animation, alpha transparency, and lossless compression. PNG means Portable Network Graphics.
```

Just not spelling it out would be fine I think, will anyone really search for the full name?

captainbrosset: Or:

```suggestion
description: The Portable Network Graphics (PNG) image format is a rater graphics file format with animation, alpha transparency, and lossless compression.
```

ddbeck: In [AVIF](https://github.com/web-platform-dx/web-features/blob/main/features/avif.yml) we do something like this:

```suggestion
description: The PNG (Portable Network Graphics) image format is a rater graphics file format with animation, alpha transparency, and lossless compression.
```

I like the idea of putting the most widely-used name first and the pedantic, official one after.

ddbeck: With Patrick's thumbs up, I applied my suggestion. @foolip does this work for you? If so, would you approve this PR?

features/png.yml
```diff
@@ -0,0 +1,10 @@
+name: PNG
+description: The PNG (Portable Network Graphics) image format is a rater graphics file format with animation, alpha transparency, and lossless compression.
```
captainbrosset:
```suggestion
description: The PNG (Portable Network Graphics) image format is a raster graphics file format with animation, alpha transparency, and lossless compression.
```

captainbrosset (approved):

Issue #3423, opened 2025-10-06
Scoped view transitions


foolip:
### Specification

There's a CSSWG resolution but no spec edit yet:
https://github.com/w3c/csswg-drafts/issues/9890

### Description

https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md

### Documentation

N/A

### Browser support

In progress in Chrome: https://chromestatus.com/feature/5109852273377280

foolip:
cc @skobes-chromium @kevers-google

skobes-chromium:
Draft spec at https://drafts.csswg.org/css-view-transitions-2/#scoped-vt

bramus:
It is currently not clear to me whether this should be a separate Web Platform Feature, or be grouped together with Nested View Transition Groups. As it currently stands, one won’t be able to implement Scoped VTs without Nested VT Groups.

ddbeck:
Philip pinged me about this over the holidays and I'm just now catching up.

Mostly notes to myself here, but I think we have something feature-shaped here. Running out of time to get to a PR on this today, but this is what I have so far:

- Name: Something like _Element view transitions_, _Element-scoped view transitions_. _Scoped view transitions_ is OK if we must but it really sounds like spec speak.
- Description: The `startViewTransition()` method of an `Element` object starts a view transition that affects only that element's DOM tree. The `contain: view-transition` CSS declaration contains the layout of the tree to be transitioned.
- Compat keys:
   - `api.Element.startViewTransition`
   - `css.properties.contain.view-transition`

This is likely to be a candidate for merging, in the distant future, into `view-transitions` proper. I should include a comment to say so.

foolip:
That SGTM @ddbeck. @skobes-chromium @kevers-google can you review the proposed wording?

bramus:
> * Name: Something like _Element view transitions_, _Element-scoped view transitions_. _Scoped view transitions_ is OK if we must but it really sounds like spec speak.

I would like us to use the term “Scoped View Transitions” as that is the name that is being “advertised” to authors. “Element-scoped view transitions” might also work.

_(Roughly translated: anything with the word “scoped” in it seems fine)_

ddbeck:
I've opened https://github.com/web-platform-dx/web-features/pull/3681 to add element-scoped view transitions, which would resolve this issue.

PR #3681, opened 2026-01-09
Add element-scoped view transitions

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3423

### DIFF
--- a/features/view-transitions-element-scoped.yml
+++ b/features/view-transitions-element-scoped.yml
@@ -7,7 +7,7 @@ spec: https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md
 group: view-transitions
 # Expecting:
 # compat_features:
-#  - api.Element.startViewTransitions
+#  - api.Element.startViewTransition
 #  - css.properties.contain.view-transition
 # This feature is subordinate to view-transitions. It should probably merge into
 # it one day.


features/view-transitions-element-scoped.yml
```diff
@@ -0,0 +1,12 @@
+name: Element-scoped view transitions
+description: >
+  The `startViewTransition()` method of an `Element` object starts a view transition that affects only that element's DOM tree.
+  The `contain: view-transition` CSS declaration contains the layout of the tree to be transitioned.
```
captainbrosset, line 4: This is an early feature, so I'll approve this anyway (no need to hold it up, we can always change the description later), but I wish this was mentioned somehow:

- transition pseudo-elements are affected by ancestor clips and transforms
- multiple transitions on separate elements can run concurrently.

ddbeck: Thanks, @captainbrosset. This is a good note.

For the second point, I think we have room to add something like "You can use this to run multiple transitions concurrently." That seems like a notable behavior that this feature distinctly offers. I'll add that in a separate suggestion and I'd welcome your review on it.

The ancestor clips and transforms is less clear to me as a "feature" we might want to describe. Document view transitions don't have ancestors (right?), so the fact that the clips and transforms apply here seems like a ramification of not being the apex of the tree. Or am I missing something here and the clips and transforms behavior has some kind of opt-in that is part of the element-scoped view transition API?

ddbeck, line 4:
```suggestion
  The `startViewTransition()` method of an `Element` object starts a view transition that affects only that element's DOM tree.
  The `contain: view-transition` CSS declaration contains the layout of the tree to be transitioned.
  You can use this to run separate elements' transitions concurrently.
```

captainbrosset: I don't think there's a way to opt in or out of this. The proposal to mention clips and transforms is more of a note than a feature. If you scope a view transition to an element only, then the way that view transition animates will depend on the transformations that might already be applied to an ancestor of that element. It may seem obvious, and again, not an optional feature, so it is fine to ignore that part (same with clip).

PR #3691, opened 2026-01-14
Add `Iterator.concat()`

o-:


features/iterator-sequencing.yaml
ddbeck, line 1: Generally, we align the feature ID to the name at the time the feature was created. So this file needs to be renamed to `iterator-concat.yml` (note the extension too, which is also causing an error right now).

features/iterator-sequencing.yaml
```diff
@@ -0,0 +1,6 @@
+name: Iterator.concat()
+description: The `Iterator.concat(iterator, ...)` creates a sequenced iterator from a list of iterators.
```
ddbeck: Some suggestions to align with our style guide (e.g., no placeholder variable names; returns not creates), but also to be a little more descriptive about what it means to consume a sequence of iterators. What do you think of this?

```suggestion
description: The `Iterator.concat()` JavaScript method returns an iterator that yields values from a sequence of iterators, exhausting each iterator before moving on to the next.
```

features/iterator-sequencing.yaml
```diff
@@ -0,0 +1,6 @@
...
+spec: https://github.com/tc39/proposal-iterator-sequencing
+group: iterators
+compat_features:
+  - javascript.builtins.Iterator.concat
```
ddbeck: Thanks for adding this. This was just recently added to BCD, so to help get this merged faster this needs to happen:

1. Merge `main` into this branch.
2. Run `npm ci && npm run dist`.
3. Commit (and push) the newly-created `.dist` file.

o-:
@ddbeck thanks for the review. ptal again.

ddbeck (approved):

ddbeck:
I'm about to push a change to the spec URL to get this into today's release. Cheers!

PR #3704, opened 2026-01-19
Add `compat_features` for `@scope` rule

dfabulich:
Not sure if I'm holding this right. I noticed that https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@scope lists no Baseline status.

https://caniuse.com/?search=%40scope lists two features, https://caniuse.com/css-cascade-scope which is Baseline Newly Available (correct!) and https://caniuse.com/mdn-css_at-rules_scope which lists no Baseline status.

Maybe this will fix it?


features/scope.yml
```diff
@@ -3,3 +3,6 @@ description: The `@scope` CSS at-rule sets the scope for a group of rules.
 spec: https://drafts.csswg.org/css-cascade-6/#scope-atrule
 group: css
 caniuse: css-cascade-scope
+compat_features:
+  - css.at-rules.scope
+  - css.selectors.nesting.at-scope
```
ddbeck, line 11: It'd be nice to keep the API for this too. (This requires re-running `npm run dist`.)

```suggestion
compat_features:
  - api.CSSScopeRule
  - api.CSSScopeRule.end
  - api.CSSScopeRule.start
  - css.at-rules.scope
  - css.selectors.nesting.at-scope
```

dfabulich:
Is this ready to merge?

dfabulich:
> But would it be worth splitting out support for `&` in another feature instead?

I think not. IMO, the old `&` behavior was a bug in the specification. https://css.oddbird.net/scope/parent-selector/

I would have been uncomfortable saying that `@scope` is fully Baseline if there were still a major browser that didn't support the updated behavior for `&`; I'd have advocated to say that it's "supported but buggy on some browsers."

Now, with `@scope` including the updated `&` behavior, I think it's fair to say that `@scope` really is Baseline newly available. The fact that the Baseline low date barely changed is bonus encouragement!

Issue #3378, opened 2025-09-24
CSS filter() function


nt1m:
### Specification

https://drafts.fxtf.org/filter-effects/#FilterCSSImageValue

### Description

`background-image: filter(url(), blur(5px))`

The filter() function takes an image and a filter, and returns a new image. This can be useful if you want to reduce the number of elements you use on your page.

### Documentation

Missing :(

WPT: https://wpt.fyi/results/css/filter-effects/filter-function?label=experimental&label=master&aligned

### Browser support

Safari 9.1+ (unprefixed, 9.0 with prefix)
https://caniuse.com/css-filter-function

captainbrosset:
@nt1m do you know if this is at least documented on BCD (browser-compat-data)? I wasn't able to find BCD keys for the filter function, but maybe I've been looking at the wrong place.
If it isn't there, it'd be great to get it added, especially if this has been available for that long (Safari 9).

nt1m:
This doesn't seem to be on BCD nor on the web docs. There's an open issue for these: https://github.com/mdn/mdn/issues/726

PR #3718, opened 2026-01-25
Add CSS `filter()` function

nt1m:
Fixes https://github.com/web-platform-dx/web-features/issues/3378

This has been supported since Safari 9, see https://caniuse.com/css-filter-function.


nt1m:
@captainbrosset @foolip Could you please take a look?

captainbrosset:
> compat data are not currently available, but will be linked once they are.

@nt1m do you know if there's already a PR opened on BCD to add the required keys for this feature? If there isn't one, would you or someone on your end be able to create one?
We can of course merge the new feature without BCD keys in the meantime, but if a BCD PR already exists (or is quick to create), it would be better to wait for it. This way, less risks of forgetting to update the feature once the keys are available.

features/filter-function.yml
```diff
@@ -0,0 +1,5 @@
+name: filter()
+description: The `filter()` CSS function applies one or more graphic effects to an CSS image and returns a new image. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects.
```
captainbrosset:
```suggestion
description: The `filter()` CSS function applies one or more graphic effects to a CSS image and returns a new image. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects.
```

captainbrosset: The name and description fields are very similar to the ones in the `filter.yml` feature. It's good for consistency, but might be a bit confusing to web developers. I'm not sure I'd be able to really tell which feature does what if I didn't already know about them, just by reading these fields.

I think what's missing for me is an example of where you use each feature. So maybe:

* In `filter.yml`:

  The `filter` CSS property applies one or more graphic effects to an element. You can use functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For example: `filter: blur(2px)`.

* In `filter-function.yml`:

  The `filter()` CSS function applies one or more graphic effects to a CSS image, such as a background image. You can use functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For example: `background: filter(url(image.png), blur(2px))`.

nt1m:
I created a BCD PR here: https://github.com/mdn/browser-compat-data/pull/28916/ . I was reluctant to create one until some MDN docs were written, but it looks like you can just create one independently!

captainbrosset:
Ah, you'll need double quotes around the descriptions. Otherwise the : characters break the yaml syntax.

features/filter-function.yml
```diff
@@ -0,0 +1,7 @@
+name: filter()
+description: The `filter()` CSS function applies one or more graphic effects to a CSS image, such as a background image. You can use functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For instance, `background: filter(url(image.png), blur(2px))`.
```
captainbrosset:
```suggestion
description: "The `filter()` CSS function applies one or more graphic effects to a CSS image, such as a background image. You can use functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For instance, `background: filter(url(image.png), blur(2px))`."
```

features/filter.yml
```diff
@@ -1,5 +1,5 @@
 name: filter
-description: The `filter` CSS property applies one or more graphic effects to an element. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects.
+description: The `filter` CSS property applies one or more graphic effects to an element. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For instance, `filter: blur(2px)`.
```
captainbrosset:
```suggestion
description: "The `filter` CSS property applies one or more graphic effects to an element. You can use filter functions, such as `blur()` and `drop-shadow()`, alone or combined to produce different effects. For instance, `filter: blur(2px)`."
```

captainbrosset:
We'll have to wait for the BCD key to be released and then bump our dependency to BCD on the repo for this to build.

PR #3726, opened 2026-01-27
Add a features/draft/proposed/ directory an exempt it from CI

foolip:


jgraham:
I don't really understand how this is going to work. If a vendor adds a draft feature here, how do they refer to the feature in e.g. an intent to prototype email or standards-position request? If the feature is updated "later" how is one supposed to handle references to the old name in immutable data (e.g. email)?

Basically I don't understand the concept of an "unpublished web feature". I think the use cases require there to be an agreed ID for early stage features, but which tools that don't care about early stage features can filter out.

If this proposal amounts to features getting an like `vendor/foo` which can be used as a real id and later renamed as normal (or marked as superseded by a different feature name), then that would make sense.

features/draft/vendor/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This directory is for features proposed by a browser vendor, to support the
+early use of web-features before there is a published web-features ID, and a
+way to transition from draft to published feature IDs.
+
+A draft vendor feature goes through these stages:
+
+- A bot creates a `.yml` file in this directory, with enough information for a
+  maintainer of web-features to follow up and act on. An `explainer` or `spec`
+  URL and a vendor-specific URL like `chromestatus` is recommended.
```
ddbeck: It would be nice to have an example file for this.

Even better, it would be nice if we set a floor for these files, to make it slightly easier to generate PRs based on them. I imagine the files would have the following properties/values:

- `kind: feature`
- `proposal` URL (e.g., explainers, spec PRs, etc.) / `spec` URL (if already in a regular spec)
-  … (optionally, anything else you want to add that's in our regular schema)
- Any vendor-specific strings/URLs in fields prefixed with `vendor_` (e.g., `vendor_chromestatus`, `vendor_bug`, etc.)

rdvnnl (approved):

foolip:
> If this proposal amounts to features getting an like `vendor/foo` which can be used as a real id and later renamed as normal (or marked as superseded by a different feature name), then that would make sense. 

That is effectively the proposal, yes. This PR doesn't say how to refer to these features externally, but `vendor/foo` SGTM.

jgraham:
> That is effectively the proposal, yes.

I think we should make that explicit, and ensure that we have group agreement on the proposal. Having a feature lifecycle that makes sense across a range of usecases feels like the important part; exactly how we represent that in the repo feels like implementation details.

features/draft/vendor/README.md
```diff
@@ -0,0 +1,18 @@
...
+- A web-features maintainer creates a real feature entry and replaces the draft
+  feature with a pointer to that feature.
```
ddbeck: It would be nice to have a stub/example file showing a mapping from a completed draft to a real feature ID.

Or we could have a JSON file showing former vendor draft paths (this would be nicer for us, to have the folder contain only un-minted features, but would require vendors to do a little more than not overwrite an existing file).

ddbeck (commented):
I'll start with my assumptions. I'm expecting a loop like this:

1. Vendor creates a draft `.yml` file for vendor-originating feature X.
2. web-features mints a real feature Y.
3. web-features deletes draft X and records somewhere that draft X became real feature Y.
4. Vendor checks that record and replaces their feature X identifier with real feature identifier Y (or however they want to handle that mapping on their end).

I'm also assuming that we do not expect vendors to do any reconciliation with each other (we'll do that when we mint a feature). For example, two standards positions requests on the same day could trigger the creation of two vendor drafts referring to the same feature and we don't care, as long as everyone's generating unique vendor draft IDs.

If that's right, then I think this is a good place to start. I've left some comments on some specifics.

ddbeck:
Also, FWIW, I have been reading "vendor" in the sense of "vendorization" and not that these things would be the sole domain of browser vendors. If, for instance, WPT or State of… surveys wanted to get in on this process, then that could happen here just as well.

foolip:
@ddbeck and I live reviewed this just now. I've updated the README to say that `vendor/a-feature` is how you can refer to these entries, and wrote an example for how I expect this to be used.

Note that I don't think we need to publish these anywhere to start out.

features/draft/vendor/README.md
ddbeck: This has been resolved below. After the live review session with Philip, I was brought around to the idea that the resolution of a draft can and should happen in-place, rather than deleting or moving the files. While this is less convenient for me personally as a maintainer, it's less complex for the bots that are going to commit this stuff (e.g., the process for creating a draft file is, more or less, "don't clobber an existing file").

features/draft/vendor/README.md
```diff
@@ -0,0 +1,44 @@
...
+When the real web-features entry is created, it is decided to consider this part
+of `<template for>`, so the draft vendor feature is updated like this:
+
+```yml
+# features/draft/vendor/html-pis.yml
+kind: moved
+redirect_target: template-for
+```
+
+Any references to `vendor/html-pis` can be updated to refer to `template-for`.
```
ddbeck: It's my expectation that we'll revise this to use a fully-realized example, once we've been through it once or twice.

ddbeck (approved):
I'm much happier with the examples here. I think this is enough for us to go on to start experimenting with this.

ddbeck (approved):
I'm much happier with the examples here. I think this is enough for us to go on to start experimenting with this.

features/draft/vendor/README.md
```diff
@@ -0,0 +1,44 @@
+# Draft vendor features
+
+This directory is for features proposed by a browser vendor, to support the
```
jgraham: Why only browser vendors?

features/draft/vendor/README.md
```diff
@@ -0,0 +1,44 @@
...
+early use of web-features before there is a published web-features ID, and a
```
jgraham: If we don't publish these anywhere we still can't reference them from e.g. standards positions without recreating all the build infrastructure to know what are valid definitions.

features/draft/vendor/README.md
```diff
@@ -0,0 +1,44 @@
...
+Write access for bots can be granted by web-features maintainers, and should
+only be used to create files in this directory.
+
+## Example
+
+The [processing instructions in HTML](https://chromestatus.com/feature/6534495085920256)
+feature is created as `vendor/html-pis.yml` file with this contents:
+
+```yml
+# features/draft/vendor/html-pis.yml
```
jgraham: I think if the subdirectory is `vendor` then there should be a vendor name for each proposal e.g. `vendor/google/html-pis`.

jgraham (dismissed):
I think this needs to consider the workflow for consumption, rather than just the publishing workflow. If these feature names are going to appear anywhere in public (explainers, s-p requests, status counters, intent emails), it needs to be possible for people to reference them just like other feature ids.

features/draft/vendor/README.md
foolip: It doesn't have to be, we can change this to allow anyone in principle. Possibly there could be a bot that looks for a specific format in WPT's WEB_FEATURES.yml files, for example.

features/draft/vendor/README.md
foolip: If you expect to reference these in standards positions, we can publish somewhere. How about a JSON file on GitHub pages that's updated for every commit to this repo?

features/draft/vendor/README.md
foolip: If it's just about the naming, could we call it `tentative` or `proposed` instead and not have any further nesting?

features/draft/vendor/README.md
jgraham: Well, I expect _you_ to reference them in standards positions (requests) :p 

But yes, publishing them separately to the main ids would be fine for me.

features/draft/vendor/README.md
jgraham: I quite like `proposed`.

features/draft/vendor/README.md
foolip: I removed "by a browser vendor" in the rewrite from vendor to proposed.

features/draft/vendor/README.md
foolip: I've made the change.

features/draft/vendor/README.md
foolip: I see, I wasn't sure how you plan to reference web-features in https://github.com/mozilla/standards-positions. If you add it to the template, then allowing proposed/* IDs there would make sense.

foolip:
What's still missing here is building `data.proposed.json` and including it as a GitHub release artifact. I'm thinking it should be an object of this shape, using the README example:

```json
{
  "html-pis": {
    "explainer": "https://github.com/WICG/declarative-partial-updates/blob/main/patching-explainer.md",
    "chromestatus": "https://chromestatus.com/feature/6534495085920256"
  }
}
```

If a file doesn't parse into an object, that should be a warning where maintainers will see it, but not block releases. An empty object would be used, so that consumers of `data.proposed.json` can always assume an object but never assume any particular properties on the object.

jgraham:
As a consumer, something that exactly matches the normal format, but with all the fields optional (for feature definitions; "metadata: entries like moves should be identical to the normal format) would be ideal.

foolip:
@jgraham that sounds reasonable, we'd just add a top-level "features" object I think, and maybe a few other things that consumers of the current `data.json` would expect.

ddbeck:
@foolip and I had a conversation today about how to make the 'all fields optional' stuff work cooperatively with the existing schema. We talked through one idea that goes something like this:

Create a new superset schema with a `kind: proposed` feature type. `kind: proposed` would be equivalent to `kind: feature`, except all fields would be optional. On each release of web-features, we'd publish `data.proposed.json` and a `schema.proposed.json`, for consumers that need `kind: proposed` data.

On publishing the mainstream `web-features` package, we'd generate `data.schema.json`, which would be the superset schema with `kind: proposed` excised at build time (i.e., exactly equivalent to today's `data.schema.json`).

foolip:
@ddbeck I've added the bits for publishing `data.proposed.json` now. The schema is hard coded.

.github/workflows/publish_web-features.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
       - run: npm ci
       - run: npm run build
-      - run: gh release upload ${{ github.ref_name }} packages/web-features/data.json data.extended.json schemas/data.schema.json
+      - run: gh release upload ${{ github.ref_name }} packages/web-features/data.json data.extended.json schemas/data.schema.json data.proposed.json schemas/data.proposed.schema.json
```
ddbeck, line 36: A nice to have thing here would be to extend this to [`.github/workflows/publish_next_web-features.yml`](https://github.com/web-platform-dx/web-features/blob/main/.github/workflows/publish_next_web-features.yml).

ddbeck (approved):
I like this. I would like to get rid of the hard-coded schema, but I'm willing to take that part on (see https://github.com/web-platform-dx/web-features/pull/3726#issuecomment-4067783986 for my proposed approach), if you don't have the time for it.

One optional suggestion in a line comment.

.github/workflows/publish_web-features.yml
foolip: Looks like it's already there, at the bottom of the file.

ddbeck:
I think this is ready to merge (I believe both my and @jgraham's concerns were addressed sufficiently). I'll leave this open for 24 hours, but if there's no commentary, then I'll go ahead and merge this.

PR #3732, opened 2026-01-29
Document the earliest Baseline (low) date

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/623.

### DIFF
--- a/docs/baseline.md
+++ b/docs/baseline.md
@@ -114,6 +114,9 @@ For each feature definition in `web-features`, an interoperable status shall be 
 If the feature has the interoperable status, then the feature’s _keystone date_ shall be set to the last release date on which a browser introduced support for the feature
 If there was more than one introduction (e.g., a feature was withdrawn then reintroduced), then only the latest date applies.
 
+> [!NOTE]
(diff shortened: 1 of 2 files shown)


captainbrosset:
> Thanks for documenting this!
>  I'm rooting for some younger browsers and we will see what happens here then :)

What do you mean by younger browsers though? That first date is when Edge got introduced, so we can't really have Baseline features before that date. 
We do have the BCD data for features that got implemented in certain other browsers before 2015. But these features were just not Baseline at the time. 
See https://web-platform-dx.github.io/web-features-explorer/release-notes/october-2011/ as an example.

Elchi3:
I guess I'm hypothetical: If Servo becomes a baseline browser, with v1 released on 01-01-2030, then the earliest_baseline_date would be 01-01-2030 according to this logic, or maybe I am missing something?

captainbrosset:
If we changed the definition of Baseline to ass Servo to the core browser set, then yes. Or, perhaps more realistically, we'd create another, parallel, definition of Baseline that accounts for a different browser set.

Issue #3707, opened 2026-01-21
<geolocation> element


rviscomi:
### Specification

https://wicg.github.io/PEPC/permission-elements.html

### Description

The `<geolocation>` HTML element is a new declarative tag that allows websites to request location access through a user-initiated button rather than traditional script-triggered prompts. This approach simplifies development by reducing boilerplate code for permission handling and improves the user experience by providing clear intent and avoiding unexpected permission blocks.

### Documentation

https://developer.chrome.com/blog/geolocation-html-element

### Browser support

Available in Chrome 144

PR #3757, opened 2026-02-11
Add `<geolocation>` HTML element feature and `geolocation` group

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3707


features/geolocation.yml
rviscomi: WDYT about renaming this "Geolocation API" to help disambiguate from the HTML element?

tomayac: `HTMLGeolocationElement` is the technically correct name.

ddbeck: Yes, I think disambiguating the Geolocation API is a good idea here. I've done that in 26724220f.

Issue #3560, opened 2025-11-21
Add feature(s) for translator and language detector APIs


ddbeck:
### Specification

[Translator and Language Detector APIs](https://webmachinelearning.github.io/translation-api/)

### Description

I learned of this via https://github.com/web-platform-dx/web-features/issues/3541, so I don't know a lot about it. From the spec:

> The translator and language detector APIs expose the ability to translate text between human languages, and detect the language of such text.

It's likely that it's actually two features, one for translation and one for language detection (given the two Chrome Platform Status entries, it's easy to imagine scenarios where you might do one but not the other).

### Documentation

- [Translator and Language Detector APIs - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Translator_and_Language_Detector_APIs)
- [webmachinelearning/translation-api: 🌏 A proposal for translator and language detector APIs](https://github.com/webmachinelearning/translation-api)

### Browser support

- [`/features/draft/spec/translation-api.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/translation-api.yml) ([dist](https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/translation-api.yml.dist))
- [Language Detector API - Chrome Platform Status](https://chromestatus.com/feature/6494349985841152)
- [Translator API - Chrome Platform Status](https://chromestatus.com/feature/5172811302961152)

ddbeck:
Not to be confused with https://github.com/web-platform-dx/web-features/issues/3502, which is for optical character recognition.

PR #3758, opened 2026-02-11
Add translator and language detector API features

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3560

### DIFF
--- a/features/translator.yml
+++ b/features/translator.yml
@@ -1,5 +1,5 @@
 name: Translator
-description: The `Translator` API translates some text from one natural languages to another, using machine translation.
+description: The `Translator` API translates some text from one natural language to another, using machine translation.
 spec: https://webmachinelearning.github.io/translation-api/#translator-api
 compat_features:
   - api.Translator


features/translator.yml
```diff
@@ -0,0 +1,15 @@
+name: Translator
+description: The `Translator` API translates some text from one natural languages to another, using machine translation.
```
ddbeck:
```suggestion
description: The `Translator` API translates some text from one natural language to another, using machine translation.
```

ddbeck:
> Should the existing features in `features/draft/spec/translation-api.yml` be removed?

@reillyeon The files in `features/draft/spec/` are automatically generated and updated, so that will get swept away after this PR merges.

features/translator.yml.dist
```diff
@@ -0,0 +1,18 @@
+# Generated from: translator.yml
+# Do not edit this file by hand. Edit the source file instead!
+
+status:
+  baseline: false
+  support:
+    chrome: "138"
```
captainbrosset, line 7: Translator is supported in Edge too, see https://learn.microsoft.com/en-us/microsoft-edge/web-platform/translator-api.
I'll file a BCD issue.

captainbrosset: See https://github.com/mdn/browser-compat-data/issues/29058

ddbeck: Successor is https://github.com/mdn/browser-compat-data/issues/28839. We'll pick up the data when it lands in BCD.

Issue #3636, opened 2025-12-17
MetaTextScale as name for <meta name="text-scale" content="scale">


davidsgrogan:
### Specification

https://drafts.csswg.org/css-fonts-5/#text-scale-meta
https://github.com/w3c/csswg-drafts/pull/13052/changes

### Description

A new set of HTML meta attribute values that tells the user agent to incorporate the user's font preferences into the CSS font-size keywords.

### Documentation

https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/meta-text-scale.md

### Browser support

- https://chromestatus.com/feature/5112244702674944
- https://issues.chromium.org/issues/430566925

ddbeck:
Thank you for filing this issue, @davidsgrogan. I've added links to a Chrome bug and status entry that I found for this feature.

Based on other the entries for `<meta name="…">` features, this will probably have the feature ID `meta-text-scale`.

I'd expect the following BCD keys for this feature:

- html.elements.meta.name.text-scale (definitely)
- html.elements.meta.name.text-scale.legacy (maybe, possibly elsewhere in the tree or not at all)
- html.elements.meta.name.text-scale.scale (maybe, possibly elsewhere in the tree or not at all)

progers:
@ddbeck , can we rename the chromium webdx feature name from `kDRAFT_MetaTextScale` to `kMetaTextScale` now, or is there more discussion needed? Chromium's internal docs say "Once the community group has agreed on the identifier name, remove DRAFT_ from the use counter name..."

ddbeck:
@progers thanks for your patience on this. I've opened https://github.com/web-platform-dx/web-features/pull/3759 to resolve this. I suggest you watch for this issue to close or that PR to merge, to start using the new ID. And if you're interested in this feature in particular, I'd welcome your review on the feature description. Cheers!

PR #3759, opened 2026-02-11
Add `<meta name="text-scale">` feature

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3636

### DIFF
--- a/features/meta-text-scale.yml
+++ b/features/meta-text-scale.yml
@@ -1,7 +1,7 @@
 name: <meta name="text-scale">
 description: >
   The `<meta name="text-scale" content="scale" />` HTML element allows the browser's initial font size to be affected by the operating system text scale settings.
-  The `<meta name="text-scale" content="legacy" />` element sets the initial font size to 16px.
+  The `<meta name="text-scale" content="legacy" />` element is the default behavior that respects only browser font-size settings.
 spec: https://drafts.csswg.org/css-fonts-5/#text-scale-meta
 # Expected:
 # compat_features:


features/meta-text-scale.yml
```diff
@@ -0,0 +1,8 @@
+name: <meta name="text-scale">
+description: >
+  The `<meta name="text-scale" content="scale" />` HTML element allows the browser's initial font size to be affected by the operating system text scale settings.
+  The `<meta name="text-scale" content="legacy" />` element sets the initial font size to 16px.
```
davidsgrogan: This part isn't exactly accurate. 'legacy' is just a no-op, it lets the browser do its default behavior, which isn't always 16px as the initial font size. Maybe you can say
`<meta name="text-scale" content="legacy" />` is the default: the browser ignores OS-level text scaling but respects UA-level font size settings

features/meta-text-scale.yml
```diff
@@ -0,0 +1,8 @@
+name: <meta name="text-scale">
+description: >
+  The `<meta name="text-scale" content="scale" />` HTML element allows the browser's initial font size to be affected by the operating system text scale settings.
```
davidsgrogan, line 3: This is good

features/meta-text-scale.yml
ddbeck: @davidsgrogan Thank you, that's helpful. Now to fit it into the length limit and style guide. Something like this, maybe?

```suggestion
  The `<meta name="text-scale" content="legacy" />` element is the default behavior that respects only browser font-size settings.
```

davidsgrogan: Yes, great.

progers:
This wording LGTM as well!

PR #3772, opened 2026-02-16
Add manifest localization feature

### LINKED ISSUE #3412: Manifest localization
### Specification

https://www.w3.org/TR/appmanifest/#x_localized-members

### Description

Supports localization of manifest members so apps can adapt their names, descriptions, icons, and shortcuts to the user’s language and region. Developers provide localized values in the web app manifest, and the browser automatically selects the appropriate resources based on the user’s language settings, introducing language support across different markets.

### Documentation

No new explainer, agreed upon changes and design in the Web Application Manifest spec.

### Browser support

_No response_

captainbrosset:
Fixes #3412 by introducing a new feature with expected BCD keys (BCD does not yet have keys for localizable manifest members).

See spec at https://www.w3.org/TR/appmanifest/#x_localized-members.


features/manifest-localization.yml
```diff
@@ -0,0 +1,10 @@
+name: Web app manifest localization
+description: The localized members of the web app manifest allow you to provide translated names, descriptions, icons, and shortcuts. The browser automatically selects the appropriate resources based on the user's language and region settings.
```
ddbeck: It'd be nice to name the entry points here or at least make them a little more identifiable. Something like this?

(also a nit: avoiding "allows")

```suggestion
description: The `_localized` suffixed members of the web app manifest set translated names, descriptions, icons, and shortcuts. The browser automatically selects resources based on the user's language and region settings.
```

Issue #3777, opened 2026-02-17
WebMCP


beaufortfrancois:
### Specification

https://webmachinelearning.github.io/webmcp/

### Description

WebMCP is a proposed web standard that exposes structured tools for AI agents on existing websites. This would replace "screen-scraping" with robust, high-performance page interaction and knowledge retrieval. WebMCP provides JavaScript and annotates HTML form elements so that agentic browsers know exactly how to interact with page features to support a user's experience.

### Documentation

https://github.com/webmachinelearning/webmcp
https://developer.chrome.com/blog/webmcp-epp

### Browser support

Chrome Canary behind a flag

https://chromestatus.com/feature/5117755740913664

PR #3779, opened 2026-02-17
Add `navigator.modelContext` (WebMCP) feature

beaufortfrancois:
FIX https://github.com/web-platform-dx/web-features/issues/3777


beaufortfrancois:
Tests will pass once https://github.com/w3c/browser-specs/issues/2365 is resolved.

features/webmcp.yml
```diff
@@ -0,0 +1,3 @@
+name: WebMCP
+description: Exposes structured tools for AI agents on existing websites.
```
ddbeck: I am suggesting describing this in terms of entrypoints: what named thing does a developer invoke to bring about what consequence? I've ended up here:

```suggestion
description: The `navigator.modelContext` API registers functions to be called by tools external to the page, such as browser extensions or in-browser generative AI models.
```

That said, I feel like this is a little speculative, since the spec is quite incomplete and the other links given in the issue don't really explain what the _API_ does (as opposed to tools that might call upon its registry).

beaufortfrancois: This is actually more than that. WebMCP provides JavaScript **and annotates HTML form elements** so that agentic browsers know exactly how to interact with page features to support a user's experience.

ddbeck: Ah, interesting! Do you know what the likely method for annotations will take? Is it an API or HTML attributes or something else?

beaufortfrancois: HTML form attributes like `toolname` and `tooldescription`.  See https://github.com/webmachinelearning/webmcp/pull/76

beaufortfrancois: @ddbeck gentle ping

ddbeck: @captainbrosset can I get your help working on the description here? I'm volunteering you since you've written about this recently. 😈 And I'm having a hard time wrapping my head around this feature. Or maybe it's multiple features. It's unclear to me.

captainbrosset: I've been focusing on the imperative part of the API only so far, and haven't looked into the declarative part (HTML form attributes).

We could create two separate features for these parts by the way.
It feels to me that early testers have mostly been using and talking about the `navigator.modelContext` API so far, so it would make sense to create a feature just for that.

I'm unclear on whether the HTML form attributes have shipped anywhere yet. If they have, then I can be convinced otherwise.
If they haven't, then I vote for keeping Daniel's suggested description changes and then merging the PR.

ddbeck: In that case, if we're going to have the potential of two or more WebMCP features, then I suggest we take a slightly different approach, so we can hold `webmcp` in reserve (to either act as an umbrella feature or the thing this is ultimately renamed to).

Rename this file to `navigator-modelcontext.yml` and do this:

```
name: navigator.modelContext
description: The `navigator.modelContext` API registers WebMCP functions to be called by tools external to the page, such as browser extensions or in-browser generative AI models.
```

beaufortfrancois: FYI Chromium has working support for both [declarative](https://webmachinelearning.github.io/webmcp/#declarative-api) and [imperative](https://webmachinelearning.github.io/webmcp/#model-context-container) shapes of the WebMCP API. 
You can try [demos](https://github.com/GoogleChromeLabs/webmcp-tools?tab=readme-ov-file#demos) in Chrome Canary with the "WebMCP for testing" flag .

captainbrosset: Ok, thank you. I had no idea. Let's make a single feature for both shapes of the API then.

captainbrosset:
```suggestion
description: The `navigator.modelContext` API registers page actions that can be called by in-browser AI agents or browser extensions. You can also register actions declaratively with the `toolname` and `tooldescription` HTML form field attributes.
```

captainbrosset: @ddbeck I'd be tempted to keep the ID as is. Because it's an early feature, I see more value in getting it in now in _some_ form rather than sweating the details.
It'll most likely change shape over the coming years, and additional sub-features will most likely get added.
But it seems to be the right first version of this feature for now.

captainbrosset: This was just discussed on the WebDX CG call, and we decided to keep this feature focused on just the imperative navigator.modelContext API. We can then create another feature for the declarative shape of the API.
That way, we can make the ID be something like `navigator-modelcontext`, as proposed by Daniel before, which will simplify our life a lot when/if the feature evolves.

@beaufortfrancois would you like to implement the recommended changes? If not, let me know, I can take it from here.

beaufortfrancois:
Should I create three features then? Something like `navigator-modelcontext` and `webmcp-tools` that would be children of `webmcp`?

captainbrosset:
I think we can start with just one feature in this PR and go from there. So just the navigator-modelcontext feature.

beaufortfrancois:
I've updated this PR.

beaufortfrancois:
Can someone unblock this?

features/modelcontext.yml
```diff
@@ -0,0 +1,4 @@
+name: WebMCP
```
captainbrosset:
```suggestion
name: navigator.modelContext
```

features/modelcontext.yml
```diff
@@ -0,0 +1,4 @@
...
+description: The `navigator.modelContext` API registers page actions that can be called by in-browser AI agents or browser extensions.
```
captainbrosset:
```suggestion
description: The `navigator.modelContext` API registers page actions that can be called by in-browser AI agents or browser extensions. Also known as WebMCP.
```

beaufortfrancois:
Shall I start another one for declarative WebMCP?

captainbrosset:
Merged. Thank you @beaufortfrancois and sorry this has been slow-going.

PR #3792, opened 2026-02-20
Add Navigation API precommit handler feature

captainbrosset:
See:

* https://chromestatus.com/feature/5134734612496384
* https://developer.mozilla.org/en-US/docs/Web/API/NavigationPrecommitController
* https://html.spec.whatwg.org/multipage/nav-history-apis.html#navigationprecommithandler

Adding this feature would make it possible to map to [one more Interop 2026 focus area](https://github.com/web-platform-tests/interop/tree/main/2026#navigation-api).

### DIFF
--- a/features/navigation-precommit-handlers.yml
+++ b/features/navigation-precommit-handlers.yml
@@ -1,5 +1,5 @@
 name: Navigation precommit handlers
-description: The `precommitHandler` option of `navigateEvent.intercept()` defers committing a navigation (updating the URL and history) until the handler resolves. This allows changing the navigation's URL, state, and history handling behavior before the commit occurs.
+description: The `precommitHandler` callback option to `NavigateEvent`'s `intercept()` method returns a promise that defers navigation until the promise resolves. You can use this to change the navigation's URL, state, and history before navigation occurs.
 spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationinterceptoptions-precommithandler
 compat_features:
   - api.NavigateEvent.intercept.precommitHandler_option


features/navigation-precommit-handlers.yml
```diff
@@ -0,0 +1,8 @@
+name: Navigation precommit handlers
+description: The `precommitHandler` option of `navigateEvent.intercept()` defers committing a navigation (updating the URL and history) until the handler resolves. This allows changing the navigation's URL, state, and history handling behavior before the commit occurs.
```
ddbeck: I read the MDN docs about this and liked how they phrased it as "just before the navigation has committed." That got me thinking about this revision, which also deals with some other style guide nits:

```suggestion
description: The `precommitHandler` option of a `NavigateEvent` object's `intercept()` method is a callback that runs just before the navigation has committed. You can use this to change the navigation's URL, state, and history before navigation occurs.
```

features/navigation-precommit-handlers.yml
```diff
@@ -0,0 +1,8 @@
...
+spec: https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationinterceptoptions-precommithandler
+compat_features:
+  - api.NavigateEvent.intercept.precommitHandler_option
+  - api.NavigationPrecommitController
+  - api.NavigationPrecommitController.addHandler
```
ddbeck, line 7: I considered that this might be a distinct feature, since it's a way to add a <b>post</b>commit handler from a precommit handler and it has a different support condition. But after thinking about it, I rejected the idea of a separate feature. Being an Interop feature, we're very likely to see this key become newly available along with the rest of the feature, so the distinction is probably of very fleeting importance.

captainbrosset: That's a good point, I was wondering about this too.
But you're right, it's in Interop, and there's a specific test for `addHandler` here: https://wpt.fyi/results/navigation-api/precommit-handler/precommitHandler-addHandler-throws.html?label=experimental&label=master&aligned&q=label%3Ainterop-2026-navigation
So it'll get interoperable at some point this year.

features/navigation-precommit-handlers.yml
captainbrosset: I like this, but I'm missing one important aspect I think:

The callback must return a promise for the Navigation API to know when to commit. This was sort of implied by using the word "resolves" in my original take.

Also, I find the "The Foo option of a Bar object's Baz method" a bit hard to parse. Would it be acceptable to go back to Bar.Baz() as in my initial description, but borrowing some of your wording around it?

```suggestion
description: The `precommitHandler` callback option of a `NavigateEvent.intercept()` method runs just before the navigation has committed, returning a promise which delays the commit until the promise has resolved. You can use this to change the navigation's URL, state, and history before navigation occurs.
```

ddbeck: Ah, that's a distinction I didn't get! Let me try this again.

Also, I agree that the "The Foo option of a Bar object's Baz method" is hard to parse. That said, our guidelines advise us to avoid non-literal code (see [this bit](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#:~:text=never%20use%20made-up%2C%20idiosyncratic%2C%20or%20non-literal%20example%20code)). And even if that weren't a barrier, I think `NavigateEvent.intercept()` implies a static method.

Maybe something like this is OK?

```suggestion
description: The `precommitHandler` callback option to `NavigateEvent`'s `intercept()` method returns a promise that defers navigation until the promise resolves. You can use this to change the navigation's URL, state, and history before navigation occurs.
```

Or maybe this:

```suggestion
description: The `precommitHandler` callback, an option to `NavigateEvent`'s `intercept()` method, returns a promise that defers navigation until the promise resolves. You can use this to change the navigation's URL, state, and history before navigation occurs.
```

PR #3800, opened 2026-02-23
`font-width`: Add `@font-face` descriptor to feature

### LINKED ISSUE #3795: Missing font-width @font-face descriptor feature
Currently the [font-width](https://github.com/web-platform-dx/web-features/blob/main/features/font-width.yml) feature only includes the CSS property. But there’s also a `font-width` descriptor. Both are supported only in Safari 18.4 since March 2025.

See BCD:

- [css/properties/font-width.json](https://github.com/mdn/browser-compat-data/blob/main/css/properties/font-width.json)
- [css/at-rules/font-face.json](https://github.com/mdn/browser-compat-data/blob/main/css/at-rules/font-face.json#L745)

Since [font-stretch](https://github.com/web-platform-dx/web-features/blob/main/features/font-stretch.yml) feature groups a property and a descriptor, I would expect the same from the font-width.

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3795

### DIFF
--- a/features/font-width.yml
+++ b/features/font-width.yml
@@ -2,7 +2,21 @@ name: font-width
 description: The `font-width` CSS property selects a font face from a font family based on width, either by a keyword such as `condensed` or a percentage.
 spec: https://drafts.csswg.org/css-fonts-4/#font-width-prop
 group: fonts
+status:
+  compute_from:
+    - css.properties.font-width
+    - css.properties.font-width.condensed
+    - css.properties.font-width.expanded
+    - css.properties.font-width.extra-condensed
+    - css.properties.font-width.extra-expanded
+    - css.properties.font-width.normal
+    - css.properties.font-width.semi-condensed
+    - css.properties.font-width.semi-expanded
+    - css.properties.font-width.ultra-condensed
+    - css.properties.font-width.ultra-expanded
+    - css.at-rules.font-face.font-width
 compat_features:
+  - api.CSSFontFaceDescriptors.fontWidth
   - css.properties.font-width
   - css.properties.font-width.condensed
   - css.properties.font-width.expanded
@@ -13,3 +27,4 @@ compat_features:
   - css.properties.font-width.semi-expanded
   - css.properties.font-width.ultra-condensed
   - css.properties.font-width.ultra-expanded
+  - css.at-rules.font-face.font-width
--- a/features/font-width.yml
+++ b/features/font-width.yml
@@ -16,6 +16,9 @@ status:
     - css.properties.font-width.ultra-expanded
     - css.at-rules.font-face.font-width
 compat_features:
+  # This one key is excluded from the `compute_from`, since it shipped much later, but doesn't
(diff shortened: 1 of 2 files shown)


features/font-width.yml
```diff
@@ -2,7 +2,21 @@ name: font-width
...
+    - css.properties.font-width.condensed
+    - css.properties.font-width.expanded
+    - css.properties.font-width.extra-condensed
+    - css.properties.font-width.extra-expanded
+    - css.properties.font-width.normal
+    - css.properties.font-width.semi-condensed
+    - css.properties.font-width.semi-expanded
+    - css.properties.font-width.ultra-condensed
+    - css.properties.font-width.ultra-expanded
+    - css.at-rules.font-face.font-width
```
captainbrosset, line 17: Wouldn't it make our life easier if we also had a `ignore_keys` field?

This feature basically uses a `compute_from` for _all_ keys but one, and it would be great if instead we could ignore that one only, with a comment.

Maybe file an issue to think about this? If you think this would be useful.

features/font-width.yml
```diff
@@ -2,7 +2,21 @@ name: font-width
...
 compat_features:
+  - api.CSSFontFaceDescriptors.fontWidth
```
captainbrosset, line 22: Can you please add a comment for this key? Something like:

"This key is excluded from the Baseline calculation using `compute_from`, above, because it shipped a little later than the rest of the keys in Safari, and it isn't worth delaying the feature for, etc...."

captainbrosset (approved):

features/font-width.yml
ddbeck: I think we'll need a more sophisticated answer for this to deal with feature merges, so the aspects "issue" is kinda representing this problem already (i.e., we'll name groups of keys to make the initial/later/related distinctions explicit and make it easier to move them around semi-automatically).

I can make a separate issue if you like, though.

features/font-width.yml
ddbeck, line 22:
```suggestion
  # This one key is excluded from the `compute_from`, since it shipped much later, but doesn't
  # represent a distinct feature. This may change if other browsers ship this feature and the
  # "floor" for this feature across browsers changes.
  - api.CSSFontFaceDescriptors.fontWidth
```

features/font-width.yml
Destop-Trang1801: ddbeck:3795-font-width

features/font-width.yml
```diff
@@ -2,7 +2,21 @@ name: font-width
 description: The `font-width` CSS property selects a font face from a font family based on width, either by a keyword such as `condensed` or a percentage.
 spec: https://drafts.csswg.org/css-fonts-4/#font-width-prop
 group: fonts
+status:
```
Destop-Trang1801, line 5: features/font-width.yml

Destop-Trang1801:
Vào Thứ 7, 7 thg 3, 2026 lúc 19:02 Ngô Trang ***@***.***> đã
viết:

PR #3813, opened 2026-02-24
Extract formData out of Fetch and into a new feature

captainbrosset:
The **Fetch uploads and ranges** Interop 2026 focus area focuses on ReadableStream in body, formData and mime-type, and the Range header (see [description](https://github.com/web-platform-tests/interop/tree/main/2026#fetch-uploads-and-ranges)).

For example, here are the formData-related tests which Interop 26 considers: https://wpt.fyi/results/fetch/api/body?label=master&label=experimental&aligned&q=label%3Ainterop-2026-fetch

The [fetch feature](https://web-platform-dx.github.io/web-features-explorer/features/fetch/) on the repo is quite big in that in contains many BCD keys, but hide most of them behind a `compute_from` directive which pins the status of the feature to the `api.fetch` BCD key.

In this PR, I'm extract 2 BCD keys from this big feature, and moving them into their own new feature. My reasons for doing so are:

1. This feature is tracked by Interop 2026.
2. Doing so doesn't regress the Baseline status of the fetch feature.
3. Being able to use `request.formData()` or `response.formData()` feels like an additive feature, not something you absolutely must use when using `Fetch()`.

### DIFF
--- a/features/fetch.yml.dist
+++ b/features/fetch.yml.dist
@@ -415,15 +415,6 @@ compat_features:
 
   # baseline: false
   # support:
-  #   firefox: "39"
-  #   firefox_android: "39"
-  #   safari: "14.1"
-  #   safari_ios: "14.5"
-  - api.Request.formData
-  - api.Response.formData
-
-  # baseline: false
-  # support:
   #   firefox: "111"
   #   firefox_android: "111"
   #   safari: "16.1"


ddbeck:
@captainbrosset can you say more about why this is in Interop? If this is a distinguishable feature from `fetch` _and_ in Interop, what behaviors (or bugs?) aren't represented in BCD? The status level (though not the date) for this new feautre is the same as `fetch`, so I'm thinking that if we're going to pluck it out, we should at least give consideration to minting a feature with a lower status.

captainbrosset:
> can you say more about why this is in Interop?

Because not all formData-related WPT tests pass on Chromium-based browsers.

```javascript
const res = new Response(new FormData());
res.formData().then(fd => console.log(fd));
```
```javascript
const req = new Request('about:blank', { method: 'POST', body: new FormData() });
req.formData().then(fd => console.log(fd));
```

The above code snippets work in Firefox and Safari, but not in Edge or Chrome. See the [WPT test results](https://wpt.fyi/results/fetch/api/body/formdata.any.html?label=master&label=experimental&aligned&q=label%3Ainterop-2026-fetch).

This inconsistent behavior is likely breaking existing sites, making it a good thing for Interop to focus on resolving this year.

> If this is a distinguishable feature from `fetch`

I don't know quite enough about how you'd use this in a real use case to 100% know if this is a distinguishable feature from `fetch`, but it sure seems like it. It feels like nice developer ergonomics to simplify submitting forms to the server. You can do it without formData, it's just harder/requires more code.

> what behaviors (or bugs?) aren't represented in BCD?

That's a great question. It does feel like a bug. The `formData()` method exists, but causes an error to be thrown in Chromium.

Maybe one of these:

* https://issues.chromium.org/issues/40065051
* https://issues.chromium.org/issues/40764130

It feels enough to mark Chrome's support as partial in BCD's `api.Request.formData` and `api.Response.formData` keys.

> if we're going to pluck it out, we should at least give consideration to minting a feature with a lower status.

If we make the BCD changes, then the feature status would be lower.

captainbrosset:
I filed the issue upstream in BCD: https://github.com/mdn/browser-compat-data/issues/29156

features/fetch-formdata.yml
```diff
@@ -0,0 +1,6 @@
+name: Request.formData() and Response.formData()
```
ddbeck: optional: What do you think of namechecking fetch, as in "Fetch form data" or "Fetch formData"?

I ask because when I searched Stack Overflow (and elsewhere, e.g., [chromestatus](https://chromestatus.com/feature/5687725978025984)), a lot of the questions mention fetch in the title.

(Depending on how this changes, the ID will need changing too.)

```suggestion
name: Fetch form data
```

features/fetch-formdata.yml
```diff
@@ -0,0 +1,6 @@
...
+description: The `Request.formData()` and `Response.formData()` methods read a request or response body and return a `Promise` that resolves to a `FormData` object. You can use it to access the fields of a form submission.
```
ddbeck: optional, wording finesse:


```suggestion
description: The `Request.formData()` and `Response.formData()` methods read a request or response body and returns a `FormData` promise. You can use it to access the fields of a form submission.
```

ddbeck (approved):
I'm happy with this in principle, but it'd be nice to align the ID with the name and maybe revise the name.

Optional suggestions in comments. Thank you!

features/fetch-formdata.yml
captainbrosset: I like it, but I think `return` needs to stay as is, since it's the Request.formData **and** Response.formData methods which return something.

```suggestion
description: The `Request.formData()` and `Response.formData()` methods read a request or response body and return a `FormData` promise. You can use it to access the fields of a form submission.
```

ddbeck:
Also, I sent https://github.com/mdn/browser-compat-data/pull/29160 upstream so we'll get a status distinction here.

captainbrosset:
The BCD update came through, so I've updated this branch and re-generated the dist file. I think we're good to go now.

PR #3816, opened 2026-02-25
Add crash report storage feature

### LINKED ISSUE #3468: Add feature for crash reporting storage API
### Specification

- https://github.com/WICG/crash-reporting/pull/37

### Description

Discovered via https://github.com/web-platform-dx/web-features/issues/3399 and Chromium use counters, so I don't actually know a lot about this. Here's my educated guess:

The `window.crashReport` object is a key-value store to record information about your application's state. If there's a crash, then the data in the key-value store is sent to your crash reporting endpoint, to help you pinpoint the cause of the crash.

### Documentation

- https://github.com/WICG/crash-reporting/blob/gh-pages/crashReport-explainer.md

### Browser support

- https://chromestatus.com/feature/6228675846209536 (145)
- https://github.com/mozilla/standards-positions/issues/1271
- https://github.com/WebKit/standards-positions/issues/528

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3468


features/reporting-crashes-storage.yml
```diff
@@ -0,0 +1,9 @@
+name: Crash report storage
+description: The `window.crashReport` object is a key-value store to record information about your application's state. If there's a crash, then the data in the key-value store is sent to your crash reporting endpoint, to help you pinpoint the cause of the crash.
```
captainbrosset, line 2: I think this API builds on top of the [Reporting API](https://web-platform-dx.github.io/web-features-explorer/features/reporting/), which should likely be named here.

captainbrosset: After re-reading a few resources, my understanding is now that:

1. The Reporting API is the higher-level feature, which basically consists in defining a `Reporting-Endpoints` header to say where the browser should send (multiple types of) reports to.
   That's https://web-platform-dx.github.io/web-features-explorer/features/reporting/

2. On top of this, the Crash Reporting API was defined. It's an extension of the Reporting API that's specifically made for reporting crashes, and which defines that reports contain data such as `reason`, `stack`, `is_top_level`, `visibility_state`, and `crash_report_api`.
   Spec: https://wicg.github.io/crash-reporting/
   We don't have a feature in the repo for this.

3. The `window.crashReport` object, which seems to be and extension of the Crash Reporting API that lets developers also send their own custom data along with a crash report, but creating a map of key/value pairs
   Spec: https://wicg.github.io/crash-reporting/#crash-report-storage
   That's the new `crash-report-storage` feature you're proposing here.

I'm not 100% confident that 2 and 3 are separate things, but they do seem this way to me.

That said, I don't know if there are BCD keys for 2.

captainbrosset: If there are no practical benefits to creating a feature for 2 _and_ 3, then we should at least revise the description to broaden the language so it encompasses the Crash Reporting API too.

captainbrosset (dismissed):

features/reporting-crashes-storage.yml
ddbeck: @captainbrosset I opened https://github.com/web-platform-dx/web-features/pull/3982 to rework the reporting features a lot. Want to take a look? Then we can come back to this one.

ddbeck: @captainbrosset https://github.com/web-platform-dx/web-features/pull/3982 has merged and now all that remains is to add the crash reporting storage feature. Want to take another look?

Elchi3 (approved):
LGTM, modulo one question: What about `http.headers.Reporting-Endpoints.crash-reporting`, is that intentionally not part of this feature?

ddbeck:
> LGTM, modulo one question: What about `http.headers.Reporting-Endpoints.crash-reporting`, is that intentionally not part of this feature?

Good question and that's correct. Actual crash reporting is part of [`reporting-crashes`](https://web-platform-dx.github.io/web-features-explorer/features/reporting-crashes/ "Crash reporting: Crash reporting sends a report to a URL nominated by the Reporting-Endpoints header when a user is unable to continue using the page because the browser or tab crashed.") ([source](https://github.com/web-platform-dx/web-features/blob/main/features/reporting-crashes.yml), [dist](https://github.com/web-platform-dx/web-features/blob/main/features/reporting-crashes.yml.dist)). This feature allows developers to add extra data to those reports.

(To ask a question not answered: I didn't combine the features because they have different browser support.)

Elchi3:
Great, makes sense. I think we can merge this by dismissing the re-review request for @captainbrosset 😋

ddbeck:
Thank you both! This is the culmination of a lot of work [in web-features](https://github.com/web-platform-dx/web-features/pull/3982), [in BCD](https://github.com/mdn/browser-compat-data/pulls?q=is%3Apr+Reporting-Endpoints+is%3Aclosed+sort%3Aupdated-desc+merged%3A2026-05-13..2026-04-08), and [in MDN](https://github.com/mdn/content/issues/43688), to make this stuff more legible to developers.

PR #3823, opened 2026-02-26
computeBaseline for WebViews - add WebViews to compat_keys

### LINKED ISSUE #3553: Add webview support data to web-features
Hi all, sorry it's taken me a few days to get this written up.

# Problem

`web-features` does not currently include data concerning the feature support of webviews. There are a subset of features in Baseline \- both newly and widely available \- that are not supported by the webview that corresponds to the browsers they are based on (Chrome Android and iOS Safari).

Webview developers thus have two possible negative outcomes when they try to use Baseline:

* They trust Baseline statuses for features and end up implementing something that doesn’t actually work in the webview they are targeting.  
* They are aware of the inconsistencies and consequently avoid using Baseline and have to depend on the same cognitively challenging workflow as before where they spend a lot of time looking at support tables.  They probably, as a result, avoid using newer features for the same reason that other web developers did before, i.e. too difficult to get a simple go/no-go decision for a feature.

I’ve created [a sheet that lists out all the features and their statuses](https://docs.google.com/spreadsheets/d/1_vHnPDgCKX55FGq8NfPTXQGiQUEbedpMQ-owDzJNnlQ/edit) as outlined below based on [a module that supplements web-features with BCD webview support data](https://github.com/tonypconway/web-features-plus-webview).

# Proposal \- Add webview data to `web-features` JSON

Adding webview support to `web-features` would make it significantly easier for technical writers and tools teams to indicate whether a feature is supported or not in webview.  
We should add a new `webview_support` objects alongside the existing `support` objects in the `web-feature` JSON output that explains the level of webview support for a given feature.  There are a few different ways we could do this.

## Feature level data

Based on my research, features exist in different states that we should reflect in the output data, broken out by `android` and `ios` support:

* Equivalent \- Chrome Android and/or Safari iOS do not support the feature and their corresponding webviews have the same level of support for all supported BCD keys in that feature, supported or not.  
  * It might make sense to break this out into two statuses (Equivalent-supported, equivalent-unsupported), or replace it with `high`, `low`, `false` based on the feature-level support.
* Partial \- Chrome Android and/or Safari iOS support the feature, and their corresponding webviews support a subset of the supported BCD keys but not all of them.  
* Unsupported \- Chrome Android and/or Safari iOS support the feature, but the webview does not support any of the keys.

In addition, the current data has an `ambiguous` status that indicates none of the above apply \- usually it applies to features where the webview supports *more* keys than the corresponding browser. Someone with more expertise would need to look at these and tell me what I’m not understanding\!

## BCD-key level data

In addition, we should add `webview_support` to each BCD key in the `by_compat_key` support object where we can.

It might also be worth including a gap if there’s a gap of more than e.g. 1 month between the release date of the webview version and browser version that first supported a feature.

## Format

Output should look something like:

```json
"feature-id": {
  ...
  "status": {
    "baseline": "high",
    "support": { ... } // browsers listed here
    "webview_support": {
      "all": "equivalent",
      "android": "equivalent",
      "ios": "equivalent",
    }
    "by_compat_key": {
      "mdn.browser.compat.key": {
        "baseline": "high",
        "support": { ... },
        "webview_support": {
          "android": "120",
          "ios": "15.4"
        }
      }
    }
  }
}
```

# Implementation details

My work in `web-features-plus-webview` is based on copying the logic of `compute-baseline` as possible \- i.e. treating `partial_implementation` and the presence of other such properties as unsupported \- without directly editing `compute-baseline`. I think it would be prudent to update `compute-baseline` to include a utility for calculating support for webviews and then mediating that to an overall status.  This could be optional when using `compute-baseline` in isolation, but always used in the `web-features` build process.

## Overrides for non-BCD features

Some web-features do not have any BCD keys, and are hard coded in a web-features `.yml` file \- specifically image formats (avif, webp) and HTTP2 and HTTP3.  At the moment, I have these set up as [overrides](https://github.com/tonypconway/web-features-plus-webview/blob/main/scripts/overrides.json) indicating full support which are [inserted and returned before the usual calculations take place](https://github.com/tonypconway/web-features-plus-webview/blob/924c2ddfb18bf6115bbdb0fab01f6c72303e037d/scripts/generate_data.js#L111-L116).  It would be worth discussing if there’s a more sustainable way to do this, either by hardcoding the same info into the `.yml` files or updating BCD to include these features.

# Other webviews

The above suggestions are based on Android and iOS webviews because they have comprehensive data in `@mdn/browser-compat-data`.  Other webviews will need more research to be fully supported, potentially involving testing or contributions from the vendors themselves:

* Windows Webview2 \- [this support article from MS](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/browser-features) details the differences between Webview2 and Edge.  It looks like we could probably handle this as an overrides subset and then copy the rest from Edge, but someone with more insight into Windows and webviews should weigh in.   
* Popular Chinese super-apps \- we should try to get folks from Tencent (WeChat), Alibaba (Alipay), Baidu and so on to contribute data on their underlying Chromium versions and what the feature delta is where possible.  
* FB/Instagram \- likewise, it would be valuable to get definitive statements from Meta on their support levels.  I may have a contact here, but any Meta folks reading this, please reach out\!

NiklasMerz:
This PR adds WebViews to the `by_compat_key` items for each web-features. For caniwebview.com this is really useful as we can now use computeBaseline for web-features to show the status on the site.

Here is a built version of the data.json file: 
[data.extended.json](https://github.com/user-attachments/files/25586256/data.extended.json)

_Should WebViews be included in the top-level browsers key and their status in the web-features?_

Closes #3553

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -7,7 +7,7 @@ import YAML from 'yaml';
 import { convertMarkdown } from "./text";
 import { GroupData, SnapshotData, WebFeaturesData } from './types';
 
-import { BASELINE_LOW_TO_HIGH_DURATION, coreBrowserSet, getStatus, parseRangedDateString } from 'compute-baseline';
+import { BASELINE_LOW_TO_HIGH_DURATION, allIdentifiers, getStatus, parseRangedDateString } from 'compute-baseline';
 import { Compat } from 'compute-baseline/browser-compat-data';
 import { assertRequiredRemovalDateSet, assertValidFeatureReference } from './assertions';
 import { isMoved, isOrdinaryFeatureData, isSplit } from './type-guards';
@@ -262,7 +262,7 @@ for (const [id, feature] of Object.entries(features)) {
 
 const compat = new Compat();
 const browsers: Partial<WebFeaturesData["browsers"]> = {};
-for (const browser of coreBrowserSet.map(identifier => compat.browser(identifier))) {
+for (const browser of allIdentifiers.map(identifier => compat.browser(identifier))) {
     const { id, name } = browser;
     const releases = browser.releases.filter(release => !release.isPrerelease()).map(release => ({
         version: release.version,
--- a/index.ts
+++ b/index.ts
@@ -7,7 +7,7 @@ import YAML from 'yaml';
 import { convertMarkdown } from "./text";
 import { GroupData, SnapshotData, WebFeaturesData } from './types';
 
-import { BASELINE_LOW_TO_HIGH_DURATION, allIdentifiers, getStatus, parseRangedDateString } from 'compute-baseline';
+import { BASELINE_LOW_TO_HIGH_DURATION, coreBrowserSet, getStatus, parseRangedDateString } from 'compute-baseline';
 import { Compat } from 'compute-baseline/browser-compat-data';
 import { assertRequiredRemovalDateSet, assertValidFeatureReference } from './assertions';
 import { isMoved, isOrdinaryFeatureData, isSplit } from './type-guards';
@@ -262,7 +262,7 @@ for (const [id, feature] of Object.entries(features)) {
 
 const compat = new Compat();
 const browsers: Partial<WebFeaturesData["browsers"]> = {};
-for (const browser of allIdentifiers.map(identifier => compat.browser(identifier))) {
+for (const browser of coreBrowserSet.map(identifier => compat.browser(identifier))) {
     const { id, name } = browser;
     const releases = browser.releases.filter(release => !release.isPrerelease()).map(release => ({
         version: release.version,
--- a/packages/compute-baseline/src/baseline/core-browser-set.ts
+++ b/packages/compute-baseline/src/baseline/core-browser-set.ts
@@ -10,6 +10,16 @@ export const identifiers = [
   "safari_ios",
 ];
 
+export const webViewIdentifiers = ["webview_android", "webview_ios"];
+
 export function browsers(compat: Compat) {
   return identifiers.map((b) => compat.browser(b));
+}
+
+export function webViews(compat: Compat) {
+  return webViewIdentifiers.map((b) => compat.browser(b));
+}
+
+export function allBrowsers(compat: Compat) {
+  return [...browsers(compat), ...webViews(compat)];
 }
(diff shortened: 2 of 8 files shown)


NiklasMerz:
Screenshot that shows webviews are added to the browser list and the support keys:

grafik

ddbeck:
Some general feedback on this idea, though I feel positively toward computing support summaries for webivews in general:

- The smaller problem is that doing this primarily via `compute-baseline` complicates the dist files: each `.dist` file gets modified to show the webview support information. Those dist files are how we review status changes to Baseline features and adding a lot of noise to those files would be… challenging. Fortunately, I think this is solvable. For example, we could refactor this to generate webviews data when we ingest dist files or when we build the release.

- This approach does reveal an assumption in the schema, which is a bigger problem. The implication is that the keys of `features["somefeature"].status.support` **are** the Baseline browser set. I don't think there's (yet) an appetite for modifying the Baseline definition to include webviews, so I imagine we'd get some pushback if we added browsers to the `support` object that do not contribute to a status.

  But since the goal of this work in the near term is to make it easier to build Baseline-inspired (or caniuse-inspired tools) that are distinct from Baseline itself, I think there are some avenues to make that happen. To me, the most straightforward seem to be:
  
  - Generate a separate parallel JSON artifact that doesn't make status claims (e.g., it generates data for every browser in BCD, but omits the `baseline`, `baseline_low_date`, and `baseline_high_date` fields).
  
  - Modify the schema, to separate the Baseline browser set from the non-Baseline browser set. For example, add something like `features["somefeature"].status.support.noncore` or `features["somefeature"].extended_support`, where data for browsers outside the core browser set goes.

NiklasMerz:
I just refactored it to our latest discussion. The support info is now not part of `support` but gets a new property for non-core browsers. It's also non part of the dist files now.

**Naming things is hard**. I settled on `ecosystem_support` but I'm not that happy. The other suggestion I have is `support_beyond` but that's worse.
@ddbeck Or should we go with `extended_support` or `support_noncore`?

ddbeck:
I think at this point I'd like to get some feedback from consumers about where they'd expect to find support data that is not used in setting a Baseline status. This might be a good topic to bring up at the next WebDX web-features call (one week from today, April 9).

NiklasMerz:
Definitely! I'll make sure to be there. I have mixed feelings if this changes is expected by users as well and would love to find out if and how we can compute Baseline for non-core browsers most effectively.

tidoust:
The group discussed integration of WebView data separately, the actual shape may be completely different. Closing this PR accordingly.

PR #3841, opened 2026-03-05
Break out `text-decoration-skip-ink` and `text-decoration-skip-ink: all` into separate features

ddbeck:
This mints two features:

- `text-decoration-skip-ink`
- `text-decoration-skip-ink-all`

I think makes more sense, since the `text-decoration` shorthand does not affect `text-decoration-skip-ink`.

`text-decoration-skip-ink: all` has a rather different support story than `auto` and `none`, so I've split it here too.

### DIFF
--- a/features/text-decoration-skip-ink.yml
+++ b/features/text-decoration-skip-ink.yml
@@ -1,5 +1,5 @@
 name: text-decoration-skip-ink
-description: The `text-decoration-skip-ink` CSS property sets whether to interrupt underlines and overlines where the line would cross a glyph.
+description: The `text-decoration-skip-ink` CSS property sets whether underlines and overlines may be interrupted where the line would cross a glyph.
 spec: https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-ink-property
 compat_features:
   - css.properties.text-decoration-skip-ink


features/text-decoration.yml.dist
```diff
@@ -127,14 +112,6 @@ compat_features:
   - css.properties.text-decoration-line.grammar-error
   - css.properties.text-decoration-line.spelling-error
```
captainbrosset, line 113: Feels like these keys could also be a separate feature.

features/text-decoration.yml.dist
captainbrosset, line 121: MDN says:

> Note: Most other browsers are converging on supporting the simpler [text-decoration-skip-ink](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-decoration-skip-ink) property.

The [spec](https://drafts.csswg.org/css-text-decor-4/#text-decoration-skipping) says:

> The CSSWG resolved to be split skipping functionality into individual properties along the lines of [text-decoration-skip-ink](https://drafts.csswg.org/css-text-decor-4/#propdef-text-decoration-skip-ink), to improve its cascading behavior. See [discussion](https://github.com/w3c/csswg-drafts/issues/843) and [resolution](https://lists.w3.org/Archives/Public/www-style/2017Feb/0049.html). This section is a rough draft and has not yet been vetted by the CSSWG

Feels like we should move these keys to a new, discouraged, feature.

features/text-decoration-skip-ink.yml
```diff
@@ -0,0 +1,7 @@
+name: text-decoration-skip-ink
+description: The `text-decoration-skip-ink` CSS property sets whether to interrupt underlines and overlines where the line would cross a glyph.
```
captainbrosset: MDN says:

> the browser _may_ interrupt underlines and overlines

Somehow preserving the optional nature of the interruption would be good, and help differentiate this from the text-decoration-skip-ink-all feature.

Not sure how though.

ddbeck: Good catch. I'll make this edit:

```suggestion
description: The `text-decoration-skip-ink` CSS property sets whether underlines and overlines may be interrupted where the line would cross a glyph.
```

features/text-decoration.yml.dist
ddbeck: For this and the other separate feature, I'll send a new PR.

features/text-decoration.yml.dist
ddbeck: Sent https://github.com/mdn/browser-compat-data/pull/29199 upstream to figure out if we can do a discouraged feature here.

features/text-decoration.yml.dist
ddbeck: Sent https://github.com/web-platform-dx/web-features/pull/3846

PR #3858, opened 2026-03-12
Add Declarative WebMCP API feature

beaufortfrancois:
FIX https://github.com/web-platform-dx/web-features/issues/3777

Here's the explainer: https://github.com/webmachinelearning/webmcp/blob/main/declarative-api-explainer.md

### DIFF
--- a/features/declarative-webmcp.yml
+++ b/features/declarative-webmcp.yml
@@ -1,4 +1,4 @@
 name: Form-associated WebMCP attributes
-description: The `toolname`, `tooldescription`, and other HTML form element attributes register actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
+description: The `toolname`, `tooldescription`, and `toolautosubmit` HTML form element attributes register actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
 spec: https://webmachinelearning.github.io/webmcp/
 group: webmcp


beaufortfrancois:
@captainbrosset PTAL

features/declarative-webmcp.yml
```diff
@@ -0,0 +1,4 @@
+name: declarative-webmcp
```
captainbrosset: This should be a human-readable name, not a repeat of the ID.
Maybe something like this:


```suggestion
name: Form-associated WebMCP attributes
```

features/declarative-webmcp.yml
```diff
@@ -0,0 +1,4 @@
...
+description: The declarative WebMCP API annotates HTML form elements as AI-agent tools that can be called by in-browser AI agents or browser extensions.
```
captainbrosset: We usually prefer to name the entry points of the feature first.
Consider:

```suggestion
description: The `toolname`, `tooldescription`, and other HTML form element attributes registers actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
```

features/declarative-webmcp.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://webmachinelearning.github.io/webmcp/
```
captainbrosset, line 3: That version of the spec does not include the HTML attributes yet. Will it soon? Or should we instead link to https://github.com/webmachinelearning/webmcp/blob/main/declarative-api-explainer.md for the time being?

beaufortfrancois: Yes. https://github.com/webmachinelearning/webmcp/blob/main/declarative-api-explainer.md will be merged soon in the spec. See https://webmachinelearning.github.io/webmcp/#declarative-api

beaufortfrancois:
@captainbrosset If that looks good to you, I'll merge https://chromium-review.googlesource.com/c/chromium/src/+/7661686

features/declarative-webmcp.yml
```diff
@@ -0,0 +1,4 @@
+name: Form-associated WebMCP attributes
+description: The `toolname`, `tooldescription`, and other HTML form element attributes registers actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
```
captainbrosset:
```suggestion
description: The `toolname`, `tooldescription`, and other HTML form element attributes register actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
```

features/declarative-webmcp.yml
```diff
@@ -0,0 +1,4 @@
+name: Form-associated WebMCP attributes
+description: The `toolname`, `tooldescription`, and other HTML form element attributes register actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
```
captainbrosset:
```suggestion
description: The `toolname`, `tooldescription`, and `toolautosubmit` HTML form element attributes register actions that can be called by in-browser AI agents or browser extensions. Also known as declarative WebMCP.
```

captainbrosset:
I'm just hesitant on the ID for this new feature. I'd prefer something like `webmcp-tool-form-attributes` but I'm also not sold on this and can't find something great.
I suggest we land this as is and then [move](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#move-a-feature-to-a-new-id) the feature when/if we find a better ID.

PR #3902, opened 2026-03-30
Demo: compat key sets authoring (successor to `compute_from`)

ddbeck:
This PR demonstrates part of the "aspects" project: naming sets of compat keys within features. This can help us make more checks on a key's membership in a feature and assign meaning (if any) that a key should have with respect to the feature's status.

As a refresher, I'm proposing three sets of keys:

- A strictest set, here called `core_name_tbc`. These keys are used to compute the headline status for the feature and gives the feature's "birthday." It's equivalent to the set of keys given by a `compat_features` array without a `compute_from`, in the existing authoring schema.
- A semi-relaxed set, here called `modifiers_name_tbc`. These keys will be required to have the same status _level_ or better as the keys in the strictest set (or the status level given by a `status` override). For example, if a feature is Baseline high overall, then these keys must all be Baseline high, even if the dates or specific version numbers do not match the headline status.
- A fully-relaxed set, here called `incidentals_name_tbc`. These keys do not contribute to the status of the feature and will not be checked against the status of the feature. It's for keys that are related but for various reasons cannot be used to enforce a status (e.g., data errors, conflicts with caniuse, etc.). Over time, we'd likely fix or move most (though not all) of these keys.

I've done a self-review on this PR to call attention to various details here. In general, you'll find that changes fall into two buckets:

- Demonstrating applications for new authoring, particularly in opening the door to new linting and validation.
- Algorithmically sorting the keys into sets happens to better *expose* information about features as they already exist. These exposures suggest ways we might improve the way we've defined features or contribute changes upstream, but this authoring makes it more obvious (and easier to find such features where work is needed).

Head to the changes tab for a guided tour.


features/absolute-positioning.yml
```diff
@@ -2,10 +2,10 @@ name: Absolute positioning
...
-  - css.properties.position.absolute
-  - css.properties.place-self.position_absolute_context
-  - css.properties.align-self.position_absolute_context
-  - css.properties.justify-self.position_absolute_context
+  core_name_tbc:
+    - css.properties.position.absolute # former compute_from
+  incidentals_name_tbc:
+    - css.properties.align-self.position_absolute_context
+    - css.properties.justify-self.position_absolute_context
+    - css.properties.place-self.position_absolute_context
```
ddbeck, line 11: The bulk of these changes will merely expose certain issues in a more findable way. Here, you can see more easily that there should be a separate feature minted. It happens that this is already in progress with https://github.com/web-platform-dx/web-features/pull/2947.

features/autofocus.yml
```diff
@@ -14,3 +14,10 @@ status:
     firefox_android: "4"
     safari: "4"
     safari_ios: "3.2"
+compat_features:
+  modifiers_name_tbc:
+    - api.HTMLElement.autofocus
+    - api.MathMLElement.autofocus
+    - api.SVGElement.autofocus
+    - html.global_attributes.autofocus
+    - svg.global_attributes.autofocus
```
ddbeck, line 23: Here's the first new application for these sets: **linting our existing overrides**. In `autofocus`, we've overridden the status to align with caniuse. By assigning the keys to the `modifiers_name_tbc` set, I'm declaring that all of the keys _must_ meet the level given by the status override.

(We _could_ have an implicit rule that a status override implies that a regular `compat_features` array *is* the `modifiers_name_tbc` set, but this seems too magical to me.)

features/backdrop.yml
```diff
@@ -2,9 +2,10 @@ name: "::backdrop"
...
-status:
-  compute_from: css.selectors.backdrop
 compat_features:
-  - css.selectors.backdrop
-  - css.selectors.backdrop.dialog
-  - css.selectors.backdrop.inherit_from_originating_element
+  core_name_tbc:
+    - css.selectors.backdrop # former compute_from
+  modifiers_name_tbc:
+    - css.selectors.backdrop.dialog
```
ddbeck, line 9: Here's another exposure: **key-sharing opportunities**. Here we have a key that should probably exist on two features: `backdrop` and `dialog`. But the influence each of the key ought to have on the status differ, depending on the feature.

features/background-sync.yml
```diff
@@ -1,17 +1,17 @@
...
+    - api.SyncEvent
+    - api.SyncEvent.SyncEvent
+    - api.SyncEvent.lastChance
+    - api.SyncEvent.tag
+    - api.SyncManager # former compute_from
+    - api.SyncManager.getTags
+    - api.SyncManager.register
+  modifiers_name_tbc:
+    - api.Permissions.permission_background-sync
+    - api.SyncManager.worker_support
```
ddbeck, line 17: Here's an application: **alignment with caniuse**. Two keys are excluded from the birthday setting here, but we can now enforce that the remaining keys should be supported in the same set of browsers as the birthday set of keys.

That said, in this case (given the negative standards positions), we could probably eliminate the `compute_from`, if we coordinated with caniuse.

features/border-radius.yml
```diff
@@ -2,23 +2,23 @@ name: border-radius
...
+    - css.properties.border-radius.4_values_for_4_corners
+    - css.properties.border-radius.elliptical_borders
+    - css.properties.border-top-left-radius
+    - css.properties.border-top-left-radius.elliptical_corners
+    - css.properties.border-top-left-radius.percentages
+    - css.properties.border-top-right-radius
+    - css.properties.border-top-right-radius.elliptical_corners
+    - css.properties.border-top-right-radius.percentages
+  modifiers_name_tbc:
+    - css.properties.border-radius.percentages
```
ddbeck, line 24: Here's another exposure: **iffy upstream data**. `css.properties.border-radius.percentage` is probably incorrect. But there's also an application here: we can accept the status level of this key and validate it against the headline status, without taking action on the upstream data immediately.

Later, we can ratchet this key into higher expectations, if and when we fix the upstream data.

features/css-object-model.yml
```diff
@@ -2,88 +2,89 @@ name: CSS object model
...
-  - api.CSSMarginRule.name
-  - api.CSSMarginRule.style
-  - api.CSSConditionRule
-  - api.CSSConditionRule.conditionText
-  - api.CSSMediaRule
-  - api.CSSMediaRule.media
-  - api.CSSPseudoElement
-  - api.CSSPseudoElement.element
-  - api.CSSPseudoElement.type
+  core_name_tbc:
```
ddbeck, line 6: Here's another exposure: **a too-large feature**. We should break this up into smaller pieces. There are maybe a dozen features like this. But in the mean time, we can declare what we know about this feature—the subset of keys that are "original", the subset that have the same status, and the subset that doesn't—and prevent _unwitting_ changes to those sets.

features/dedicated-workers.yml
```diff
@@ -3,37 +3,38 @@ description: The `Worker()` constructor runs a script in its own thread, which c
...
+    - api.DedicatedWorkerGlobalScope.name
+    - api.Worker.Worker.options_name_parameter
+    - api.WorkerGlobalScope.importScripts.mime_checks
+    - api.WorkerGlobalScope.navigator
+    - api.WorkerGlobalScope.rejectionhandled_event
+    - api.WorkerGlobalScope.self
+    - api.WorkerGlobalScope.unhandledrejection_event
+    - api.WorkerLocation.origin
+  incidentals_name_tbc:
+    - api.Worker.Worker.mime_checks
```
ddbeck, line 39: This demonstrates a good application for `incidentals_name_tbc`: associating a key with the most-relevant feature, even though that key cannot contribute to the headline status.

In this case, `api.Worker.Worker.mime_checks` isn't a "feature" that a developer can _use_, but it is relevant restriction that developers will probably want to respect going forward.

See `link-selectors` for a closely-related situation.

features/hyphens.yml
```diff
@@ -5,71 +5,72 @@ description: The `hyphens` CSS property controls when long words are broken by l
...
-  - css.properties.hyphens.language_telugu
-  - css.properties.hyphens.language_turkish
-  - css.properties.hyphens.language_turkmen
-  - css.properties.hyphens.language_ukrainian
-  - css.properties.hyphens.language_upper_sorbian
-  - css.properties.hyphens.language_welsh
+  core_name_tbc:
+    - css.properties.hyphens
+    - css.properties.hyphens.auto # former compute_from
+    - css.properties.hyphens.language_english
```
ddbeck, line 12: Here's another application: ignoring data we can't really trust.

It turns out that I'm the original author of the majority of the `hyphens.language_*` data… almost ten years ago, when I migrated it from the MDN wiki. 😭  I do not really trust that this data is up-to-date (I'm not even sure it was terribly accurate at the time).

There's a subset here I _do_ trust (this `core_name_tbc` set), but the rest… well, the one thing I can say here definitively is that these keys _are_ related to the `hyphens` feature. So we might as well make that relationship explicit (and eventually burndown the incidentals list).

features/link-rel-preload.yml
```diff
@@ -3,13 +3,13 @@ description: The `rel="preload"` attribute for the `<link>` HTML element request
...
-  - html.elements.link.rel.preload.as-track
+  core_name_tbc:
+    - html.elements.link.rel.preload # former compute_from
+    - html.elements.link.rel.preload.as-fetch
+    - html.elements.link.rel.preload.as-font
+    - html.elements.link.rel.preload.as-image
+    - html.elements.link.rel.preload.as-script
+    - html.elements.link.rel.preload.as-style
+  incidentals_name_tbc:
+    - html.elements.link.rel.preload.as-track
```
ddbeck, line 15: Here's another application: ignoring very minor issues.

There's a very poor case for breaking up `link-rel-preload` over `as="track"` support. The way I see it, the lack of support for `as="track"` is one browser's minor bug. Instead of making the platform seem much more complicated than it is by breaking this feature into 3 to 6 features, let's overlook this one key.

features/link-selectors.yml
```diff
@@ -3,13 +3,13 @@ description: The `:link` CSS pseudo-class matches unvisited links, `:visited` ma
...
-  - css.selectors.visited.privacy_measures
+  core_name_tbc:
+    - css.selectors.any-link # former compute_from
+    - css.selectors.link
+    - css.selectors.visited
+    - css.selectors.visited.privacy_measures
+  modifiers_name_tbc:
+    - css.selectors.any-link.not_match_link
+    - css.selectors.link.not_match_link
+    - css.selectors.visited.not_match_link
```
ddbeck, line 15: Here's another application: capturing behavior evolution for **reverse features**, where the platform took abilities away from developers.

In this case, these keys do not represent a "feature" at all. Instead, they represent a privacy limitation and a way this feature interacts with another feature (the `<link>` element), or rather how they _do not_ interact with each other. It would be silly to mint a feature for "Not selecting `<link>` link elements with `:link`", so `modifiers_name_tbc` lets us sidestep doing something silly.

features/list-elements.yml
```diff
@@ -7,23 +7,20 @@ spec:
...
+    - api.HTMLUListElement
+    - html.elements.li # former compute_from
+    - html.elements.li.value
+    - html.elements.ol # former compute_from
+    - html.elements.ol.start
+    - html.elements.ol.type
+    - html.elements.ul # former compute_from
+  modifiers_name_tbc:
+    - api.HTMLOListElement.reversed
+    - html.elements.ol.reversed
```
ddbeck, line 26: The last one here helps queue up the future by exposing **candidates for (retroactive) feature merging**. Once we implement feature merging, the `reversed` attribute might make sense as a separate feature that has merged into `list-elements`. Here's it's exposed as a possibility that we might author later.

features/link-rel-preload.yml
malchata: So, this falls into a domain I have some expertise on, and forgive me if the question seems naïve: Is it really incidental if the `as="track"` value has a small bug in one browser? If the intent is to fix the bug and bring it in line, should it remain under `incidentals_name_tbc`, or will it be eventually moved into `core_name_tbc`? Sorry if this question is a bit of a waste of time, but it does help my understanding!

malchata: Or rather, it could be moved into `modifiers_name_tbc`. The intent of the question is to see if these things eventually move into other categories when bugs surface or are fixed.

features/list-elements.yml
malchata: Is there prior art for telegraphing things that may be features? It is a naïve question, but I'm genuinely curious.

features/link-rel-preload.yml
ddbeck: @malchata good question! In this case, I was under the impression that `<link rel="preload" …>` works for the overwhelmingly common uses (styles and scripts, though if I'm wrong about this I'm open to correction). With that assumption, I figured that means there's _some_ preload feature that is Baseline.

That leaves us with `as="track"`. Either we temporarily ignore it (as I've done) or mint a separate feature. If we ignore it as I've done, then — when `as="track"` is fixed — we'd promote the `as-track` key to the `modifiers` set (where it would be required to always have the same _status_ as the `core` keys, but not necessarily the same exact ship date or versions).

Does that help explain the idea better?

features/list-elements.yml
ddbeck: There are some comments in existing files, but mostly this kind of information shows up in PRs. If you go look at whatever PR created most features with a `compute_from` or any feature with `discouraged` but no `alternatives`, then you're likely to find some discussion about how something _might_ be a separate feature but it's inconvenient or confusing to mint one. But there's never been a conventional comment or structural cue for this.

I think there's a possibility to follow this up with annotations on BCD keys (e.g., to flag keys as candidates for break up and the like). I think it would be a nice thing to have.

malchata: Got it. I agree. I think making this information part of the data structure makes it more obvious!

features/link-rel-preload.yml
malchata: This was more about an edge case—I do think that `as="track"` is not going to be nearly as common as say, `as="script"` or `as="style"`. But it's a good example for what you're trying to communicate.

This does help explain the idea better, and thank you for taking the time to explain it! It gives me a bit of pause about how we think about features like this. Is `rel="preload"` in general one feature? Is it a feature that has several subfeatures (various `as` attribute values)? Or is each resource type its own feature? I don't pretend to be an expert, therefore are rhetorical questions, not questions you're obligated to answer!

ddbeck: > Is `rel="preload"` in general one feature?

If this were a new feature _today_, I'd take one of two possible approaches:

- Mint a preload feature for every `as="…"` value
- Mint one preload feature that would not be Baseline until `as="track"` ships (with the option to break it up later, if it became clear that implementation would not happen across the board or at the same time)

In this case, preload predates web-features and Baseline. When we've minted existing features, we have tried to be consistent with established understandings of support (e.g., caniuse has [a feature entry](https://caniuse.com/link-rel-preload) that overlooks `as="track"`). So we've followed caniuse's lead here, on the assumption that if the data has never been faulted on caniuse that developers don't actually care if `as="track"` works.

If developer understanding were to change (e.g., web-features or caniuse got complaints from developers or an Interop proposal on this subject succeeded), we might change the status here (i.e., by moving the `as-track` key to `core_name_tbc`) or break the feature up, probably in coordination with caniuse and MDN.

ddbeck:
This is now implemented as of https://github.com/web-platform-dx/web-features/pull/4015.

PR #3917, opened 2026-04-01
Migrate `latest`-tagged `web-features` releases to trusted publishing

ddbeck:
This is the sequel to https://github.com/web-platform-dx/web-features/pull/3582.

Unfortunately, npm allows nominating only a single workflow per package as the "trusted publishing" workflow, so this consolidates the next and latest workflows into a single file.

Upon merging, the packages need configuration in npmjs.com (see [this section of the npm docs](https://docs.npmjs.com/trusted-publishers#for-github-actions) and [this area](https://www.npmjs.com/package/web-features/access) in the npm UI, if you have access). If you're not ready to do this yourself, then please approve and leave unmerged. Thank you!


captainbrosset:
You can only bless one workflow file to do trusted publishing. Until this PR, we had separate workflows. Now they have to be together. This is why the workflow is more complex than it used to be. It's consolidation.
Also changed a couple little things.

.github/workflows/publish_next_web-features.yml
```diff
@@ -1,123 +0,0 @@
...
-  push:
-    branches:
-      - "main"
-    paths:
-      - features/**
-      - groups/**
-      - packages/web-features/**
-      - snapshots/**
-      - index.ts
-      - scripts/build.ts
```
captainbrosset, line 13: Why don't we need this filter anymore in the new version?

.github/workflows/publish_web-features.yml
```diff
@@ -2,54 +2,160 @@ name: Publish web-features
...
           cache: npm
+      - run: npm install -g 'npm@>=11.5.1'  # required for trusted publishing
       - run: npm ci
       - run: npm test
-  artifacts:
-    if: github.repository == 'web-platform-dx/web-features'
+
+  publish_latest:
+    name: Publish release
+    if: ${{ github.repository == 'web-platform-dx/web-features' && case(startsWith(github.ref, 'refs/tags/'), 'latest', 'next') == 'latest' }}
```
captainbrosset, line 44: Maybe add a comment saying how the workflow runs on each push (because 'next' needs it) but this line filters only tags, for official releasing.

Also, can you explain why we need to filter on github.repository?

.github/workflows/publish_web-features.yml
```diff
@@ -2,54 +2,160 @@ name: Publish web-features
...
+      - run: gh release upload "$GITHUB_REF_NAME" $ARTIFACTS
         env:
           GH_TOKEN: ${{ github.token }}
-  publish:
-    if: github.repository == 'web-platform-dx/web-features'
+          ARTIFACTS: ${{ env.artifacts }}
+
+  publish_next:
+    name: Publish prerelease
+    if: ${{ github.repository == 'web-platform-dx/web-features' && case(startsWith(github.ref, 'refs/tags/'), 'latest', 'next') == 'next' }}
```
captainbrosset, line 74: A comment here explaining the filtering logic would be good here too.

.github/workflows/publish_web-features.yml
```diff
@@ -2,54 +2,160 @@ name: Publish web-features
...
+        run: gh release edit --draft "$TAG"
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ env.dist_tag }}
+
+      - name: Update the tag
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
captainbrosset, line 137: For my own knowledge, where does this bot user email address come from? The magic number part, I mean.

captainbrosset (commented):
Leaving a few comments first. I'm fine with the change, but first want to make sure I understand all the pieces.

PR #3942, opened 2026-04-09
`user-agent-sniffing`: add more compat keys

ddbeck:
This will remain in draft until BCD's next release (https://github.com/mdn/browser-compat-data/pull/29408).

(In an ideal world, this "feature" would be discouraged, but WHATWG was unwilling to make the spec changes needed to permit those changes in BCD and on MDN. See https://github.com/mdn/browser-compat-data/pull/29438 for background and more links.)


features/user-agent-sniffing.yml
```diff
@@ -1,6 +1,25 @@
 name: User agent sniffing
-description: "The `navigator.userAgent` property is the user agent string for the current browser. Selectively showing content based on the user agent string is unreliable. Consider using feature detection instead."
+description: >
+  The `navigator.userAgent` property is the user agent string for the current browser.
+  This property and several others loosely identify a browser.
+  Selectively showing content based on the user agent string, better known as user agent sniffing, is unreliable.
+  Consider using feature detection instead.
```
captainbrosset: Shouldn't we also suggest using user agent client hints?

features/user-agent-sniffing.yml
```diff
@@ -1,6 +1,25 @@
...
 spec: https://html.spec.whatwg.org/multipage/system-state.html#the-navigator-object
 compat_features:
   - api.Navigator.userAgent
+  - api.Navigator.appCodeName
+  - api.Navigator.appName
+  - api.Navigator.appVersion
+  - api.Navigator.javaEnabled
+  - api.Navigator.mimeTypes
```
captainbrosset: These are the only two that don't fit super well with this feature. They don't really detect browsers as much ad they detect features. 
Maybe leave them to another feature?

ddbeck: I re-read the specs for both of these and they're both a bit odd:

- `javaEnabled` always returns `false`, so I don't think it really detects anything. I put this in the browser sniffing feature thinking that it's similar to many of the other properties here, which more less tell lies to developers. 😅 
- `mimeTypes` is similar, except it _can_ give the same information (in a different format) as `navigator.pdfViewerEnabled`. I'll move that key there.

features/user-agent-sniffing.yml
```diff
@@ -1,6 +1,23 @@
 name: User agent sniffing
-description: "The `navigator.userAgent` property is the user agent string for the current browser. Selectively showing content based on the user agent string is unreliable. Consider using feature detection instead."
+description: >
+  The `navigator.userAgent` property and several others loosely identify a browser.
+  Selectively showing content based on the user agent string, also known as user agent sniffing, is unreliable.
+  Consider using feature detection or user agent client hints instead.
```
ddbeck, line 5: @captainbrosset wrote:

> Shouldn't we also suggest using user agent client hints?

Yes, but length limits make this a little challenging. I've rewritten the description to make it fit. Can I get you take a look at this again?

features/user-agent-sniffing.yml
captainbrosset: Maybe we leave the Java key alone, as a forgotten and useless thing of the past web? 
Or we keep it in ua sniffing. Your call.

captainbrosset (approved):

features/user-agent-sniffing.yml
ddbeck: Yes, I think we leave it here as a thing of the past. The alternative would be to mint a feature for it (`java-detection`?), which is probably not actually useful to anyone. But if I'm wrong about that, we can change our minds later.

PR #3951, opened 2026-04-14
Revert XSLT's regression from Baseline widely available

ddbeck:
~~I merged https://github.com/web-platform-dx/web-features/pull/3946 but had forgotten to push this comment first.~~

This regression for `xslt` might be premature, so this freezes the status as it was before the regression.


Elchi3 (approved):
I created https://github.com/mdn/browser-compat-data/pull/29484 but approving this PR in case you need to move things around faster.

captainbrosset (approved):

PR #3969, opened 2026-04-20
Assign compat keys from BCD 7.3.12 for existing features

ddbeck:
This assigns new keys from `@mdn/browser-compat-data@7.3.12` (https://github.com/web-platform-dx/web-features/pull/3966) where there's an existing feature.

### DIFF
--- a/features/declarative-shadow-dom.yml
+++ b/features/declarative-shadow-dom.yml
@@ -13,8 +13,13 @@ compat_features:
   - api.HTMLTemplateElement.shadowRootMode
   - api.HTMLTemplateElement.shadowRootDelegatesFocus
   - api.HTMLTemplateElement.shadowRootSerializable
-  - api.HTMLTemplateElement.shadowRootSlotAssignment
   - html.elements.template.shadowrootmode
   - html.elements.template.shadowrootclonable
   - html.elements.template.shadowrootdelegatesfocus
   - html.elements.template.shadowrootserializable
+
+  # This key was added later, but it's not a distinct feature. Rather, a spec
+  # contributor regarded it as an unintended "omission" from the original
+  # specification. See the intent-to-ship for background:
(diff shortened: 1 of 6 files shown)


features/declarative-shadow-dom.yml
```diff
@@ -13,6 +13,7 @@ compat_features:
   - api.HTMLTemplateElement.shadowRootMode
   - api.HTMLTemplateElement.shadowRootDelegatesFocus
   - api.HTMLTemplateElement.shadowRootSerializable
+  - api.HTMLTemplateElement.shadowRootSlotAssignment
```
ddbeck, line 16: This is an "omitted" attribute from the initial specification. See: https://groups.google.com/a/mozilla.org/g/dev-platform/c/8qxAHE9-Xdo

features/hanging-punctuation.yml
```diff
@@ -6,5 +6,6 @@ compat_features:
   - css.properties.hanging-punctuation
   - css.properties.hanging-punctuation.allow-end
   - css.properties.hanging-punctuation.first
+  - css.properties.hanging-punctuation.force-end
```
ddbeck, line 9: This is a backfill key from BCD. See: https://redirect.github.com/mdn/browser-compat-data/pull/29497

features/declarative-shadow-dom.yml
captainbrosset: Ok so this is a later addition to the spec, but not one that we'd want to create a new feature for, because it was an omission. Might be good to add a comment about this in the file, above that key, so we know it's ok for it to have a much lower status than all the other keys.

captainbrosset (approved):
Looks good. One suggestion to add a comment for posterity.

features/declarative-shadow-dom.yml
ddbeck: Good idea. I'll do this and merge it. Thank you!

PR #3988, opened 2026-04-27
`view-transitions-element-scoped`: add compat keys

3ru:
Updates `view-transitions-element-scoped` now that scoped view transitions are specified in CSS View Transitions Level 2 and the corresponding BCD keys are available

Changes:
- Replace the stale `contain: view-transition` placeholder with current BCD compat keys.
- Move the scoped view transitions keys out of the CSS View Transitions Level 2 draft bucket.

References:
- CSSWG issue: https://github.com/w3c/csswg-drafts/issues/9890
- CSSWG spec PR for `view-transition-scope`: https://github.com/w3c/csswg-drafts/pull/13390
- BCD Chrome 147 PR: https://github.com/mdn/browser-compat-data/pull/29244
- Spec: https://drafts.csswg.org/css-view-transitions-2/#scoped-vt

### DIFF
--- a/features/view-transitions-element-scoped.yml
+++ b/features/view-transitions-element-scoped.yml
@@ -1,13 +1,16 @@
 name: Element-scoped view transitions
 description: >
   The `startViewTransition()` method of an `Element` object starts a view transition that affects only that element's DOM tree.
-  The `contain: view-transition` CSS declaration contains the layout of the tree to be transitioned.
+  The `view-transition-scope` CSS property limits `view-transition-name` discovery to the element and its flat-tree descendants.
   You can use this to run separate elements' transitions concurrently.
-spec: https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md
+spec: https://drafts.csswg.org/css-view-transitions-2/#scoped-vt
 group: view-transitions
-# Expecting:
-# compat_features:
-#  - api.Element.startViewTransition
-#  - css.properties.contain.view-transition
+compat_features:
+  - api.Element.activeViewTransition
+  - api.Element.startViewTransition
+  - api.ViewTransition.transitionRoot
+  - css.properties.view-transition-scope
+  - css.properties.view-transition-scope.all
+  - css.properties.view-transition-scope.none
 # This feature is subordinate to view-transitions. It should probably merge into
 # it one day.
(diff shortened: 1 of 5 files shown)


features/view-transitions-element-scoped.yml
```diff
@@ -1,13 +1,16 @@
 name: Element-scoped view transitions
 description: >
   The `startViewTransition()` method of an `Element` object starts a view transition that affects only that element's DOM tree.
-  The `contain: view-transition` CSS declaration contains the layout of the tree to be transitioned.
+  The `view-transition-scope` CSS property limits `view-transition-name` discovery to this element's subtree.
```
ddbeck: I'm finding the "limit[ing] … discovery" bit to be quite challenging. I've read the spec and I'm not really understanding what exactly "discovery" is or where it's defined. 

When I originally wrote this description, I tried to capture the entry point to this feature, with the understanding that `contain: view-transition` was going to be a requirement for web developers to use `startViewTransition()`. It's not clear to me that `view-transition-scope` is now part of that requirement.

If it's required, then I might suggest something like "The `view-transition-scope: all` CSS declaration limits transitions to those in the element's subtree" (if that's correct—again I'm not certain I've grasped the change since I wrote the original description), to avoid the novel discovery language.

If it's not required, we might just drop this whole line? If `startViewTransition()` is the only guaranteed entry point into element-scoped view transitions, then avoiding going into too much detail within our 300 character limit might be a better way to go.

ddbeck (changes requested):
Thanks for filing this PR, @3ru! I'm happy with the other changes here, but I've left a long comment on the change to the description, which I'm still trying to get my head around. Thank you again!

features/view-transitions-element-scoped.yml
3ru: @ddbeck Thanks! I agree that dropping this sentence is the least misleading option.

My understanding is that authors do not need to set `view-transition-scope: all` before calling `Element.startViewTransition()`. The spec notes that starting a transition implies that the element has `view-transition-scope: all` for the duration of the transition:
https://github.com/w3c/csswg-drafts/blob/c2e25638b06b5e8ff0adad3e23d13baa83ffad7b/css-view-transitions-2/Overview.bs#L987-L988

The WICG explainer also says that if the scope does not have `view-transition-scope: all`, it acquires that behavior while the transition is running:
https://github.com/WICG/view-transitions/blob/4d9d9b44fbe3fd0602d4a73070b51d461df37868/scoped-transitions.md?plain=1#L208-L211

So I’ll remove this sentence for now. If the spec and developer understanding around here mature further, we can add text back as needed.

ddbeck (approved):
Thank you, @3ru!

Issue #3820, opened 2026-02-26
WebDX Community Group logo


captainbrosset:
I'm proposing that we adopt a logo for our community group.

This would help us be more recognizable as the entity behind the web-features package, our dev research efforts, and the Baseline definition.

We could use the logo on our github org's page, on [our website](https://web-platform-dx.github.io/web-features/), on [the explorer](https://web-platform-dx.github.io/web-features-explorer/), etc. Basically anywhere we need.

If you disagree that we should have a logo, please comment.
If you have logo ideas, please add them below.

captainbrosset:
Here is my proposal:

<img width="751" height="477" alt="A series of little black dots which converge to a single line that leads to a green checkmark. WebDX Community Group is written below." src="https://github.com/user-attachments/assets/a8241f23-ba0e-45b9-bf79-54c657dea63b" />

The dots represent features, dev signals, surveys, etc. and are all leading to a common line (the baseline).

estelle:
https://codepen.io/estelle/pen/ZYpOEVO

<img width="250" height="225" alt="Stylized W and X letters, over each other, to create an abstract black and green sign. WebDX Community Group appears below it." src="https://github.com/user-attachments/assets/81abbf20-c74f-4a44-8c2f-87f2eb98adcc" />

captainbrosset:
@meyerweb also proposed an idea:

<img width="818" height="401" alt="WDX letters. The W is black and green, the left part is black and the right part is green, making it look like a check mark. D is black. X is orange, and looks like a across. The checkmark and cross signs are circled in matching color circles." src="https://github.com/user-attachments/assets/336eb3de-1307-417c-a0da-82c6bf32d71a" />

tidoust:
> [@meyerweb](https://github.com/meyerweb) also proposed an idea:

I like that it aligns with the Baseline logos. I wish it wouldn't create a new acronym on top of "WebDX", but inserting "eb" would probably break the overall visual balance.

captainbrosset:
We could try to add it in there. The logo isn't really balanced anyway, with part of the W sticking out of the green circle. I might try to take a look.

dragontheory:
Sorry I am late to the party. Flexible. Can be in a line or broken up as shown. Reflects the dev/coding side of things.


<img width="362" height="378" alt="Web Developer Experience Community Group" src="https://github.com/user-attachments/assets/2195362b-f1c7-486e-b8a2-0807d1441bf2" />

[...]

PR #3993, opened 2026-04-27
Remove the website in favor of the new org-level site

captainbrosset:
- We have a new logo, see #3820.
- We decided that the site should be hosted at the org level, so I moved the source code to the https://github.com/web-platform-dx/web-platform-dx.github.io/ repo, which is now published at https://web-platform-dx.github.io/
- I also made some modifications to the site:
  - Add the new logo in header + favicon.
  - Made the WebDX CG page be the landing page.
  - Added a dark theme.
  - Made some minor CSS improvements.
  - Made some minor content improvements.

Currently, the `/web-features/` path still goes to _the current_ repo.
Once we merge this PR, the current repo will stop hosting the site, and the `/web-features/` path will start being served by the org-level gh-pages site.
At the org-level site, the `/web-features/` path is redirected to `/baseline/`. This is so that existing links aren't broken, and improve the site. The `/web-features/` path was a technical one only, due to the our repo's name. In reality, web-features and baseline are 2 different things, and the new `/baseline/` makes a lot more sense here.


captainbrosset:
@sebazelonka thank you very much for your [logo proposal](https://github.com/web-platform-dx/web-features/issues/3820#issuecomment-4194741227)! I took the liberty to modify it a tiny bit, as you can see above.

I'd love for our group to adopt this logo. Are you good with this?

captainbrosset:
Note for myself: revert the change where I made the CG page be the homepage instead of the Baseline page. We don't want to do this in this PR, to avoid breaking existing links.
Future plan: move the CG homepage to the github org level, and keep this repo for the baseline page.

sebazelonka:
@captainbrosset I love the gradient and how the proportions are adjusted. TBH, I don't like too much where the WebDX is, I like it more just beside the Icon, but if the idea is to have something bigger in the same width, I think is the right alternative.

NiklasMerz (approved):
The new logo and website is looking sharp

tidoust:
> At the org-level site, the `/web-features/` path is redirected to `/baseline/`

Not sure how Eleventy manages redirect-like pages, but it seems to me that the [code in the org-level site](https://github.com/web-platform-dx/web-platform-dx.github.io/blob/main/src/web-features-redirect.html) will only redirect `/web-features/`, and not subpages such as [`/web-features/supported-browsers/`](https://web-platform-dx.github.io/web-features/supported-browsers/). I do not know whether there are sources that currently target these subpages but they may be worth redirecting too?

The approach looks good to me, and the structure makes more sense. I propose to re-add the Baseline wordmark in the [`/baseline/`](https://web-platform-dx.github.io/baseline/) page, see https://github.com/web-platform-dx/web-platform-dx.github.io/pull/1

captainbrosset:
> Not sure how Eleventy manages redirect-like pages, but it seems to me that the [code in the org-level site](https://github.com/web-platform-dx/web-platform-dx.github.io/blob/main/src/web-features-redirect.html) will only redirect /web-features/, and not subpages such as [/web-features/supported-browsers/](https://web-platform-dx.github.io/web-features/supported-browsers/). I do not know whether there are sources that currently target these subpages but they may be worth redirecting too?

Correct, only the /web-features/ path is redirected, as this is the most important one. I'm much less worried about the other ones, but sure, we can add redirect pages to handle them too.

> The approach looks good to me, and the structure makes more sense. I propose to re-add the Baseline wordmark in the [/baseline/](https://web-platform-dx.github.io/baseline/) page, see https://github.com/web-platform-dx/web-platform-dx.github.io/pull/1

Done! Thanks for the PR.

captainbrosset:
> > Not sure how Eleventy manages redirect-like pages, but it seems to me that the [code in the org-level site](https://github.com/web-platform-dx/web-platform-dx.github.io/blob/main/src/web-features-redirect.html) will only redirect /web-features/, and not subpages such as [/web-features/supported-browsers/](https://web-platform-dx.github.io/web-features/supported-browsers/). I do not know whether there are sources that currently target these subpages but they may be worth redirecting too?
> 
> Correct, only the /web-features/ path is redirected, as this is the most important one. I'm much less worried about the other ones, but sure, we can add redirect pages to handle them too.

Done, see the *-redirect.html files in https://github.com/web-platform-dx/web-platform-dx.github.io/tree/main/src

tidoust (approved):

PR #3999, opened 2026-04-29
Map more CSS sub features (currentColor, logical property values)

Elchi3:
No baseline status should be affected here, just addition of many more keys that are sub features.


features/currentcolor.yml.dist
```diff
@@ -27,3 +38,17 @@ compat_features:
...
+  # ⬇️ Same status as overall feature ⬇️
+  # baseline: high
+  # baseline_low_date: 2015-07-29
+  # baseline_high_date: 2018-01-29
+  # support:
+  #   chrome: "1"
+  #   chrome_android: "18"
+  #   edge: "12"
+  #   firefox: "1.5"
+  #   firefox_android: "14"
```
captainbrosset: Weird, `css.properties.border-left.currentColor` is supported on firefox for android 4, but `css.properties.border-right.currentColor` only on version 14?
That smells like a typo in BCD.

captainbrosset (approved):
LGTM apart from what's probably a typo in BCD, but which doesn't need to block this PR.

features/currentcolor.yml.dist
Elchi3: I've sent https://github.com/mdn/browser-compat-data/pull/29575. If merged and package upgraded, the dist should correct itself.

ddbeck (approved):
Thank you!

Issue #3510, opened 2025-11-06
Generate new statistics


ddbeck:
Since the beginning of the year, we've been (mostly) complete in backfilling features from the history of the web, so the stats in #788 no longer have a lot of meaning for progress in web-features. We ought to adopt new stats that help us understand web-features in this more stable and maintenance-oriented era of the project.

New stats ought to help us:

- Highlight material changes in coverage of BCD keys and caniuse mappings (for example, a fractional percent change in BCD or caniuse coverage does not speak very loudly)
- Encourage us to direct our attention or resources to areas that will make material improvements to the data (for example, presently all BCD keys have the same weight in our statistics, even though I know they do not have equal utility or interest for developers)

ddbeck:
I've got ideas for the following new metrics:

- Cumulative shipping days of unmapped BCD keys (smaller is better)

  Calculated as the sum of days since each unmapped BCD key has shipped in each browser in the core browser set. Keys which have not yet shipped in a stable release would count as 0. Partial implementations and prefixes would count toward this sum, but prefixes and flags would not.

  Unmapped keys shipped in more browsers would increase the metric more than unmapped keys shipped in fewer browsers. My intent here would be that the metric would go up more for keys developers are more likely to have encountered. Consequently, this metric would encourage us to prioritize mapping keys that have shipped more widely before keys that have shipped less widely (or not at all).

- Absolute unmapped BCD keys remaining (smaller is better)

  This would succeed (and invert) the existing "compat keys mapped to web-features" and BCD coverage metrics.

  Calculated as the number of BCD keys that have not been assigned to a feature in web-features. The target value would be 0 (though this would probably be an infrequent event, at best).

  If this number trends upward, it indicates that we're not assigning keys to keep up with BCD and may need more attention or resources. If this number trends downward, then it means that we're keeping up with BCD (and, presumably, the platform itself).

  This number would jump sometimes (e.g., if BCD adds a lot of backfill data, as it did in [v7.1.18](https://github.com/mdn/browser-compat-data/releases/tag/v7.1.18)) but that would be a good signal that existing features need revision.

- Absolute unmapped caniuse IDs (smaller is better)

  This would succeed (and invert) the existing "caniuse IDs in web-features" and "caniuse IDs" coverage metric.

   Calculated as the number of caniuse IDs that do not have a mapping in web-features. The target value would be 0. Since caniuse now ingests web-features entries, it's possible that we might reach 0 and stay there.

[...]

PR #4000, opened 2026-04-29
Refactor `caniuse.ts` to support reusing code for stats

ddbeck:
This is to support generating stats as described in https://github.com/web-platform-dx/web-features/issues/3510#issuecomment-3497758593.

To do _that_, I need the list of caniuse keys and our coverage on it. Since `caniuse.ts` already does this, I've modified it to make the relevant map and set exported and stuffed the rest of the script into a `main()` method.

While I was here, I turned on prettier for this file.

Using `diff`, I confirmed that the output of the script before and after these changes are the same.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,6 +16,5 @@ towards-features.md
 /docs
 !/docs/publishing.md
 /features/draft
-/scripts/caniuse.ts
 /scripts/schema.ts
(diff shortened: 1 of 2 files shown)


Elchi3 (approved):
Thanks Daniel! 👍

PR #4001, opened 2026-04-30
Add a workflow to help with BCD updates

jgraham:
This attempts to automate many of the existing manual steps:
* Checks out the branch
* Checks for removed BCD keys
* Checks for features with an empty definition
* Checks for baseline status changes

Where possible it automatically updates to removed BCD keys. So far it doesn't attempt to figure out the correct replacement for removed keys.

It is designed to output a GitHub comment indicating the changes, to speed up human review.

### DIFF
--- a/.github/workflows/bcd_upgrade.yml
+++ b/.github/workflows/bcd_upgrade.yml
@@ -1,0 +1,90 @@
+name: BCD upgrade automation
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  check-metadata:
+    name: Check if BCD upgrade
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    outputs:
+      is-bcd-upgrade: ${{ steps.metadata.outputs.is-bcd-upgrade }}
+    steps:
+      - name: Fetch Dependabot metadata
+        id: fetch-metadata
+        continue-on-error: true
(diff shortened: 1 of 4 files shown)


XavierMP14 (approved):

XavierMP14 (approved):

XavierMP14 (approved):

ddbeck:
Thanks for making an attempt at this, @jgraham. I tried this with a recent BCD upgrade and it silently did nothing. This is probably operator error, but overall I'm excited to see this developed more, but also decomposed a bit. If you're keen, I was thinking something along these lines:

- It'd be nice to have a standalone script that just compares statuses between what's checked out and another `data.json` (e.g., by default it could fetch the `next` `data.json` from GitHub or [unpkg](https://unpkg.com/)). I imagine this could be useful for _any_ PR that modifies existing features, not just dependabot PRs. Even nicer would be a markdown output for wiring up to issues and PRs and a JSON output for combining with other tools (I took this sort of approach for [my completely non-portable shell script](https://github.com/ddbeck/web-features/blob/__utils__/bin/x-release-statuses.fish) to do some of this work).
- Likewise, it'd be nice to have a script that only strips deleted keys from authored YAML files and prints what it did (again as JSON or Markdown). I've been talking to MDN about BCD prereleases, so we could use this not just for Dependabot ugprades but also for (for example) a dashboard issue that warns about upcoming removals/deprecations/etc. before the appear in a Dependabot PR.

By splitting this up, we could wring a bit more use out of it. What do you think?

PR #4019, opened 2026-05-07
`anchor-positioning`: add more compat keys

ddbeck:
Toward https://github.com/web-platform-dx/web-features/issues/3558

### DIFF
--- a/features/anchor-positioning.yml
+++ b/features/anchor-positioning.yml
@@ -136,6 +136,7 @@ compat_features:
   - css.properties.position-anchor
   - css.properties.position-anchor.auto
   - css.properties.position-anchor.none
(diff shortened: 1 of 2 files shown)


Elchi3 (approved):
At some point I'd like alphabetization or some sort of ordering of lists. Right now, you have like multiple chunks of e.g. `css.properties.position-try-fallbacks.*` list items and I think it would be better if they would appear altogether and sorted. Not going to block on it but if you have the same feeling we could maybe have an issue about this.

ddbeck:
@Elchi3 That's a good idea. I filed https://github.com/web-platform-dx/web-features/issues/4041 to track it. If you thoughts on what the canonical order should be, a comment there would be welcome.

PR #4020, opened 2026-05-08
`webauthn-public-key-easy`: pin status to common algorithm

### LINKED ISSUE #3995: Reconsideration: allow `webauthn-public-key-easy` to regress to limited availability
In https://github.com/web-platform-dx/web-features/pull/3991, I'm putting a hold on [`webauthn-public-key-easy`](https://web-platform-dx.github.io/web-features-explorer/features/webauthn-public-key-easy/ "Web authentication easy public key access: The getAuthenticatorData(), getPublicKey(), and getPublicKeyAlgorithm() methods of AuthenticatorAttestationResponse access credential data inside attestationObject without the need to parse it.") ([source](https://github.com/web-platform-dx/web-features/blob/main/features/webauthn-public-key-easy.yml), [dist](https://github.com/web-platform-dx/web-features/blob/main/features/webauthn-public-key-easy.yml.dist)) from regressing. We should decide how to handle this status change.

## Background

In https://github.com/mdn/browser-compat-data/pull/29458, Safari's implementation of [AuthenticatorAttestationResponse: getPublicKey()](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAttestationResponse/getPublicKey) was marked as partial due to reported missing support for certain algorithms.

This was filed with WebKit as [bug 312081](https://webkit.org/b/312081). It's currently unconfirmed.

The [spec says](https://w3c.github.io/webauthn/#ref-for-dom-authenticatorattestationresponse-getpublickey%E2%91%A1:~:text=null%2E-,User,Ed25519%29%2E) support for the missing algorithms are required with "must" language.

In https://github.com/web-platform-dx/web-features/pull/3991, this partial cascaded to the status for the `webauthn-public-key-easy` feature. At the time the feature was frozen, this feature was Baseline low.

```yaml
status:
  baseline: low
  baseline_low_date: 2023-10-24
  support:
    chrome: "85"
    chrome_android: "85"
    edge: "85"
    firefox: "119"
    firefox_android: "119"
    safari: "16"
    safari_ios: "16"
```

Without the status freeze this feature would regress to limited availability. Without the partial, this feature would have advanced to Baseline high.

## Options

- **Regress.** Agree that the bug is a serious lack of support, undo the editorial override, and allow the regression to happen.
- **Advance** with changes upstream to unmark this partial (needs support from BCD maintainers too).
- **Advance** with a `compute_from` ignoring `webauthn-public-key-easy`. Silently ignore BCD. Note that this does not override per-key statuses.
- **Hold**. Leave override in place. Silently ignore BCD. Note that this does not override per-key statuses. We should choose this option if we cannot come to an agreement about other solutions.

ddbeck:
This PR resolves the status problem described in https://github.com/web-platform-dx/web-features/issues/3995 by pinning to the commonly-supported algorithms. See the issue for a fuller description of this approach.

~~This PR is blocked until the data changes proposed in https://github.com/web-platform-dx/web-features/issues/3995 is accepted. It may change and I'll leave it in draft until there's a BCD release with such data.~~

It's possible in the future that we might break this feature up, minting for the algorithms themselves. Since this is a long-standing API with apparently little developer interest in the other algorithms, I don't think this is necessary today.

## Related issues

- Fixes: https://github.com/web-platform-dx/web-features/issues/3995
- Depends on: https://github.com/mdn/browser-compat-data/pull/29637


ddbeck:
In the WebDX chat, @captainbrosset asks:

> why do we need a `compute_from` statement if all keys are in it?

The answer is that we're excluding two keys:

- `api.AuthenticatorAttestationResponse.getPublicKey.algorithm_eddsa`
- `api.AuthenticatorAttestationResponse.getPublicKey.algorithm_rs256`

The full list of keys (from a tag in BCD) is in the expanded diff in the `.dist` file.

captainbrosset (approved):
Approved, keeping in mind our conversation in chat, just now, about the fact that features like this which only list a few keys, while other keys are tagged in BCD, are a little harder to work with.

PR #4034, opened 2026-05-13
Allow overlapping compat keys

ddbeck:
Towards https://github.com/web-platform-dx/web-features/issues/1173 and a prerequisite to feature merging.

This PR allows authors to allow BCD keys to be shared between features, by adding them to an allow list. It provides guidelines for when to do this (rarely) and demonstrates it by resolving one of the commented cases where we wished to do this.

This PR is not _strictly_ breaking, but we should announce it in the [upcoming changes](https://github.com/web-platform-dx/web-features/discussions/2613) discussion thread and emphasize this change in the release notes.


docs/guidelines.md
```diff
@@ -423,3 +423,35 @@ When you set a `discouraged` block in a feature file, do:
...
   Guide developers to the most relevant features that would help them stop using the discouraged feature.
+
+## Overlapping `compat_features` keys
+
+Typically, an `@mdn/browser-compat-data` (BCD) key is assigned to one and only one feature.
+Rarely, two or more features have equal claim to a key and their sets of `compat_features` keys must overlap.
+You can add a single key to two or more `compat_features` lists when independent features share some interface and one of these situations applies:
+
+- No feature has any claim to being first and the shared interface does not represent a useful feature on its own.
+  This sometimes happens when two independent features share a common interface.
```
foolip: I think such features are very rarely independent, and more commonly are closely related:

```suggestion
  This sometimes happens when two features share a common interface.
```

ddbeck: Ah, good catch. I think was trying to cover the nuance of something like 'two features with a non-dependent relationship' (e.g., if you have feature `abc` and feature `abc-later-addition` then you wouldn't add all of `abc`'s keys to `abc-later-addition`). But that's not something likely to be understood in one word anyway, so let's not try!

features/_overlap_allowlist.yml
```diff
@@ -0,0 +1,54 @@
+- features:
```
captainbrosset, line 1: We have an allowlist for specs that aren't really specs, and in it, we have comments for why we need them.
Could we do the same here? I feel like this would bring a lot of value over time.

docs/guidelines.md
```diff
@@ -423,3 +423,35 @@ When you set a `discouraged` block in a feature file, do:
...
+
+## Overlapping `compat_features` keys
+
+Typically, an `@mdn/browser-compat-data` (BCD) key is assigned to one and only one feature.
+Rarely, two or more features have equal claim to a key and their sets of `compat_features` keys must overlap.
+You can add a single key to two or more `compat_features` lists when independent features share some interface and one of these situations applies:
+
+- No feature has any claim to being first and the shared interface does not represent a useful feature on its own.
+  This sometimes happens when two features share a common interface.
+  For example, the `HTMLMediaElement` interface is required for both the `<video>` and `<audio>` elements, but there's no application for the `HTMLMediaElement` alone.
```
ddbeck, line 435:
```suggestion
  For example, the `HTMLMediaElement` interface is required for both the `<video>` and `<audio>` elements, but there's no application for the `HTMLMediaElement` interface alone.
```

features/_overlap_allowlist.yml
ddbeck: @captainbrosset yes, this is a good idea. Would a high-level comment field work for you, something like this?

```suggestion
- reason: "The `<audio>` and `<video>` elements share the `HTMLMediaInterface`, unlike most other elements' interfaces (e.g., `HTMLDivElement` is exclusive to `<div>` elements)."
  features:
```

Or are you expecting that we would justify each feature ID and key individually, with a more deeply nested structure?

captainbrosset: A high-level comment works for me!

PR #4038, opened 2026-05-14
Add meta refresh feature

ddbeck:
This maps our oldest unmapped, undeprecated compat key.

This is an odd corner of the platform—the "header" here [isn't really an HTTP feature](https://lists.w3.org/Archives/Public/ietf-http-wg/2019JanMar/0197.html). I named this feature on the basis of the [wikipedia page](https://en.wikipedia.org/wiki/Meta_refresh). I extracted it from the [`meta`](https://web-platform-dx.github.io/web-features-explorer/features/meta/ "<meta>: The <meta> element represents metadata about the page used by the browser or search engines, including description, keywords, and character sets.") feature since it's quite different than applying metadata.


Elchi3 (approved):
LGTM 👍

PR #4046, opened 2026-05-18
Add css gap decorations BCD keys

captainbrosset:
Chromium 149 supports CSS gap decorations. The relevant BCD keys now exist and are available to web-features.

This PR adds these BCD keys to gap-decorations.yml.

This PR also moves the gap-decorations-related keys off of multi-column.yml and into gap-decorations.yml.
The multi-column.yml feature uses a `compute_from` field, so this does not impact its status.


features/gap-decorations.yml.dist
```diff
@@ -4,3 +4,174 @@
...
+  # baseline_low_date: 2017-03-07
+  # baseline_high_date: 2019-09-07
+  # support:
+  #   chrome: "50"
+  #   chrome_android: "50"
+  #   edge: "12"
+  #   firefox: "52"
+  #   firefox_android: "52"
+  #   safari: "9"
+  #   safari_ios: "9"
```
ddbeck: It seems weird to me that properties that are over 10 years old are now under a headline status with no browser support at all. Can you elaborate on this?

(I'm wondering if there's some sort data issue here, like missing `flex_context` keys, or else multiple features, something like `gap-decoration-multicol`, `gap-decoration-flex`, etc.)

captainbrosset: Gap decorations started with the CSS multicol feature, a long time ago. At the time, it was only possible to decorate gaps in a multicol layout, and only vertical gaps (i.e. columns). At the time, also, gap decorations wasn't really considered as a separate feature, but instead was part of multicol.

Starting with Chrome/Edge 149, gap decorations is now its own, separate, feature. It now can be used both vertically and horizontally, and with grid, flex, and multicol.
Additionally, the feature now has a bunch of new properties to define exactly where the decorations start and end, how they cross, or where they appear at all.

My first idea was to only add the brand new BCD keys to gap-decorations.yml and leave other features untouched.
Then I thought it would make more sense to move the gap decorations-related keys that are in multi-column.yml to gap-decorations.yml.
This way, multi-column.yml is only about, well, multi-column (i.e. the column-width and column-count properties), and gap-decorations.yml contains the complete list of keys that make up the feature.

I did this mostly because multi-column.yml uses `compute_from` to pin its status to just one property.

I can revert if the historical fact that _column_ gap decorations _for multicol only_ seems more important to preserve.

ddbeck: I guess now that I've had more time to think on this, I think we do in fact have missing data.

If we merged this PR, then if I look up the compatibility information for this feature, it'll imply that I can set `column-rule` on a CSS grid container and get rules. But that's not true. It seems like something is missing from BCD, if that's the case (perhaps a key for `<gap-rule-list>` or a `grid_context` key). I think we need to file one or more bugs upstream, to decide how to represent that.

But even if we had that stuff, gap decorations are _conceptually_ different from multicol. Maybe there's a case to be made for a third reason to allow compat keys to be shared between two features (see https://github.com/web-platform-dx/web-features/pull/4034/changes#diff-1dd2fbbb7774410e5b66e9a554df689ef71981117b2c8fdd6570432cf0a13c59), when a key is an entry point into two or more features.

captainbrosset: You're right. `column-rule` would work on a CSS grid, only if `row-rule` is also supported. The addition of `row-rule` (as well as `rule`, and a bunch of other long-hand properties) is sign that the "new" gap decorations are supported.

Maybe the best thing to do here is to keep `column-rule` where it was (in multicol) and only add the new BCD keys to the new feature.

ddbeck: Do you want to make the change here, then we can merge this? If so, I'd be willing to file an issue upstream to figure out how to structure this information in BCD, so we can get the non-multicol coverage there.

captainbrosset: @ddbeck done

jgraham (approved):

Elchi3 (commented):
There are a lot more keys in https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/css-gaps-1.yml. I had added more to BCD. Maybe take another look?

captainbrosset:
> There are a lot more keys in https://github.com/web-platform-dx/web-features/blob/main/features/draft/spec/css-gaps-1.yml. I had added more to BCD. Maybe take another look?

Oh my, this never stops :) 
Now that you mention it, I've heard of these latest additions. Adding them now! Thank you.

PR #4047, opened 2026-05-18
Add connection allowlists feature

### LINKED ISSUE #3834: Connection Allowlists
### Specification

https://wicg.github.io/connection-allowlists/

### Description

Connection Allowlists is a feature designed to provide explicit control over external endpoints by restricting connections initiated via the Fetch API or other web platform APIs from a document or worker.

### Documentation

https://github.com/WICG/connection-allowlists

### Browser support

_No response_

ddbeck:
Fixes: https://github.com/web-platform-dx/web-features/issues/3834


ddbeck:
Blocked by https://github.com/w3c/browser-specs/issues/2253. See https://github.com/w3c/browser-specs/issues/2253#issuecomment-4477933334.

features/connection-allowlists.yml
```diff
@@ -0,0 +1,8 @@
+name: Connection allowlists
+description: The `Connection-Allowlist` HTTP response header limits what resources that a page may request using fetch and other APIs.
```
Elchi3:
```suggestion
description: The `Connection-Allowlist` HTTP response header controls which requests a page may initiate through Fetch and other web platform APIs.
```

Elchi3 (commented):
I think the description can be better. Also, do we want to add this to the "security" group. This is like CSP basically but just for request controlling.

features/connection-allowlists.yml
ddbeck: @elchi3 Can you say more about what you're trying to achieve with this suggestion? I'm scratching my head a little here. "Controls" strikes me as less specific. I'm not clear on why _initiating_ a request is important to emphasize. And our scope is such that I think "web platform" doesn't really need to be stated.

But maybe I'm missing some important context!

Though this could certainly be tightened up:

```suggestion
description: The `Connection-Allowlist` HTTP response header limits which resources a page may request using fetch and other APIs.
```

Elchi3: hm, not sure what June-@Elchi3 was thinking. 

On "initiate" vs "request": I guess my thinking was that a focus on requests would be good, so "limits requests", and then "requesting requests" would be weird wording so I went with "initiate". But anyway I think your last suggestion is great and would work, too.

```suggestion
description: The `Connection-Allowlist` HTTP response header limits which requests a page may initiate using fetch and other APIs.
```

ddbeck: Ooh, I like that version best. Thank you!

Elchi3 (approved):
Approving but not merging in case you want to wait for the BCD keys to be available.

ddbeck:
Thank you, @Elchi3. The tests will now fail, then pass when the BCD upgrade hits.

(Also, I pushed the change adding this to the security group. Cheers!)

PR #4065, opened 2026-05-22
Add variadic `color-mix()` feature

### LINKED ISSUE #4044: `color-mix()` with 3+ colors
### Specification

https://drafts.csswg.org/css-color-5/#funcdef-color-mix

### Description

An extension of `color-mix()` to 3+ colors

### Documentation

MDN mentions multiple arguments: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/color-mix#browser_compatibility

### Browser support

Firefox. Testcase: https://codepen.io/leaverou/pen/azBpNpg

ddbeck:
Fixes: https://github.com/web-platform-dx/web-features/issues/4044

I'm not completely certain about this approach. Arguably, this is incremental syntax change and doesn't merit an independent feature. We disposed of a feature for single-color gradients, though this has more actual utility over its antecedent feature.


features/color-mix-variadic.yml
```diff
@@ -0,0 +1,6 @@
+# Merge candidate into `color-mix`
+name: color-mix() with one or more colors
+description: The `color-mix()` CSS function is variadic, mixing one or more colors.
```
foolip: Can "variadic" be explained? The main new feature here is _three_ or more colors.

foolip (approved):

features/color-mix-variadic.yml
LeaVerou: +1. "Variadic" to me implies that `color-mix()` with a single color would also be valid.

features/color-mix-variadic.yml
```diff
@@ -0,0 +1,6 @@
+# Merge candidate into `color-mix`
+name: color-mix() with one or more colors
```
LeaVerou:
```suggestion
name: color-mix() with three or more colors
```

features/color-mix-variadic.yml
ddbeck: It is indeed the case that a single color is valid (see [the spec text](https://drafts.csswg.org/css-color-5/#color-mix:~:text=the%20color%2Dmix%28%29%20function%20takes%20a%20list%20of%20one%20or%20more%20%3Ccolor)), which is what is tracked in BCD. But the feature for developer is ≥3, so I'll rework this description.

I kinda like keeping `variadic` in the ID though, since it's pithy. Any thoughts on that?

ddbeck:
```suggestion
description: The `color-mix()` CSS function accepts three or more colors.
```

LeaVerou (approved):

PR #3879, opened 2026-03-19
Migrate tsconfig.json from ESNext to NodeNext

caugner:
Migrates `tsconfig.json` from `module: esnext` to `module: nodenext`.

This implies the following changes:

- Add file extensions on imports.
- Use the named Ajv export.

Fixes https://github.com/web-platform-dx/web-features/issues/3878.

PR #4072, opened 2026-05-26
Upgrade and simplify root tsconfig for Node.js 22

ddbeck:
## Summary

This turns on some JavaScript language features and cleans up some longstanding quirks in our TypeScript config.

## Background

[When we upgraded to Node.js 22](https://github.com/web-platform-dx/web-features/pull/4005), we didn't change the TypeScript config, so the type checker won't actually let you use any of the new JavaScript niceties in this version of Node.

On top of that, our root `tsconfig.json` file is a little unconventional (there are other PRs and issues open about this, though this PR doesn't address all of it).

## What's new here

This PR does a few little things:

- To minimize weirdness, uses Microsoft's recommended base configurations from [tsconfig/bases](https://github.com/tsconfig/bases) for Node.js 22 and Node.js + Typescript, so the things that we're doing that are unusual stand out in the `tsconfig.json` file. Then, as a consequence, the three following things.
- Uses `.ts` import extensions, exclusively, in our project scripts.
- Uses explicit `type` imports, where applicable. This is more explicit and will help us drop a dependency in the future.
- Fixes the remaining type errors that resulted.

Finally, this _only_ touches the root `tsconfig.json`. There is some related work to do in the `packages/` directories, but to make this PR easier to review, I elected to do those in follow up PRs.

## Related PRs

- This unblocks https://github.com/web-platform-dx/web-features/pull/4030.
- This will eliminate some work seen in other PRs, especially https://github.com/web-platform-dx/web-features/pull/3879.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,14 +1,14 @@
-import { DefinedError } from "ajv";
+import type { DefinedError } from "ajv";
 import stringify from "fast-json-stable-stringify";
+import { fdir } from "fdir";
 import { execSync } from "node:child_process";
 import fs from "node:fs";
 import path from "node:path";
 import winston from "winston";
-import yargs from "yargs";
-import * as data from "../index.js";
-import { validate, validateProposed } from "./validate.js";
-import { fdir } from "fdir";
 import YAML from "yaml";
+import yargs from "yargs";
+import * as data from "../index.ts";
+import { validate, validateProposed } from "./validate.ts";
 
 const logger = winston.createLogger({
   format: winston.format.combine(
--- a/scripts/inspect-feature.ts
+++ b/scripts/inspect-feature.ts
@@ -1,4 +1,9 @@
 import { coreBrowserSet, getStatus } from "compute-baseline";
+import {
+  Compat,
+  feature,
+  SupportStatement,
+} from "compute-baseline/browser-compat-data";
 import escapeHtml from "escape-html";
 import fs from "node:fs";
 import path from "node:path";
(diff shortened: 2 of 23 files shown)


scripts/build.ts
```diff
@@ -1,14 +1,14 @@
...
 import path from "node:path";
 import winston from "winston";
-import yargs from "yargs";
-import * as data from "../index.js";
-import { validate, validateProposed } from "./validate.js";
-import { fdir } from "fdir";
 import YAML from "yaml";
+import yargs from "yargs";
+import * as data from "../index.ts";
+import { validate, validateProposed } from "./validate.ts";
```
ddbeck, line 11: My editor sorted these imports. Let me know if you want me to revert this.

scripts/inspect-feature.ts
```diff
@@ -94,7 +98,7 @@ function getNotes(compatKey: string): Map<string, string[]> {
   const f = feature(compatKey);
   for (const browserKey of coreBrowserSet) {
     const statements = f
-      .supportStatements(defaultCompat.browser(browserKey))
+      .supportStatements(compat.browser(browserKey))
```
ddbeck, line 101: The package doesn't actually export  `defaultCompat`, which makes the type checker mad. These two change fix that problem.

tsconfig.json
```diff
@@ -1,14 +1,12 @@
...
-    "moduleResolution": "Bundler",
-    "esModuleInterop": true,
-    "resolveJsonModule": true,
     "noEmit": true,
-    "lib": ["ES2023"],
     "types": ["node", "mocha"],
     "strict": false
   },
-  "exclude": ["packages/", "gh-pages/"]
+  "exclude": ["packages/"]
```
ddbeck, line 11: `gh-pages/` doesn't exist anymore.

tsconfig.json
```diff
@@ -1,14 +1,12 @@
 {
+  "extends": [
+    "@tsconfig/node22/tsconfig.json",
+    "@tsconfig/node-ts/tsconfig.json"
+  ],
```
ddbeck, line 5: This is the big difference. We only set the options specific to this project (e.g., `noEmit`), then accept recommended defaults for our runtime.

ddbeck (commented):
My guided tour for the actual reviewer.

foolip (approved):

captainbrosset (approved):

PR #4077, opened 2026-05-27
Add feature for responsive iframes

### LINKED ISSUE #4036: Responsively-sized <iframe>
### Specification

https://drafts.csswg.org/css-sizing-4/#responsive-iframes

### Description

Allow sites to opt into iframes having responsive sizing (sizing the <iframe> element in the parent document to the iframe document's layout overflow sizing, so that scrolling in the child document is avoided).

### Documentation

https://github.com/w3c/csswg-drafts/blob/main/css-sizing-4/responsive-iframes-explainer.md

### Browser support

https://chromestatus.com/feature/5108373464547328

ddbeck:
Adds a feature for responsive iframes with the `frame-sizing` CSS property, `<meta name="responsive-embedded-sizing">` element, and `window.requestResize()` API.

Fixes https://github.com/web-platform-dx/web-features/issues/4036.


features/frame-sizing.yml
```diff
@@ -0,0 +1,13 @@
+name: Responsive iframes
+description: >
+  The `frame-sizing` CSS property sets the size of an `<iframe>` element to that of it contents, to avoid scrolling in the iframe.
+  Embedded documents allow this with the `<meta name="responsive-embedded-sizing">` element and request resizing in the embedding page by calling `window.requestResize()`.
+spec: https://drafts.csswg.org/css-sizing-4/#propdef-frame-sizing
+# Expected
+# compat_features:
+#   - css.properties.frame-sizing
+#   - css.properties.frame-sizing.auto
+#   - css.properties.frame-sizing.content-width
```
Elchi3: C.f. https://collector.openwebdocs.org/tests/css/properties/frame-sizing
```suggestion
#   - css.properties.frame-sizing.content-height
#   - css.properties.frame-sizing.content-width
```

features/frame-sizing.yml
```diff
@@ -0,0 +1,13 @@
...
+#   - css.properties.frame-sizing.content-block-size
+#   - css.properties.frame-sizing.content-inline-size
+#   - api.Window.requestResize
```
Elchi3: Maybe also this?

```suggestion
#   - api.Window.requestResize
#   - html.elements.meta.name.responsive-embedded-sizing
```

Elchi3 (approved):
Description looks great to me! Just two thoughts on expected keys.

features/frame-sizing.yml
```diff
@@ -0,0 +1,13 @@
+name: Responsive iframes
+description: >
+  The `frame-sizing` CSS property sets the size of an `<iframe>` element to that of it contents, to avoid scrolling in the iframe.
+  Embedded documents allow this with the `<meta name="responsive-embedded-sizing">` element and request resizing in the embedding page by calling `window.requestResize()`.
+spec: https://drafts.csswg.org/css-sizing-4/#propdef-frame-sizing
+# Expected
+# compat_features:
+#   - css.properties.frame-sizing
```
malchata: Question for @ddbeck: The linked spec in the feature here mentions the frame's "internal intrinsic size"—would this warrant linking the `min-content` and `max-content` CSS keywords in this feature? I'm referring specifically to the intrinsic size concept, as it figures in pretty heavily here for `<iframe>` contents.

On the other side, I suppose it could be argued that you'd then have to add references for these keywords to many features' `compat_features` field.

ddbeck: @malchata thanks for your patience while I got back to this. We have a data policy to not add keys that are not immediately related to the feature or is necessary for the computed status to make sense (or else there would be a lot of duplication if we captured all of the "dependencies" on every feature). Hope that helps!

PR #4091, opened 2026-06-03
Add TextModules feature

o-:


o-:
Closing, since #579 seems to indicate that import attributes are not tracked separately.

PR #4115, opened 2026-06-17
Create service-workers-static-routing feature

jgraham:
This is a distinct part of the ServiceWorker API that isn't supported in all browsers.


features/service-workers-static-routes.yml
```diff
@@ -0,0 +1,8 @@
+name: Service workers static routing
+description: A declarative routing API for service workers that allows performance optimisation through bypassing the service worker for specified routes.
```
captainbrosset: An attempt to name the entry point and be a little more consistent with other descriptions:

```suggestion
description: "The service worker install event's `addRoutes` method declares resource paths that should either be fetched from the network or the cache, to bypass the service worker and improve performance."
```

captainbrosset (approved):
This does seem like a separate feature, and it doesn't impact the baseline status of the parent feature. LGTM modulo some description changes.

PR #4116, opened 2026-06-17
Create a separate feature for anchor position transforms.

jgraham:
This is a specific case where browsers have different levels of support, which we should surface through web-features. It's also a case where authors could avoid the interaction between the features if they know it can cause problems.

Presumably later once both features are baseline we'll want to re-merge this into the main anchor positioning feature.


features/anchor-positioning-transforms.yml
```diff
@@ -0,0 +1,5 @@
+name: Anchor positioning transforms
+description: Anchor positioned elements take CSS transforms into account.
```
jamesnw:
```suggestion
description: Anchor positioned elements take CSS transforms on their anchor elements into account.
```
There is also a likely future feature to be able to control whether or not to take CSS transforms into account. I don't know if we want to make it clear that this is always taking it into account, to help differentiate a theoretical future feature from this?

jamesnw (approved):
LGTM!

captainbrosset (approved):
LGTM.

This made me want to look into what's blocking Chrome/Edge from supporting.

I found this set of keys:

- css.properties.position-anchor
- css.properties.position-anchor.auto
- css.properties.position-anchor.none
- css.properties.position-anchor.normal

And

- css.properties.position-visibility.anchors-valid

For the first set (`position-anchor`), this seems mostly due to https://issues.chromium.org/issues/501399489 which was fixed last week. So this should get resolved soon.

For the second (`position-visibility`), it looks like Chromium doesn't yet have support for the `anchors-valid` value, and that's tracked in https://issues.chromium.org/issues/333421963. The first comment in that bug goes 

> This was implemented in https://crrev.com/1280441 but later reverted in https://crrev.com/1283614 due to open spec issues (see: [github discussion](https://github.com/w3c/csswg-drafts/issues/7758#issuecomment-2038106525) and issues on [anchors-valid spec](https://drafts.csswg.org/css-anchor-position-1/#position-visibility)).

which is making me feel like we might also want to extract `position-visibility` to its own feature. @jamesnw @jgraham thoughts?

PR #4138, opened 2026-06-26
Add heading offset

Th3S4mur41:
This adds the `headingoffset`and `headingreset` attributes as specified in https://html.spec.whatwg.org/multipage/sections.html#heading-levels-&-offsets

The feature is currently being prototyped in Firefox and Chromium: 
- https://bugzilla.mozilla.org/show_bug.cgi?id=1974383
- https://issues.chromium.org/issues/333628468

Matching PR in BCD: https://github.com/mdn/browser-compat-data/pull/29931


features/heading-offset.yml
```diff
@@ -0,0 +1,4 @@
+name: Heading offset
```
captainbrosset, line 1: Many HTML attribute features on the repo just use the name of the attribute as the `name` field. So this could be `headingoffset`.
That said, this feature is actually about 2 attributes. So your proposal could work too (for example, we have the `Interest invokers` feature, which is also about multiple attributes).

features/heading-offset.yml
```diff
@@ -0,0 +1,4 @@
...
+description: The `headingoffset` attribute offsets the heading level of descendant `<h1>`–`<h6>` elements. The `headingreset` boolean attribute prevents heading offset computation from traversing beyond the element.
```
captainbrosset: Other HTML attribute features mention "HTML" in their description. We should do this too here, for consistency:

```suggestion
description: The `headingoffset` HTML attribute offsets the heading level of descendant `<h1>`–`<h6>` elements. The `headingreset` HTML attribute prevents heading offset computation from traversing beyond the element.
```

features/heading-offset.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://html.spec.whatwg.org/multipage/sections.html#heading-levels-&-offsets
+group: html
```
captainbrosset, line 4: I noticed that BCD now has keys for this feature, let's add them in:


```suggestion
group: html
compat_features:
  - html.global_attributes.headingoffset
  - html.global_attributes.headingreset
  - api.HTMLElement.headingOffset
  - api.HTMLElement.headingReset
```

features/heading-offset.yml.dist
```diff
@@ -0,0 +1,6 @@
+# Generated from: heading-offset.yml
```
captainbrosset, line 1: You'll need to run `npm run dist features/heading-offset.yml` again after adding the keys.

captainbrosset (changes requested):

features/heading-offset.yml
Th3S4mur41: I've bumped @mdn/browser-compat-data to 8.0.5 to add support

features/heading-offset.yml
Th3S4mur41: Yes, I actually followed the pattern from `Popover` for this one, so I'll leave it as it is, if it's ok with you?

package.json
```diff
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@eslint/js": "^10.0.1",
     "@js-temporal/polyfill": "^0.5.1",
-    "@mdn/browser-compat-data": "8.0.4",
+    "@mdn/browser-compat-data": "8.0.5",
```
captainbrosset: Please remove this change (and also revert package-lock.json), I've bumped web-features to use the latest BCD in https://github.com/web-platform-dx/web-features/pull/4157.
(there's a whole process to it, and we do it in separate PRs)

captainbrosset (approved):

PR #4149, opened 2026-07-02
Initial submission for mixins.

thejohnjansen:
I am creating a ChromeStatus entry for CSS-Mixins. Development is just starting on this at Microsoft. The ChromeStatus feature entry requests a link to the Web Feature ID, so I am submitting this PR for that purpose.


features/mixins.yml
```diff
@@ -0,0 +1,4 @@
+name: CSS mixins
+description: "CSS mixins let you define reusable style fragments and invoke them with custom values, reducing repeated declarations across stylesheets."
```
captainbrosset: We usually lead descriptions with the entry point a developer would use. This description is currently only prose, with no syntax. Here is another proposal:

```suggestion
description: "The `@mixin` and `@apply` CSS at-rules define reusable style declaration fragments and apply them inside rules, optionally passing arguments.
```

features/mixins.yml
```diff
@@ -0,0 +1,4 @@
+name: CSS mixins
```
captainbrosset: We don't usually prefix names with the technology, and we could probably just use the syntax token here, for clarity, and consistency with `@function`:

```suggestion
name: "@mixins"
```

features/mixins.yml
```diff
@@ -0,0 +1,4 @@
...
+description: "CSS mixins let you define reusable style fragments and invoke them with custom values, reducing repeated declarations across stylesheets."
+spec: https://drafts.csswg.org/css-mixins-1/
```
captainbrosset: The spec covers functions and mixins, let's point at the mixins section anchor:

```suggestion
spec: https://drafts.csswg.org/css-mixins-1/#mixin-rule
```

features/mixins.yml
```diff
@@ -0,0 +1,4 @@
...
+group: css
```
captainbrosset: For features that aren't yet covered in the browser-compat-data project, we usually try to add _expected_ BCD keys. Here is what I would add here:

```suggestion
group: css
# Expected:
# compat_features:
#   - css.at-rules.mixin
#   - css.at-rules.mixin.contents
#   - css.at-rules.mixin.result
#   - css.at-rules.apply
```

captainbrosset (changes requested):
Thank you @thejohnjansen for the PR. Here are some suggestions you could apply.

captainbrosset (approved):

PR #4165, opened 2026-07-10
supports-at-rule: map at-rule() compat key

3ru:
The BCD key for it now exists, so this wires it up and lets the status be computed


captainbrosset (approved):
Thank you!

PR #4166, opened 2026-07-10
languagemodel: map expected compat keys

3ru:
The authoring comment listed the event as `oncontextoverflow_event`; the actual BCD key is `contextoverflow_event` (BCD names DOM events `<name>_event`, without the on handler prefix), so updated


captainbrosset (approved):
Thanks!

PR #4180, opened 2026-07-15
Declarative adopted stylesheets for DSD

### LINKED ISSUE #4148: shadowrootadoptedstylesheets on HTMLTemplateElement
### Specification

PR here: https://github.com/whatwg/html/pull/12339

### Description

`shadowrootadoptedstylesheets` allows for setting `adoptedStyleSheets` on declarative shadow DOM. It accepts a space-separated list of specifiers that are fetched and added to the generated shadow root's `adoptedStyleSheets` array in specified order.

### Documentation

https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ShadowDOM/explainer.md

### Browser support

This is currently only supported in Chromium browsers behind Experimental Web Platform Features. We are planning on shipping it in Chromium 152. More details are under https://chromestatus.com/feature/4790543041298432.

captainbrosset:
Fixes #4148.

This introduces a new feature for the `shadowrootadoptedstylesheets` attribute on `<template>` elements.
This attribute allows sharing stylesheets in declarative shadow DOMs.


features/shadowrootadoptedstylesheets.yml
```diff
@@ -0,0 +1,7 @@
+name: '<template shadowrootadoptedstylesheets>'
```
captainbrosset: I opted for the formal tag+attribute name here in the name field, because I couldn't find anything I was very happy with. But, other features in the same area tend to use "feature" names instead. For example, "declarative shadow dom" doesn't use `<template shadowrootmode>` as its name.

PR #4183, opened 2026-07-15
cookie-store: move non-standard CookieListItem properties to spare

bakulf:


jgraham:
I didn't notice this one before creating https://github.com/web-platform-dx/web-features/pull/4189. I somewhat prefer my approach since it makes it clear that some browsers are implementing nonstandard properties rather than just hiding that fact, but I'm happy for either to land.

bakulf:
I prefer your approach!

PR #4189, opened 2026-07-16
Split nonstandard properties from cookie-store feature

jgraham:
Getting cookie metadata from CookieStore.get was removed from the specification, since it creates new compatibility problems (browsers lose freedom to override the metadata) and potential new privacy problems (since it provides access to data that wasn't previously available via document.cookie).

Put these properties into a new discouraged feature.


features/cookie-store-get-metadata.yml
```diff
@@ -0,0 +1,13 @@
+name: Cookie store get metadata
```
captainbrosset, line 1:
```suggestion
name: cookieStore.get() metadata
```

I think this is simpler with the entry point written like this.

features/cookie-store-get-metadata.yml
```diff
@@ -0,0 +1,13 @@
...
+description: The `domain`, `expires`, `path`, `sameSite` and `secure` properties returned from CookieStore.get
```
captainbrosset, line 2:
```suggestion
description: The `cookieStore.get()` method returns cookie metadata, such as the `domain`, `expires`, `path`, `sameSite` and `secure` properties.
```

captainbrosset: Should `partitioned` be in that list of properties too?

captainbrosset (changes requested):

captainbrosset (approved):
LGTM. Just unsure about pointing to the PR URL in `according_to`. If possible, I think we should point to language within the spec.

jgraham:
Not sure what we could point to in the spec to show that something isn't present any more. But happy to fix that in a future PR if we find a better reference.

PR #4212, opened 2026-07-29
Add iterator-join feature

o-:


o-:
Not sure if this makes sense. Every TC39 feature will first appear as not part of the ecma262 repo, because features are merged to the main repo *after* browsers ship them. Shouldn't we rather allow all proposals under the tc39 organization?

o-:
looks like this tc39 feature was already auto imported

PR #4226, opened 2026-08-05
Add text-modules feature

o-:
Still not sure if this should be added. However, I now realized that module-json exists.


features/text-modules.yml
```diff
@@ -0,0 +1,4 @@
+name: Text import attributes
+description: 'Module `import … with { type: "text" }` statements load text data.'
```
ddbeck: I think it would be nice to explain what happens to the imported data here:

```suggestion
description: 'Module `import … with { type: "text" }` statements load text data as a string value.'
```

features/text-modules.yml
```diff
@@ -0,0 +1,4 @@
...
+spec: https://github.com/tc39/proposal-import-text
```
ddbeck: If you link to the specification text, then you don't need the spec exception:

```suggestion
spec: https://tc39.es/proposal-import-text/
```

scripts/specs.ts
```diff
@@ -243,6 +243,10 @@ const defaultAllowlist: allowlistItem[] = [
     [
         "https://github.com/whatwg/html/pull/9546",
         "Allowed because the <input type=\"checkbox\" switch> feature is not yet in a formal spec. Remove this exception when a formal spec is available."
+    ],
+    [
+        "https://github.com/tc39/proposal-import-text",
+        "Allowed because import text proposal is not yet merged into ecma262."
     ]
```
ddbeck: This is not required if you use a URL known to the `browser-specs` package ([source](https://github.com/w3c/browser-specs/blob/a42d148a02f419c0408409b5ef8d25bc5401c4f2/specs.json#L462)).

features/text-modules.yml
```diff
@@ -0,0 +1,4 @@
...
+group: text
```
ddbeck: `text` is a subgroup of `css`, so this probably not the right choice. I'd suggest this grouping instead:


```suggestion
group: js-modules
```

features/text-modules.yml
ddbeck, line 1: If you append this and regenerate the dist, we'll get a more accurate support status:

```
compat_features:
  - javascript.statements.import.import_attributes.type_text
```

ddbeck (dismissed):
Thanks for opening this, @o-. I've made a few suggestions in line comments, but I think this is worth minting a feature for. Thanks again!

o-:
Thanks for the help. Can you have another look?

captainbrosset (approved):

PR #4237, opened 2026-08-11
Add soft-navigations and interaction-contentful-paint

tunetheweb:
Adds the `soft-navigations` and `interaction-contentful-paint` features as part of the [Soft Navigations and Interaction Contentful Paint spec](https://wicg.github.io/soft-navigations/) supported [by default in Chrome 151](https://chromestatus.com/feature/5144837209194496).

There is an argument that these are one feature, instead of two, since they are in the same spec, were released at the same time, and the main use case for `interaction-contentful-paint` at present is for measuring LCP for soft navigations, however I have decided to split this into two features for a number of reasons:

- `interaction-contentful-paint` is likely to be split into it's own spec in the future.
- `interaction-contentful-paint` is emitted even when no soft navigation occurs and has use cases beyond soft navs.

Let me know if you disagree.

Depends on https://github.com/w3c/browser-specs/pull/2593 landing first.

Note the BCD data depends on this PR: https://github.com/mdn/browser-compat-data/pull/30195 so is commented out for now.


features/interaction-contentful-paint.yml
```diff
@@ -0,0 +1,12 @@
+name: Interaction Contentful Paint
```
captainbrosset:
```suggestion
name: Interaction contentful paint
```

features/interaction-contentful-paint.yml
```diff
@@ -0,0 +1,12 @@
...
+description: The `interaction-contentful-paint` performance entry and the `InteractionContentfulPaint` API is emitted by the browser when an interaction causes one or more contentful paints. It is useful to measure LCP for SPA-style soft navigations or for measuring contentful paints due to an interaction.
```
captainbrosset: The phrase "The `e` performance entry and the `A` API **is** emitted by the ..." feels incorrect. We're talking about two things, so it should be **are** emitted, but that also feels wrong.

Is my understanding correct that the API is for the performance entry object which you receive when listening for the event?

We could focus on the event name only and not mention the interface name (which I feel would be fine, since the event name is really the entry point here):

**The `interaction-contentful-paint` performance entry is emitted by the browser when an interaction [rest unchanged]**

Or we could rephrase as:

**The `interaction-contentful-paint` performance entry, which returns a `InteractionContentfulPaint` instance, is emitted by the browser when an interaction causes one or more contentful paints. [rest unchanged]**

(or something like this).

I'm realizing that our other performance entry features are also weirdly formatted. E.g. `largest-contentful-paint`: "The `largest-contentful-paint` performance entry and the `LargestContentfulPaint` API measure**s** the time ..."
We should fix those as a separate PR.

features/soft-navigations.yml
```diff
@@ -0,0 +1,13 @@
+name: Soft Navigations
```
captainbrosset: Would it make sense to use the word "timing" here, like in the [navigation timing feature](https://web-platform-dx.github.io/web-features-explorer/features/navigation-timing/). So maybe:

```suggestion
name: Soft navigation timing
```

It feels easier to understand/find to me. But let me know what you think.

features/soft-navigations.yml
```diff
@@ -0,0 +1,13 @@
...
+description: The `soft-navigation` performance entry and the `PerformanceSoftNavigation` API is emitted by the browser when an SPA-style soft navigation occurs. This is defined as when a trusted user interaction causes a URL update and a visible contentful paint to the screen.
```
captainbrosset: Similar comment than in the other file.

captainbrosset (changes requested):

features/interaction-contentful-paint.yml
tunetheweb: BTW not sure I totally agree with this btw since this is a proper name. But seems to be the convention here so 🤷‍♂️

features/interaction-contentful-paint.yml
tunetheweb: Yeah that's where I based these upon. The last one SGTM.


```suggestion
description: The `interaction-contentful-paint` performance entry, which returns a `InteractionContentfulPaint` instance, is emitted by the browser when an interaction causes one or more contentful paints. It is useful to measure LCP for SPA-style soft navigations or for measuring contentful paints due to an interaction.
```

features/soft-navigations.yml
tunetheweb: The entry does include timing information it's true, but some may be interested just the mere fact that a soft navigation occurred and not necessarily the timing. Plus, unlike navigation timing it's not an explicit part of the API name. So think I'd prefer to keep it as is.

features/soft-navigations.yml
tunetheweb:
```suggestion
description: The `soft-navigation` performance entry, which returns a `PerformanceSoftNavigation` instance, is emitted by the browser when an SPA-style soft navigation occurs. This is defined as when a trusted user interaction causes a URL update and a visible contentful paint to the screen.
```

features/soft-navigations.yml
tunetheweb: Though I see what you mean now about it reading a little difficult without something.

WDYT about this?

```suggestion
name: Soft navigations performance entries
```

tunetheweb: OR:

```suggestion
name: Soft navigations measurement
```

features/interaction-contentful-paint.yml
```diff
@@ -0,0 +1,12 @@
+name: Interaction contentful paint
+description: The `interaction-contentful-paint` performance entry, which returns a `InteractionContentfulPaint` instance, is emitted by the browser when an interaction causes one or more contentful paints. It is useful to measure LCP for SPA-style soft navigations or for measuring contentful paints due to an interaction.
```
captainbrosset: We're above the 300 chars limit unfortunately.
How about:

```suggestion
description: The `interaction-contentful-paint` performance entry, which returns an `InteractionContentfulPaint` instance, is emitted when an interaction causes one or more contentful paints. Use it to measure LCP for soft navigations in single page apps or to measure contentful paints due to an interaction.
```

features/soft-navigations.yml
```diff
@@ -0,0 +1,13 @@
+name: Soft navigations performance entries
+description: The `soft-navigation` performance entry, which returns a `PerformanceSoftNavigation` instance, is emitted by the browser when an SPA-style soft navigation occurs. This is defined as when a trusted user interaction causes a URL update and a visible contentful paint to the screen.
```
captainbrosset: Thought I'd expand SPA here too, for more clarity:


```suggestion
description: The `soft-navigation` performance entry, which returns a `PerformanceSoftNavigation` instance, is emitted when soft navigation occurs in a single page app. This is defined as when a trusted user interaction causes a URL update and a visible contentful paint to the screen.
```

captainbrosset (approved):
Approved with 2 suggestions to fully define SPA and stay under the 300 chars limit.

tunetheweb:
Thanks @captainbrosset ! I think I still need the web spec change mentioned in the first comment though. Will comment here once that’s merged and released.

captainbrosset:
> Thanks @captainbrosset ! I think I still need the web spec change mentioned in the first comment though. Will comment here once that’s merged and released.

Correct, that's why the test CI job is failing with:

`error: URL for interaction-contentful-paint not in web-specs: https://wicg.github.io/soft-navigations/#sec-interaction-contentful-paint`

and

`error: URL for soft-navigations not in web-specs: https://wicg.github.io/soft-navigations/#sec-soft-navs`

features/soft-navigations.yml
```diff
@@ -0,0 +1,13 @@
+name: Soft navigations performance entries
```
captainbrosset:
```suggestion
name: Soft navigation performance entries
```

tunetheweb:
@captainbrosset think this can merge now!

ddbeck (approved):
I'm still a little unsatisfied with the descriptions here, but I'm going to merge this to get it into today's release and send a follow up PR. Thank you!

Issue #4073, opened 2026-05-26
Out of order streaming with <template for>


noamr:
### Specification

https://github.com/whatwg/html/pull/11818 (will end up in HTML standard)

### Description

Allow updating exiting DOM using a template element

### Documentation

https://github.com/WICG/declarative-partial-updates/blob/main/patching-explainer.md

tunetheweb:
Suggest we use `template-for` as the feature id? Can we reserve that?

tunetheweb:
Oh digging in I see that @foolip used this feature as an example in #3726 with the new ["proposed" directory and README](https://github.com/web-platform-dx/web-features/tree/main/features/draft/proposed), temporarily named it `html-pis.yml` but did show of an example when this could be renamed to `template-for`.

@foolip (or anyone else), not sure if that was just an example, but any thoughts on which of those two names we want to use for this particular example? @noamr needs it for the I2S and I need it for the MDN docs.

ddbeck:
Thanks for filing this. I'm taking some notes here to help someone author the YAML file for this feature. I'd welcome a review of this information from @noamr and @tunetheweb, but it's not required.

## Name and ID

The name for this feature is probably something like _Out-of-order HTML streaming_.

Since an entry point is an attribute for the `<template>` element (feature ID `template`), `template-for` seems like a natural ID for this, though it would probably never be a candidate for merging back into `template`.

That said, because the processing instructions are _also_ a kind of entrypoint for this, we might instead consider shadowing the name `out-of-order-html-streaming` or doing something like `html-patching-processing-instructions` to avoid suggesting that it's merely an extension of `<template>`'s behavior alone.

I'm still mulling this over and would welcome impressions on this.

## Description

This is pretty novel, so it's a tough one to write. Here's my rough draft:

> The `<template for="…">` HTML element replaces the content at a named range or insertion point, marked by the processing instructions `<?start name="…">` and `<?end>` or `<?marker name="…">`. You can use this to stream partial updates into the document out of order.

## Compat data

I'm expecting these BCD keys:

```yaml
- html.elements.template.for
- api.ProcessingInstruction
- api.ProcessingInstruction.ProcessingInstruction
- api.ProcessingInstruction.target
- api.ProcessingInstruction.hasAttributes
- api.ProcessingInstruction.getAttributeNames
- api.ProcessingInstruction.getAttribute
- api.ProcessingInstruction.setAttribute
- api.ProcessingInstruction.removeAttribute
[...]

PR #4238, opened 2026-08-11
Add HTML processing instructions feature

tunetheweb:
Chrome 150 supports Processing Instructions in HTML documents.

Processing instructions such as `<?xml>` were originally only supported in XML documents, and treated as comments when present in HTML documents. Browsers have started supporting them in HTML documents for out-of-order streaming (using `<?start>`, `<?end>`, and `<?marker>`) and other use cases.

There is an argument to include this as part of out-of-order streaming/template-for and while that use case was why it was originally enabled in Chrome, there are use cases beyond that (e.g. JavaScript frameworks can use this for their own patching) so I think this should be a separate feature.


tunetheweb:
Depends on #4245 landing first

Edit: Done.

ddbeck:
Also noting that this is towards https://github.com/web-platform-dx/web-features/issues/4073.

features/html-processing-instructions.yml
```diff
@@ -0,0 +1,13 @@
+draft_date: 2026-08-11
+name: HTML processing instructions
+description: Processing instructions such as `<?xml>` were originally only supported in XML documents, and treated as comments when present in HTML documents. Browsers have started supporting them in HTML documents for out-of-order streaming (using `<?start>`, `<?end>`, and `<?marker>`) and other use cases.
```
ddbeck: In our style, we'd probably write something along these lines, so it's a bit more oriented towards developers who are maybe unconcerned with the history of a feature. This is a bit hard to write about though, since it's quite low level (compared to `template-for`). This is probably not ideal, but maybe gives you an idea of where to go from here?

```suggestion
description: An HTML processing instruction is text enclosed by `<?` and `>`, which marks a target for further processing. They may appear anywhere in a document and are treated like comments when they're not recognized for some purpose.
```

features/html-processing-instructions.yml
ddbeck, line 1: The `features/draft/spec` folder is exclusively for generated files.

We don't need a draft here—we can publish this or near enough to it. To fix this, move both files to `features/` and remove the `draft_date` line. (You might have to refresh the dist, though I doubt it.)

ddbeck (changes requested):
Thanks for opening this, @tunetheweb! A couple of small suggestions for you.

features/html-processing-instructions.yml
tunetheweb: I went with a slight variant:

```suggestion
description: An HTML processing instruction is text enclosed by `<?` and `>`, which marks a target for further processing. They may appear anywhere in a document. They are ignored like comments when they're not recognized for a specific purpose.
```

features/html-processing-instructions.yml
tunetheweb: Moved!

tunetheweb:
Thanks @ddbeck ready for another look when you get a chance.

I also have #4265 for the second part of this if really keen 😁

ddbeck (approved):
Thank you!

tunetheweb:
Thanks @ddbeck . Do you happen to know if that feature id gets copied across to MDN BCD using those compat keys? Or do I need to do that manually?

ddbeck:
@tunetheweb that ought to happen automatically — your job is done here. 🙂

PR #4265, opened 2026-08-24
Add `<template for>`

### LINKED ISSUE #4073: Out of order streaming with <template for>

tunetheweb:
Closes #4073 

Note dependent on https://github.com/mdn/browser-compat-data/pull/30292 for BCD keys


ddbeck (approved):
Great, very nice.

@tunetheweb I'll wait to merge this until the next BCD/web-features release is imminent. If the BCD keys land in that release, we can uncomment the keys in this PR (it looks like I can push to the PR branch too, so there's no blocking here). Otherwise, we'll merge without the keys to get the feature minted in the next web-features release and take up assigning the keys on the next release.

Thank you!

tunetheweb:
SGTM!

I've been struggling to understand if the preferred route is web-feature first, BCD second or vice versa since they both reference each other! Do you know? Or have a preference/recommendation?

ddbeck:
My preference would be to get out from under the backlog of feature entry requests for features that have already shipped, then be in a world where we're minting features just as implementation work is in progress, so we'd be ahead of BCD. In that world, we'd assign keys to features after they've already been minted.

That said, _right now_ I'm intent on catching up, so my attention is currently on shipping features where BCD keys already exist (or ought to).

features/template-for.yml
```diff
@@ -0,0 +1,6 @@
+name: <template for>
+description: The `<template for="…">` HTML element replaces the content at a named range or insertion point, marked by the processing instructions `<?start name="…">` and `<?end>` or `<?marker name="…">`. This is used this to stream or "patch" partial updates into the document out of order.
```
ddbeck: oh I just noticed the typo here ("this" twice), which lead to…

```suggestion
description: The `<template for="…">` HTML element replaces the content at a named range or insertion point, marked by the processing instructions `<?start name="…">` and `<?end>` or `<?marker name="…">`. You can use this to stream or "patch" partial updates into the document out of order.
```

tunetheweb:
@ddbeck think we can merge this now!

PR #4270, opened 2026-08-27
Add anchor positioning animations and transitions feature

ddbeck:
This creates a subordinate anchor positioning feature to capture animation and transition behaviors in the way `anchor()` and `anchor-size()` are resolved. See https://github.com/web-platform-dx/web-features/issues/3558#issuecomment-5358733231 for further discussion.

For the description, I cribbed a little from [this note](https://drafts.csswg.org/css-anchor-position-1/#resolvable-anchor-function:~:text=transitions%20or%20animations%20of%20a%20property%20using%20an%20anchor%20function%20will%20work%20%22as%20expected%22) in the specification. Unfortunately, it appears far away from the most relevant spec text that I could link to.

Tests will fail when I open this. It'll remain blocked by https://github.com/mdn/browser-compat-data/pull/30339 and a subsequent BCD release.


features/anchor-positioning-animations.yml
```diff
@@ -0,0 +1,6 @@
+name: Anchor positioning animations and transitions
+description: The `anchor()` and `anchor-size()` CSS functions animate or transition on changes such as the anchor box moving, anchor elements being added or removed from the document, the anchor element changing.
```
jamesnw:
```suggestion
description: The `anchor()` and `anchor-size()` CSS functions animate or transition on changes such as the anchor box moving or the anchor element changing.
```
This could be simplified, since the effect of the anchor elements being added or removed is that the anchor element changes.

jamesnw (approved):

PR #4273, opened 2026-08-27
Add `window-drag` feature

### LINKED ISSUE #3948: window-drag
### Specification

Intended to be added to the [CSS UI 4](https://drafts.csswg.org/css-ui-4/) spec as a rename of `app-region`.
CSSWG resolution: https://github.com/w3c/csswg-drafts/issues/13102
Draft PR: https://github.com/w3c/csswg-drafts/pull/13801

### Description

The `window-drag` CSS property allows web content to designate regions of an installed desktop web app’s UI that should behave as draggable window titlebar areas. When applied, user pointer interactions (e.g., click-and-drag) on that region move the top-level application window rather than triggering normal page interaction.

### Documentation

Developer documentation for using `app-region` with WCO: https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps/how-to/window-controls-overlay

### Browser support

`window-drag` is a rename of `app-region`, which is available on [these browsers](https://caniuse.com/?search=app-region).

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/3948.


features/window-drag.yml
```diff
@@ -0,0 +1,12 @@
+name: window-drag
+description: >
+  The `window-drag: move` CSS declaration sets an element as a window drag handle, like a device's app title bars.
```
captainbrosset:
```suggestion
  The `window-drag: move` CSS declaration sets an element as a window drag handle, like a device's app title bar.
```

features/window-drag.yml
```diff
@@ -0,0 +1,12 @@
...
+  You can use this to add click-and-drag areas to a progressive web app, often alongside window controls overlay.
```
captainbrosset:
```suggestion
  You can use this to add click-and-drag areas to an installed progressive web app, often alongside window controls overlay.
```

Do you think this addition is needed? You can use a PWA without installing it, in which case window-drag does nothing.

captainbrosset (approved):

features/window-drag.yml
ddbeck: I think this helps make it more clear what's going on here, which is to make installed apps seem more native. This works for me.

PR #4275, opened 2026-08-27
Add `alpha()` CSS feature

### LINKED ISSUE #4086: CSS Relative Alpha Colors
### Specification

https://drafts.csswg.org/css-color-5/#relative-alpha

### Description

Relative alpha colors refer to an origin color, and only change the alpha channel. The meaning of alpha channels is defined in CSS Color 4 § 4.2 Representing Transparency in Colors: the <alpha-value> syntax.

### Documentation

https://bsky.app/profile/nerdy.dev/post/3mmsrchlp6k2o
https://github.com/mozilla/standards-positions/issues/1396#issuecomment-4363280736

### Browser support

https://chromestatus.com/feature/5070160203481088

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/4086.


features/alpha.yml
```diff
@@ -0,0 +1,6 @@
+name: alpha()
+description: The `alpha()` CSS function computes a color value by adjusting the opacity of a color relative to an origin color.
```
captainbrosset: "adjusting the opacity of a color relative to an origin color" confuses me. It makes it sound like we're passing 2 colors to the function.

How about:


```suggestion
description: The `alpha()` CSS function computes a new color value by adjusting the opacity of an origin color.
```

captainbrosset (changes requested):

features/alpha.yml
ddbeck: I was trying to shoehorn the word "relative" (as in "relative color syntax") in to the description, so it would search nicely. That didn't work.

The original version was this:

```suggestion
description: The `alpha()` CSS function computes a color value by adjusting the opacity of an origin color.
```

What do you think of something like this instead?

```suggestion
description: The `alpha()` CSS function computes a relative color value by adjusting the opacity of an origin color.
```

captainbrosset (approved):

julienw:
Hey folks,

How is the .dist file was generated? Firefox has this since v155 (see https://github.com/mdn/browser-compat-data/blob/main/css/types/color.json -- where it was added 2 weeks ago, so before this PR was created).

ddbeck:
@julienw The dist files only reflect stable browser releases. Since Firefox 155 is still in beta, we don't show that information. Changing this is on our radar (see https://github.com/web-platform-dx/web-features/issues/955), but when BCD updates Firefox 155's status to "current" (see [`firefox.json`](https://github.com/mdn/browser-compat-data/blob/12dd89bce05a531172ee8dfe1f56ed980c851dee/browsers/firefox.json#L1113)), we'll pick that up in the web-features dist file and release that follows. I hope that helps!

Edit: I see that today might be the release date for 155? If so, I expect that to flip in BCD's release (and web-features) on Thursday of this week. Again, this is an area that needs some improvement. 😅

julienw:
Yeah Firefox 155 shipped today :-) So I guess it's just the matter of getting everything synchronized!

I was surprised because in https://caniuse.com/?search=alpha it doesn't show v155 as being supported, using the data from this project, despite that the [child feature](https://caniuse.com/mdn-css_types_color_alpha) does.

ddbeck:
That's right. The child features are from BCD, which is a distinct data source that caniuse has long shown (with the MDN filter). caniuse's ingestion of web-features entries is relative new, so this quirk was not previously apparent. Again, doing something about #955 will help here.

PR #4278, opened 2026-08-27
Revise performance timing entries' names and descriptions

ddbeck:
A follow up to https://github.com/web-platform-dx/web-features/pull/4237.

### DIFF
--- a/features/element-timing.yml
+++ b/features/element-timing.yml
@@ -1,5 +1,5 @@
 name: Element timing
-description: The `element` performance entry and the `PerformanceElementTiming` API measures the time to render elements that have the `elementtiming` attribute.
+description: An `element` performance entry measures the time to render elements that have the `elementtiming` attribute.
 spec: https://w3c.github.io/element-timing/
 group: performance
 compat_features:
--- a/features/element-timing.yml
+++ b/features/element-timing.yml
@@ -1,4 +1,4 @@
-name: Element timing
+name: Element timing performance entries
 description: An `element` performance entry measures the time to render elements that have the `elementtiming` attribute.
 spec: https://w3c.github.io/element-timing/
 group: performance
--- a/features/layout-instability.yml
+++ b/features/layout-instability.yml
@@ -1,5 +1,5 @@
 name: Layout instability
-description: The `layout-shift` performance entry and `LayoutShift` API measures the layout stability of web pages based on movements of the elements on the page. They're used to calculate Cumulative Layout Shift (CLS), a common metric for perceived stability.
+description: A `layout-shift` performance entry measures the layout stability of web pages based on movements of the elements on the page. They're used to calculate Cumulative Layout Shift (CLS), a common metric for perceived stability.
 spec: https://wicg.github.io/layout-instability/
 group: performance
 compat_features:
--- a/features/layout-instability.yml
+++ b/features/layout-instability.yml
@@ -1,5 +1,5 @@
-name: Layout instability
-description: A `layout-shift` performance entry measures the layout stability of web pages based on movements of the elements on the page. They're used to calculate Cumulative Layout Shift (CLS), a common metric for perceived stability.
+name: Layout instability performance entries
+description: A `layout-shift` performance entry measures the layout stability of web pages based on movements of the elements on the page. They're used to calculate cumulative layout shift (CLS), a common metric for perceived stability.
 spec: https://wicg.github.io/layout-instability/
 group: performance
 compat_features:
--- a/features/navigation-timing.yml
+++ b/features/navigation-timing.yml
@@ -1,5 +1,5 @@
 name: Navigation timing
-description: The `navigation` performance entry and the `PerformanceNavigationTiming` API measures navigation events, such as loading time or the number of redirects.
+description: A `navigation` performance entry measures navigation events, such as loading time or the number of redirects.
 spec: https://w3c.github.io/navigation-timing/
 group: performance
 status:
--- a/features/navigation-timing.yml
+++ b/features/navigation-timing.yml
@@ -1,4 +1,4 @@
-name: Navigation timing
+name: Navigation timing performance entries
 description: A `navigation` performance entry measures navigation events, such as loading time or the number of redirects.
 spec: https://w3c.github.io/navigation-timing/
 group: performance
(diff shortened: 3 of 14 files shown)


features/element-timing.yml
```diff
@@ -1,5 +1,5 @@
 name: Element timing
```
tunetheweb, line 1:
```suggestion
name: Element Timing performance entries
```

features/layout-instability.yml
```diff
@@ -1,5 +1,5 @@
 name: Layout instability
```
tunetheweb, line 1:
```suggestion
name: Layout Instability performance entries
```

features/navigation-timing.yml
```diff
@@ -1,5 +1,5 @@
 name: Navigation timing
```
tunetheweb, line 1:
```suggestion
name: Navigation Timing performance entries
```

features/paint-timing.yml
```diff
@@ -1,5 +1,5 @@
 name: Paint timing
```
tunetheweb, line 1:
```suggestion
name: Paint Timing performance entries
```

features/resource-timing.yml
```diff
@@ -1,5 +1,5 @@
 name: Resource timing
```
tunetheweb, line 1:
```suggestion
name: Resource Timing performance entries
```

features/soft-navigations.yml
```diff
@@ -1,5 +1,5 @@
 name: Soft navigation performance entries
```
tunetheweb, line 1:
```suggestion
name: Soft Navigation performance entries
```

features/resource-timing.yml
```diff
@@ -1,5 +1,5 @@
...
-description: "`PerformanceResourceTiming` entries report when network events happen while loading a resource, such as when connections start and end. You can use this information to measure loading times."
+description: A `PerformanceResourceTiming` performance entry records when a network event happens while loading a resource, such as when a connection starts and ends. You can use this information to measure loading times.
```
tunetheweb, line 2:
```suggestion
description: A `resource` performance entry records network event timings, such as when a connection starts and ends or loading time, for resources requested by a page.
```

features/page-visibility-state.yml
```diff
@@ -1,5 +1,5 @@
 name: Page visibility state
```
tunetheweb, line 1:
```suggestion
name: Page Visibility State Entry performance entry
```

I think this is especially important to rename to differentiate it from [`page-visibility.yml`](https://github.com/web-platform-dx/web-features/blob/main/features/page-visibility.yml) (which is related but different - one just gets the current state, the other monitors for changes).

tunetheweb (commented):
Mostly LGTM but if we're capitalising some, then I'd capitalise them all. The specs mostly treat these as proper nouns (though not in all places I've noticed!).

And similarly I'd suggest adding "peformance entries" consistently.

I've also suggested changes to Visibility State and Resource Timing to make them clearer.

ddbeck:
Thanks for the early review, @tunetheweb!

I'm still not sure about capitalization. I agree we should capitalize proper nouns, but I'm not inclined to defer to specs as a rule. They're almost German in their tendency towards capitalization and, as you've noted, not consistently edited. "Paint timing" is a good example where even the spec is seemingly unconvinced that it's a proper noun, shedding the capitalization after the early sections.

Instead, I'd suggest we follow capitalization where the named measures have a well-known-to-developers initialism or it's likely to be known that way. Or, more generally, if there's a term of art that is likely to be capitalized by developers, but I don't think we have a good example in this batch.

In this case, I think that means we capitalize:

- Interaction Contentful Paint (ICP)
- Largest Contentful Paint (LCP)

And though they're not mentioned in feature names, in the description for the `paint-timing` feature we should continue to write:

- First Paint (FP)
- First Contentful Paint (FCP)

Are there any others you want to make a case for?

(Otherwise, consistency across these features would be good. I should also pluralize consistently here, with "performance entries.")

tunetheweb:
Yeah it's a tricky one!

Some more thoughts:

I think to explain the spec choices (somewhat) there is a difference between using an API (e.g. "The page uses the Paint Timing API") and the concept ("this is used to measure paint timings - i.e. the timing of paints"). But the concept could also be worded in terms of the API ("this is used to measure the Paint Timing (as defined by the spec) of the page"). But that nuance is tricky to comprehend.

However for web features the concept is more important than the API IMHO so we probably shouldn't capitalise unless there's a well-known acronym/initialism as you say. That also fits with other examples of features like "View transitions" which is not capitalised, despite there being a "View Transition API" and a ["View Transitions" spec](https://drafts.csswg.org/css-view-transitions-2/). So plus one to that. And agree FP and FCP deserve that capitalism while others (e.g. "Navigation timing") do not.

When it comes to acronyms, there is another subtlety in that we have APIs (`LargestContentfulPaint`) and metrics ("Largest Contentful Paint (LCP)") and I agree the latter is what people know it as. The difference between the API and the metric is often razor thin (for FP and FCP they are pretty much the same as the API emits the metric, for LCP it's different as the LCP metric is calculated from a number of `LargestContentfulPaint` entries). Anyway, TLDR: when there is a well-known metric initialism that the API is primarily used for, I agree that should be the name and capitalism therefore makes sense.

But this introduces another couple of oddities:
- Interaction to Next Paint (INP) is the metric measured by the Event Timing API. But the feature is currently called "Event timing" and not "Interaction to Next Paint (INP)". Though that is included in the description. I think this is fine as is, and it's been out for a while, but is a slight deviation from what I said above.
- "Interaction Contentful Paint" is not a metric, and so not that well-known by it's initialism. In fact I've been careful to avoid calling it ICP in case people presume it's a metric. Plus it's easy to mistake for LCP when you see it written depending on the font and how easy it is to distinguish I from L. So I'd say this probably should stay "Interaction contentful paint" feature, rather than "Interaction Contentful Paint", but that might be jarring since it's inconsistent with "Largest Contentful Paint" and "First Contentful Paint" (which are both APIs AND metrics). Maybe we should just avoid capitilization even when there is a well-known initialism?

So, at the end of that meandering thought, I think I'm back to NOT capitalising them after all! Even if there's a well-known initialism.

ddbeck:
@tunetheweb thank you for your thoughtful input on this! I've done another pass on the performance feature entries and I'd welcome your take on the latest revisions. Cheers!

tunetheweb (approved):
LGTM!

captainbrosset (approved):

PR #4264, opened 2026-08-24
Add stats to release artifacts

ddbeck:
Towards https://github.com/web-platform-dx/web-features/issues/3510.

This PR adds `stats.json` as release artifacts (for both `next`-tagged releases and the regular stable releases).

PR #4281, opened 2026-08-27
Fix artifacts upload on production releases

ddbeck:
I broke the upload with https://github.com/web-platform-dx/web-features/pull/4264. 😬


Elchi3 (approved):
Great! 😄

PR #4290, opened 2026-09-01
`stats.ts`: load data lazily

ddbeck:
This makes it so running `stats.ts` just for `--help` doesn't take many seconds. There's more refactoring to make `index.ts` play nicer in this regard, but that's a bigger refactor, so I'm doing this now as a stopgap.

~~Depends on https://github.com/web-platform-dx/web-features/pull/4289.~~

See also: the sequel https://github.com/web-platform-dx/web-features/pull/4291.

### DIFF
--- a/scripts/stats.ts
+++ b/scripts/stats.ts
@@ -3,10 +3,7 @@ import assert from "node:assert/strict";
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import yargs from "yargs";
-import { features, groups } from "../index.ts";
 import { isOrdinaryFeatureData } from "../type-guards.ts";
-import { caniuseToWebFeaturesId } from "./caniuse.ts";
-import { compatFeaturesToCumulativeDaysShipped } from "./unmapped-compat-keys.ts";
 
 interface ResultBase {
   featuresCount: number; // `kind: feature` ID count (ignores other `kind` values)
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Can confirm this speeds up `--help`.

Without this PR:
`npm run stats -- --help  9.87s user 0.56s system 111% cpu 9.369 total`
With this PR:
`npm run stats -- --help  0.87s user 0.18s system 108% cpu 0.960 total`


`rpm run stats` (without `--help`) is slow on this PR still.

ddbeck:
Yes, that's right. Briefly, the main sources of slowness are:

- `compute-baseline` file loading (we load and parse the several megabytes of BCD, every time `new Compat()` is called)
- `compute-baseline` status computations (there's no memoization on expensive multi-key calculations and no caching between runs)
- `index.ts` file loading (we have to read 2k+ YAML files, every time, no matter if there are any changes)
- `index.ts` status computations (every time we import `index.ts`, we recalculate everything even if the underlying data hasn't changed, which means that our various scripts repeat this effort many times)
- Status computations in `unmapped-compat-keys.ts`, which `stats.ts` imports (this duplicates a lot of what `index.ts` does)

This PR mostly _defers_ all the slowness, unless it's actually needed. Much more performance work to be done. But later, after I have a stats pipeline. 😄

PR #4291, opened 2026-09-01
`stats.ts`: fetch previous release stats for comparisons

ddbeck:
This makes it pretty quick and relatively easy to run stats against a previous release (that has stats precomputed) or the `next` release (i.e., versus the `main` branch).

You can try it by running one of these:

```shell
npm run stats -- --previous-release=latest
npm run stats -- --previous-release=v3.36.0
npm run stats -- --previous-release=next
```

Right now these all produce the same results—a bunch of zeros—except for `next` which shows the change caused by fixing the miscalculation in https://github.com/web-platform-dx/web-features/pull/4289.

~~Depends on https://github.com/web-platform-dx/web-features/pull/4289 and https://github.com/web-platform-dx/web-features/pull/4290.~~

### DIFF
--- a/scripts/stats.ts
+++ b/scripts/stats.ts
@@ -5,6 +5,10 @@ import { fileURLToPath } from "node:url";
 import yargs from "yargs";
 import { isOrdinaryFeatureData } from "../type-guards.ts";
 
+// A release tag after "v3.5.1" or "next"
+type SupportedReleaseTag =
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
LGTM. I tested the error paths and I do see the change from the miscalculation. 👍 
Will be more interesting when there are actual changes but I trust that it is going to work then :)

PR #4293, opened 2026-09-02
`stats.ts`: fix error-able versions matching

ddbeck:
The previous iteration of this code matched versions like `2.34` but _not_ `2.36`. It should reject _all_ versions before v3.35.

Prequel to https://github.com/web-platform-dx/web-features/pull/4294.


Elchi3 (approved):
Good catch! 👍

PR #4294, opened 2026-09-02
Statistics: generate a markdown report from `stats.ts`

ddbeck:
This generates a Markdown report from `stats.ts`, for a comment suitable for a successor to #788.

You can see a rendered example in https://github.com/ddbeck/web-features/issues/37#issuecomment-5512270198.

You can try this yourself by running a command line `npm run --silent stats -- --previous-release=latest | npx tsx ./scripts/stats-report.ts -`. That said, I don't really expect any humans to run this; it'll be used by GitHub Actions workflows, coming soon in follow up PRs.

Depends on ~~https://github.com/web-platform-dx/web-features/pull/4293 and~~ https://github.com/web-platform-dx/web-features/pull/4346.

Towards #3510.


Elchi3 (commented):
This seems to work. Two questions:

- Is "Net" supposed to be a number like "+42"? We don't want percentages anymore?
- The "Before" and "After" headings have no date anymore, leaving the reader wondering: before and after _what_?

ddbeck:
> * The "Before" and "After" headings have no date anymore, leaving the reader wondering: before and after _what_?

Heh, it turns out fixing this is easier said than done. I've opened https://github.com/web-platform-dx/web-features/pull/4346 underneath this change, to generate more data and modify/break the types involved.

ddbeck:
@Elchi3 I've had another go at the tables. You can see an example in https://github.com/ddbeck/web-features/issues/37#issuecomment-5683484905, which shows an actual change (albeit a small one). Here's what's new:

- The first table shows the dates in the header.
- At the end of the comment, it shows the start and end date, duration (number of days), and links to the compare view between the hash at the start and the hash at the end of the period under consideration.
- The tables now shows the relative change (i.e., how much did the values change versus the previous run). Since the WebDX roadmap is focused on the coverage _gap_, I chose not to include total coverage percentages.
- I updated some text to improve the layout wrapping of the tables and emphasize the coverage gap angle that's the focus here.

Let me know what you think!

ddbeck:
I pushed two more changes:

- I noticed a typo, now fixed.
- The stats script is modified to use the committer date, not the author date. My demo comment's dates are a little wonky (the before date is after the after date); this is because I used the _authoring_ date of the commit, not the _committer_ date. This shouldn't really make a difference in practice, since we use squash merges on this repo, but it does break the demo. And making this change now means that, if we ever alter our merge policy the stats script will continue to work.

scripts/stats-report.ts
```diff
@@ -0,0 +1,303 @@
...
+
+function report(stats: Result): string {
+  const startDate = Temporal.Instant.from(
+    stats.change.timestamp,
+  ).toZonedDateTimeISO("Etc/UTC");
+  const endDate = Temporal.Instant.from(stats.timestamp).toZonedDateTimeISO(
+    "Etc/UTC",
+  );
+  const duration = endDate.since(startDate);
+  const revisions = `[\`${stats.change.hash.slice(0, 8)}..${stats.hash.slice(0, 8)}\`](https://github.com/web-platform-dx/web-features/compare/${stats.change.hash}..${stats.hash})`;
```
Elchi3, line 51: Maybe I'm holding it wrong, but I'm getting undefined errors for `stats.change.timestamp` and `stats.change.hash`. Maybe you want them from the `stats` object?

```suggestion
    stats.timestamp,
  ).toZonedDateTimeISO("Etc/UTC");
  const endDate = Temporal.Instant.from(stats.timestamp).toZonedDateTimeISO(
    "Etc/UTC",
  );
  const duration = endDate.since(startDate);
  const revisions = `[\`${stats.hash.slice(0, 8)}..${stats.hash.slice(0, 8)}\`](https://github.com/web-platform-dx/web-features/compare/${stats.hash}..${stats.hash})`;
```

ddbeck: Ah, yes, did you run `npm run --silent stats -- --previous-release=latest | npx tsx ./scripts/stats-report.ts -`? This fails for me too. It fails because the latest release asset still doesn't have the new statistics with the hash and timestamp.

Running `npm run --silent stats -- --previous-release=next | npx tsx ./scripts/stats-report.ts -` does work (note the **`--previous-release=next`**). The `latest` should start working on Thursday (or whenever the next BCD release happens), when we cut a new web-features release.

(I can also backfill stats on previous releases, if you like. But it's a bit of a manual process.)

Elchi3: ah got it, thank you! :) 

Backfilling is not needed, imo.

scripts/stats-report.ts
```diff
@@ -0,0 +1,303 @@
...
+    "",
+    "This shows feature entry coverage for fine-grained compatibility data. For unmapped keys, fewer is better.",
+    "",
+    reportCompatCoverage(stats, startDate, endDate),
+    "",
+    "### Cumulative shipping days",
+    "",
+    "This is is a time- and browser-weighted measure of unmapped BCD keys. The numbers grow for each day since a key first shipped, per browser. Smaller is better.",
+    "",
+    reportCumulativeShippingDays(stats),
```
Elchi3, line 64: Nit: do you want to startDate and endDate here as well? (as well is in reportCaniuseCoverage)?

Elchi3 (approved):
Thanks Daniel, this seems to work! :) Thank you for going the extra miles for the dates.

scripts/stats-report.ts
ddbeck: Personally, I thought it added a lot of visual noise to repeat the dates in the headers of each and every table. It's mostly a style preference, so I'm happy to add them in if they're wanted. But my inclination here is to merge this and start getting feedback from the WebDX CG sooner rather than fuss over the table layout further from here.

PR #4318, opened 2026-09-04
Add HTML setters features

### LINKED ISSUE #4117: HTML enhanced setter and streaming methods
### Specification

https://github.com/WICG/declarative-partial-updates/blob/main/dynamic-markup-revamped-explainer.md#resulting-api

Various discussions in HTML spec:
https://github.com/whatwg/html/issues/11669
https://github.com/whatwg/html/issues/2142
https://github.com/whatwg/html/pull/12495
https://github.com/whatwg/html/pull/12753
https://github.com/whatwg/html/pull/12756
https://github.com/whatwg/html/pull/12758

### Description

Chrome is launching a new set of static and streaming HTML APIs as described here:
https://developer.chrome.com/blog/declarative-partial-updates#a_new_set_of_static_and_streaming_apis

This includes enhancements to `setHTML` (covered in [sanitizer.yml](https://github.com/web-platform-dx/web-features/blob/main/features/sanitizer.yml)), `setHTMLUnsafe` (covered in [parse-html-unsafe.yml](https://github.com/web-platform-dx/web-features/blob/main/features/parse-html-unsafe.yml.dist)), and various new positional methods (replace, before, prepend, append, after) as well as streaming versions of these.

As these are likely to be implemented together if they roll out to other browsers, but the static and streaming versions are somewhat separate and could even be implemented at different times. So we in the Chrome team are suggesting two new features for this:

- HTML setter methods
- HTML stream methods

For now, since there is existing browser support we'll keep `setHTML` and `setHTMLUnsafe` separate, but longer term I could see those merging into "HTML setter methods" if all browsers converge on these (as we hope they will).

### Documentation

https://developer.chrome.com/blog/declarative-partial-updates#a_new_set_of_static_and_streaming_apis
https://github.com/WICG/declarative-partial-updates/blob/main/dynamic-markup-revamped-explainer.md#resulting-api
https://github.com/WICG/declarative-partial-updates/tree/main/examples

### Browser support

Available behind a flag from Chrome 148. Shipping in Chrome 154:
https://chromestatus.com/feature/5054329641893888

Firefox: Positive - https://github.com/mozilla/standards-positions/issues/1370
WebKit: No signal yet - https://github.com/WebKit/standards-positions/issues/629

tunetheweb:
Fixes #4117

Due for launch in Chrome 154

Spec updates still WIP but have added commented out lines for them in the meantime.


features/html-setters.yml
```diff
@@ -0,0 +1,44 @@
+name: HTML positional setters
+description: The setHTML(), replaceWithHTML(), beforeHTML(), prependHTML(), appendHTML() and afterHTML() methods, and unsafe equivalents, insert HTML into an existing tree.
+spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
```
tunetheweb, line 3: This is actually waiting on https://github.com/whatwg/html/pull/12753 before it's added to this section of the HTML spec. But that's pretty close now and is just waiting on some final reviews.

features/html-streaming-setters.yml
```diff
@@ -0,0 +1,43 @@
+name: HTML streaming setters
+description: The streamHTML(), streamReplaceWithHTML(), streamBeforeHTML(), streamPrependHTML(), streamAppendHTML() and streamAfterHTML() methods, and unsafe equivalents, return writeable streams that incrementally insert HTML into an existing tree.
+spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
```
tunetheweb, line 3: Same for this one that's waiting on https://github.com/whatwg/html/pull/12758 also pretty close to landing

features/html-setters.yml
```diff
@@ -0,0 +1,44 @@
+name: HTML positional setters
+description: The setHTML(), replaceWithHTML(), beforeHTML(), prependHTML(), appendHTML() and afterHTML() methods, and unsafe equivalents, insert HTML into an existing tree.
```
jamesnw:
```suggestion
description: The `setHTML()`, `replaceWithHTML()`, `beforeHTML()`, `prependHTML()`, `appendHTML()` and `afterHTML()` methods, and unsafe equivalents, insert HTML into an existing tree.
```

jamesnw: Do we need to say what these are methods on? 

Also, there seems to be some overlap with `sanitizer` and `parse-html-unsafe` - some of the methods mentioned here are in the other features. I'm not quite sure what the solution is there.

jamesnw (dismissed):

features/html-setters.yml
tunetheweb: There was a bit of discussion on this in the issue (#4117) with @ddbeck .

Only the `setHTMLUnsafe` and `setHTMLUnsafe` overlap, which is why I've only added `api.Element.setHTMLUnsafe.options_runscripts_parameter` here. Think of this feature as expanding on that initial implementation, both in those methods and in creating additional, adjacent methods.

Personally I disagree with `setHTML` being part of Sanitzer. It was the first use of Sanitizing so added as part of that, but it's more a user of it, than part of the API IMHO (as are these).

Ultimately I'd like to see both those methods brought in here, but while this is early, and Chrome-only, and those are supported by other browsers that's likely to cause more confusion. So, while I agree this is not ideal, I think it's the best option for now.

tunetheweb: > Do we need to say what these are methods on?

Sorry forgot to answer this bit. That was @ddbeck 's original suggestion in https://github.com/web-platform-dx/web-features/issues/4117#issuecomment-5344842976.

Personally I think this is a bit verbose:

> The `setHTML()`, `replaceWithHTML()`, `beforeHTML()`, `prependHTML()`, `appendHTML()` and `afterHTML()` methods of the CharacterData, Element, DocumentType, and ShadowRoot (and their unsafe equivalents), insert HTML into an existing tree.

It's also not accurate since only half of them apply to all and the other half to just `Element` and `ShadowRoot`.

I think it's fairly obvious these are DOM methods and apply to the appropriate places and MDN, the spec or other documentation sites are the best place to get this level of detail.

tunetheweb: We could say "DOM methods" if preferred?

> The setHTML(), replaceWithHTML(), beforeHTML(), prependHTML(), appendHTML() and afterHTML() DOM methods, and their unsafe equivalents, insert HTML into an existing tree.

tunetheweb: We also also take a slightly more aggressive stance here and:

- Move `setHTML` out of the [sanitizer.yml](https://github.com/web-platform-dx/web-features/blob/main/features/sanitizer.yml) feature and into here (leaving the rest of the Sanitzer API in there). As I say it's not quite right being in there IMHO. Firefox would still show as supporting the Sanitizer API, but no longer show support of `setHTML` since it would now be part of this wider feature that they haven't fully implemented.
- Move `setHTMLUnsafe` out of [`parse-html-unsafe`](https://github.com/web-platform-dx/web-features/blob/main/features/parse-html-unsafe.yml.dist) into here (leaving `api.Document.parseHTMLUnsafe_static` in there). Firefox and Safari would still show as supporting `parseHTMLUnsafe` but no longer show support of `setHTMLUnsafe` since it would now be part of this wider feature that they haven't fully implemented.

That would be less confusing and ultimately where we want to go longer term.

Then longer, longer term this feature gets folded into [`dom`](https://github.com/web-platform-dx/web-features/blob/main/features/dom.yml.dist) anyway.

WDYT?

jamesnw: I think the feature composition as you have it makes sense- sorry for rehashing existing conversations!



> We could say "DOM methods" if preferred?
> 
> > The setHTML(), replaceWithHTML(), beforeHTML(), prependHTML(), appendHTML() and afterHTML() DOM methods, and their unsafe equivalents, insert HTML into an existing tree.

This sounds good to me. And you're right that this isn't the place for the amount of specificity needed to full describe the API.

tunetheweb: Updated description to "DOM methods".

I'm actually more and more thinking we should just bite the bullet and move `setHTML` and `setHTMLUnsafe`. Especially as we are merging these into the HTML spec, so this is not just some Chrome proposal. I've added that to this PR in https://github.com/web-platform-dx/web-features/pull/4318/commits/de89f56598c3656d2dbcb464d74ce786388e5a14 so you can see what that looks like. Let's also see what @ddbeck thinks here.

features/html-setters.yml
```diff
@@ -0,0 +1,48 @@
+name: HTML positional setters
```
ddbeck: Previously, in https://github.com/web-platform-dx/web-features/issues/4117, the proposed name was "HTML setter methods" but this has since changed.

I'm wondering how wedded you are to including "positional." First, I think it might be slightly misleading, for `setHTML()`. And it sounds a little… off to me; "Positional HTML setters" or "Positional HTML setter methods" sounds more 'right' to my ear, but it emphasizes "positional" which is maybe not the goal.

In a vacuum, I'd say let's go with "HTML setter methods" or "HTML setters" but I'm not the expert on what this thing is or how you've been talking to developers about it.

features/html-setters.yml
```diff
@@ -0,0 +1,48 @@
...
+description: The `setHTML()`, `replaceWithHTML()`, `beforeHTML()`, `prependHTML()`, `appendHTML()` and `afterHTML()` DOM methods, and unsafe equivalents, insert HTML into an existing tree.
```
ddbeck, line 2: "DOM methods" is a good workaround to the other problems. Nice work!

features/parse-html-unsafe.yml
```diff
@@ -1,8 +1,5 @@
 name: Unsanitized HTML parsing methods
```
ddbeck: Let's finesse the name here, so we're not lying about what's inside. 😄

```suggestion
name: Unsanitized HTML parsing
```

features/parse-html-unsafe.yml
```diff
@@ -1,8 +1,5 @@
...
-description: The `Document.parseHTMLUnsafe()` static method parses HTML into a DOM tree, while the `setHTMLUnsafe()` method of `Element` and `ShadowRoot` parses and inserts HTML into an existing tree. No sanitization applies to these methods, so never call them with user-provided HTML strings.
-spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#unsafe-html-parsing-methods
-# TODO: https://github.com/web-platform-dx/web-features/issues/1971
-# Status changed: https://github.com/web-platform-dx/web-features/pull/3200
-# 2025-08-06 - low → false - In Safari, if there are custom elements in a declarative shadow root in the parsed HTML, then the elements' constructors and `connectedCallback()` methods are not called.
-# References:
-# - https://bugs.webkit.org/show_bug.cgi?id=296420
+description: The `Document.parseHTMLUnsafe()` static method parses HTML into a DOM tree, in a similar way that `setHTMLUnsafe` and similar DOM method inserts HTML into the DOM. No sanitization applies to these methods, so never call them with user-provided HTML strings.
```
ddbeck: Let's finesse this a little to make it less wordy and preserve the spirit of the original. 😄

```suggestion
description: The `Document.parseHTMLUnsafe()` static method parses HTML into a DOM tree, much like the `setHTMLUnsafe()` DOM method inserts HTML into an existing tree. No sanitization applies, so never call it with user-provided strings.
```

features/parse-html-unsafe.yml
```diff
@@ -1,8 +1,5 @@
...
+spec: https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
+compat_features:
+  - api.Document.parseHTMLUnsafe_static
```
ddbeck, line 5: For status consistency, we should restore the original set of compat keys to the `core` or `modifier` set later, when we've merged https://github.com/web-platform-dx/web-features/pull/4034.


```suggestion
  - api.Document.parseHTMLUnsafe_static
  # TODO: overlap `setHTMLUnsafe()` with `html-setters` and this feature, to guarantee status consistency
  # See: https://github.com/web-platform-dx/web-features/issues/1173
```

features/sanitizer.yml
```diff
@@ -1,14 +1,8 @@
 name: Sanitizer API
-description: The `Document.parseHTML()` static method and the `setHTML()` method of `Element` and `ShadowRoot` objects parse and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. The `Sanitizer` API can customize the sanitization process.
+description: The `Sanitizer` API can customize the sanitization process used to parse and insert HTML into the DOM in a way that can prevent cross-site scripting attacks. `Document.parseHTML()` and other DOM methods like `setHTML()` can use the Sanitizer API on the provided HTML.
```
ddbeck: Let's address the reader more directly here:

```suggestion
description: The `Sanitizer` API customizes the sanitization of HTML parsed and inserted into the DOM in a way that can prevent cross-site scripting attacks. You can use it with DOM APIs that accept a sanitizer option, such as `Document.parseHTML()` or HTML setters such as `setHTML()`.
```

features/sanitizer.yml
ddbeck, line 22:
```suggestion
  - api.Sanitizer.setDataAttributes
  # TODO: overlap `setHTML()` with `html-setters` and this feature, to guarantee status consistency
  # See: https://github.com/web-platform-dx/web-features/issues/1173
```

features/html-setters.yml
ddbeck: Do you still expect this to land quickly? The formally correct thing to do is link to the PR and put an exception in `specs.ts` to allow it. But it's maybe not worth the trouble, if the exception is going to last just a couple days.

ddbeck (changes requested):
Hi @tunetheweb, thanks for your patience while I came back to this!

I have some mostly nitpicky suggestions and questions in line comments (though with one serious question about the name of the main setters feature). That said, I'm happy with the direction of this and I don't expect you to labor on it much longer. 😄

features/html-setters.yml
tunetheweb: Oh I think this was just an old name I forgot to update. "HTML setter methods" LGTM

```suggestion
name: HTML setter methods
```

features/parse-html-unsafe.yml
tunetheweb: Did you see the above comment about whether we should just bite the bullet and move these now?

https://github.com/web-platform-dx/web-features/pull/4318#discussion_r3935892872

tunetheweb: Oh wait. This is about coping it BACK when we can, so seems you're in agreement about moving it to the new features? Sorry misunderstood there.

I'm not sure there's a need to move this back. It seems weird to me that the `parse-html-unsafe` feature covered `setHTMLUnsafe()` in the first place.

features/html-setters.yml
tunetheweb: It's so very, very close 😞 But Noam's on holiday for a couple of weeks. So suspect it'll be weeks rather than days.

Will add a `specs.ts` override for now.

features/sanitizer.yml
tunetheweb: Same comment above about whether we should move this back, even when we can. `setHTML` is no more part of the Sanitzer HTML IMHO than the new setters are. They are all just methods that kinda use the Sanitizer. And I think the Sanitizer keys we've left in this feature cover it enough.

features/html-setters.yml
tunetheweb: Added the overrides for now.

features/html-setters.yml
tunetheweb: Ha when I went to edit my local copy, I had an unsaved change of this name. So just forgot to save and commit and push! Anyway, all sorted now.

features/parse-html-unsafe.yml
ddbeck: Yes, this is about having the keys in _both_ features later. For this one I'm less fussed about it. In the description, we mention the `setHTMLUnsafe()` method, so it seems slightly weird for it to be possible for this feature to have a different support condition than that of `setHTMLUnsafe()`. But that also seems like a very unlikely turn of events (e.g., some new showstopper bug appears in `setHTMLUnsafe()` that would shift the Baseline date forward in time but one that doesn't _also_ affect `parseHTMLUnsafe()`), so we don't have to do it. I'm going to go ahead and mark this as resolved.

features/sanitizer.yml
ddbeck: This one I feel more nervous about. No developer is going to use `Sanitizer` in isolation because it's configuration for something else. For it to be a meaningful feature, a developer needs to be able to do something useful with it.

When we minted this feature we figured `setHTML()` was the most likely place a developer would use a `Sanitizer` instance. As a I recall, it was the ur-application for sanitization. For example, [this sanitizer demo](https://sanitizer-api.dev/) describes `setHTML()` as "[t]he core of the API."

That said, this feature includes `Document.parseHTML()` and it **is** a thing a developer can use with `Sanitizer`, though I suspect a less-commonly used one.

This is a judgment call and I'm putting it in your hands. If you think `Document.parseHTML()` is a good enough, then we can drop this comment. Otherwise, let's keep it.

ddbeck (commented):
OK, one last question for you regarding `sanitizer`, but consider this a provisional approval. Thank you again for your work on this, @tunetheweb!

features/sanitizer.yml
tunetheweb: It's a tricky one and I'm not 100% certain my suggestion is the right one. But let me explain so more so you at least understand where I'm coming from:

I agree that Sanitizer is a new primitive and something needs to use it for it to be useful. And `setHTML` (and to a lesser point `parseHTML`) was the first (and basically only!) way of using this until now.

However now we have all these, that use the Sanitizer API:

- `afterHTML`
- `appendHTML`
- `beforeHTML`
- `prependHTML`
- `replaceHTML`
- `replaceWithHTML`
- `setHTML`

So why does `setHTML` deserve special treatment to be included as part of the Sanitizer API but the rest not? Because it launched at the same time? That's not a bad argument to be fair, but I'm not totally convinced...

I guess I'd put it another way: if a browser launched the Sanitizer API with `afterHTML` only but not `setHTML` (I don't see why they'd do that btw, but straw manning here), then would we say the Sanitizer API is supported? Only partially supported? Does it matter since it's very unlikely they'd do that? But maybe it's more likely in the future as more users of this API come on board...

I'm also not 100% convinced that `parseHTML` belongs under the Sanitizer API either but, unlike `setHTML` I've no better place to put that for now.

Now I don't think we should put ALL the above methods in the Sanitizer API either. Firefox supports the Sanitizer API (including `setHTML`) but doesn't support the new additional methods ([yet](https://bsky.app/profile/tunetheweb.com/post/3mv3lfpd7a22s))! So I think it's fair to say they support the Sanitizer API, but not `html-setters`.

Anyway, I'm not totally against adding this comment, and adding `setHTML` to this feature too when web features supports it. But think I've a slight preference to not doing that for the reasons above. But happy to go with your opinion here if you really feel it should.

tunetheweb: Pinged @jakearchibald to see if he has opinions here so it's not just "Chrome" deciding here.

tunetheweb: OK I had a good chat with @jakearchibald about this off line and he'd prefer to keep `setHTML` in `sanitizer` for now as a bare minimum method that shows the API is supported, with a duplication to `html-setters` once web features supports that. I can live with that so I reverted to that.

ddbeck: I'm happy with this and thank you for bringing it to Jake's attention.

FWIW, it's not that I'm supposing `setHTML()` is part of the Sanitizer API _per se_, just that it's a good measure of support in the world that actually exists. In your straw man scenario, I suppose I'd have to contemplate setting a status manually or implementing some new status calculation logic (as in 'all of these keys, plus one or more of these other keys').

Anyway, all of this will be more coherent when we can overlap the keys. I'll try to push that ahead sooner rather than later. Thanks again!

ddbeck (approved):

ddbeck:
At least five people contributed to understanding this and https://github.com/web-platform-dx/web-features/issues/4117. Thank you, @tunetheweb, @jamesnw, @jakearchibald, and @noamr!

Issue #84, opened 2023-03-07
Validate any valid spec URL


ddbeck:
via https://github.com/web-platform-dx/feature-set/pull/78

## What I want

- [ ] To be able to link to any valid page (or fragment) within a specification known to `web-specs`
- [ ] To not be able to link an invalid page (or fragment) within a specification known to `web-specs`
- [ ] To not be able to link to a specification unknown to `web-specs`

## How it works now

Right now, the tests allow linking only to the single URL known to `web-specs`. For example `https://html.spec.whatwg.org/multipage/` is a valid `spec` URL, but `https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements` is not.

## How it could work in the future

We could change the tests to allow any valid derivative URL (i.e., any URL where the scheme and domain are the same and the path starts with the path known to `web-specs`).

If we wanted to go further, we could periodically scrape spec URLs to fully enumerate all of the pages and anchors and validate against them (I don't think this would be too bad—there's not _that_ many specs, but not exactly zero effort).

foolip:
@ddbeck do you think we need to do more than we're already doing for this?

ddbeck:
@foolip We don't _need_ to do anything, but we should probably log the idea somewhere that we wanted to validate the URL fragments (see also: https://github.com/w3c/browser-specs/issues/931). More nice-to-have than need-to-have (at least for now).

ddbeck:
> Perhaps `browser-specs` and `web-specs` itself should have a validation function?

Maybe? I'm not sure how that interacts with `web-ref`, which knows about the fragments themselves.

But it seems to me that a reverse lookup would make the most sense, given the goals browser- and web-specs: you give it a complete URL and you get back the corresponding `spec` object. We'd just check for success or failure.

tidoust:
> Perhaps browser-specs and web-specs itself should have a validation function?

Not sure if that's meant to cover fragments. If so, I don't think we'll include them in browser-specs directly. We can look into providing the information in a more directly usable way in Webref though. Webref effectively contains browser-specs, so it should be possible to get all the validation info you need relying on Webref rather than on browser-specs.

foolip:
Ah, right, it was fixed for pages, not fragment identifiers...

I've filed https://github.com/w3c/webref/issues/1198 for that.

schalkneethling:
[...]

PR #4320, opened 2026-09-04
Validate spec URL fragments with @webref/xref

schalkneethling:
Progress on #84.

Checks the fragment of every spec URL against the anchors [w3c/webref](https://github.com/w3c/webref) crawls, using the same rules as BCD's `spec_url` linter (mdn/browser-compat-data#23958): skip fragment-less URLs, strip `:~:text=` fragments, `lookup(url, { series: true, standing: "good" })`.

- Unknown fragments **warn without failing the build**; flip `fragmentErrorsAreFatal` in `scripts/specs.ts` to enforce once the data is clean.
- `npm run test:specs -- --report` writes a self-contained HTML report of every fragment check to `reports/spec-fragments.html`.

Currently 32 of 1,356 spec URLs warn. Per [@tidoust's suggestion](https://github.com/web-platform-dx/web-features/issues/84#issuecomment-5537331291), I'll relink those to current ids in a follow-up PR rather than asking webref to carry legacy alias ids.


scripts/spec-fragments.ts
```diff
@@ -0,0 +1,60 @@
...
+    initialized = true;
+  }
+
+  let target = new URL(url);
+
+  if (!target.hash) {
+    return { status: "skipped", reason: "no-fragment" };
+  }
+
+  // A text fragment (`#:~:text=…`) can't be validated, but a section id
```
tidoust, line 44: I think Web-features does not use any text fragment for now, and it seems easier to avoid them altogether in web-features than it is in BCD (where BCD keys may be very fine-grained). Given that these fragments can easily break with spec updates, I would suggest to forbid them in web-features.

schalkneethling: Thank you for the feedback. I will update the code and the PR.

scripts/specs.ts
```diff
@@ -291,6 +307,95 @@ function suggestSpecs(bad: URL): void {
...
+    const options = sortedCategories
+        .map(([category]) => `<option>${escape(category)}</option>`)
+        .join("\n");
+    const tableRows = rows
+        .map((row) => {
+            const category = fragmentCategory(row);
+            const specs = row.result.status === "valid"
+                ? [...new Set(row.result.matches.map((m) => m.entry.spec))].join(", ")
+                : "";
+            return `<tr data-category="${escape(category)}"><td>${escape(row.feature)}</td><td><a href="${escape(row.url)}">${escape(row.url)}</a></td><td>${escape(category)}</td><td>${escape(specs)}</td></tr>`;
```
tidoust: Pretty minor but, while playing with the report, I found it useful to also link to the explorer here to be one click away from further links such as the source/dist files:

```suggestion
            return `<tr data-category="${escape(category)}"><td><a href="https://web-platform-dx.github.io/web-features-explorer/features/${escape(row.feature)}/">${escape(row.feature)}</a></td><td><a href="${escape(row.url)}">${escape(row.url)}</a></td><td>${escape(category)}</td><td>${escape(specs)}</td></tr>`;
```

schalkneethling:
Thank you for the review, @tidoust - I finally got around to addressing the remaining item. Apologies for the delay.

PR #4321, opened 2026-09-04
Relink spec URL fragments that webref no longer knows

schalkneethling:
Progress on #84. Companion to #4320 — updates the 28 spec URLs whose fragments are unknown to [w3c/webref](https://github.com/w3c/webref), relinking to current ids rather than asking webref to carry legacy aliases, [as suggested](https://github.com/web-platform-dx/web-features/issues/84#issuecomment-5537331291) by @tidoust.

Every replacement was verified against `@webref/xref` `lookup()` with BCD's options. Buckets:

- **Renamed/moved anchors** — e.g. `css-anchor-position-1/#anchoring` → `#anchor-positioning`, `text-fit` moved from css-text-4 to css-text-5, css-contain-3 content moved back to css-contain-2.
- **ecma262 legacy `@@`-style ids** → current `%Symbol.…%` ids (e.g. `#sec-map.prototype-@@iterator` → `#sec-map.prototype-%25symbol.iterator%25`).
- **Fragments dropped** where no meaningful anchor exists: `javascript` and `intl` intro links, GIF's `spec-gif89a.txt#top` (a plain-text file), and `typed-arrays`' `#table-49` (removed; the `#sec-typedarray-constructors` link on the next line already covers it).
- The lone `#ref-for-…` id (`wasm-bigint`) → the `"i64"` ValueType dfn.

4 warnings intentionally remain: `template-for` (the dfn exists in the live HTML spec; webref's weekly crawl hasn't caught up), `soft-navigations`/`interaction-contentful-paint` (spec standing is `pending`, so `standing: "good"` excludes it regardless of fragment), and `html-wrapper-methods` (webref doesn't extract headings for ecma262 annex pages).


tidoust:
> (webref doesn't extract headings for ecma262 annex pages).

And that's a bug ;)
I filed https://github.com/w3c/reffy/issues/2154

PR #4322, opened 2026-09-04
Add link parameters feature

captainbrosset:
Linked Parameters allow developers to pass CSS values to external SVG resources. This can be done in 3 ways:

```css
img {
    link-parameters: param(--color, blue);
}
```

```html
<img src="example.svg#:~:param(--color,blue)">
```

```css
.icon {
  background-image: url("example.svg", param(--color, blue));
}
```

ChromeStatus: https://chromestatus.com/feature/5095153430822912
Spec: https://drafts.csswg.org/css-link-params/
WPT: https://wpt.fyi/results/svg/styling/css-linked-parameters?label=experimental&label=master&aligned


features/linked-parameters.yml
```diff
@@ -0,0 +1,9 @@
+ame: Linked parameters
+description: The `link-parameters` CSS property, `param()` CSS function, and the `:~:param()` URL directive set CSS custom environment variables on an external resource, such as an external SVG image, allowing the external resource to adapt its presentation based on the values provided.
```
captainbrosset, line 2: @ddbeck for an early review of the description before I go further.

jamesnw:
Link or linked? I think it's link parameters on a linked resource. The spec name doesn't match the wording in the spec.

features/link-parameters.yml
```diff
@@ -0,0 +1,9 @@
+ame: Linked parameters
+description: The `link-parameters` CSS property, `param()` CSS function, and the `:~:param()` URL directive set CSS custom environment variables on an external resource, such as an external SVG image, allowing the external resource to adapt its presentation based on the values provided.
```
ddbeck: - In [`scroll-to-text-fragment`](https://web-platform-dx.github.io/web-features-explorer/features/scroll-to-text-fragment/ "Scroll to text fragment: Text fragments are URL fragments on the form #:~:text=snippet and link to a snippet of text within a page. The browser may scroll, highlight, or otherwise bring that text to the reader's attention.") ([source](https://github.com/web-platform-dx/web-features/blob/main/features/scroll-to-text-fragment.yml), [dist](https://github.com/web-platform-dx/web-features/blob/main/features/scroll-to-text-fragment.yml.dist)), we call the `#:~:` notation _fragments_. Should we do the same here? Or adopt the specs' "fragment directives" language in both? In the suggestion below, I've followed the other feature, but I won't insist on this, if there's a good reason to go the other way.
- I think we should show the `#` in the literal text here, since that's a probably familiar hook for reader understanding.
- `param()` is a URL "modifier" for `url()` (not an independent value-producing CSS function), so maybe we can hint at this here?


```suggestion
description: The `link-parameters` CSS property, `param()` modifier for the `url()` CSS function, and the `#:~:param(…)` URL fragment set custom CSS environment variables on an external resource, such as an SVG image, allowing the external resource to adapt its presentation based on the variables.
```

features/link-parameters.yml
```diff
@@ -0,0 +1,9 @@
+ame: Link parameters
```
ddbeck:
```suggestion
name: Link parameters
```

ddbeck (approved):
Thank you!

PR #4323, opened 2026-09-04
Add `toJSON` methods to `soft-navigations` and `interaction-contentful-paint` features

tunetheweb:
The `toJSON` methods were added late to MDN so weren't originally included here but should be.


ddbeck (approved):
Thank you!

Issue #4329, opened 2026-09-08
Prepare `attr()` for Baseline status change


captainbrosset:
This single BCD key is going to hold the `attr()` feature back: https://github.com/web-platform-dx/web-features/blob/21d9b8ea048734277a5a7b30054b81450c9743f7/features/attr.yml.dist#L47

It isn't supported anywhere, and is marked as experimental in BCD.

I propose to split it off into its own feature.

The other keys are either Baseline low, or very close to it. Safari TP 242 supports the new `attr()` function (and it is part of Interop 2026). By the end of the year, we can expect `attr()` to be reasonably usable by developers on all major browsers. So we should expect the feature to progress to Baseline low.

However, if we keep the above key, it won't.

ddbeck:
I've opened https://github.com/web-platform-dx/web-features/pull/4334 to partly(?) address this. I suspect the right thing here is to:

- Merge that PR, bringing the data in line with expectations.
- Open a new issue (or update this issue) to use the feature template, then sort out a name, description, etc. for the `<url>` type handling in `attr()`.

captainbrosset:
https://drafts.csswg.org/css-values-5/#typedef-attr-type says something interesting:
> Note: `<url>` is banned because, for now, [attr()](https://drafts.csswg.org/css-values-5/#funcdef-attr) values can’t be used in URLs at all due to security concerns.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/attr#attr-type also says:
> Note: For [security reasons](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/attr#limitations_and_security) [`<url>`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/url_value) is not allowed as an attr() data type.

And links to https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/attr#limitations_and_security

ddbeck:
@captainbrosset interesting! This leads to me to two thoughts:

- Any (reasonable) use of a `<url>` CSS value goes through the `url()` CSS function or is the return value of the `url()` function or both.
- Notwithstanding the [spec defining `url` as possible type name](https://drafts.csswg.org/css-values-5/#typedef-syntax-type-name) for `type(<…>)`, the "banned" spec text you quoted effectively forbids the use of `<url>` in `type(…)`.

[...]

PR #4334, opened 2026-09-10
Don't let `attr()`'s status block on `<url>` type

ddbeck:
The `<url>` type is probably a distinct feature. But since that's going to require a distinct name, description, spec, etc. I thought it would be nice to partly address https://github.com/web-platform-dx/web-features/issues/4329 right away by taking the `css.types.attr.type_function.url` key out of the critical path.


captainbrosset (approved):
Really nice use of `spare`. Happy that we now have this ability.

PR #4341, opened 2026-09-10
Add feature for extended-lifetime shared workers

jgraham:


features/shared-workers-extended-lifetime.yml
```diff
@@ -0,0 +1,6 @@
+name: Shared worker extended lifetime
+description: Allows shared workers extra time to perform operations after all their owner globals have disappeared.
```
ddbeck: A couple possible versions of this. Narrowly, we must mention the entry point to this feature:

```suggestion
description: The `extendedLifetime` option of the `SharedWorker()` constructor gives shared workers extra time to perform operations after all their owner globals have disappeared.
```

But we might consider adopting part of [MDN's phrasing](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker/SharedWorker#extendedlifetime:~:text=the%20shared%20worker%20is%20allowed%20to%20remain%20alive%20for%20a%20short%20period%20after%20all%20pages%20using%20it%20have%20been%20navigated%20away%20from%20or%20closed) here, which I think hints at possible applications a little better:

```suggestion
description: The `extendedLifetime` option of the `SharedWorker()` constructor gives shared workers extra time to perform operations after all pages using it have been navigated away from or closed.
```

features/shared-workers-extended-lifetime.yml
```diff
@@ -0,0 +1,6 @@
...
+spec: https://html.spec.whatwg.org/#dom-sharedworkeroptions-extendedlifetime
```
ddbeck: I think the tests will pass if you use the multipage URL, like BCD does:

```suggestion
spec: https://html.spec.whatwg.org/multipage/workers.html#dom-sharedworkeroptions-extendedlifetime
```

ddbeck (changes requested):
A couple of little things for your consideration. Thank you!

ddbeck (approved):

PR #4343, opened 2026-09-11
Add caniuse links to hyphenate-character and hyphenate-limit-chars

bartveneman:
Noticed these were missing while doing some frontend archeology.

There are also dozens of other caniuse links missing. What would be the preferred way to update these? Add them to this PR or create several smaller PRs?

### DIFF
--- a/features/hyphenate-character.yml
+++ b/features/hyphenate-character.yml
@@ -1,6 +1,7 @@
 name: Hyphenate character
 description: The `hyphenate-character` CSS property sets the character or string to use at the end of a line before a line break.
 spec: https://drafts.csswg.org/css-text-4/#hyphenate-character
+caniuse: wf-hyphenate-character
 group: text-wrap
 compat_features:
   - css.properties.hyphenate-character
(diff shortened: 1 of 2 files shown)


ddbeck:
Hi @bartveneman, thanks for your interest in helping out! The `wf-` features on caniuse are actually imported _from_ web-features, so these links end up being self-referential. I don't think we want to create those links, but that's not currently communicated anywhere. I've opened https://github.com/web-platform-dx/web-features/pull/4345 for a better error message.

There are other features where mappings might exist, but it's not exactly straightforward. We have some [guidance in our docs](https://github.com/web-platform-dx/web-features/blob/main/docs/guidelines.md#caniuse-values) to help you learn how to set those values. There are two further complications worth mentioning:

1. Some caniuse features are currently unmappable because the web-features ID represents a larger (or smaller) feature than caniuse covers (especially for older features). To map those, we need to be able to merge features and mint historical features; this is currently blocked by https://github.com/web-platform-dx/web-features/pull/4083 and related work.

2. There are a couple of WIP PRs, which you'll probably want to avoid duplicating:

   - https://github.com/web-platform-dx/web-features/pull/3301
   - https://github.com/web-platform-dx/web-features/pull/3300

If you're still interested in helping, running `npx tsx ./scripts/caniuse.ts` from the root of the project can show you a Markdown formatted list of caniuse features and whether they're mapped. If you're still keen to contribute, a PR of any size is welcome, though small PRs (yes, even one liners) tend to get merged more quickly. Limiting yourself to 3-5 open PRs at a time is nice too.

If there are questions, issues (or PRs with guesses) are also welcome. Thank you!

PR #4346, opened 2026-09-11
`stats.ts`: add commit hash and timestamp to output

ddbeck:
To make it easier to figure out _what_ is being compared between runs, I've added two new fields to the stats report:

- `hash` (the Git commit hash at HEAD when running the script)
- `timestamp` (the Git commit's authoring date)

Since this is going to break old stats reports anyway, I've also decided to correct a mistake I made previously, appending `Change` to the key names in the `change` object. Having to type `change.groupsCountChange` felt a bit redundant and this PR removes that redundancy.


Elchi3 (approved):
Thanks Daniel! 👍

PR #4358, opened 2026-09-11
Update `wasm-multi-value` spec URL

gsnedders:
Multi-value has been merged into the core WebAssembly spec.

### DIFF
--- a/features/wasm-multi-value.yml
+++ b/features/wasm-multi-value.yml
@@ -1,6 +1,6 @@
 name: Multi-value (WebAssembly)
 description: Instructions and blocks can produce multiple result values.
-spec: https://github.com/WebAssembly/spec/blob/main/proposals/multi-value/Overview.md
+spec: https://webassembly.github.io/spec/core/bikeshed/
 group: webassembly
 caniuse: wasm-multi-value
 compat_features:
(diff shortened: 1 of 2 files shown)


features/wasm-multi-value.yml
```diff
@@ -1,6 +1,6 @@
 name: Multi-value (WebAssembly)
 description: Instructions and blocks can produce multiple result values.
-spec: https://github.com/WebAssembly/spec/blob/main/proposals/multi-value/Overview.md
+spec: https://webassembly.github.io/spec/core/bikeshed/
```
ddbeck, line 3: Like the other PRs, is there a more specific section you can link to?

ddbeck (changes requested):

features/wasm-multi-value.yml
gsnedders: https://webassembly.github.io/spec/core/bikeshed/#release-20 would says:

> [Function types](https://webassembly.github.io/spec/core/bikeshed/#syntax-functype) allow more than one result
> 
> [Block types](https://webassembly.github.io/spec/core/bikeshed/#syntax-blocktype) can be arbitrary function types

So I guess we could link to those two sections?

PR #4359, opened 2026-09-11
Update `wasm-non-trapping-float-to-int` spec URL

gsnedders:
Non-trapping float-to-int has been merged into the core WebAssembly spec.

### DIFF
--- a/features/wasm-non-trapping-float-to-int.yml
+++ b/features/wasm-non-trapping-float-to-int.yml
@@ -1,6 +1,6 @@
 name: Non-trapping float-to-int conversion (WebAssembly)
 description: Saturating floating-point to integer conversion operators return the maximum or minimum integer value on overflow instead of trapping.
-spec: https://github.com/WebAssembly/spec/blob/main/proposals/nontrapping-float-to-int-conversion/Overview.md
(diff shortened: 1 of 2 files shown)


features/wasm-non-trapping-float-to-int.yml
```diff
@@ -1,6 +1,6 @@
 name: Non-trapping float-to-int conversion (WebAssembly)
 description: Saturating floating-point to integer conversion operators return the maximum or minimum integer value on overflow instead of trapping.
-spec: https://github.com/WebAssembly/spec/blob/main/proposals/nontrapping-float-to-int-conversion/Overview.md
+spec: https://webassembly.github.io/spec/core/bikeshed/
```
ddbeck, line 3: I'm not a subject matter expert here, but would it make sense to link to something more specific? (e.g., skimming both the proposal and the new spec, maybe something like https://webassembly.github.io/spec/core/bikeshed/#numeric-instructions%E2%91%A6 is reasonable?)

ddbeck (changes requested):
Thanks for all these PRs, @gsnedders. A question on this one.

PR #4360, opened 2026-09-11
Update `wasm-string-builtins` spec URL

gsnedders:
Proposal now has a rendered spec instead of a bare markdown doc.


jgraham (approved):

features/wasm-string-builtins.yml
```diff
@@ -1,6 +1,6 @@
 name: String builtins (WebAssembly)
 description: The WebAssembly builtin string functions mirror a subset of the JavaScript `String` API and adapt it to be efficiently callable without JavaScript glue code.
-spec: https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md
+spec: https://webassembly.github.io/js-string-builtins/js-api/
```
ddbeck, line 3: I don't know a lot about wasm, but would it make sense to link to something more specific here? (e.g., skimming both documents, maybe https://webassembly.github.io/js-string-builtins/js-api/#builtins-js-string is reasonable?)

ddbeck (changes requested):

PR #4364, opened 2026-09-14
Switch from Mocha to `node:test` test runner

ddbeck:
The migration eliminates 5 dependencies and many transitive dependencies, which should reduce noise in web-features Dependabot updates.

I'm following the trend set by https://github.com/mdn/browser-compat-data/pull/28992 and https://github.com/openwebdocs/mdn-bcd-collector/pull/3274.


Elchi3 (approved):
LGTM, nice one!

If you want to go more miles, c8 can be removed with node:coverage. I did that in https://github.com/openwebdocs/mdn-bcd-collector/pull/3287. Still behind a flag, though: https://nodejs.org/learn/test-runner/collecting-code-coverage

ddbeck:
Thanks for the review and pointer, @Elchi3! I'll follow up with a coverage PR. 👍

PR #4365, opened 2026-09-14
Add `initialPermissionStatus` discouraged feature

### LINKED ISSUE #4305: `initialPermissionStatus` (discouraged)
### Specification

Chrome is the only implementer, so this satisfies:

https://chromestatus.com/feature/5113729407647744

### Description

From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLGeolocationElement/initialPermissionStatus):

> … an enumerated value representing the permission status for the geolocation feature when the page first loads.

### Documentation

https://developer.mozilla.org/en-US/docs/Web/API/HTMLGeolocationElement/initialPermissionStatus

### Browser support

- BCD: `api.HTMLGeolocationElement.initialPermissionStatus`

ddbeck:
See also: https://github.com/mdn/browser-compat-data/pull/30519.

Fixes https://github.com/web-platform-dx/web-features/issues/4305.


features/initialpermissionstatus.yml
```diff
@@ -0,0 +1,8 @@
+name: initialPermissionStatus
```
Elchi3: I'm rusty with the conventions here, but should this indicate "geolocation" somehow in the name?

Elchi3 (approved):
Just a question re the name of it and wondering if we add deprecated features to groups? If we don't then this is fine, otherwise I would expect the geolocation group to be set as well with the new feature.

One or both of these things would help me tie this to "geolocation", otherwise LGTM! 👍

features/initialpermissionstatus.yml
ddbeck: Yes, you're absolutely right. Other discouraged features like this name the interface (e.g., `audiolistener-setposition-setorientation`). Also in the ID should go with it's "root" feature.

```suggestion
name: HTMLGeolocationElement initialPermissionStatus
```

features/geolocation-element-initialpermissionstatus.yml
```diff
@@ -0,0 +1,8 @@
+name: initialPermissionStatus
+description: The `initialPermissionStatus` property of the `HTMLGeolocationElement` API represents whether `geolocation` was previously granted, denied, or neither, when the page first loads.
+spec: https://wicg.github.io/PEPC/geolocation-element.html#dom-powerfulfeatureobserver-initialpermissionstatus
```
ddbeck, line 3:
```suggestion
spec: https://wicg.github.io/PEPC/geolocation-element.html#dom-powerfulfeatureobserver-initialpermissionstatus
group: geolocation
```

Elchi3 (approved):
Great, thank you! 👍

PR #4367, opened 2026-09-15
`select`: add `HTMLOptionsCollection` compat keys

ddbeck:
Assigns the keys for `HTMLOptionsCollection` to the `<select>` feature. This API is the thing you get from `HTMLSelectElement.prototype.options`.


Elchi3 (approved):
Thanks Daniel! 👍

PR #4371, opened 2026-09-16
Split out type=week|month input types from input-date-time

jgraham:
These aren't yet supported on all desktop browsers so it's misleading to claim they're part of a baseline:high feature


features/input-date-time-month.yml
```diff
@@ -0,0 +1,8 @@
+name: <input> type month
```
ddbeck: `input-date-time` is a bit weird with its name, this would make it more consistent with other inputs:

```suggestion
name: '<input type="month">'
```

features/input-date-time-week.yml
```diff
@@ -0,0 +1,8 @@
+name: <input> type week
```
ddbeck: Like before:

```suggestion
name: '<input type="week">'
```

ddbeck (commented):
Little consistency suggestions for the names. Thank you!

ddbeck (approved):

PR #4373, opened 2026-09-17
Add feature for multiple import maps

jgraham:


features/import-maps-multiple.yml
```diff
@@ -0,0 +1,5 @@
+name: Multiple import maps
+description: Multiple `<script type="importmap">` HTML element that are merged to produce the overall import map.
```
ddbeck: number agreement nit + maybe a simplification? I don't really 'get' this feature so this suggestion is presented weakly. 😅 

```suggestion
description: Multiple `<script type="importmap">` HTML elements merge to produce the overall import map.
```

ddbeck (changes requested):

ddbeck (approved):

PR #4383, opened 2026-09-21
Add textStream to html-streaming-setters

tunetheweb:
The `textStream()` convenience method was added as part of adding HTML streaming APIs so we should add that to this feature.

It would also help spot if browsers only implement the streaming APIs and not this which could cause interop issues.

More details:
https://developer.chrome.com/docs/web-platform/declarative-partial-updates#textstream_convenience_method


ddbeck:
A genuine question I have: is this convenience method _only_ (or mostly) to the benefit of the streaming setters? Or is this plausibly a standalone feature (e.g., anybody who uses `fetch()` for a purpose that is not directly changing the DOM)?

If the latter, I can still see assigning the keys to _two_ features—particularly if it's a strong interoperability signal to developers about whether or not the convenience method exists—but I'm just kinda testing the concept here, of whether this does not merit an independent separate feature entry.

tunetheweb:
It's a good question. As it's name suggests it's for streaming text and basically just wraps `.pipeThrough(new TextDecoderStream())`.

So in theory it can be used for any text streaming required. So I think there is an argument it shouldn't be too tightly couple like I'm proposing here, and maybe even should be it's own feature. I'm not sure it really belongs in `fetch` either like it currently is, before this PR, as it's what's happens before and after the fetch. I notice `TextDecoderStream` itself is part of the [`text-encoding`](https://github.com/web-platform-dx/web-features/blob/main/features/text-encoding.yml) feature to maybe it belongs there? Though I'm reluctant to knock that back from it's Baseline status for this new addition.

In practice however I suspect the primary use case, by a long shot, will be for streaming HTML with the new `html-stream-setter` methods. In fact that was the reason we added this, as no one had asked for it before that was introduced. I'd really like it to be part of that (even if it's also part of something else) to ensure interop there as it's an easy to miss this, when implementing the rest of these streaming methods.

Much like our `setHTML` sanitizer argument, there's a balance between grouping features how they practically will be used versus how they are strictly specced.

WDYT?

ddbeck (changes requested):
Agreed that we shouldn't (retroactively) add these convenience methods to an existing Baseline feature and send it backwards. It's definitely a new thing.

I'm inclined to let these keys remain with `html-streaming-setters` feature, with the possibility of later minting a parallel feature or separating them into a wholly distinct feature (if—for whatever unlikely reason—these develop into having wildly divergent support stories).

But let's leave some breadcrumbs for ourselves. I'm going to suggest moving the three keys to the bottom of the list with a comment like this:

```yaml
  # These convenience methods are possibly a parallel or independent feature,
  # but we choose to include them here in the spirit of feature parsimony. For
  # background, see https://github.com/web-platform-dx/web-features/pull/4383/
  - api.Blob.textStream
  - api.Request.textStream
  - api.Response.textStream
```

(Sorry, I couldn't get GitHub to do the suggestion thing properly.)

tunetheweb:
Works for me! Though I tweaked the wording slightly - see https://github.com/web-platform-dx/web-features/pull/4383/commits/00862999c22a70c76b05e6bcd42bd80c9480d6a5

ddbeck (approved):
Great. I like your version of the comment better. Thank you!

PR #4386, opened 2026-09-21
Add discouraged feature for `trimLeft()` and `trimRight()`

### LINKED ISSUE #4324: `trimLeft` and `trimRight` (discouraged)
### Specification

- https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#String.prototype.trimleft
- https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#String.prototype.trimright

### Description

They are aliases to `trimStart()` and `trimEnd()`, respectively.

For discouragement, this in Annex B. They're legacy aliases and shouldn't be used.

### Documentation

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd#browser_compatibility

Discovered in the course of evaluating https://github.com/web-platform-dx/web-features/issues/4300.

### Browser support

This data is under `javascript.builtins.String.trim{Start,End}` (e.g., [MDN table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd#browser_compatibility)). The first step here will be to break that data up (separately spec'd features should have separate keys in BCD, not `alternative_name` values).

ddbeck:
Fixes https://github.com/web-platform-dx/web-features/issues/4324


captainbrosset (approved):

ddbeck:
@captainbrosset Oof it was really broken when you reviewed. Want to take another quick look, with the new keys, dist, and alternative feature?

captainbrosset (approved):
Sorry about not catching the bad keys in my previous review. I focused specifically on the description, according_to and reason fields, and too quickly glanced at the rest. Looks good to me.

